PR #286, opened 2019-03-12
HARP-4087: Update threejs to v102.

zbigg:
Notable change:
* we use typings directly from `three` package
* `TrackballControls` and `TransformControls` do not have typings
* we must support stateful render target because WebGLRenderer.render doesn't support render target`

Ref: https://github.com/mrdoob/three.js/wiki/Migration-Guide


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1222,7 +1215,7 @@ export class MapView extends THREE.EventDispatcher {
      */
     setFovCalculation(fovCalculation: FovCalculation) {
         this.m_options.fovCalculation = fovCalculation;
-        this.calculateFocalLength(this.m_renderer.getSize().height);
+        this.calculateFocalLength(this.m_renderer.getSize(new THREE.Vector2()).height);
```
Astrak: Could we have a unique vector2 in the file that is reused there everytime instead?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1696,7 +1689,7 @@ export class MapView extends THREE.EventDispatcher {
      * concerns, this is the preferred approach.
      */
     private getFrameBufferSize() {
-        const { width, height } = this.m_renderer.getSize();
+        const { width, height } = this.m_renderer.getSize(new THREE.Vector2());
```
Astrak: same

@here/harp-mapview/lib/MapViewFog.ts
```diff
@@ -6,14 +6,14 @@
 
 import { Theme } from "@here/harp-datasource-protocol";
 import { HighPrecisionLineMaterial, SolidLineMaterial } from "@here/harp-materials";
-import { Fog, Material, Mesh } from "three";
+import { Fog, Material, Mesh, OrthographicCamera, PerspectiveCamera, Scene } from "three";
```
Astrak: Do we need to keep the direct imports?

@here/harp-mapview/lib/MapViewFog.ts
```diff
@@ -6,14 +6,14 @@
...
 
 /**
  * Manages the fog display in [[MapView]].
  */
 export class MapViewFog {
     private m_enabled: boolean = true;
-    private m_fog: THREE.Fog = new Fog(0x000000); // Default color asked by DefinitelyTyped.
+    private m_fog: Fog = new Fog(0x000000); // Default color asked by DefinitelyTyped.
```
Astrak: Now that we can rely on ThreeJS, this may be removed

@here/harp-mapview/lib/Utils.ts
```diff
@@ -6,7 +6,16 @@
...
+import {
+    Camera,
+    Math as MATH,
+    Matrix4,
+    PerspectiveCamera,
+    Plane,
+    Quaternion,
+    Raycaster,
+    Vector3
+} from "three";
```
Astrak: Do we need the direct imports or should we not just import THREE?

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -964,7 +964,7 @@ export class TextElementsRenderer {
         }
 
         const pixelRatio = this.m_mapView.pixelRatio;
-        const screenSize = this.m_mapView.renderer.getSize();
+        const screenSize = this.m_mapView.renderer.getSize(new THREE.Vector2());
```
Astrak: Can we keep a single Vector2 instance somewhere in this file?

@here/harp-mapview/test/UtilsTest.ts
```diff
@@ -168,6 +168,6 @@ describe("map-view#Utils", function() {
             const cube = new THREE.Mesh(geometry, material);
             scene.add(cube);
         }
-        expect(MapViewUtils.estimateObjectSize(scene)).to.be.equal(4078548);
+        expect(MapViewUtils.estimateObjectSize(scene)).to.be.equal(4082552);
```
Astrak: Why do these tests need to change?

@here/harp-materials/lib/CirclePointsMaterial.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Color, ShaderMaterial, Uniform } from "three";
+import { Color, IUniform, ShaderMaterial, ShaderMaterialParameters, Uniform } from "three";
```
Astrak: My bad, I did these direct `import from "three"`. Can we change that to the peer dependency as in the other files?

@here/harp-mapview/lib/composing/MSAARenderPass.ts
```diff
@@ -178,18 +178,19 @@ export class MSAARenderPass extends Pass {
...
                 renderer.setClearColor(0x000000);
+                renderer.clear();
             }
-            renderer.render(
-                this.m_quadScene,
-                this.m_localCamera,
-                this.renderToScreen ? undefined : writeBuffer,
-                i === 0
-            );
+            renderer.render(this.m_quadScene, this.m_localCamera);
```
Astrak: The performance / antialias example does not work anymore when using this internal MSAA, it breaks in the TryMe. Your code seems ok, so I tried to download it but got an error at downloading, so cannot check immediately,

@here/harp-mapview/lib/MapViewFog.ts
zbigg: @Astrak: It's you code, i'm just following what we have in code ;)

But i agree `import * ...` is more convenient.

@here/harp-mapview/lib/MapViewFog.ts
zbigg: @Astrak:
I don't follow.
What exactly can be removed ?

@here/harp-mapview/lib/composing/MSAARenderPass.ts
zbigg: Thanks, will check this.

@here/harp-mapview/lib/MapViewFog.ts
Astrak: The default color. @types/three required a colour to initialize the fog, but THREE.Fog actually does not. Depends on the implementation in the ThreeJS'typings.

@here/harp-mapview/test/UtilsTest.ts
zbigg: Because internal representation of 3JS structures changed ...
Just this.
This test measures estimated size of objects in bytes, looks like 3JS added some members here and there.

@here/harp-mapview/lib/MapViewFog.ts
zbigg: Actually, Fog requires color ...
https://github.com/mrdoob/three.js/blob/dev/src/scenes/Fog.d.ts#L14

Astrak: Interesting. I wonder if they have just copied the types from @types/three then. Because the argument is only used to create a `Color` object, where the argument is optional in the constructor.

@here/harp-mapview/lib/composing/MSAARenderPass.ts
zbigg: Apparently, 3JS has bug as it requires specifies:
```setRenderTarget(v: WEBGLRendereTarget | undefined)```

But throws exception, when you pass `undefined`.
`null` works as it was always working.

See https://github.com/heremaps/harp.gl/pull/286/commits/f8f96cc850e444e8eb93a2f18cb2d5658e556b9f#diff-8fbcf5dde65d97a3aed881c2d420db2cR143

File: @here/harp-text-canvas/lib/rendering/GlyphTextureCache.ts [hunk 10]
```diff
@@ -289,10 +289,20 @@ export class GlyphTextureCache {
...
+        let oldRenderTarget: THREE.RenderTarget | undefined;
+
+        const willClearGeometry = this.m_clearGeometry.drawCount > 0;
+        const willCopyGeometry = this.m_copyGeometry.drawCount > 0;
+
+        if (willClearGeometry || willCopyGeometry) {
+            oldRenderTarget = renderer.getRenderTarget();
+            renderer.setRenderTarget(this.m_rt);
+        }
         if (this.m_clearGeometry.drawCount > 0) {
```
amsgames94:
```suggestion
        if (willClearGeometry) {
```

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1222,7 +1217,7 @@ export class MapView extends THREE.EventDispatcher {
      */
     setFovCalculation(fovCalculation: FovCalculation) {
         this.m_options.fovCalculation = fovCalculation;
-        this.calculateFocalLength(this.m_renderer.getSize().height);
+        this.calculateFocalLength(this.m_renderer.getSize(new THREE.Vector2()).height);
```
amsgames94:
```suggestion
        this.calculateFocalLength(this.m_renderer.getSize(tmpVector).height);
```

PR #292, opened 2019-03-12
HARP-4826: Add sphere projection and OBBs.

robertoraggi:


@here/harp-examples/src/projection_sphere.ts
```diff
@@ -0,0 +1,187 @@
+
```
haraldF: License header missing

PR #296, opened 2019-03-13
HARP-4900: ZoomLevel alignment.

amsgames94:
Changed the ``zoomLevel`` calculation to align with other map renderers.

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -487,7 +487,6 @@ export class MapView extends THREE.EventDispatcher {
      */
     readonly mapRenderingManager: IMapRenderingManager;
 
-    private m_zoomLevelBias: number = 1;
     private m_createdLights?: THREE.Light[];
     private m_skyBackground?: SkyBackground;
     private readonly m_screenProjector = new ScreenProjector();
@@ -1070,6 +1069,13 @@ export class MapView extends THREE.EventDispatcher {
     }
 
     /**
+     * The distance (in pixels) between the screen and the camera.
+     */
+    get focalLength(): number {
+        return this.m_focalLength;
+    }
+
+    /**
      * The position in geo coordinates of the center of the scene.
      */
     get geoCenter(): GeoCoordinates {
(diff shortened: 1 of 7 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -492,7 +492,7 @@ export class MapView extends THREE.EventDispatcher {
      */
     readonly mapRenderingManager: IMapRenderingManager;
 
-    private m_zoomLevelBias: number = 1;
+    private m_zoomLevelBias: number = -1;
```
zbigg: Previously `zoomLevelBias` was multiplier, now you changed it to offset.

I don't have opinion which is good or bad, but it's a change.

Document at least in `get/set` zoomLevelBias that it's offset (or multiplier if you revert logic).

After thining, I think that offset is better than multiplier is hard to control. 
It's just much more easier to reason what is impact of -1, +1 than 0.9, 1.1

@here/harp-mapview/lib/Utils.ts
```diff
@@ -318,27 +313,12 @@ export namespace MapViewUtils {
...
-         * 3. That in most cases the curvature of the earth won't have a large impact. In the case
-         *    of zooming out to view the whole globe with a sphere projection, it will subdivide a
-         *    bit more than it would otherwise, but it shouldn't be much of an issue.
-         */
-        const halfCircumference = EarthConstants.EQUATORIAL_CIRCUMFERENCE / 2;
-        const targetSize = visibleHeight / targetSubdivisions;
-
-        const zoomLevel = Math.log2(halfCircumference / targetSize);
-        return geoUtils.MathUtils.clamp(zoomLevel, mapView.minZoomLevel, mapView.maxZoomLevel);
+        const tileSize = (256 * height) / mapView.focalLength;
```
nzjony: 256 should be documented, otherwise it isn't clear what this is doing.

nzjony: You will get different zoom levels for different size displays, I'm not sure this is what you want.

PR #298, opened 2019-03-13
HARP-4545: Implement animated extrusion of the buildings

SergiiKulakovskyi:
WIP: working on method that decides which extruded object should be animated and which not.

### DIFF
--- a/@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
+++ b/@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
@@ -1,0 +1,353 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { TileKey } from "@here/harp-geoutils";
+import { ExtrusionFeature } from "@here/harp-materials";
+import { MathUtils } from "@here/harp-utils";
+import { MapView, MapViewEventNames, RenderEvent } from "./MapView";
+import { Tile } from "./Tile";
+
+import * as THREE from "three";
+
+/**
+ * Animation states for extrusion effect
+ */
+export enum AnimatedExtrusionState {
+    None,
+    Started,
+    Playing,
+    Finished
+}
+
+/**
+ * Handles animated extrusion effect of the buildings in [[MapView]].
+ */
+export class AnimatedExtrusionHandler {
+    /**
+     * Animate the extrusion of the buildings if set to `true`.
+     */
+    enabled: boolean = true;
+    /**
+     * Duration of the building's extrusion in milliseconds
+     */
+    duration: number = 750;
+    /**
+     * If `forceEnabled` is set to `true` then `animateExtrusion` and `animateExtrusionDuration`
+     * values from [[extrudedPolygonTechnique]] will be ignored in [[Tile]] and
+     * `AnimatedExtrusionHandler.enabled` with `AnimatedExtrusionHandler.duration` will be used
+     */
+    forceEnabled: boolean = false;
+
+    private m_zoomLevelPrevious: number;
+    private m_tileHandlerMap: Map<Tile, AnimatedExtrusionTileHandler> = new Map();
+    private m_zoomDirection: number = 0;
+    private m_forceAnimatedExtrusion: boolean | undefined;
+    private m_forceAnimatedExtrusionDuration: number | undefined;
+
+    /**
+     * Creates an [[AnimatedExtrusionHandler]] in [[MapView]].
+     *
+     * @param m_mapView Instance of [[MapView]] that passes `zoomLevel`
+     * through the `zoom` poperty update
+     */
+    constructor(private m_mapView: MapView) {
+        this.m_zoomLevelPrevious = this.m_mapView.zoomLevel;
+    }
+
+    /**
+     * Returns direction of the last zoom event.
+     * Zoom in - positive value, zoom out - negative value
+     */
+    get zoomDirection(): number {
+        return this.m_zoomDirection;
+    }
+
+    /**
+     * [[MapView]] updates zoom level. Then [[AnimatedExtrusionTileHandler]] calculates actual
+     * extrusion ratio and trigger animation
+     */
+    set zoom(zoomLevel: number) {
+        // if zoomLevel has been changed since last render
+        if (this.m_zoomLevelPrevious !== zoomLevel) {
+            this.m_tileHandlerMap.forEach(tileHandler => {
+                this.m_zoomDirection = zoomLevel > this.m_zoomLevelPrevious ? 1 : -1;
+                tileHandler.zoomLevelChanged(this.m_zoomDirection);
+            });
+            this.m_zoomLevelPrevious = zoomLevel;
+        }
+    }
+
+    /**
+     * Checks whether animated extrusion effect was forcibly turned off/on in [[MapView]]
+     */
+    get forceAnimatedExtrusion(): boolean | undefined {
+        return this.m_forceAnimatedExtrusion;
+    }
+
+    /**
+     * If set to `true` it will force the animation extrusion effect to run.
+     * In case it was set to `false` animation extrusion effect will be switched off.
+     */
+    set forceAnimatedExtrusion(animatedExtrusion: boolean | undefined) {
+        this.m_forceAnimatedExtrusion = animatedExtrusion;
+    }
+
+    /**
+     * Returns `animatedExtrusionDuration` value that was set in [[MapView]].
+     */
+    get forceAnimatedExtrusionDuration(): number | undefined {
+        return this.m_forceAnimatedExtrusionDuration;
+    }
+
+    /**
+     * If value is set, it will overlap in [[Tile]] with
+     * `animatedExtrusionDuration` property from [[ExtrudedPolygonTechnique]]
+     * and `DEFAULT_DURATION` from [[AnimatedExtrusionTileHandler]].
+     */
+    set forceAnimatedExtrusionDuration(extrusionDuration: number | undefined) {
+        this.m_forceAnimatedExtrusionDuration = extrusionDuration;
+    }
+
+    /**
+     * Adds an [[AnimatedExtrusionTileHandler]] to [[AnimatedExtrusionHandler]]
+     */
+    add(tileHandler: AnimatedExtrusionTileHandler): void {
+        this.m_tileHandlerMap.set(tileHandler.tile, tileHandler);
+    }
+
+    /**
+     * Removes tile from the list subscribed for extrusion ratio updates
+     */
+    removeTile(tile: Tile): void {
+        this.m_tileHandlerMap.delete(tile);
+    }
+
+    /**
+     * Returns first [[AnimatedExtrusionTileHandler]] existed from the list of [[Tile]]s
+     */
+    find(tileKeys: Array<TileKey | undefined>): AnimatedExtrusionTileHandler | undefined {
+        for (const tileHandler of this.m_tileHandlerMap) {
+            for (const tileKey of tileKeys) {
+                if (
+                    tileKey !== undefined &&
+                    tileHandler[0].tileKey.mortonCode() === tileKey.mortonCode()
+                ) {
+                    return tileHandler[1];
+                }
+            }
+        }
+        return undefined;
+    }
+}
+
+/**
+ * Implements animated extrusion effect for the extruded objects in the [[Tile]]
+ */
+export class AnimatedExtrusionTileHandler {
+    /**
+     * Minimum ratio value for extrusion effect
+     */
+    static DEFAULT_RATIO_MIN: number = 0.001;
+    /**
+     * Maximum ratio value for extrusion effect
+     */
+    static DEFAULT_RATIO_MAX: number = 1;
+
+    private m_extrudedObjects: THREE.Object3D[] = [];
+    private m_animatedExtrusionRatio: number = AnimatedExtrusionTileHandler.DEFAULT_RATIO_MAX;
+    private m_animatedExtrusionState: AnimatedExtrusionState = AnimatedExtrusionState.None;
+    private m_animatedExtrusionStartTime: number | undefined = undefined;
+    private m_mapView: MapView;
+    private m_animatedExtrusionHandler: AnimatedExtrusionHandler;
+
+    constructor(
+        private m_tile: Tile,
+        extrudedObjects: Array<{ object: THREE.Object3D; materialFeature: boolean }>,
+        private m_animatedExtrusionDuration: number
+    ) {
+        this.m_mapView = m_tile.mapView;
+        this.m_animatedExtrusionHandler = this.m_mapView.animatedExtrusionHandler;
+
+        extrudedObjects.forEach(extrudedObject => {
+            if (extrudedObject.materialFeature) {
+                ExtrusionFeature.addRenderHelper(extrudedObject.object);
+            }
+            this.m_extrudedObjects.push(extrudedObject.object);
+        });
+
+        this.startExtrusionAnimationIfNeeded(this.m_animatedExtrusionHandler.zoomDirection);
+    }
+
+    /**
+     * Set an extrusion ratio value for the materials [[MapMeshBasicMaterial]]
+     * and [[EdgeMaterial]]. Controlled by [[AnimatedExtrusionHandler]]
+     * for extrusion animation effect.
+     */
+    set extrusionRatio(value: number) {
+        this.m_animatedExtrusionRatio = value;
+
+        this.m_extrudedObjects.forEach(object => {
+            const material = (object as (THREE.Mesh | THREE.LineSegments))
+                .material as ExtrusionFeature;
+            material.extrusionRatio = this.m_animatedExtrusionRatio;
+        });
+    }
+
+    /**
+     * Returns the [[Tile]] related to [[AnimatedExtrusionTileHandler]]
+     */
+    get tile(): Tile {
+        return this.m_tile;
+    }
+
+    /**
+     * Return the current state of animated extrusion effect
+     */
+    get animationState(): AnimatedExtrusionState {
+        return this.m_animatedExtrusionState;
+    }
+
+    /**
+     * Cancel animation and remove from [[AnimatedExtrusionHandler]]
+     */
+    dispose() {
+        this.stopExtrusionAnimation();
+        this.m_animatedExtrusionHandler.removeTile(this.m_tile);
+    }
+
+    /**
+     * Start / Stop extrusion animation if zoom level was changed
+     */
+    zoomLevelChanged(zoomDirection: number) {
+        if (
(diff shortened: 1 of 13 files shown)


Astrak:
@heremaps-bot ok to test

@here/harp-materials/lib/Utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    insertedShaderName: string,
+    addTab?: boolean
+): string {
+    const tabChar = addTab === true ? "\t" : "";
+
+    const result = shaderContent.replace(
+        `#include <${shaderName}>`,
+        `
+#include <${shaderName}>
+// << Patched fading shader START >>
```
zbigg: > fading

Not relevant anymore

@here/harp-materials/lib/Utils.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Special function to insert shader includes after another shader include.
```
zbigg: Remove special, remove `function` as this adds no value, it's just:

> Insert shader includes after another shader include.

@here/harp-mapview/lib/Extrusion.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    private stopExtrusionAnimation(): void {
+        this.mapView.removeEventListener(MapViewEventNames.AfterRender, this.animate);
+
+        this.m_tilesSet.forEach(
+            animationParams => (animationParams.animationStartTime = undefined)
+        );
+    }
+
```
Astrak: `startExtrusionAnimation` and `stopExtrusionAnimation` are then common to all tiles? I can see that all buildings are going up at the same time. When a new tile is loaded in between, `stopExtrusionAnimation` is called, which brings the building down, etc. until all tiles are loaded, and finally the buildings can reach their correct height.

It seems instead that these function should be independent between tiles.

In some previous reviews there were worries that adding up `requestAnimationFrame` for each tile was a problem from the framerate, but in fact it is just a callback that the browser calls when a frame is ready. When you add the unique listener to `AfterRender`, it is the same, and we could have as many as these small callbacks as there are tiles in the view. It should not be a performance problem.

zbigg:
@heremaps-bot ok to test

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Check if extrusion animation should be done for this tile
+     */
+    shoudAnimate(dataSource: DataSource, mortonCode: number) {
+        // console.log(this.m_tileKeysAnimatedSet.size);
+        // if (this.m_zoomLevelPrevious > this.m_zoomLevel) {
+        //     return false;
+        // }
```
Astrak: Please remove inline code

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
```diff
@@ -0,0 +1,132 @@
...
+ */
+export class AnimatedExtrusionHandler {
+    /**
+     * Animate the extrusion of the buildings if set to `true`.
+     */
+    static DEFAULT_ENABLED = true;
+    /**
+     * Duration of the building's extrusion in milliseconds
+     */
+    static DEFAULT_DURATION = 400;
```
Astrak: Sometimes the tiles pop in so fast on the corners of the screen that we don't see the animation. When trying locally I find that a duration of 1sec gives much more smoothness to the pan experience. That's personal though, as you prefer.

Astrak:
Forgot to write why I am requesting changes while my 2 comments are minors: as sent in DM, there is a bug that appears in the rendering sometimes, I haven't identified where it comes from though. It needs to be ironed out before merging.

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
```diff
@@ -0,0 +1,132 @@
...
+     */
+    static DEFAULT_RATIO_MAX: number = 1;
+
+    private m_zoomLevel: number;
+    private m_zoomLevelPrevious: number;
+    private m_tilesSet: Set<Tile> = new Set();
+    private m_tileKeysAnimatedSet: Set<TileKey> = new Set();
+
+    /**
+     * Constructs a `MapViAnimatedExtrusionHandlerewFog` instance.
```
zbigg: copy pasta

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
```diff
@@ -0,0 +1,132 @@
...
+ */
+export enum AnimatedExtrusionState {
+    None,
+    Started,
+    Playing,
+    Finished
+}
+
+/**
+ * Creates `AnimatedExtrusionHandler` that controls animated extrusion effect for
```
zbigg: It doesn't create anything, it 'is'.

More about architecture,

When i looked at this class i envisioned it will handle extrusion in whole MapView or in Tile.

Now, i see this one manages extrusion in MapView, but it's just a simple list.

I don't like that `Tile` which is already huge class gets new almost ten members and methods for this feature, i think that we need separate class also for `AnimatedExtrusionTileHandleer`.

And ofc, describe difference and relationship between those two classes.

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -407,6 +413,183 @@ export class FadingFeatureMixin implements FadingFeature {
...
+     * @param additionalCallback If defined, this function will be called before the function will
+     *          return.
+     */
+    export function addRenderHelper(object: THREE.Object3D) {
+        object.onBeforeRender = chainCallbacks(
+            object.onBeforeRender,
+            ExtrusionFeature.onBeforeRender
+        );
+    }
+
```
zbigg: Add doc like this:

Handles animated extrusion on each frame. Should be installed as respective `Object3D.onBeforeRender` of meshes which use animated extusion feature.

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -407,6 +413,183 @@ export class FadingFeatureMixin implements FadingFeature {
...
+                return this.getExtrusionHeight();
+            },
+            set: val => {
+                this.setExtrusionHeight(val);
+            }
+        });
+    }
+
+    /**
+     *
```
zbigg: Extra empty line in comment ?

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -407,6 +413,183 @@ export class FadingFeatureMixin implements FadingFeature {
...
+     * Apply the extrusionRatio value from the parameters to the respective properties.
+     *
+     * @param params `ExtrusionMeshBasicMaterial` parameters.
```
zbigg: if you there is no way to write meaningful  param doc and signature says it all, just don't document it ...

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -407,6 +413,183 @@ export class FadingFeatureMixin implements FadingFeature {
...
+    /**
+     * @see [[ExtrusionFeature#extrusion]]
+     */
+    protected getExtrusionHeight(): number {
+        return this.m_extrusion;
+    }
+    /**
+     * @see [[ExtrusionFeature#extrusion]]
+     */
+    protected setExtrusionHeight(value: number) {
```
zbigg: leftover from renaming height -> ratio ?

(same with getter above)

@here/harp-utils/lib/Functions.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+type chainFunction<T> = (...args: T[]) => T;
+
+/**
+ * Chains two functions for further assigning as one wrapped callback function
+ */
+export function chainCallbacks<T extends chainFunction<T>>(f1: any, f2: any): any {
```
zbigg: why generic T if not used at all ?

I know it's hard to write proper typings for this, at least shows the intent in typing albeit impl is not type-safe:
```
function chainCallbacks<T extends (this: unknown, ...args: any[]) => any>(
    f1: T | null | undefined,
    f2: T
): T {
    return function(this: any, ...args: any[]): ReturnType<T> {
        if (f1) {
            f1.apply(this, args);
        }
        return f2.apply(this, args);
    } as T;
}
```

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
```diff
@@ -0,0 +1,132 @@
...
+ * extruded objects in [[Tile]] when they appear on the scene.
```
zbigg: not in `Tile` but in `MapView`.

Albeit as written above, this code begs for `AnimatedExtrusionTileHandler`

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
```diff
@@ -0,0 +1,132 @@
...
+        this.m_zoomLevelPrevious = this.m_mapView.zoomLevel;
+        this.m_zoomLevel = this.m_mapView.zoomLevel;
+    }
+
+    /**
+     * [[MapView]] updates zoom level. Then calculate actual extrusion ratio and trigger animation
+     */
+    set zoom(zoomLevel: number) {
+        // if zoomLevel has been changed since last render
+        if (this.m_zoomLevel !== zoomLevel) {
```
zbigg: If you use `this.m_zoomLevel` like this, then `...Previous` is not needed.

Actually, `m_zoomLevel` is not needed ans thou shall use only `m_zoomLevelPrevious`

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Check if extrusion animation should be done for this tile
+     */
+    shoudAnimate(dataSource: DataSource, mortonCode: number) {
+        // console.log(this.m_tileKeysAnimatedSet.size);
+        // if (this.m_zoomLevelPrevious > this.m_zoomLevel) {
+        //     return false;
+        // }
+
+        for (const animatedTileKey of this.m_tileKeysAnimatedSet) {
```
zbigg: That "one level" only "down" search smells with:

 - re-animation if we have extr-buildings at levels 14,15, 16 and - 16 will be animated again
 - animation of level 15, if initial view was at 16

Am i right ?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1828,6 +1839,8 @@ export class MapView extends THREE.EventDispatcher {
         }
         ++this.m_frameNumber;
 
+        this.m_animatedExtrusionHandler.zoom = this.m_zoomLevel;
```
zbigg: Too early.
`this.m_zoomLevel` for current frame is calculated in `this.updateCameras` call.

Current implementation always lags one render frame behind actual zoom level change.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -434,6 +437,14 @@ export class Tile implements CachedResource {
 
     private m_colorMap: Map<string, THREE.Color> = new Map();
 
+    private m_extrudedObjects: Array<THREE.Mesh | THREE.LineSegments> = [];
```
zbigg: All this stuff should be moved to already mentioned `AnimatedExtrusionTileHandler` as commented above.

With related methods of course.

@here/harp-mapview/lib/Extrusion.ts
```diff
@@ -0,0 +1,182 @@
...
+        this.m_tilesSet.delete(tile);
+    }
+
+    set zoom(zoomLevel: number) {
+        this.m_zoomLevelPrevious = this.m_zoomLevel;
+        this.m_zoomLevel = zoomLevel;
+
+        if (this.m_zoomLevelPrevious !== this.m_zoomLevel) {
+            if (
+                this.m_zoomLevel >= Extrusion.DEFAULT_EXTRUSION_ANIMATION_ZOOM_LEVEL &&
```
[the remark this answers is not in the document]
SergiiKulakovskyi: removed DEFAULT_EXTRUSION_ANIMATION_ZOOM_LEVEL and zoom level check with any default values

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
SergiiKulakovskyi: replaced with 750ms as a compromise

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
SergiiKulakovskyi: Refactored. As you suggested, now it has two classes AnimatedExtrusionHandler and AnimatedExtrusionTileHandler

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
SergiiKulakovskyi: Made a deep level check for both up & down directions

@here/harp-utils/lib/Functions.ts
SergiiKulakovskyi: I understand it wrongly at the beginning. Thank you for the function code with typings

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1413,13 +1420,23 @@ export class Tile implements CachedResource {
                     }
                 }
 
+                const extrudedObjects: Array<{
+                    object: THREE.Object3D;
+                    materialFeature: boolean;
```
ninok: Please add some comment what materialFeature means.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1413,13 +1420,23 @@ export class Tile implements CachedResource {
...
+                }> = [];
+
                 const renderDepthPrePass =
                     technique.name === "extruded-polygon" && isRenderDepthPrePassEnabled(technique);
 
                 if (renderDepthPrePass) {
                     const depthPassMesh = createDepthPrePassMesh(object as THREE.Mesh);
                     objects.push(depthPassMesh);
 
+                    extrudedObjects.push({
```
ninok: The depthPassMesh and the edgeObj are always added to the extrudedObjects whereas the object is added only if:
`                    isExtrudedPolygonTechnique(technique) &&
                    animatedExtrusionHandler !== undefined`

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1461,10 +1478,50 @@ export class Tile implements CachedResource {
...
+                if (
+                    isExtrudedPolygonTechnique(technique) &&
+                    animatedExtrusionHandler !== undefined
+                ) {
+                    extrudedObjects.push({
+                        object,
+                        materialFeature: true
+                    });
+
+                    const extrudedPolygonTechnique = technique as ExtrudedPolygonTechnique;
```
ninok: This alias is not needed. Typescript is smart enough to detect that the type of technique is ExtrudedPolygonTechnique b/c you checked it above

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -0,0 +1,20 @@
...
+uniform float extrusionRatio;
+`,
+    extrusion_vertex: `
+transformed.z = transformed.z * extrusionRatio;
+`,
+    extrusion_pars_fragment: `
+uniform float extrusionRatio;
+`,
+    extrusion_fragment: `
+gl_FragColor.a *= smoothstep( 0.0, 0.25, extrusionRatio );
```
ninok: Why do we need to animate the alpha as well? Why is the factor at the end of the animation 0.25?

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export default {
+    extrusion_pars_vertex: `
+uniform float extrusionRatio;
+`,
+    extrusion_vertex: `
+transformed.z = transformed.z * extrusionRatio;
```
ninok: This will not work for the SphereProjection, instead you have to multiply the normal of the tile withe the height of the vertex and the extrusionRatio: 
```suggestion
float height = position.w; 
transformed = transformed + normal * height * extrusionRatio;
```

@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
```diff
@@ -0,0 +1,353 @@
...
+ */
+export class AnimatedExtrusionHandler {
+    /**
+     * Animate the extrusion of the buildings if set to `true`.
+     */
+    enabled: boolean = true;
+    /**
+     * Duration of the building's extrusion in milliseconds
+     */
+    duration: number = 750;
```
Astrak: Please initialize the value through a new constant outside of the class: `const DEFAULT_ANIMATION_DURATION = 750;`.

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
SergiiKulakovskyi: I decided to add alpha to make animation a bit smoothier. The appearance of extruded objects at one moment, before it starts extruding, disharmonize with the smooth animation of extrusion.
Adding alpha transition in the first quarter of animation ( 0.0, 0.25 ) makes user better experience. On my taste, I think it's better.

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
SergiiKulakovskyi: Thank you for pointing out!

@here/harp-mapview/lib/Tile.ts
SergiiKulakovskyi: You are right. It shouldn't be like that. I'll correct it

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
Astrak: I agree with Sergii, and it's also what 2gis is doing, however I'm not sure it's in the scope of the task given that it may conflict with the fading in of tiles? Or that could be kept and dealt with especially if Sergii works on this other feature.

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
Astrak: Could that work if the position attribute always remains at 3 components?

PR #302, opened 2019-03-14
HARP-5108: Improved PointLabels VertexBuffer allocations.

amsgames94:
Modified how ``TextBufferObjects`` are computed and stored for ``point`` labels to minimize and delay VB allocations as much as possible.


amsgames94:
GPU tests please

PR #303, opened 2019-03-15
HARP-4667: Make tile cache computation based on actual memory usage

StefanDachwitz:

Make cache usage computation based on memory estimation, not number of tiles. Since the visible tiles will not be discarded from the cache, the cache size may be larger than the cache capacity


@here/harp-lrucache/lib/LRUCache.ts
```diff
@@ -60,21 +60,15 @@ export class LRUCache<Key, Value> {
...
-
     /**
      * Creates a new instance of `LRUCache`. The optional sizeFunction can be used
      * fine tune the size required to cache that item.
      *
      * @param cacheCapacity The maximum number of entries to store in the cache.
      * @param sizeFunction A function determining the size per element.
      */
-    constructor(cacheCapacity: number, sizeFunction: (v: Value) => number = () => 1) {
+    constructor(cacheCapacity: number, public sizeFunction: (v: Value) => number = () => 1) {
```
haraldF: Any reason this should be public? At least, make it readonly as I don't think it's a good idea to replace the `sizeFunction` at runtime

@here/harp-lrucache/lib/LRUCache.ts
```diff
@@ -159,6 +153,26 @@ export class LRUCache<Key, Value> {
         this.evict();
     }
 
+    /**
+     * Updates the size of all elements in this cache. If their aggregated size is larger than the
+     * capacity, items will be evicted until the cache shrinks to fit the capacity.
+     */
+    checkCapacity(): void {
```
haraldF: This needs a better name as it actually mutates the state. `shrinkToSize` or something?

@here/harp-lrucache/lib/LRUCache.ts
```diff
@@ -159,6 +153,26 @@ export class LRUCache<Key, Value> {
...
+        let size = 0;
+        const sizeFunction = this.sizeFunction;
+
+        let entry = this.m_newest;
```
haraldF: Any reason you iterate from newest to oldest rather than just iterate over the internal `m_map`?

@here/harp-lrucache/lib/LRUCache.ts
StefanDachwitz: can do

@here/harp-lrucache/lib/LRUCache.ts
StefanDachwitz: It seemed to be the least overhead compared to the alternatives, of which there seem to be many. Do you have a specific loop in mind?

@here/harp-lrucache/lib/LRUCache.ts
StefanDachwitz: I can modify it to be a single function for all cases. Good idea.

@here/harp-lrucache/lib/LRUCache.ts
haraldF: The other functions use `m_map.forEach()`, you could also use `for (let [key, value] of this.m_map)`

StefanDachwitz: As I see it, none of your suggested approaches are comparable in performance or simplicity. Saving a few TypeScript chars to replace them with many JS/machine code bytes does not look like an improvement to me.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1193,6 +1213,13 @@ export class MapView extends THREE.EventDispatcher {
         return this.canvas.height;
     }
 
+    /**
```
zbigg: Not relevant.

zbigg: Not relevant.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -446,6 +460,229 @@ export namespace MapViewUtils {
...
+                        estimateTextureSize(uniform, isVisible, objectSize);
+                    }
+                }
+            }
+        } else if (
+            material instanceof THREE.MeshBasicMaterial ||
+            material instanceof MapMeshBasicMaterial
+        ) {
+            const meshMaterial = material;
+            estimateTextureSize(meshMaterial.map, isVisible, objectSize);
```
zbigg: Will it count one texture attached to several objects in several meshes as once ?

I don't see any tracking of already counted objects.

Just highlighting issue, not sure if it's iportant at all at this stage.

(we don't have shared textured anywhere, only in ImageCache)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -789,6 +789,26 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_visibleTiles.getDataSourceCacheSize();
     }
 
+    /**
+     * Checks if cache size should be in MB (mega bytes). If `false` the size is in number of
+     * tiles. The default is `true`.
+     *
+     * @returns {boolean} `true` if the size is in MB, `false` otherwise.
+     */
+    getCacheSizeInMb(): boolean {
```
zbigg: To me it reads 'get cache size' - so like count bytes in cache.

Maybe, `{get,set)ResourceUsageMode(mode: 'basic' | 'memory'`) ?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -446,6 +460,229 @@ export namespace MapViewUtils {
...
+        if (attribute.array.BYTES_PER_ELEMENT !== undefined) {
+            bytesPerElement = attribute.array.BYTES_PER_ELEMENT;
+        }
+        if (attribute instanceof THREE.InterleavedBufferAttribute) {
+            const interAttr = attribute;
+            if (interAttr.offset !== 0) {
+                // Array is shared between attributes, count it only once
+                return;
+            }
+            attrBytes = bytesPerElement * attribute.count * attribute.itemSize;
```
zbigg: > `* attribute.itemSize`

Shouldn't it be `attribute.data.stride` ?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -446,6 +460,229 @@ export namespace MapViewUtils {
...
+    }
+
+    /**
+     * Computes estimate for size of a single THREE.Mesh or THREE.Line object. Shared materials
+     * and/or attributes will be counted multiple times.
+     *
+     * @param object The mesh object to evaluate
+     * @param size The [[MemoryUsage]] to update.
+     */
+    export function estimateMeshSize(object: THREE.Object3D, objectSize: MemoryUsage): void {
```
zbigg: Without `ArrayBuffer` tracing, this one in particular with double memory usage for any re-used geometry like extruded buildings depth and pre-pass or future line foreground and background.

Again, this may be minor and room for enchancement.

@here/harp-lrucache/lib/LRUCache.ts
StefanDachwitz: Changed

@here/harp-lrucache/lib/LRUCache.ts
StefanDachwitz: implemented as requested

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: Relevant in an accompanying internal file that estimates memory usage. The memory used on the GPU depends on this setting.

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: Right, bad name. Modifying it.

@here/harp-datasource-protocol/lib/TileInfo.ts
```diff
@@ -92,6 +92,23 @@ export class FeatureGroup {
...
+        return (
+            (this.featureIds.length +
+                this.techniqueIndex.length +
+                this.textIndex.length +
+                this.positionIndex.length +
+                this.positions.length +
+                (this.layerIndex !== undefined ? this.layerIndex.length : 0) +
+                (this.classIndex !== undefined ? this.classIndex.length : 0) +
+                (this.typeIndex !== undefined ? this.typeIndex.length : 0)) *
+            4
```
amsgames94:
```suggestion
            8
```
Should be 8, since all ``number`` in javascript are actually doubles.

@here/harp-datasource-protocol/lib/TileInfo.ts
```diff
@@ -122,6 +139,16 @@ export class LineFeatureGroup extends FeatureGroup {
...
      */
     userData?: Array<{} | undefined> = [];
+
+    getNumBytes(): number {
+        return (
+            super.getNumBytes() +
+            ((this.segmentIds !== undefined ? this.segmentIds.length : 0) +
+                (this.segmentStartOffsets !== undefined ? this.segmentStartOffsets.length : 0) +
+                (this.segmentEndOffsets !== undefined ? this.segmentEndOffsets.length : 0)) *
+                4
```
amsgames94:
```suggestion
                8
```
Same as above.

@here/harp-mapview/lib/Utils.ts
StefanDachwitz: It looks pretty much like THREE.JS is doing it, no?:

	/**
	 * @param {Array<THREE.BufferGeometry>} geometry
	 * @return {number}
	 */
	estimateBytesUsed: function ( geometry ) {

		// Return the estimated memory used by this geometry in bytes
		// Calculate using itemSize, count, and BYTES_PER_ELEMENT to account
		// for InterleavedBufferAttributes.
		var mem = 0;
		for ( var name in geometry.attributes ) {

			var attr = geometry.getAttribute( name );
			mem += attr.count * attr.itemSize * attr.array.BYTES_PER_ELEMENT;

		}

		var indices = geometry.getIndex();
		mem += indices ? indices.count * indices.itemSize * indices.array.BYTES_PER_ELEMENT : 0;
		return mem;

	},

@here/harp-datasource-protocol/lib/TileInfo.ts
```diff
@@ -178,6 +205,18 @@ export class PolygonFeatureGroup extends FeatureGroup {
...
+
+    getNumBytes(): number {
+        return (
+            super.getNumBytes() +
+            ((this.outerRingStartIndex !== undefined ? this.outerRingStartIndex.length : 0) +
+                (this.innerRingIsOuterContour !== undefined
+                    ? this.innerRingIsOuterContour.length
+                    : 0) +
+                (this.innerRingStartIndex !== undefined ? this.innerRingStartIndex.length : 0)) *
+                4
```
amsgames94:
```suggestion
                8
```
Same as above.

@here/harp-datasource-protocol/lib/TileInfo.ts
```diff
@@ -249,6 +293,36 @@ export class ExtendedTileInfo implements TileInfo {
             this.typeCatalog = new Array<string>();
         }
     }
+
+    /**
+     * Compute the memory footprint caused by objects owned by the `ExtendedTileInfo`.
+     */
+    getNumBytes(): number {
+        let numBytes = 100;
```
amsgames94: Why do you start with a memory consumption of 100 bytes?

@here/harp-mapview/lib/Utils.ts
StefanDachwitz: I will add code to count attributes/arraybuffers and textures only once.

@here/harp-datasource-protocol/lib/TileInfo.ts
```diff
@@ -249,6 +293,36 @@ export class ExtendedTileInfo implements TileInfo {
...
+        for (const str of this.textCatalog) {
+            numBytes += 2 * str.length;
+        }
+
+        numBytes += this.techniqueCatalog.length * 100;
```
amsgames94: And this 100 here? If this numbers come from somewhere, please define them as constants as the beginning of the file for better readability.

@here/harp-mapview/lib/Utils.ts
StefanDachwitz: I will add code to count attributes/arraybuffers and textures only once. (see above) Counting the buffers twice for extruded buildings with outlines is actually really significant.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -305,9 +305,9 @@ export interface MapViewOptions {
     /**
      * Size of a tile cache in bytes.
```
amsgames94: Outdated documentation (it's a boolean now, cannot be a ``size``).

@here/harp-mapview/lib/Tile.ts
```diff
@@ -114,6 +115,26 @@ export interface TileFeatureData {
...
 }
 
+/**
+ * Compute the memory footprint of `TileFeatureData`.
+ */
+export function getFeatureDataSize(featureData: TileFeatureData): number {
+    let numBytes = 100;
+
+    if (featureData.ids !== undefined) {
+        numBytes += featureData.ids.length * 4;
```
amsgames94:
```suggestion
        numBytes += featureData.ids.length * 8;
```
Double.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -114,6 +115,26 @@ export interface TileFeatureData {
...
+    }
+    if (featureData.starts !== undefined) {
+        numBytes += featureData.starts.length * 4;
```
amsgames94:
```suggestion
        numBytes += featureData.starts.length * 8;
```
Double.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -114,6 +115,26 @@ export interface TileFeatureData {
     objInfos?: Array<{} | undefined>;
 }
 
+/**
+ * Compute the memory footprint of `TileFeatureData`.
+ */
+export function getFeatureDataSize(featureData: TileFeatureData): number {
+    let numBytes = 100;
```
amsgames94: Again, this initial estimation seems kind of weird. Please use constants in all similar places (will not add more comments for these).

@here/harp-mapview/lib/Tile.ts
```diff
@@ -160,6 +181,33 @@ export interface RoadIntersectionData {
...
+ */
+function getRoadIntersectionDataSize(intersectionData: RoadIntersectionData): number {
+    let numBytes = 100;
+
+    // 4 bytes per techniqueIndex
+    // 4 bytes per start
+    // 4 bytes per width
+    // 4 bytes per position
+    // 100 (estimated) bytes per technique
+    const bytesPerEntry = 4 + 4 + 4 + 4 + 100;
```
amsgames94: Again, fix all references to numbers as 8 bytes instead of 4 bytes.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1647,6 +1704,51 @@ export class Tile implements CachedResource {
...
+        let numTextElements = 0;
+        let numUserTextElements = 0;
+
+        const aggregatedObjSize = {
+            heapSize: 0,
+            gpuSize: 0
+        };
+
+        for (const object of this.objects) {
+            if (object.visible) {
```
amsgames94: Even if objects are not visible, they're still in memory, right?

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1647,6 +1704,51 @@ export class Tile implements CachedResource {
...
+
+        for (const group of this.textElementGroups.groups) {
+            numTextElements += group[1].elements.length;
+        }
+        numUserTextElements = this.userTextElements.length;
+
+        // 216 was the shallow size of a single TextElement last time it has been checked, 312 bytes
+        // was the minimum retained size of a TextElement that was not being rendered. If a
+        // TextElement is actually rendered, the size may be _much_ bigger.
+        heapSize += (numTextElements + numUserTextElements) * 312;
```
amsgames94: Except from the picking info, we can make a pretty good worst-case estimation of how much memory a ``TextElement`` consumes. Ping me if you need help with this.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -446,6 +460,229 @@ export namespace MapViewUtils {
...
+    ) {
+        if (texture === null || texture.image === undefined) {
+            return;
+        }
+
+        // May be HTMLImage or ImageData
+        const image = texture.image;
+        // Assuming RGBA
+        objectSize.heapSize += 4 * image.width * image.height;
+        if (isVisible) {
```
amsgames94: Again, even if it's not visible, the texture resource is resident on the GPU.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -446,6 +460,229 @@ export namespace MapViewUtils {
...
+            attrBytes = bytesPerElement * attribute.count * attribute.itemSize;
+        } else if (attribute instanceof THREE.BufferAttribute) {
+            const bufferAttr = attribute;
+            attrBytes = bytesPerElement * bufferAttr.count * bufferAttr.itemSize;
+        } else {
+            logger.warn("estimateMeshSize: unidentified attribute: ", attrName);
+        }
+
+        objectSize.heapSize += attrBytes;
+        if (isVisible) {
```
amsgames94: Same as above, even if it's not visible, resources will be resident on the GPU.

@here/harp-text-canvas/lib/rendering/FontCatalog.ts
```diff
@@ -502,6 +503,34 @@ export class FontCatalog {
...
 
+    /**
+     * Update the info with the memory footprint caused by objects owned by the `FontCatalog`.
+     *
+     * @param info The info object to increment with the values from this `FontCatalog`.
+     */
+    updateMemoryUsage(info: MemoryUsage) {
+        let numBytes = 0;
+
+        for (const block of this.unicodeBlocks) {
```
amsgames94: This is not really an accurate estimation, as ``unicodeBlocks`` only contains info on which blocks are covered, but a single block could be covered by multiple fonts .

I think the ``FontCatalog`` memory estimation should be something like:
* Sum of all the info passed on construction (like ``unicodeBlocks`` and ``fonts``).
* ``GlyphTextureCache`` size estimation (already accurate).
* Sum of all the arays contained in all ``m_loadedJson``.
* Sum of all the textures in ``m_loadedPages`` (already accurate).
* Sum of all ``m_loadedGlyphs``.

@here/harp-text-canvas/lib/rendering/FontCatalog.ts
```diff
@@ -502,6 +503,34 @@ export class FontCatalog {
...
+        }
+
+        let textureBytes =
+            this.m_glyphTextureCache.textureSize.x * this.m_glyphTextureCache.textureSize.y * 3;
+
+        for (const page in this.m_loadedPages.entries) {
+            if (this.m_loadedPages.get(page) !== undefined) {
+                const loadedPage = this.m_loadedPages.get(page);
+                if (loadedPage !== undefined) {
+                    textureBytes += loadedPage.image.width * loadedPage.image.height * 3;
```
amsgames94:
```suggestion
                    textureBytes += loadedPage.image.width * loadedPage.image.height * 4;
```

Loaded textures are ``.png``, and thus probably contain an alpha channel.

@here/harp-datasource-protocol/lib/TileInfo.ts
StefanDachwitz: Done

@here/harp-datasource-protocol/lib/TileInfo.ts
StefanDachwitz: Done

@here/harp-datasource-protocol/lib/TileInfo.ts
StefanDachwitz: Done

@here/harp-datasource-protocol/lib/TileInfo.ts
StefanDachwitz: Every Object is a map that uses a lot of bytes for itself, and then add its properties. An object that actually takes up just 100 bytes would not be very big...

@here/harp-mapview/lib/Utils.ts
zbigg: Yes, but this:
```
if (interAttr.offset !== 0) {
```
causes that we count only first attribute, so the first one should report memory of whole buffer.

THREE.Js is iterating over all the attributes.
...

BTW, why are we writing all this stuff by ourselves, if we have something is already in THREE.JS ?
Is it so unusable ?

@here/harp-text-canvas/lib/rendering/FontCatalog.ts
StefanDachwitz: They are actually RGB, whiy would they need an alpha channel? The PNG format does not require it...

@here/harp-mapview/lib/Utils.ts
StefanDachwitz: Added handling of shared Geometries, Materials, Textures and Attributes.

@here/harp-datasource-protocol/lib/TileInfo.ts
StefanDachwitz: Done

@here/harp-datasource-protocol/lib/TileInfo.ts
StefanDachwitz: Done

@here/harp-datasource-protocol/lib/TileInfo.ts
StefanDachwitz: Added constants

@here/harp-mapview/lib/Tile.ts
StefanDachwitz: Decided to keep current estimation, as TextElement handling will be revised thoroughly in the future.

@here/harp-mapview/lib/Utils.ts
StefanDachwitz: Removed the "isVisible" test, because the handling of shared resources (like THREE.Materials and THREE.Attributes) does not support handling the visibility state of the resource.

@here/harp-mapview/lib/Utils.ts
StefanDachwitz: Removed the "isVisible" test, because the handling of shared resources (like THREE.Materials and THREE.Attributes) does not support handling the visibility state of the resource.

@here/harp-text-canvas/lib/rendering/FontCatalog.ts
StefanDachwitz: Fixed

@here/harp-datasource-protocol/lib/TileInfo.ts
StefanDachwitz: Added constants

@here/harp-datasource-protocol/lib/TileInfo.ts
StefanDachwitz: Done

@here/harp-text-canvas/lib/rendering/FontCatalog.ts
StefanDachwitz: The main memory hogs are already accounted for. For improved precision, we can modify later.

@here/harp-mapview/lib/Utils.ts
StefanDachwitz: Fixed the issue with counting only the first usage. I made some wrong assumptions... 

The function estimateBytesUsed is only implemented for Geometries, which is only a small fraction of the features we would like to know about.

PR #304, opened 2019-03-15
HARP-4807: Flipped inverted path labels.

amsgames94:


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1801,14 +1801,16 @@ export class TextElementsRenderer {
...
+                tempScreenPosition.copy(screenPoints[0]);
+                for (let i = 0; i < screenPoints.length - 1; ++i) {
+                    textPath.add(new SimpleLineCurve(screenPoints[i], screenPoints[i + 1]));
+                }
+                // Flip the path if the label is gonna be rendered downwards.
+                if (textPath.getPoint(0.5).x - textPath.getPoint(0.51).x > 0) {
                     tempScreenPosition.copy(screenPoints[screenPoints.length - 1]);
+                    textPath = new THREE.Path();
                     for (let i = screenPoints.length - 1; i > 0; --i) {
                         textPath.add(new SimpleLineCurve(screenPoints[i], screenPoints[i - 1]));
```
nzjony: Could this be a performance regression, because all paths which are negative have their path added above cleared and then readded.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1801,14 +1801,16 @@ export class TextElementsRenderer {
...
-                    }
-                } else {
+
+                let textPath = new THREE.Path();
+                tempScreenPosition.copy(screenPoints[0]);
+                for (let i = 0; i < screenPoints.length - 1; ++i) {
+                    textPath.add(new SimpleLineCurve(screenPoints[i], screenPoints[i + 1]));
+                }
+                // Flip the path if the label is gonna be rendered downwards.
+                if (textPath.getPoint(0.5).x - textPath.getPoint(0.51).x > 0) {
```
nzjony: what does 0.5 mean, is this the center of the first path? or the center of the whole line?

amsgames94: I know, but the current issue is that just based on our current screen projected points, there's no way to know where the center of the final path will be (since every ``SimpleLineCurve`` might have different lengths). The path provide us the center of all the combined curves, which is why we use it for evaluation.

We could compute the length of every ``SimpleLineCurve`` segments as we project points to the screen, but this would happen for all labels, not only for the flipped ones.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
amsgames94: I know, but the current issue is that just based on our current screen projected points, there's no way to know where the center of the final path will be (since every ``SimpleLineCurve`` might have different lengths). The path provide us the center of all the combined curves, which is why we use it for evaluation.

We could compute the length of every ``SimpleLineCurve`` segments as we project points to the screen, but this would happen for all labels, not only for the flipped ones.

nzjony: makes sense

PR #306, opened 2019-03-15
HARP-4946: Fix empty Tiles

pit-rpg:


zbigg:
@heremaps-bot ok to test

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -236,6 +236,12 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
             } else {
                 // empty tiles are traditionally ignored and don't need decode
                 tile.forceHasGeometry(true);
```
StefanDachwitz: The call "forceHasGeometry(true)" should go after the "if (tileLoader.decodedTile !== undefined) {}" and be:

// Force to be handled as if tile has geometry, in case nothing has been extracted from the decodedTile.
tile.forceHasGeometry(tileLoader.decodedTile === undefined || !tile.hasGeometry());

If done like this, the "force" flag is only set if there is really no geometry, which may still be the case.

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -255,8 +260,15 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
 
                 tile.setDecodedTile(tileLoader.decodedTile);
             } else {
-                // empty tiles are traditionally ignored and don't need decode
-                tile.forceHasGeometry(true);
+                if (tileLoader.decodedTile !== undefined) {
+                    // to render at least an empty surface
```
StefanDachwitz: Please adhere to the coding style and general language rules:
- fill line length of 100 chars and avoid comments spread over multiple short lines
- no comma at the end of a sentence, but a colon.
- start of a sentence looks better with uppercase...
- these rules have not been applied to all parts of the code, but please follow them in new code.

pit-rpg:
the fix is does not work with the warp/OLP datasources

PR #308, opened 2019-03-18
HARP-4697: Refactored Theme and Technique.

robertoraggi:
Also, add the script 'generate-json-schema' to harp-datasource-protocol
to generate a JSON schema from the theme types.

---
This change is [Reviewable](https://reviewable.io/reviews/heremaps/harp.gl/308)

### DIFF
--- a/@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
+++ b/@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
@@ -1,0 +1,365 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { LoggerManager } from "@here/harp-utils";
+
+import { Expr, MapEnv } from "./Expr";
+import { isInterpolatedPropertyDefinition } from "./InterpolatedProperty";
+import { InterpolatedPropertyDefinition, InterpolationMode } from "./InterpolatedPropertyDefs";
+import { IndexedTechnique, Technique } from "./Techniques";
+import { Style, StyleSet } from "./Theme";
+
+export const logger = LoggerManager.instance.create("StyleSetEvaluator");
+
+interface StyleInternalParams {
+    /**
+     * Optimization: Lazy creation and storage of expression in a style object.
+     */
+    _whenExpr?: Expr;
+
+    /**
+     * Optimization: Index into table in StyleSetEvaluator.
+     * @hidden
+     */
+    _index?: number;
+}
+
+type InternalStyle = Style & Partial<StyleInternalParams>;
+
+/**
+ * Combine data from datasource and apply the rules from a specified theme to show it on the map.
+ */
+export class StyleSetEvaluator {
+    readonly styleSet: InternalStyle[];
+
+    private readonly m_renderOrderBiasGroups: Map<string, number> = new Map();
+    private readonly m_techniques: IndexedTechnique[] = [];
+
+    constructor(styleSet: StyleSet) {
+        let techniqueRenderOrder = 0;
+        const cloneStyle = (style: Style): Style => {
+            return {
+                ...style,
+                styles:
+                    style.styles !== undefined
+                        ? style.styles.map(subStyle => cloneStyle(subStyle))
+                        : undefined
+            };
+        };
+        this.styleSet = styleSet.map(style => cloneStyle(style));
+        const computeDefaultRenderOrder = (style: InternalStyle): void => {
+            if (style.renderOrderBiasGroup !== undefined) {
+                const renderOrderBiasGroupOrder = style.renderOrderBiasGroup
+                    ? this.m_renderOrderBiasGroups.get(style.renderOrderBiasGroup)
+                    : undefined;
+                if (
+                    style.renderOrderBiasRange !== undefined &&
+                    renderOrderBiasGroupOrder === undefined
+                ) {
+                    if (style.renderOrder !== undefined) {
+                        logger.warn(
+                            "WARN: style.renderOrder will be overridden if " +
+                                "renderOrderBiasGroup is set:",
+                            style
+                        );
+                    }
+                    const [minRange, maxRange] = style.renderOrderBiasRange;
+                    style.renderOrder =
+                        minRange < 0
+                            ? techniqueRenderOrder + Math.abs(minRange)
+                            : techniqueRenderOrder;
+                    techniqueRenderOrder += Math.abs(minRange) + maxRange;
+                    if (style.renderOrderBiasGroup) {
+                        this.m_renderOrderBiasGroups.set(
+                            style.renderOrderBiasGroup,
+                            style.renderOrder
+                        );
+                    }
+                    techniqueRenderOrder++;
+                } else if (renderOrderBiasGroupOrder) {
+                    if (style.renderOrder !== undefined) {
+                        logger.warn(
+                            "WARN: style.renderOrder will be overridden if " +
+                                "renderOrderBiasGroup is set:",
+                            style
+                        );
+                    }
+                    style.renderOrder = renderOrderBiasGroupOrder;
+                }
+            }
+            // search through child styles
+            if (style.styles !== undefined) {
+                for (const currStyle of style.styles) {
+                    computeDefaultRenderOrder(currStyle);
+                }
+            } else {
+                if (style.technique !== undefined && style.renderOrder === undefined) {
+                    style.renderOrder = techniqueRenderOrder++;
+                }
+            }
+        };
+        for (const style of this.styleSet) {
+            computeDefaultRenderOrder(style);
+        }
+    }
+    /**
+     * Find all techniques that fit the current objects' environment.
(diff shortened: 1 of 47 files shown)


robertoraggi:
The JSON schema can be generated with the script `generate-json-schema` of `@here/harp-datasource-protocol`. For example, any of the following commands should create the JSON schema `@here/harp-datasource-protocol/theme.schema.json`:

`$ lerna run generate-json-schema --scope=@here/harp-datasource-protocol`

`$ yarn workspace @here/harp-datasource-protocol generate-json-schema`

Some work is still needed, before merging this change:
  - some of the internal properties like of `Technique` (like `_index`, and `_renderOrderAuto`) should be removed
  - the properties of the techniques that are allowed to interpolate must be explicitly typed (e.g. like we do for lineWidth)

blazejkroll:

blazejkroll:

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -1079,8 +948,205 @@ export interface TextTechnique extends BaseTechnique {
...
+     * Clip the line outside the tile if `true`.
+     */
+    clipping?: boolean;
+    /**
+     * The name of the property containing the text.
+     */
+    textLabel?: string;
+}
+
+/*
```
amsgames94: Why license header here?

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -665,17 +554,15 @@ export interface BasicExtrudedLineTechnique extends BaseTechnique, PolygonalTech
      * Style of both end caps. Possible values: `"None"`, `"Circle"`. A value of undefined maps to
      * `"Circle"`.
      */
-    caps?: string;
+    caps?: "None" | "Circle";
```
amsgames94: Just as a general case, it would be a good moment to decide on letter casing policy for our themes, since we use ``value`` instead of ``Value`` in a lot of places, I would propose ``lowerCamelCase`` as our default.

I know this might make the PR more complex, so don't take this comment too strictly.

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -1079,8 +948,205 @@ export interface TextTechnique extends BaseTechnique {
...
+ */
+export interface LineMarkerTechnique extends MarkerTechniqueParams {
+    /**
+     * Name of technique. Is used in the theme file.
+     */
+    name: "line-marker";
+    /**
+     * Minimal distance between markers in screen pixels.
+     */
+    minDistance?: number;
```
amsgames94: Kind of weird that this is a member or ``LineMarkerTechnique`` and not of ``MarkerTechniqueParams`` (specially when this is the only technique where this happens). Wouldn't it make more sense to store this value inside of ``MarkerTechniqueParams``, or create ``LineMarkerTechniqueParams extends MarkerTechniqueParams``?

@here/harp-mapview/lib/DepthPrePass.ts
```diff
@@ -21,7 +21,7 @@ export const DEPTH_PRE_PASS_STENCIL_MASK = 0x01;
  *
  * @param technique [[BaseStandardTechnique]] instance to be checked
  */
-export function isRenderDepthPrePassEnabled(technique: BaseStandardTechnique) {
+export function isRenderDepthPrePassEnabled(technique: BaseStandardTechniqueParams) {
```
amsgames94:
```suggestion
export function isRenderDepthPrePassEnabled(techniqueParams: BaseStandardTechniqueParams) {
```

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1923,7 +1931,7 @@ export class Tile implements CachedResource {
     /**
      * Gets the fading parameters for several kinds of objects.
      */
-    private getFadingParams(technique: BaseTechnique): FadingParameters {
+    private getFadingParams(technique: BaseTechniqueParams): FadingParameters {
```
amsgames94:
```suggestion
    private getFadingParams(techniqueParams: BaseTechniqueParams): FadingParameters {
```

@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
```diff
@@ -0,0 +1,409 @@
...
+                technique.renderOrderBiasGroup = currStyle.renderOrderBiasGroup;
+            }
+            if (currStyle.attr !== undefined) {
+                Object.getOwnPropertyNames(currStyle.attr).forEach(property => {
+                    // check for valid attr keys
+                    // TODO: Do proper validation (somewhere else). See HARP-732
+                    if (this.validate) {
+                        if (property === "technique") {
+                            logger.warn(
+                                "WARNING: technique defined in attr " + "(deprecated):",
```
ninok: Remove the validation?

@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
```diff
@@ -0,0 +1,409 @@
...
+                                currStyle
+                            );
+                        } else if (property === "renderOrder") {
+                            logger.warn(
+                                "WARNING: renderOrder defined in attr " + "(deprecated):",
+                                currStyle
+                            );
+                        }
+                    }
+                    if (currStyle.attr !== undefined) {
```
ninok: This is already checked above

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -4,47 +4,44 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { MathUtils } from "@here/harp-geoutils";
-import * as THREE from "three";
+import {
```
ninok: What are we doing here? We are in Technique.ts and include sth. from Technique.ts? Luckily the compiler is not ending up in an infinite loop.

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -1089,8 +962,166 @@ export interface TextTechnique extends BaseTechnique {
...
+    /**
+     * The name of the property containing the text.
+     */
+    textLabel?: string;
+}
+
+/**
+ * Runtime representation of [[SquaresStyle]] as parsed by [[StyleSetEvaluator]].
+ */
+export interface SquaresTechnique extends PointTechniqueParams {
```
ninok: Maybe a good time to split this huge file into two, one for the params and one for the actual techniques

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -1089,8 +962,166 @@ export interface TextTechnique extends BaseTechnique {
...
+ * Technique used to render a mesh geometry with textures.
+ */
+export interface StandardTexturedTechnique extends StandardTexturedTechniqueParams {
+    name: "standard-textured";
+}
+
+/**
+ * Possible techniques that can be used to draw a geometry on the map.
+ */
+export type Technique =
```
ninok: This is really ugly and not extendable. Users that implement a datasource can not implement their own techniques. Instead we should have some interface for techniques (or separate techniques into geometry processing and materials), but I guess this will come with Robertos style rework.

zbigg: Agree, but you either have closed / properly typing or open interface.
It was always closed, and I choose to leave it as is as we have no clear vision where to go ...

Custom technique would require extension points in so many places, that it's rather separate task.

@here/harp-datasource-protocol/lib/Techniques.ts
zbigg: Sorry you must've hit some bad revision i've accidentaly pushed some time ago.

@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
zbigg: Fixed, thanks.

@here/harp-datasource-protocol/lib/Techniques.ts
zbigg: Done.

@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
zbigg: Agreed removed validation completly as it's of no use.

zbigg:
___
*[@here/harp-omv-datasource/test/OmvDecodedTileEmitterTest.ts, line 62 at r1](https://reviewable.io/reviews/heremaps/harp.gl/308#-LdQ-t_W48Eil-TBljPq:-LdcD-IrC46JVzcqJjeG:b-5013es) ([raw file](https://github.com/heremaps/harp.gl/blob/73e38f0c383f4adab56e199a0334eafe7f5e95bc/@here/harp-omv-datasource/test/OmvDecodedTileEmitterTest.ts#L62)):*
> IndexedTechnique


StandardTexturedTechnique is extension of all the `Techniques` with internal params, like _index.

Also, `dynamic-dispatch` in `Technique` (let's call it interface) is based on `name` field, so this one is a proper `StandardTexturedTechnique` with extension from `IndexedTechnique`...

PR #309, opened 2019-03-18
HARP-5103: Reduce TextCanvas memory usage.

amsgames94:
* Optimized ``GlyphTextureCache`` internal buffers to only store attributes relevant to glyph copy/clear operations.
* Modified ``TextGeometry`` to be dinamically resized, also reducing the initial allocation size.


amsgames94:
GPU tests please

PR #310, opened 2019-03-18
HARP-4923: Add default controls buttons.

Astrak:


musculman:
I am fine with this code as is, but I am not an expert on it. Perhaps you want to get a second opinion

PR #314, opened 2019-03-19
HARP-2497: Removed ZoomLevel from decoding.

amsgames94:


@here/harp-omv-datasource/lib/IGeometryProcessor.ts
```diff
@@ -98,14 +94,12 @@ export interface IGeometryProcessor {
      * @param layerName The name of the enclosing layer.
      * @param geometry The list of polygons.
      * @param env The enviroment containing the properties of the geometry.
-     * @param storageLevel The storage level of the data.
-     * @param displayZoomLevel The desired display zoom level.
+     * @param storageLevel The storage level of the data
```
StefanDachwitz: missing colon

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1187,6 +1187,17 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_zoomLevel;
     }
 
+    /**
+     * Returns the storage level for the given camera setup.
```
StefanDachwitz: Please add something like this: "Actual storage level of the rendered tile data also depends on [[Tile.storageLevelOffset]]."

@here/harp-omv-datasource/lib/IGeometryProcessor.ts
amsgames94: Good catch. Will fix.

@here/harp-mapview/lib/MapView.ts
amsgames94: Will do.

amsgames94:

PR #317, opened 2019-03-20
HARP-4924 Adding TileProxy

nzjony:
Note, the text rendering isn't yet working, I will add that in a follow up commit.
Tests are missing, these will be added also.


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1984,13 +1988,25 @@ export class MapView extends THREE.EventDispatcher {
         this.dispatchEvent(DID_RENDER_EVENT);
     }
 
-    private renderTileObjects(tile: Tile, zoomLevel: number) {
+    // Helper method to extract the list of objects from the supplied Tile or TileProxy
+    private getObjects(tileOrProxy: Tile | TileProxy) {
+        return tileOrProxy instanceof Tile ? tileOrProxy.objects : tileOrProxy.clonedObjects;
```
StefanDachwitz: if you rename "clonedObjects" to "objects", this whole function would not be required.

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -210,11 +210,17 @@ export class PickHandler {
             planeIntersectPosition.add(this.mapView.worldCenter);
 
             const cameraPos = this.mapView.worldCenter.clone().add(this.mapView.camera.position);
-
-            this.mapView.forEachVisibleTile(tile => {
+            const worldExtents = this.mapView.projection.worldExtent(0, 0).max.x;
+            this.mapView.forEachVisibleTile(tileOrProxy => {
+                // This removes any offset from the camera position, such that the intersection
+                // query takes place as if the camera is over the tile at offset 0
```
StefanDachwitz: a colon at the end of a sentence is always nice in a comment, especially if it starts in upper case, like a proper sentence.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -256,6 +257,46 @@ interface PolygonFadingParameters extends FadingParameters {
     lineFadeFar?: number;
 }
 
+/**
+ * Proxies a Tile with a given offset (a numeric value to represent the longitudinal shift of the
+ * tile (a multiple of 360 degrees)). The given offset should be non zero, because we don't need to
+ * proxy the Tile at offset 0 (a Tile is by definition at offset 0).
+ *
```
StefanDachwitz: Please explain the uses case in a bit more detail, since it messes up the code quite a bit: 
* Why was Tile.tile added
* Why is the proxy needed
* Why do we use cloned objects
* Why are cloned TextElements/objects acceptable (not many objects at that level)

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -59,30 +60,32 @@ export class TileEntry {
...
+ * Represents a unique TileKey and the area it takes up on screen.
  */
 class TileKeyEntry {
-    constructor(public tileKey: TileKey, public area: number) {}
+    constructor(public tileKey: TileKey, public offset: number, public area: number) {}
 }
 
 /**
- * Missing Typedoc
+ * Stores Tile's or TileProxy's in an LRUCache
```
StefanDachwitz: Stores [[Tiles]]s or [[TileProxy]]s in the [[LRUCache]].

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -132,13 +135,76 @@ export interface DataSourceTileList {
...
     /**
      * List of tiles that will be rendered. This includes tiles that are not in the visibleTiles
      * list but that are used as fallbacks b/c they are still in the cache.
      */
-    renderedTiles: Tile[];
+    renderedTiles: Array<Tile | TileProxy>;
+}
+
+/**
+ * Creates a unique key based on the supplied parameters. Note, the uniqueness
```
StefanDachwitz: Please fill the line length of 100 chars in all comments.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -132,13 +135,76 @@ export interface DataSourceTileList {
...
+ * is bounded by the bitshift. The mortonCode supports currently up to 26
+ * levels, the bitshift reduces this accordingly, so given the default bitshift
+ * of four, we support up to 22 levels. Given the current support up to level 19
+ * this should be fine.
+ * @param uniqueTileKey The unique tile key from which to compute the unique key.
```
StefanDachwitz: empty line before @param is standard, and indentation with two tabs for the following line looks nicer...

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -132,13 +135,76 @@ export interface DataSourceTileList {
...
+ * @param bitshift How much space we have to store the offset. The default of 4 means
+ * we have enough space to store 16 unique tiles.
+ */
+function getKeyForTileKeyAndOffset(tileKey: TileKey, offset: number, bitshift: number = 4) {
```
StefanDachwitz: Are these bitshift things not better located in MathUtils? And the morton code specialities in TileKey.ts?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -246,78 +313,96 @@ export class VisibleTileSet {
...
-                            ];
-
-                            contour.push(contour[0]);
-
-                            const n = contour.length;
-
-                            for (let p = n - 1, q = 0; q < n; p = q++) {
-                                subTileArea +=
-                                    contour[p].x * contour[q].y - contour[q].x * contour[p].y;
+                tilingScheme.getSubTileKeys(tileKey).forEach(childTileKey => {
```
StefanDachwitz: Please add a comments that motivate the changes due to the use case

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -526,18 +646,23 @@ export class VisibleTileSet {
...
+     * Computes the list of tiles that are to be rendered on screen.
+     *
      * Search cache to replace visible but yet empty tiles with already loaded siblings in nearby
      * zoom levels.
      *
      * Useful, when zooming in/out and when "newly elected" tiles are not yet loaded. Prevents
      * flickering by rendering already loaded tiles from upper/higher zoom levels.
      */
-    private fillMissingTilesFromCache() {
+    private computeTilesToRender() {
```
StefanDachwitz: I actually found the old name more descriptive, no?

@here/harp-mapview/lib/Tile.ts
zbigg: > Why do we use cloned objects

I don't thnk that 3JS can render one object in two places in scene.
Otherwise, we would need several render passes - each one for each "globe" when each globe has different x-offset.

On the other hand, we may reuse underlying geometry & materials so these clones should be smart.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -256,6 +257,46 @@ interface PolygonFadingParameters extends FadingParameters {
...
+ * The TileObject array and the TextElement array are cloned and stored in this object.
+ */
+export class TileProxy {
+    // Copies of the objects in the Tile.
+    clonedObjects: TileObject[] = [];
+
+    // Track if the Tile has been cloned, note we can't clone in the constructor because the tile
+    // may not yet have the data decoded
+    isCloned: boolean = false;
```
zbigg: > note we can't clone in the constructor because the tile
    // may not yet have the data decoded

So simple tracking doesn't work too (also looking at where 'clone' is actually called).

Note, that tile can be updated and current implementation doesn't support it at all.
You should just "listen" for original tile geometry changes and clone everything.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -246,78 +313,96 @@ export class VisibleTileSet {
...
-
-                            const n = contour.length;
-
-                            for (let p = n - 1, q = 0; q < n; p = q++) {
-                                subTileArea +=
-                                    contour[p].x * contour[q].y - contour[q].x * contour[p].y;
+                tilingScheme.getSubTileKeys(tileKey).forEach(childTileKey => {
+                    // Here we attempt to find unique Tiles with +/- 5 offsets (which means
+                    // there is a total of 11 possible tiles rendered)
+                    const longitudeOffset = Math.round(worldGeoPoint.longitudeInDegrees / 360);
```
zbigg: Intuition says, that we have search for different offsets only at root level.
If you happen to hit different offset root tile, all the children will be selected automatically.

So i think that current way of searching is correct, but have to be parametrized with offset and wrapped in something like this:
And just call it like this:
```
newGetTiles(...) {
   for (offsetCandidate in range[-5,5]) {
      oldGetTiles(..., offsetCandidate)
  }
}
```

Just thinking, i may be wrong.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -256,6 +257,46 @@ interface PolygonFadingParameters extends FadingParameters {
...
 }
 
+/**
+ * Proxies a Tile with a given offset (a numeric value to represent the longitudinal shift of the
+ * tile (a multiple of 360 degrees)). The given offset should be non zero, because we don't need to
+ * proxy the Tile at offset 0 (a Tile is by definition at offset 0).
+ *
+ * The TileObject array and the TextElement array are cloned and stored in this object.
+ */
+export class TileProxy {
```
zbigg: I concur with @StefanDachwitz comment and think that proxy is not needed at all.
All we need is a proper tile interface (`ITile` and two implementation
 - old one , that owns whole geometry and has `offset` = 0
 - proxy, that 
    - clones geometry of original tile when it's updated
    - have`offset != 0`
    - carefully disposes itself, so it doesn't dispose parent tile buffers and materials

Actually, it might be reasonable to refererence-count buffers and materials.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -625,15 +760,22 @@ export class VisibleTileSet {
...
-
+                            const childTileCode = getKeyForTileKeyAndOffset(childTileKey, offset);
                             checkedTiles.add(childTileCode);
-                            if (childTile !== undefined && childTile.hasGeometry) {
+
+                            const childTileOrProxy = tileCache.get(childTileCode);
+                            if (
+                                childTileOrProxy !== undefined &&
+                                childTileOrProxy.tile.hasGeometry &&
+                                !childTileOrProxy.tile.disposed
```
zbigg: > !childTileOrProxy.tile.disposed

This!
Why not allowing clone - which should be independent from original tile - to live if parent has been disposed?

Another pros for reference-counting underlying buffers & geometries.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -59,30 +60,32 @@ export class TileEntry {
...
     constructor(options: VisibleTileSetOptions) {
-        this.tileCache = new LRUCache<number, Tile>(options.tileCacheSize);
-        this.tileCache.evictionCallback = (_, tile) => {
-            if (tile.tileLoader !== undefined) {
+        this.tileCache = new LRUCache<number, Tile | TileProxy>(options.tileCacheSize);
+        this.tileCache.evictionCallback = (_, tileOrProxy) => {
+            if (tileOrProxy.tile.tileLoader !== undefined) {
                 // Cancel downloads as early as possible.
-                tile.tileLoader.cancel();
+                tileOrProxy.tile.tileLoader.cancel();
```
zbigg: This will allow proxies to cancel loading of `parent` visible tiles when proxy is being evicted.

@here/harp-mapview/lib/Tile.ts
zbigg: One proposal to track if geometry of `proxy` is corrent, is to introduce `Tile.geometryVersion`, which increased with each update of geometry.

Proxy, then can check in `willRender` (assuming it adheres to same interface as Tile!), will check if it has geometry which is up to date and if not, clone it ...

@here/harp-mapview/lib/Tile.ts
nzjony: - Tile.tile is described below, see get tile(), but can add here if you think that is necessary.
- Will update the comment to otherwise fix what you mentioned (include what @zbigg mentioned)

@here/harp-mapview/lib/Tile.ts
nzjony: You mean that I could add a call back into the Tile createGeometries function?

This is probably the cleanest way to do this and avoids storing an extra variable to track the geometry version as mentioned.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Thanks

@here/harp-mapview/lib/Tile.ts
zbigg: Not sure if callbacks are cleanest solution. 
You have to keep track of them in some array, allow addition, removal etc, add callbacks methods here and there.

Side note, immediate clone of geometry in callback would add performance penalty as it might be too much CPU work for one frame, so it's better to postpone it a little.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: No, because canEvict in this case would return false (proxies parent tile is still Visible) and prevent the eviction happening.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Done

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Done, missed this one unfortunately

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Agreed, have noted down to move this.

nzjony: These also are missing unit tests.

@here/harp-mapview/lib/VisibleTileSet.ts
zbigg: > canEvict in this case would return false

So this means that all proxies will live as long as their parent tile is visible and this is wrong too.
Proxies should have separate and completly independent lifecycles.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Good point, there is no need to test so many offsets I will fix.

nzjony: Good point, there is no need to test so many offsets I will fix.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: The old name is misleading IMO and caused some confusion until I had read the code, because this function computes the list of tiles to be rendered, as opposed to just filling missing tiles (which it also does yes).
Maybe one idea is to split this into two functions, one called computeTilesToRender, and another fillMissingTilesFromCache which takes the output of the first function and fills in the gaps, that could make the code a bit easier to read.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: > Why not allowing clone - which should be independent from original tile - to live if parent has been disposed?

Technically the clone is still alive if the parent has been disposed, and that is the problem. The TileProxy and the Tile are two separate entries in the cache. It is possible that the Tile has been evicted but not the TileProxy. In that case, if we don't make a check for it, we will try to render something which Proxies nothing, and therefore we get a blank area (this was visible when zooming in / out in the offsetted region).

The problem is that the Tile has no reference to the TileProxy, so when the Tile is evicted, it isn't easy to inspect the cache and remove all associated TileProxys, but I don't think this is a major, they will eventually be removed anyway.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Yes, this is required, I will move this into a separate function I think to make it clear

@here/harp-mapview/lib/MapView.ts
nzjony: Agreed

@here/harp-mapview/lib/PickHandler.ts
nzjony: You mean "full stop"? Agreed

@here/harp-mapview/lib/Tile.ts
nzjony: Ok, thanks for response, if that is the easiest way to do it, it sounds good.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: True, but I don't think this is a big issue, to get this specific behaviour, the user would need to jump a multiple of 360 longitudinal degrees to keep the parent tile visible and have a proxy which is then redundant. However, as soon as this tile is removed (Pan / zoom / rotate), both TileProxy's could be removed. Even when zoomed out, yes, there may be many TileProxy's visible, but they would be removed as soon as the user zooms in.

Or is there a specific use case where this could be a problem?

@here/harp-mapview/lib/Tile.ts
nzjony: @zbigg, I spoke with @StefanDachwitz after the Berlin standup.

I will look into see how much effort such an interface could be. It will reduce the mental effort required for sure, it could however be a lot of work to implement the interface for the proxy. Will investigate however.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Still missing, will work on this tomorrow.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Still open

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Still open

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Still open

@here/harp-mapview/lib/Tile.ts
nzjony: @zbigg, I have added geometryVersion, please see my latest commit.

@here/harp-mapview/lib/Tile.ts
nzjony: @zbigg , @StefanDachwitz , I have implemented the ITile interface and two concrete sub tiles: Tile and TileProxy, please review at your next convenience.

@here/harp-mapview/lib/Tile.ts
nzjony: I added these comments to the TileProxy class, see: https://github.com/heremaps/harp.gl/pull/317/commits/e9d3bc08a9761f80566d8a26d101f672f7f4ae34#diff-bec3ef09a66148da8cb27c7acb7625b7
So marking as resolved.

@here/harp-mapview/lib/VisibleTileSet.ts
zbigg: > but I don't think this is a big issue

That worries me so much, because we will rely on very tricky behaviors.
If we would have complete different lifecycle of tile/proxytile this wouldn't be tricky at all.

In general, i don't even know why proxy should reference original `Tile` object at all.
Why shouldn't it just call `dataSource.getTile(t)` on it's own ?

On high levels, cost of decoding and extra geometry is small.

Optimization - i.e reuse of geometry between normal tiles and offset-ed tiles can be added as next step.

zbigg: > On high levels, cost of decoding and extra

On low level, you usually wouldn't even have "original" non-offseted "parent" in cache at all!

nzjony: > That worries me so much, because we will rely on very tricky behaviors.

The idea isn't that particularly complicated, the tile cache has two types, Tiles and TileProxys which reference the Tile. A tile can be evicted, in which case the TileProxy should be evicted at the next convenience, because it is pointing at something useless, it is similar to a std::weak_ptr, we can see if what it is pointing to is still alive or not.

> If we would have complete different lifecycle of tile/proxytile this wouldn't be tricky at all.
> 
> In general, i don't even know why proxy should reference original `Tile` object at all.
> Why shouldn't it just call `dataSource.getTile(t)` on it's own ?

Yes, I had tried this originally, see the comment here: https://github.com/heremaps/harp.gl/pull/264#discussion_r262151610 however it was deemed a "no-go". I also would prefer this simpler approach to avoid this large change, but that might require some further discussion.

> On high levels, cost of decoding and extra geometry is small.

Agreed, but deemed not acceptable as in previous comment.

> Optimization - i.e reuse of geometry between normal tiles and offset-ed tiles can be added as next step.

Ok, given that you call dataSource.getTile(t) twice, how should one tile reuse the geometry of another Tile. This sounds potentially more complicated than the current idea of using a proxy.

nzjony: > > On high levels, cost of decoding and extra
> 
> On low level, you usually wouldn't even have "original" non-offseted "parent" in cache at all!

Not following, if the Tile or TileProxy is visible, then it will be in the cache.

@here/harp-mapview/lib/TileProxy.ts
```diff
@@ -0,0 +1,399 @@
...
+    /**
+     * @inheritdoc
+     */
+    get disposed(): boolean {
+        return this.tile.disposed;
+    }
+
+    /**
+     * @inheritdoc
+     */
```
nzjony: Remove

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: Moved to Utils.ts

@here/harp-mapview/lib/TileProxy.ts
```diff
@@ -0,0 +1,268 @@
...
+                clone.onAfterRender = tileObject.onAfterRender;
+                this.objects.push(clone);
+            }
+        }
+        // Todo: Clone text elements also.
+        this.geometryVersion = this.tile.geometryVersion;
+    }
+
+    /**
+     * Checks if the [[TileProxy]] can be cloned.
```
zbigg: Actually "should be cloned" ....
...

On the other hand, why is this even public ?

Shouldn't `TileProxy.prepareForRender` check if it's geometry copy is up to date ?

@here/harp-mapview/lib/TileProxy.ts
```diff
@@ -0,0 +1,268 @@
...
+    get numFramesVisible(): number {
+        return this.tile.numFramesVisible;
+    }
+
+    set numFramesVisible(val: number) {
+        this.tile.numFramesVisible = val;
+    }
+
+    get isVisible(): boolean {
+        return this.tile.isVisible;
```
zbigg: Sharing `isVisible` and `frameNumVisible` is not good idea, as we end-up with 
* `isVisible` but not actually visible, because out-of-frustum "parent Tiles"
* `IsVisible` but not actually visible and evictable! proxies, because some of it's parent proxies is visible...

This `isVisible` should be either renamed or well documented.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2277,6 +2283,10 @@ export class MapView extends THREE.EventDispatcher {
         let result: CopyrightInfo[] = [];
         for (const tileList of this.m_visibleTiles.dataSourceTileList) {
             for (const tile of tileList.renderedTiles) {
+                if (tile.isProxy) {
```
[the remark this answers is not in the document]
nzjony: Good point, I will fix.

@here/harp-mapview/lib/TileProxy.ts
nzjony: Clone is done in VisibleTileSet, hence being public.

However this could be a good alternative, I will take a look.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1984,13 +1985,18 @@ export class MapView extends THREE.EventDispatcher {
...
+        // World space longitudinal offset
+        const worldOffsetX = this.projection.worldExtent(0, 0).max.x * tile.offset;
         if (tile.willRender(zoomLevel)) {
             for (const object of tile.objects) {
                 object.position.copy(tile.center);
                 if (object.displacement !== undefined) {
                     object.position.add(object.displacement);
                 }
+                if (worldOffsetX !== 0) {
+                    object.position.x += worldOffsetX;
```
nzjony: @ninok, this is the translation that I was mentioning

@here/harp-mapview/lib/Utils.ts
```diff
@@ -481,3 +482,107 @@ export namespace MapViewUtils {
...
+     *
+     * Note, because bit shifting in JavaScript works on 32 bit integers, we use powers of 2 to set
+     * the high bits instead.
+     *
+     * @param offset Offset to pack into the high bits.
+     * @param bitshift How many bits to use to pack the offset.
+     */
+    export function getShiftedOffset(offset: number, bitshift: number = 3) {
+        let result = 0;
+        const maxOffset = Math.pow(2, bitshift);
```
sasha240100: Why not using bitwise shidting operator `<<` which is 2x times faster rather than `Math.pow(2, x)` ? 
http://jsfiddle.net/awncj81m/

haraldF: `<<` changes every number (double precision float) to a 32 bit unsigned integer, which is maybe not what we want here. In geoutils' `TileKey` I have some workaround for the slow Math.pow while still keeping the 53 bits precision of double

nzjony: @haraldF, thanks, yes that would be good, I will take a look.
@sasha240100 , I will add a comment to make it clear for those that aren't of Javascripts bitshift eccentricities

nzjony:
Closing because we found a better way to do this.

@here/harp-mapview/lib/Utils.ts
nzjony: sorry @sasha240100 , this shouldn't be reviewed, hence I have closed it, however the code in particular you reviewed, this I will keep, so hopefully soon I will have something to review.

PR #319, opened 2019-03-20
HARP-4946: Fix empty Tiles

pit-rpg:
Signed-off-by: Stepan <11714928+pit-rpg@users.noreply.github.com>

### DIFF
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -122,6 +122,14 @@ const MINIMUM_SMALL_OBJECT_SIZE_ESTIMATION = 16;
 const MINIMUM_OBJECT_SIZE_ESTIMATION = 100;
 
 /**
+ * A default empty [[DecodedTile]], for tiles that must be rendered but do not have any objects.
+ */
+const defaultEmptyDecodedTile: DecodedTile = {
+    techniques: [],
+    geometries: []
+};
+
+/**
  * Compute the memory footprint of `TileFeatureData`.
  */
 export function getFeatureDataSize(featureData: TileFeatureData): number {
@@ -593,10 +601,6 @@ export class Tile implements CachedResource {
      * Called by [[VisibleTileSet]] to mark that [[Tile]] is visible and it should prepare geometry.
      */
     prepareForRender() {
-        if (!this.m_decodedTile || this.m_disposed) {
-            return;
-        }
-
         // If the tile is not ready for display, or if it has become invisible while being loaded,
         // for example by moving the camera, the tile is not finished and its geometry is not
         // created. This is an optimization for fast camera movements and zooms.
@@ -604,13 +608,23 @@ export class Tile implements CachedResource {
             return;
         }
 
+        // Prevents of unnecessary re creation of objects
+        if (this.m_disposed || this.m_decodedTile === undefined) {
+            return;
+        }
         const decodedTile = this.m_decodedTile;
+        this.m_decodedTile = undefined;
+
+        // creates an empty plane if there are no objects on it, but it must be rendered
+        if (this.m_forceHasGeometry && this.dataSource.addTileBackground) {
+            this.clear();
+            this.addGroundPlane();
+            return;
+        }
 
         if (decodedTile.tileInfo !== undefined) {
             this.roadIntersectionData = this.dataSource.mapView.pickHandler.registerTile(this);
         }
-
-        this.m_decodedTile = undefined;
 
         setTimeout(() => {
             const stats = PerformanceStatistics.instance;
@@ -791,6 +805,9 @@ export class Tile implements CachedResource {
      * @param value A new value for the [[hasGeometry]] flag.
      */
     forceHasGeometry(value: boolean) {
+        if (value) {
+            this.setDecodedTile(defaultEmptyDecodedTile);
+        }
         this.m_forceHasGeometry = value;
     }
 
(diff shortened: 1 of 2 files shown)


@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -246,6 +246,11 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
         }
 
         tileLoader.loadAndDecode().then(() => {
+            // if decodedTile once set on tile - do nothing
+            if (tile.decodedTile !== undefined) {
```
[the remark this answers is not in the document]
pit-rpg: I tried to prevent toggling of m_forceHasGeometry when 
tile.forceHasGeometry(tileLoader.decodedTile === undefined || !tile.hasGeometry));
from 
https://github.com/heremaps/harp.gl/pull/306

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -255,8 +255,17 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
...
                 tile.setDecodedTile(tileLoader.decodedTile);
             } else {
-                // empty tiles are traditionally ignored and don't need decode
-                tile.forceHasGeometry(true);
+                if (tileLoader.decodedTile !== undefined) {
+                    // to render at least an empty surface
+                    // when the data is loaded
+                    // tile requires "decodedTile" object,
+                    tile.setDecodedTile(tileLoader.decodedTile);
+                }
```
nzjony: This to me sounds like a hack, we are giving the tile empty data to force it to render.
Please fix the root cause and not add a work around like this to "fix" the issue.

pit-rpg: A lot of logic depend on "decodedTile" property of the Tile class.
Including 
`tile.prepareForRender() {
    tile.createGeometries(decodedTile) {
        tile.createObjects(decodedTile, this.objects) {
            tile.createPlane(...);
        }
    }
 }`
call stack.
if the tile has no objects some decoders return empty decodedTile with no geometrys, and some of them return "unifuned".
the idea was to universalize behavior.
If the tile is not loadet the tile.decodedTile = undefined. Then when we get the data, but without any objects for the tile, then tile.decodedTile  = {default empty obj};
so behavior will be more predictable, and there will be less "if else" statements in the logic

zbigg: > the idea was to universalize behavior

This is good idea, but relying on `tile.decodedTile` is wrong because we want to forget `decodedTile` (and let GC kick-in) as soon as geometry is created.
This is the reason why `Tile.prepareForRender` resets `this.m_decodedTile` as soon geometry is created.

...

That's why we have `hasGeometry` flag.
This is definitive marker for rest of code, that this tile should be rendered.

On the other hand, w.r.t to you patch.
All this fiddling with flags can be replaced by:

Either this:
```
 tileLoader.loadAndDecode().then(() => {
     const decodedTile = tileLoader.decoderTile || defaultEmptyDecodedTile;
     // copyright stuff
     tile.setDecodedTile(decodedTile);
}
```
or this:

```
// Tile.forceHasGeometry
    forceHasGeometry() {
        this.createGeometries(defaultEmptyDecodedTile)
    }
```

Both of these approaches should fix issue.

nzjony: I like the idea that the forceHasGeometry calls createGeometry.... or maybe rather not call createGeometries, but some other simpler method, createDefaultPlane

@here/harp-mapview/lib/Tile.ts
```diff
@@ -604,6 +604,15 @@ export class Tile implements CachedResource {
             return;
         }
 
+        // creates an empty plane if there are no objects on it,
```
StefanDachwitz: Now I am writing for the third time:
Please fill the line length on 100 chars in the comments.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -593,7 +593,7 @@ export class Tile implements CachedResource {
      * Called by [[VisibleTileSet]] to mark that [[Tile]] is visible and it should prepare geometry.
      */
     prepareForRender() {
-        if (!this.m_decodedTile || this.m_disposed) {
+        if (this.m_disposed) {
```
zbigg: Are you aware that `prepareForRendered` is called before EVERY frame for particular tile ?

And ...

@here/harp-mapview/lib/Tile.ts
```diff
@@ -604,6 +604,14 @@ export class Tile implements CachedResource {
             return;
         }
 
+        // creates an empty plane if there are no objects on it, but it must be rendered
+        if (this.m_decodedTile === undefined) {
+            if (this.dataSource.addTileBackground && this.m_forceHasGeometry) {
+                this.addPlane();
```
zbigg: ... it looks like you're adding new `plane` (what plane?) every frame ...

Please check it, i may miss some fact.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1168,21 +1176,7 @@ export class Tile implements CachedResource {
...
-                this.center,
-                this.dataSource.tileBackgroundColor === undefined
-                    ? this.mapView.clearColor
-                    : this.dataSource.tileBackgroundColor,
-                this.dataSource.tileBackgroundIsVisible
-            );
-
-            this.registerTileObject(groundPlane);
-            objects.push(groundPlane);
+            this.addPlane();
```
zbigg: `addBackgroundPlane` or `addGroundPlane`

Just `plane` is too generic.

@here/harp-mapview/lib/Tile.ts
pit-rpg: yes, this is an old behavior, I did not modify this logic, only separate this function so that I could render empty tiles.

pit-rpg: as well as in each frame all geometry is removed from the video memory and then created again in the "createGeometries" method,
I understand it should be refactored but not in the context of this task

nzjony: You changed the logic to the entrance of the function as @zbigg mentioned above. m_decodedTile is always undefined, it is for one frame set, then it is cleared again in.

PR #323, opened 2019-03-22
MINOR: Limit default number of decoders to maximum of 4

StefanDachwitz:
* Reduce heap usage, since all workers have a copy of code and data

### DIFF
--- a/@here/harp-mapview/lib/ConcurrentWorkerSet.ts
+++ b/@here/harp-mapview/lib/ConcurrentWorkerSet.ts
@@ -9,6 +9,7 @@ import {
     IWorkerChannelMessage,
     LoggerManager,
     LogLevel,
+    MathUtils,
     WORKERCHANNEL_MSG_TYPE
 } from "@here/harp-utils";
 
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/ConcurrentWorkerSet.ts
```diff
@@ -157,7 +158,8 @@ export class ConcurrentWorkerSet {
         const workerCount = getOptionValue(
             this.m_options.workerCount,
             typeof navigator !== "undefined" && navigator.hardwareConcurrency !== undefined
-                ? Math.max(1, navigator.hardwareConcurrency) // we need to have at least one worker
+                ? // We need to have at least one worker
+                  MathUtils.clamp(navigator.hardwareConcurrency - 1, 1, 4)
```
nzjony: Why do you subtract one from the concurrency?

StefanDachwitz: So one thread is left for rendering.

PR #331, opened 2019-03-26
Support MapView on off-screen canvas without client dimensions.

zbigg:


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2460,4 +2461,28 @@ export class MapView extends THREE.EventDispatcher {
...
+            clientWidth === 0 ||
+            clientHeight === 0 ||
+            typeof clientWidth !== "number" ||
+            typeof clientHeight !== "number"
+        ) {
+            const pixelRatio = this.m_renderer.getPixelRatio();
+            return {
+                width: Math.round(this.canvas.width / pixelRatio),
+                height: Math.round(this.canvas.height / pixelRatio)
+            };
```
Astrak: Why not always return this value? Is there a benefit to use `clientWidth` over `width / pixelRatio`? ThreeJS automatically puts the CSS width to `width * pixelRatio`, via the DOM, so as far as I understand they should be equal and any client-side CSS should not be valid, unless set outside... but everything should break then, likely many ThreeJS features as well.

nzjony: clientWidth is good to use because the CSS engine could have shrunk the height / width, and using the canvas width or height, we won't get the actual size of the canvas that the user sees.

zbigg: I would say, that this is some heuristics.
Width and height is is used to
1) calculate aspect ratio
2) calculate focal length
3) convert screen/css/pixel space to other spaces

If `clientWidth/Height` is defined, then they are good for representing what client actually sees for all cases.
If it's not defined, `width`/`height` work well for 1, 2 and 3 actually doesn't matter at all becuse there are no UI interactions with off-screen canvas.

So, imho client size with fallback to adjusted canvas size is good enough, although i don't know if there is smarter way

Astrak: My concern is this: the only advantage of reading the dimensions ourselves via `clientWidth` and `clientHeight` is if the user sets a custom CSS value for the canvas. However one will unavoidably use [`setSize`, which overrules the stylesheet for inline CSS in the HTML](https://github.com/mrdoob/three.js/blob/master/src/renderers/WebGLRenderer.js#L379). So the only way to have unpredictable CSS dimensions (`clientWidth` / `clientHeight`) is to have a user who messes up the canvas' size, with javascript, after each `setSize` call. I can't much think of a use case for that thus far. So if we want to agree on the assumption that this will not happen (maybe there are use cases for it indeed?), we could then do the following:

We can use the ThreeJS methods `getSize` and `getDrawingBufferSize`. `getSize` returns the size on the page. [It is initialized with the canvas' dimensions](https://github.com/mrdoob/three.js/blob/master/src/renderers/WebGLRenderer.js#L152). And according to the above we can trust that it will be the actual size in CSS. The advantage of it, is that it does not query the DOM, but the last dimensions requested by the user. `getDrawingBufferSize` is explicit enough, it is the actual `canvas.width` and `canvas.height`, but without querying the DOM either, so it should support offscreen use cases.

nzjony: @astrak, currently the `MapView::resize()` method calls `setSize` with `updateStyle` to `false`, so any user set css should stay in place.

Astrak: Oh right... This maintains the `width: 100%; height: 100%;` in the stylesheet. But why do we do that instead of setting it in JS? It means that we ask the user to set, and maintain, the dimensions in the CSS. All the examples in ThreeJS use the JS instead. I think it's more flexible and more expected as well.

nzjony: Probably we should create a ticket to clean this up and specify a suggested best practice.

Astrak: I'd like to send a minor with several kind of ThreeJS alignments, working on it

zbigg:
>  But why do we do that instead of setting it in JS? 

I don't think that we should even try to touch CSS style of canvas. It's under control of mapview client and (s)he should be able work with any weird CSS width/height setting.
So in my opionion that `updateStyle=false` is good and should stay.

I'm not sure what is conclusion now though ...

PR #335, opened 2019-03-28
Reset TravisCI Deployment only for Releases.

amsgames94:


haraldF:
Your commit message is weird...

PR #341, opened 2019-04-01
MapView: fix first FrameComplete event.

zbigg:
TextElementsRenderer is fully initialized and can be used for
label layout only when all font catalogs are loaded.

Introduce `#ready` property and use it to prevent initial
label placement until in MapView until font catalog is loaded.

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -189,6 +189,7 @@ export class TextElementsRenderer {
 
     private m_tmpVector = new THREE.Vector2();
     private m_overloaded: boolean = false;
+    private m_catalogsLoading: number = 0;
 
     /**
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -189,6 +189,7 @@ export class TextElementsRenderer {
 
     private m_tmpVector = new THREE.Vector2();
     private m_overloaded: boolean = false;
+    private m_catalogsLoading: number = 0;
```
Astrak: Why is it assumed that loading a catalog will trigger render calls next frame? What if you get a 404? What if it takes 2sec to load the catalog?

Astrak: > Why is it assumed that loading a catalog will trigger render calls next frame

Sorry in that line I was mislead with the title of the PR, I understood that it referred to something else.

As it happens, by chance the other questions are still relevant I think!

PR #342, opened 2019-04-01
Performance improvements for generation of lines geometries.

keaukraine:
Removed unused `N` array.
According to benchmark (https://jsbench.me/igjttyjttb), preallocating array of known size is about 75% faster in Chrome, while performance improvement in FF is negligible.

Signed-off-by: Oleksandr Popov <ext-oleksandr.popov@here.com>

### DIFF
--- a/@here/harp-lines/lib/Lines.ts
+++ b/@here/harp-lines/lib/Lines.ts
@@ -34,21 +34,19 @@ export function createLineGeometry(
     const V = geometry.vertices;
     const I = geometry.indices;
 
-    const N: number[] = [];
-    const L: number[] = [];
-    const tangents: Array<[number, number]> = [];
-    const baseVertex = geometry.vertices.length / 12;
+    const L = new Array<number>(pointCount);
+    const tangents = new Array<[number, number]>(pointCount - 1);
+    const baseVertex = V.length / (highPrecision ? HP_LINE_STRIDE_SIZE : LINE_STRIDE_SIZE);
 
     let sum = 0;
-    L.push(0);
+    L[0] = 0;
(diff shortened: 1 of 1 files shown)


musculman:
ok to test

@here/harp-lines/lib/Lines.ts
```diff
@@ -34,21 +34,19 @@ export function createLineGeometry(
     const V = geometry.vertices;
     const I = geometry.indices;
 
-    const N: number[] = [];
-    const L: number[] = [];
-    const tangents: Array<[number, number]> = [];
-    const baseVertex = geometry.vertices.length / 12;
+    const L = new Array<number>(pointCount);
+    const tangents = new Array<[number, number]>(pointCount - 1);
+    const baseVertex = V.length / (highPrecision ? 14 : 12);
```
nzjony: Was this a bug? Why do high precision points start at 14?

nzjony: Was this a bug? Why do high precision points start at 14?

keaukraine: Stride size for high precision lines is 14, for regular ones is 12. They cannot be mixed within one geometry so this does look like a bug to me.

nzjony: Ok, that makes sense. 
Can you please either add a:
"TODO: take away magic numbers"
Or add some const numbers to the top of this file? I notice that the `static createGeometry(` function also relies on these numbers.

keaukraine: Added constants instead of these magic numbers.

PR #347, opened 2019-04-02
HARP-5248: Implement API for custom features.

Astrak:


@here/harp-features-datasource/lib/FeaturesDataSource.ts
```diff
@@ -0,0 +1,166 @@
...
+                ...feature.properties,
+                __mapViewUuid: feature.uuid
+            }
+        };
+        this.m_featureCollection.features.push(geojsonFeature);
+    }
+
+    private removeFeature(feature: MapViewFeature) {
+        // Remove geojson feature from the root FeatureCollection.
+        const features = this.m_featureCollection.features;
```
nzjony: why not use features.find((value,index, obj) => {
...?
});

Astrak: We need the index to then splice the array, so did as you suggested but with `findIndex`

@here/harp-geojson-datasource/lib/GeoJsonDataProvider.ts
```diff
@@ -24,8 +24,8 @@ export class GeoJsonDataProvider implements DataProvider {
      *
      * @returns New `GeoJsonDataProvider`.
      */
-    constructor(readonly name: string, readonly input: URL | GeoJson) {
-        this.m_tiler = ConcurrentTilerFacade.getTiler("omv-tiler");
+    constructor(readonly name: string, public input: URL | GeoJson, workerTilerUrl?: string) {
```
zbigg: > public input: URL | GeoJson 

I would say that input should be only parameter and not even kept in this context.

Let it be freely GC-ed when unneeded in this context.

@here/harp-mapview-decoder/lib/TilerService.ts
```diff
@@ -79,31 +81,45 @@ export class TilerService extends WorkerService {
...
-                json = await response.json();
+            await this.writeIndex(message);
+        }
+        return { response: {} };
+    }
+
+    private async handleUpdateIndexRequest(
+        message: WorkerTilerProtocol.UpdateIndexRequest
+    ): Promise<WorkerServiceResponse> {
+        await this.writeIndex(message);
```
zbigg: You can basically `return this.writeIndex(...)`.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1335,9 +1335,7 @@ export class MapView extends THREE.EventDispatcher {
     addDataSource(dataSource: DataSource): Promise<void> {
         const twinDataSource = this.getDataSourceByName(dataSource.name);
         if (twinDataSource !== undefined) {
-            throw new Error(
-                `A DataSource with the name "${dataSource.name}" already exists in this MapView.`
-            );
+            return new Promise(resolve => resolve());
```
zbigg: Isn't change discussed in #424 ? I would remove it from this patch because this as already discussed this approach is broken.

@here/harp-features-datasource/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+    "compilerOptions": {
+        "target": "es5",
```
zbigg: All other packages are ` "target": "es6",`

@here/harp-features-datasource/lib/Features.ts
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * The type of the feature. The extended class should initialize this value. It defaults to
+     * "Point" in order to avoid allowing `null` or `undefined`.
+     */
+    type: FeatureGeometry["type"] = "Point";
+
+    /**
+     * A string identifying this feature.
+     */
+    uuid: string = THREE.Math.generateUUID();
```
zbigg: Is that `uuid` really needed?
I see it's only used for searching in current list of features.

Wouldb't matching be simpler when just comparing identity of objects?

@here/harp-geojson-datasource/lib/GeoJsonDataProvider.ts
Astrak: Names are used in the Facade-s and inputs are stored to be used later in `connect` and `updateIndex`?

zbigg: Actually i care only about `input`.

Input should be forgotten as soon as it's sent to tiler. There is no reason to keep reference.

So ...

> this.input = input;

Should be removed from `updateIndex`
And `connect` should nullify `this.input` immediately after it has been sent.

GEOJSON will be usually "just loaded" throw-away data that is handed-over to other context, so it's not used anymore.

Recently allocated objects are usually in so called "young" generation pool which is best candidate for GC, so we shouldn't deny GC possibility to recycle it.

@here/harp-features-datasource/lib/FeaturesDataSource.ts
```diff
@@ -0,0 +1,204 @@
...
+        });
+    }
+
+    /**
+     * Creates a polygon feature.
+     *
+     * @param coordinates The geometry of the feature.
+     * @param style The style to draw the feature with.
+     */
+    createPolygon(coordinates: Polygon["coordinates"], properties?: {}): MapViewPolygonFeature {
```
nzjony: these create* methods don't need to be part of this DataSource do they? I think it would be clearer to have them as free functions separate from the data source

@here/harp-examples/src/features_custom.ts
```diff
@@ -0,0 +1,440 @@
...
+ *
+ * We then add the feature manually:
+ * ```typescript
+ * [[include:harp_demo_features_example_3.ts]]
+ * ```
+ *
+ * The rest of the code is example-dependent logic to handle the countries sets, with the joining
+ * and leaving mechanism, that adds and removes datasources.
+ */
+export namespace DemoFeaturesExample {
```
nzjony: this is a great example, but it is quite complex, I think it would be good to have a very simple demo which just adds a single geo json object, for those who may not be so advanced.

@here/harp-features-datasource/test/FeaturesTest.ts
```diff
@@ -0,0 +1,47 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* import { assert } from "chai";
+import { FeaturesDataSource } from "../lib/FeaturesDataSource"; */
+
+// tslint:disable:only-arrow-functions
+//    Mocha discourages using arrow functions, see https://mochajs.org/#arrow-functions
+
+describe("features-datasource", function() {
```
nzjony: please remove this file, the example you have shows the interface

Astrak:
@heremaps-bot Retest this please

PR #350, opened 2019-04-02
HARP-1355: Reusable SolidLine geometries.

amsgames94:


amsgames94:
GPU tests please

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -732,6 +732,19 @@ export interface SolidLineTechnique extends BaseTechnique, PolygonalTechnique {
...
     clipping?: boolean;
+    /**
+     * Color of a line's outline in hexadecimal or CSS-style notation, for example: `"#e4e9ec"`,
+     * `"#fff"`, `"rgb(255, 0, 0)"`, or `"hsl(35, 11%, 88%)"`.
+     */
+    outlineColor?: string;
+    /**
+     * Width of a line's outline in `metricUnit`s for different zoom levels.
+     */
+    outlineWidth?: number;
```
zbigg: From this doc alone, it may me not clear if `outlineWidth` is 'width' of geometry that draws outline or is it `width` added around main part of line.

So, strictly speaking if line with with `B` has outline with width `B`, then total visible width should be `A+B+B` (i.e main line +2 outlines from both sides).
This way of defining property reveals impl detail that outline is independent line with independent width.

So, either clarify doc, or change PR, so `outlineWidth` a _pure outline_ width.

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -732,6 +732,19 @@ export interface SolidLineTechnique extends BaseTechnique, PolygonalTechnique {
...
+    /**
+     * The render order of the outline object created using this technique.
+     */
+    outlineRenderOrder?: number;
```
zbigg: > Why is this needed at all ?

Answered by myself, leaving it for others who may ask this question too ...

Actually, for roads all outlines should be in one range and all foregrounds have to be in distinct ranges we have to do it manually.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1508,6 +1513,66 @@ export class Tile implements CachedResource {
...
+                    outlineMaterial.uniforms.diffuse.value = outlineColor;
+                    const outlineObj = new ObjectCtor(bufferGeometry, outlineMaterial);
+
+                    if (outlineTechnique.outlineRenderOrder !== undefined) {
+                        outlineObj.renderOrder = outlineTechnique.outlineRenderOrder;
+                    } else {
+                        if (technique._renderOrderAuto === undefined) {
+                            throw new Error("Technique has no renderOrderAuto");
+                        }
+                        outlineObj.renderOrder = technique._renderOrderAuto - 0.1;
```
zbigg: Yeah, it's questionable at at least in themes out there in wild, we are already defined values at magnitude of 0.1.
Make it smaller, like 0.00001 or something that is literally impractical for user.

I know it's controversial, but as for now only solution. In _future_, we should define uprecision of user supplied renderOrder so we can do hacks like this.

@here/harp-examples/resources/dayTwoFonts.json
```diff
@@ -986,29 +987,21 @@
...
-   },
-   {
-    "description": "borders foreground",
-    "when": "$layer == 'admin' && ($layer == 'admin')",
-    "technique": "solid-line",
-    "attr": {
-     "color": "#5C5C5C",
-     "lineWidth": {
+     },
+     "outlineWidth": {
```
zbigg: It's just discussion, by no means try to blocki patch.

I generally like this patch, but  for future designers of themes/styles/techniques i would like to define `outlineWidth` for as percentage, so in general case I don't have to interpolate at all.

@here/harp-mapview/lib/Tile.ts
amsgames94: Will change this.

@here/harp-examples/resources/dayTwoFonts.json
amsgames94: Agree that percentage is more intuitive, but don't want to change the units the designers work with with not a strong enough reason.

@here/harp-datasource-protocol/lib/Techniques.ts
amsgames94: I agree that specifying a secondary renderOrder is really weird, but it's the only current way to support this with the current API status (things might be different when a proper renderOrder management system is in place).

@here/harp-datasource-protocol/lib/Techniques.ts
amsgames94: Will change the prefix of these properties to ``secondary``, since actually ``outline`` doesn't really define the behaviour of the second geometry as you pointed out.

PR #351, opened 2019-04-02
HARP-5263: Fix high GPU usage with WebTileDataSource.

zbigg:
* ensure that webtiles texture is `disposed` when we're disposing Tile
* invalidate memory resource usage after texture is loaded, so tile cache can has up-to-date view of tiles size

### DIFF
--- a/@here/harp-webtile-datasource/lib/WebTileDataSource.ts
+++ b/@here/harp-webtile-datasource/lib/WebTileDataSource.ts
@@ -225,6 +225,16 @@ const hereCopyrightInfo: CopyrightInfo = {
     label: "HERE"
 };
 
+class WebTile extends Tile {
+    material?: THREE.MeshBasicMaterial;
+    clear() {
+        if (this.material !== undefined && this.material.map) {
+            this.material.map.dispose();
+        }
+        super.clear();
+    }
+}
+
 /**
  * Instances of `WebTileDataSource` can be used to add Web Tile to [[MapView]].
  *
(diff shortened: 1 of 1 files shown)


@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -225,6 +225,16 @@ const hereCopyrightInfo: CopyrightInfo = {
     label: "HERE"
 };
 
+class WebTile extends Tile {
+    material?: THREE.MeshBasicMaterial;
+    clear() {
+        if (this.material !== undefined && this.material.map) {
```
ninok: Wouldn't it makes sense to move this to Tile.clear (i.e. when iterating through the object materials)
I just realized that my TerrainDataSource has the same problem so it would be good if disposing of textures would be done in a generic fashion. It could be problematic though if textures are shared across tiles, but I don't know if there is a usecase for that.

zbigg: > It could be problematic though if textures are shared across tiles,

That's the exact reason why i didn't moved it into shared code.
In WebTiles we understand lifecycle of texture and know when to delete, it's not the case in abstract Tile ...

So i would leave it as it is ?

amsgames94: The material needs to also be disposed to free the WebGL program resources. The best way to handle disposal in general, is to give ownership over the resources (check ``TextCanvas.material`` setter) or to have them referenced counted in a centralized place and free them whenever they're no longer used.

This is a bigger task anyways, but we should tackle this at some point to handle our resources allocations/dellocations properly.

zbigg: > The material needs to also be disposed to free the WebGL program resources

Material is already disposed in `Tile.clear`:
https://github.com/heremaps/harp.gl/blob/master/%40here/harp-mapview/lib/Tile.ts#L845

PR #352, opened 2019-04-03
HARP-5295: Support map, normalMap and displacementMap in StandardTexturedTechnique

ninok:


@here/harp-mapview/lib/Tile.ts
```diff
@@ -1231,7 +1231,8 @@ export class Tile implements CachedResource {
                         {
                             technique,
                             level: Math.floor(displayZoomLevel),
-                            fog: this.mapView.scene.fog !== null
+                            fog: this.mapView.scene.fog !== null,
+                            skipExtraProps: ["map", "normalMap", "displacementMap"]
```
zbigg: Shouldn't `(map|normalMap|...)Properties` be exluded too ?

ninok: True, good catch

ninok:

PR #354, opened 2019-04-03
HARP-5128: Fix zoom for small steps.

amsgames94:
Modify the ``easeOutCubic`` function used in ``MapControls`` to snap to desired values when different between desired and current value is smaller than ``EPSILON``.

### DIFF
--- a/@here/harp-map-controls/lib/MapControls.ts
+++ b/@here/harp-map-controls/lib/MapControls.ts
@@ -85,9 +85,9 @@ const USER_INPUTS_TO_CONSIDER = 5;
 const DEFAULT_MAX_PITCH_ANGLE = Math.PI / 4;
 
 /**
- * Epsilon value to rule out when a number can be considered 0 (used for pitch angle in radian).
+ * Epsilon value to rule out when a number can be considered 0.
  */
-const EPSILON_PITCH = 0.01;
+const EPSILON = 0.01;
 
 /**
  * This map control provides basic map-related building blocks to interact with the map. It also
@@ -488,7 +488,7 @@ export class MapControls extends THREE.EventDispatcher {
      */
     toggleTilt(): void {
         this.m_startPitch = this.currentPitch;
-        const aimTilt = this.m_startPitch < EPSILON_PITCH || this.m_tiltState === TiltState.Down;
+        const aimTilt = this.m_startPitch < EPSILON || this.m_tiltState === TiltState.Down;
         this.m_pitchRequested = aimTilt ? this.tiltAngle : 0;
         this.m_tiltState = aimTilt ? TiltState.Tilted : TiltState.Down;
         this.m_tiltAnimationStartTime = performance.now();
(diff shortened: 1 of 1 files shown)


@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -85,9 +85,9 @@ const USER_INPUTS_TO_CONSIDER = 5;
 const DEFAULT_MAX_PITCH_ANGLE = Math.PI / 4;
 
 /**
- * Epsilon value to rule out when a number can be considered 0 (used for pitch angle in radian).
+ * Epsilon value to rule out when a number can be considered 0.
  */
-const EPSILON_PITCH = 0.01;
+const EPSILON = 0.01;
```
Astrak: This is an epsilon in radian used to define when toggling the 2D/3D view should be tilted (if the camera is already almost vertical, the user will not expect the animation to go for the last missing epsilon to vertical, but rather to the fully tilted view, it is used for this specific problem). The ease method below interpolates unitless values, all between 0 and 1. Likely the typedoc is missing these information in the first place. This epsilon should be kept for the pitch, and another one created for the other purpose.

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -645,7 +645,9 @@ export class MapControls extends THREE.EventDispatcher {
     }
 
     private easeOutCubic(startValue: number, endValue: number, time: number): number {
-        return startValue + (endValue - startValue) * (--time * time * time + 1);
+        return Math.abs(endValue - startValue) < EPSILON
```
Astrak: I think the check should instead occur in the zoom implementation, and on the absolute values and not the relative ones used here. The reason is that `easeOutCubic` is used for both zooming and panning, and that `endValue` and `startValue` are always between 0 and 1.

@here/harp-map-controls/lib/MapControls.ts
amsgames94: I get your reasoning, but its a private constant and in both cases we can use the same value. If we want to define different values in the future, I'd split, but I don't see why it's better to use 2 specific epsilons with the same value rather than one generic one.

@here/harp-map-controls/lib/MapControls.ts
amsgames94: I guess it's a matter of taste, but since this problem can also appear with every action that uses this function (if you limit the positions you can pan in the X axis, for example), I believe that having a ``snap-to-value`` behaviour is quite acceptable.

@here/harp-map-controls/lib/MapControls.ts
ninok: I agree with Andres, if we need to different values we can split later.

@here/harp-map-controls/lib/MapControls.ts
ninok: I as well think this logic should not be in easeOutCubic, unless you rename that function.
Btw., can we move this function to some utils b/c I saw that Sergii was re-implementing this function already for the building animation.

amsgames94: Ok, I'll keep the change outside of the function. And maybe we can move it outside to Utils, but a bit outside of the scope of this bugfix ticket.

ninok: okay

PR #363, opened 2019-04-04
HARP-5324: Properly dispose textures that are created/owned by Tile

ninok:
Also fixed missing textures in memory estimation for MeshStandardMaterial

### DIFF
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -71,7 +71,7 @@ export interface MaterialOptions {
  */
 export function createMaterial(
     options: MaterialOptions,
-    materialUpdateCallback?: () => void
+    textureReadyCallback?: (texture: THREE.Texture) => void
 ): THREE.Material | undefined {
     const technique = options.technique;
     const Constructor = getMaterialConstructor(technique);
@@ -126,8 +126,8 @@ export function createMaterial(
                 (material as any)[texturePropertyName] = texture;
                 texture.needsUpdate = true;
                 material.needsUpdate = true;
-                if (materialUpdateCallback) {
-                    materialUpdateCallback();
+                if (textureReadyCallback) {
+                    textureReadyCallback(texture);
                 }
             };
 
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -7,7 +7,6 @@ import {
     BaseTechnique,
     BufferAttribute,
     DecodedTile,
-    ExtendedTileInfo,
     ExtrudedPolygonTechnique,
     FillTechnique,
     Geometry,
@@ -19,7 +18,6 @@ import {
     isExtrudedLineTechnique,
     isExtrudedPolygonTechnique,
     isFillTechnique,
-    isInterpolatedProperty,
     isSolidLineTechnique,
     isSquaresTechnique,
     isStandardTechnique,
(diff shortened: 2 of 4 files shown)


@here/harp-mapview/lib/Tile.ts
```diff
@@ -447,6 +445,9 @@ export class Tile implements CachedResource {
 
     private m_resourceInfo: TileResourceInfo | undefined;
 
+    // List of owned textures for disposal
+    private m_ownedTextures: Set<string> = new Set();
```
zbigg: This can be `Set<THREE.Texture>`.
IMHO, it's more straighforard and probably a little bit  efficient as we search will compare by reference.

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -71,7 +71,7 @@ export interface MaterialOptions {
  */
 export function createMaterial(
     options: MaterialOptions,
-    materialUpdateCallback?: () => void
+    materialUpdateCallback?: (texture: THREE.Texture) => void
```
zbigg: Not sure why this is needed.
BTW. both landmark and should call this `materialUpdateCallback` right after their texture is ready, but ... they shoula also call `<TIle>this.addOwnedTexture(t)`, so i don't see reason to leak this texture in this particular place.

Btw, `WebTileDataSource` should be updated to remove "my" WebTile and just call `addOwnedTexture`.

@here/harp-mapview/lib/Tile.ts
ninok: I wanted to avoid holding a reference to the texture, but since we own the texture it should be fine to use the texture directly.

@here/harp-mapview/lib/DecodedTileHelpers.ts
zbigg: > Not sure why this is needed.

Ok, i get it, it's needed because `createMaterial` creates textures.
However, this `materialUpdateCallback` is more like `textureCreatedCallback` ...

ninok: The problem is, that createMaterial does not know anything about Tile, so I can only add the ownership of the texture to the tile in the callback and I have to do the same for the LandmarkTile (and WebTile).
The only way I see to get around this would be to make createMaterial a method of Tile but that would mean moving almost the whole DecodedTileHelpers code into Tile.

zbigg: first ;) 
Thanks i've already found explanation in companion patch.

However, `WebTileDataSource` should be updated to use `addOwnedTexture`.

@here/harp-mapview/lib/Tile.ts
zbigg: >  wanted to avoid holding a reference to the texture,

For this case you have WeakMap/Set [1] in your toolbox. But i agree, in this case there is no special reason to use WeakSet, normal set is enough.

[1]https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet

ninok: Ah, good to know. I think I will use WeakSet b/c it makes more clear that textures are not held in this set but in the material.
Maybe for some weird reason somebody disposes the material and the texture from outside it would be even safer.

zbigg: Good point ... 
+2

PR #364, opened 2019-04-05
HARP-5346: Support providing direct GeoJson in TilerService.

Astrak:


@here/harp-mapview-decoder/lib/TilerService.ts
```diff
@@ -81,14 +81,17 @@ export class TilerService extends WorkerService {
...
-                logger.error(`${message.url} Status Text:  ${response.statusText}`);
-                return { response: {} };
+            let rawJSON = message.input;
+            if (typeof message.input === "string") {
+                const response = await fetch(message.input);
+                if (!response.ok) {
+                    logger.error(`${message.input} Status Text:  ${response.statusText}`);
+                    return { response: {} };
+                }
+                const answer = await response.text();
```
zbigg: I would use `response.json()`.

This enables browser to parse json immediately while receiving without need to buffer whole string.
May be more efficient on platforms that want to leverage this opportunity.

@here/harp-geoutils/lib/coordinates/GeoCoordinates.ts
```diff
@@ -154,4 +154,30 @@ export class GeoCoordinates implements GeoCoordinatesLike {
...
+     * Update this `GeoCoordinates` by assigning it the southwesternmost position between it and
+     * another `GeoCoordinates`.
+     *
+     * @param other The other `GeoCoordinates` to compare to this one.
+     */
+    min(other: GeoCoordinates) {
+        this.latitude = Math.min(this.latitude, other.latitude);
+        this.longitude = Math.min(this.longitude, other.longitude);
+        this.altitude =
+            this.altitude && other.altitude ? Math.min(this.altitude, other.altitude) : undefined;
```
ninok: Might be a bit counter-intuitive if this has an altitude and other has none that the result will also have none.
In general this method does not make much sense as a member of GeoCoordinates but only in combination with GeoBox, so consider to move the code there.

@here/harp-geoutils/lib/coordinates/GeoBox.ts
```diff
@@ -198,6 +198,16 @@ export class GeoBox {
         return new GeoBox(this.southWest, this.northEast);
     }
 
+    /**
+     * Update the bounding box by considering a given point.
+     *
+     * @param point The point that may expand the bounding box.
+     */
+    expandByPoint(point: GeoCoordinates) {
```
ninok: in harp-native this method is called growToContain. Please rename to be consistent

@here/harp-geoutils/lib/coordinates/GeoCoordinates.ts
```diff
@@ -154,4 +154,30 @@ export class GeoCoordinates implements GeoCoordinatesLike {
...
+     * Update this `GeoCoordinates` by assigning it the northeasternmost position between it and
+     * another `GeoCoordinates`.
+     *
+     * @param other The other `GeoCoordinates` to compare to this one.
+     */
+    max(other: GeoCoordinates) {
+        this.latitude = Math.max(this.latitude, other.latitude);
+        this.longitude = Math.max(this.longitude, other.longitude);
+        this.altitude =
+            this.altitude && other.altitude ? Math.max(this.altitude, other.altitude) : undefined;
```
zbigg: > this.altitude && other.altitude

This is false-y. `0 && 5` will result in undefined.

@here/harp-geoutils/lib/coordinates/GeoBox.ts
nzjony: I would say this should be renamed in a separate commit

@here/harp-datasource-protocol/lib/WorkerTilerProtocol.ts
```diff
@@ -24,7 +25,7 @@ export namespace WorkerTilerProtocol {
     export interface RegisterIndexRequest extends WorkerServiceProtocol.ServiceRequest {
         type: Requests.RegisterIndex;
         id: string;
-        url: string;
+        input: string | GeoJson;
```
nzjony: should this be URL | GeoJson? That at least seems to be used elsewhere.

@here/harp-geoutils/lib/coordinates/GeoCoordinates.ts
nzjony: Maybe it could be a util method which returns the combined min of two GeoCoordinates

@here/harp-geoutils/lib/coordinates/GeoBox.ts
nzjony: expandByPoint is used in THREE.js, so actually it wasn't chosen randomly.

Considering that we don't share code with HARP, I think the original name is better, but maybe that is just a personal opinion.

@here/harp-datasource-protocol/lib/WorkerTilerProtocol.ts
Astrak: From the worker, the only need is to make the call to the URL, and we can pass the ready url to it as a string instead of the URL class. What would be the benefit of doing it the other way around?

nzjony: Ok, I would do it just for readability, but anyway minor.

zbigg: > would do it just for readability

That wouldn't help either, because this is definition of "message on wire" (where "wire data" is a limited set of Javascript as defined here: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).

So, here we can have string or any other representation of URL, like complety decomposed IURLinterface like `interface IURL {
    hash?: string; host?: string; hostname?: string; href?: string ... }` _without_ any methods, expectations about being a objects of actual `URL` class etc.

So, `string` as a common canonical representation of URL/I is good enough and best in this context.

@here/harp-geoutils/lib/coordinates/GeoBox.ts
```diff
@@ -198,6 +198,35 @@ export class GeoBox {
         return new GeoBox(this.southWest, this.northEast);
     }
 
+    /**
+     * Update the bounding box by considering a given point.
+     *
+     * @param point The point that may expand the bounding box.
+     */
+    growToContain(point: GeoCoordinates) {
```
nzjony: I find it weird that we have growToContain in GeoBox, but Box2 in THREE has expandByPoint, I think this will just cause confusion for our users, @ninok do we need to have compatibility with HARP Native?

ninok: harp-geoutils are not only used in harp.gl, but also e.g. in Hype.
Hype doesn't know anything about three.js and it would be weird for Hype developers to see differences in the web and native foundation libraries that are motivated by three.js.

nzjony: True, but I would argue that the github community using harp.gl is (will be) bigger than the number of hype developers, so we are making it easier to understand for a few but missing the audience.

nzjony: But anyway, @Astrak, merge it anyway, and the community can later decide to rename if necessary, I don't want to block.

PR #366, opened 2019-04-08
HARP-5323: Added ElevationRangeSource and use it in VisibleTileSet to compute proper bounding boxes

ninok:


@here/harp-geoutils/lib/projection/EarthConstants.ts
```diff
@@ -10,4 +10,16 @@ export class EarthConstants {
...
+     * Lowest point on dry land: Shore of the Dead Sea.
+     * https://en.wikipedia.org/wiki/Extreme_points_of_Earth
+     */
+    static MIN_ELEVATION: number = -432.65;
+
+    /**
+     * Mount Everest.
+     * https://en.wikipedia.org/wiki/Extreme_points_of_Earth
+     */
+    static MAX_ELEVATION: number = 8.848;
```
nzjony: I think you don't mean to have the . there ?

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -195,6 +196,18 @@ export abstract class DataSource extends THREE.EventDispatcher {
      */
     abstract getTilingScheme(): TilingScheme;
 
+    /**
+     * Returns a [[ElevationRangeSource]] so that `MapView` can compute proper bounding boxes that
```
nzjony: [[MapView]]

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -195,6 +196,18 @@ export abstract class DataSource extends THREE.EventDispatcher {
...
+     * take the elevation range of the data into account.
+     * There can be only one [[ElevationRangeSource]] per `MapView`. `MapView` will take the
```
nzjony: replace ` with [[*]]

@here/harp-mapview/lib/DataSource.ts
nzjony: "MapView can compute" is incorrect, this should be VisibleTileSet?

Perhaps mentioning where it is computed is not needed.

@here/harp-mapview/lib/ElevationRangeSource.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface ElevationRange {
+    minElevation: number;
+    maxElevation: number;
+}
+
+export enum CalculationStatus {
+    // Calculated approximately. A more precise result may be available later.
+    PendingApproximate = 1,
+    // Calculation completed. The result is final, won't improve upon retrying.
+    FinalPrecise = 3
```
nzjony: is 3 special for some reason?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -269,15 +270,29 @@ export class VisibleTileSet {
...
+        let allBoundingBoxesFinal: boolean = true;
 
         if (this.options.extendedFrustumCulling) {
             this.m_mapTileCuller.setup();
         }
 
+        let elevationRangeSource: ElevationRangeSource | undefined;
+        for (const dataSource of dataSources) {
+            elevationRangeSource = dataSource.getElevationRangeSource();
+            if (elevationRangeSource !== undefined) {
```
nzjony: I notice you check for the tiling scheme below, does it make more sense to check here? Else you could get an ElevationRangeSource which won't match.

@here/harp-geoutils/lib/projection/EarthConstants.ts
ninok: good catch

@here/harp-mapview/lib/DataSource.ts
ninok: VisibleTileSet is an internal thing but MapView and DataSource are public API that's why I put it like this.

@here/harp-mapview/lib/ElevationRangeSource.ts
ninok: No, I had more states before and removed them. Will fix.

@here/harp-mapview/lib/VisibleTileSet.ts
ninok: Assuming we have multiple datasources some with webMercator tiling scheme and some with equirectangular tiling scheme. The current behaviour will be that all webMercator data sources will use proper bounding boxes and all equirectangular data sources will have flat bounding boxes.
You mean to change the behavior so that also the webMercator data sources get flat bounding boxes if there is any equirectangular(or whatever) data source

@here/harp-mapview/lib/DataSource.ts
nzjony: Good point.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -269,15 +270,29 @@ export class VisibleTileSet {
...
+            }
+        }
+
         for (const dataSource of dataSources) {
             const displayZoomLevel = dataSource.getDisplayZoomLevel(zoomLevel);
 
             const tilingScheme = dataSource.getTilingScheme();
+            const useElevationRangeSource: boolean =
+                elevationRangeSource !== undefined &&
+                elevationRangeSource.getTilingScheme() === tilingScheme;
```
zbigg: > `elevationRangeSource.getTilingScheme() === tilingScheme`

I tdon't think check belongs here. I would change contract of `DataSource.getElevationRangeSource` that it returns only _compatible_ elevation source.
If there is no compatible elevation source, it should just return undefined.

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -195,6 +196,18 @@ export abstract class DataSource extends THREE.EventDispatcher {
      */
     abstract getTilingScheme(): TilingScheme;
 
+    /**
+     * Returns a [[ElevationRangeSource]] so that [[MapView]] can compute proper bounding boxes that
+     * take the elevation range of the data into account.
+     * There can be only one [[ElevationRangeSource]] per [[MapView]]. [[MapView]] will take the
```
zbigg: > There can be only one [[ElevationRangeSource]] per [[MapView]]. (...)

Why? That means that we shouldn't even add `ElevationRangeSource` to DataSource interface, but add it to `MapView` ... (at least at this stage).

On the other hand, at this stage, MapView doesn't even use `ElevationRangeSource` and i see no reason why two separate DS would have two different elevation datasources working properly with two separate `ElevationRangeSource`s supporting different tiling schemes.

@here/harp-geoutils/lib/projection/EarthConstants.ts
```diff
@@ -10,4 +10,16 @@ export class EarthConstants {
 
     /** The equatorial radius in meters. */
     static EQUATORIAL_RADIUS: number = 6378137.0;
+
+    /**
+     * Lowest point on dry land: Shore of the Dead Sea.
```
zbigg: While these constants are interesting i'm wondering why we even define them here ?

While max is rather stable, `MIN_ELEVATION` doesn't matter if we choose to visualize depth of oceans ...

@here/harp-mapview/lib/ElevationRangeSource.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export enum CalculationStatus {
+    // Calculated approximately. A more precise result may be available later.
+    PendingApproximate,
+    // Calculation completed. The result is final, won't improve upon retrying.
+    FinalPrecise
+}
+
+export interface ElevationRangeSource {
+    getElevationRange(tileKey: TileKey, range: ElevationRange): CalculationStatus;
```
zbigg: Hmm,
1) it's not documented that `range` is filled with result
2) IMHO, result type should be
```
ElevationRangeResult {
    minElevation: number;
    maxElevation: number;
    calculationStatus?: CalculationStatus;
}
```
Rationale: I just don't think it's good idea to return function results through two channels like here - output variable and return value.
It should be either return value or `out` variable if performance matters.

@here/harp-mapview/lib/ElevationRangeSource.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export enum CalculationStatus {
+    // Calculated approximately. A more precise result may be available later.
+    PendingApproximate,
+    // Calculation completed. The result is final, won't improve upon retrying.
+    FinalPrecise
+}
+
+export interface ElevationRangeSource {
```
zbigg: This will be quite important interface in future, please typedoc at least basic assumptions (whole file).

@here/harp-geoutils/lib/projection/EarthConstants.ts
ninok: If we would render ocean elevation, I would assume we would just render shading but not real geometry(except if somebody implements a submarine simulator).
Nevertheless I agree that this is more a data-source specific thing. I will move it.

zbigg: Maybe use them in implementations of `Projections.worldExtent` ?

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: What does a compatible elevation source mean? There is probably only one elevation data source (unless you want to somehow overlay different terrain sets)

zbigg: `getElevationRangeSource` has builtin tiling scheme, so in this case _compatibility_ is defined as "same tiling scheme".

What i mean, that if `DS.getElevationRangeSource` returns something, then it should be usable.
You shouldn't even have incompatible elevation source attached to `DS`.

>  There is probably only one elevation data 

Why?
I understand that there is probably one data set and service that serves elevation data, but each datasource may have different one javascript object instance serving elevation data for different tiling schemes.

> There is probably only one elevation data source

Note, that here we have elevation *range* source, which is only useful for tiling and is heavily bound to tiling scheme.
For general elevation data (i assume) we will need more fine grained interface like `getElevation(lat,long)` - imagine picking or laying out roads on terrain.

...

So, this discussion revolves about question "is elevation range source" property of DS or MapView, so it goes back to this comment: https://github.com/heremaps/harp.gl/pull/366#discussion_r274878015

If it's DS property, then yes, DS shouldn't return incompatible one but on the other hand, we should support many of them

If it's MapView property, then move it to MapView and check tilng scheme here.

nzjony: > > There is probably only one elevation data
> 
> Why?

What use case do you have for two differently enabled tiled terrain sources? What do you want to show the user?

> I understand that there is probably one data set and service that serves elevation data, but each datasource may have different one javascript object instance serving elevation data for different tiling schemes.

The `DataSource` may not know about elevation data, but we know that the terrain data source knows about elevation data, it just extracts the data from the tile and makes it usable for other data sources.

> 
> > There is probably only one elevation data source
> 
> Note, that here we have elevation _range_ source, which is only useful for tiling and is heavily bound to tiling scheme.
> For general elevation data (i assume) we will need more fine grained interface like `getElevation(lat,long)` - imagine picking or laying out roads on terrain.

I think this is a separate topic.

> 
> ...
> 
> So, this discussion revolves about question "is elevation range source" property of DS or MapView, so it goes back to this comment: [#366 (comment)](https://github.com/heremaps/harp.gl/pull/366#discussion_r274878015)
> 
> If it's DS property, then yes, DS shouldn't return incompatible one but on the other hand, we should support many of them

Do you want to force each DS to return a valid `ElevationRangeSource`?

> 
> If it's MapView property, then move it to MapView and check tilng scheme here.

How does the MapView get this data?

ninok: @zbigg Maybe there is a misunderstanding in the behaviour: The ElevationRangeSource is coming from the terrain and is applied to all data sources that have the same tiling scheme(e.g. OMV but also terrain itself) b/c the will be overlayed on top of terrain. The other data sources don't have an elevation range source. The case that another data source (e.g. OMV) has 3d data and an elevation source is currently not supported.
The reason why I moved the ElevationRangeSource to DataSource and not MapView is that data sources can be disabled. In my first implementation the terrain data source was setting the elevation range source to the MapView, but this is wrong if the terrain data source is disabled.

As mentioned by @nzjony, getting elevation values for arbitrary geo coordinates and ray casting are not part of that change. This will be handled by a different class (e.g. HeightProvider) that will basically use the TerrainTiles that are in the TileCache.

@here/harp-geoutils/lib/projection/EarthConstants.ts
ninok: As far as I can see worldExtent is only used for the projection in the tiling scheme, not for the target projection used by MapView(except for the FlatBoundingBoxGenerator that should stay flat).

@here/harp-mapview/lib/VisibleTileSet.ts
zbigg: Thanks for explanation and sorry for questioning change, it's just adding  interfaces without clear doc & example implementation is confusing and begs for questions.

More implementation and/or use-case for this interfaces would help use understand why it is really needed, but to the point.

After rethinking this again, what i see it that for this particular change - namely, "support culling of tiles with some min/max altitut"e it's enough to add:
```
DataSource::getTileElevationRange(tileKey): ElevationRangeResult | undefined;
```
With basic premise that datasources who does not support elevation data or are not attached to any will just return `undefined`. and be done with it.

Less code, less burden.
(This of course is true, if we don't have any other use case for `ElevationRangeSource` which i'm not aware of yet).

ninok: How should that work for the data sources that don't have an elevation range for a given key (e.g. OMV)? Should I for these data sources loop through all other data sources and call getTileElevationRange on them?

This all depends on a bigger discussion that did not happen yet, i.e. who is responsible for overlaying vector data:
1. MapView
2. The vector data source itself

I was already tending to go into the direction of 1. but maybe we should have this discussion first (i.e. finish the terrain vector overlay solutioning)

@here/harp-mapview/index.ts
```diff
@@ -10,6 +10,7 @@ export * from "./lib/MapViewPoints";
 export * from "./lib/composing";
 export * from "./lib/DataSource";
 export * from "./lib/DecodedTileHelpers";
+export * from "./lib/ElevationRangeSource";
```
zbigg: How this even passed CI ?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -13,6 +13,7 @@ import { IMapAntialiasSettings, IMapRenderingManager, MapRenderingManager } from
 import { ConcurrentDecoderFacade } from "./ConcurrentDecoderFacade";
 import { CopyrightInfo } from "./CopyrightInfo";
 import { DataSource } from "./DataSource";
+import { ElevationRangeSource } from "./ElevationRangeSource";
```
zbigg: How this even passed CI ?
It should be rejected as not-used symbol.

ninok: Hm, strange. It was also not cached when I ran tslint locally. I will do a follow-up commit to fix it.

zbigg: I'm already looking at CI scripts to check what have happened.

PR #367, opened 2019-04-08
HARP-5286: Fix MapView.pixelToWorld ratio calculation.

kkostecki:
When using pixel units for styling via themes definition, such as attribute
metric is defined as:

     "metricUnit": "Pixel",

we need to calculate ratio between screen space pixel size and world space
distance. This is done via MapView.pixelToWorld() method call. This ratio
should be performed in exact center of projection (canvas center) and
independently from the projection type and camera tilt.

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -1452,10 +1452,14 @@ export class MapView extends THREE.EventDispatcher {
      */
     get pixelToWorld(): number {
         if (this.m_pixelToWorld === undefined) {
-            const { width, height } = this.getCanvasClientSize();
-            const center = this.getWorldPositionAt(width * 0.5, height);
-            const unitOffset = this.getWorldPositionAt(width * 0.5 + 1.0, height);
-
+            const size = this.getCanvasClientSize();
+            const { centerX, centerY } = { centerX: size.width * 0.5, centerY: size.height * 0.5 };
+            // Calculate canvas middle point (center of projection) world position.
+            const center = this.getWorldPositionAt(centerX, centerY);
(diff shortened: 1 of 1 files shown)


zbigg:
@heremaps-bot ok to test

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1452,10 +1452,14 @@ export class MapView extends THREE.EventDispatcher {
...
-
+            const size = this.getCanvasClientSize();
+            const { centerX, centerY } = { centerX: size.width * 0.5, centerY: size.height * 0.5 };
+            // Calculate canvas middle point (center of projection) world position.
+            const center = this.getWorldPositionAt(centerX, centerY);
+            // Calculate world point that is offset by one pixel in x-axis (or camera right vector)
+            // direction, this ensures that we are perspective/tilt indepenent.
+            const unitOffset = this.getWorldPositionAt(centerX + 1.0, centerY);
+            // Our ratio equals the distance between points in the world space.
             this.m_pixelToWorld =
```
ninok: Couldn't we derive the ratio from the zoomlevel instead of doing raycasting at the screen center?
It might happen that we don't hit the earth surface especially if you tilt using middle mouse button and also later when we use the SphereProjection.

kkostecki: This approach is tilt independent, as also will work with spherical projection. Of course we have the problem if tilt constraints would be removed (possibility to hit outer space instead of terrain), for that case we have discussed to use far plane as the ray cast limit, but there is another problem is if roll would be possible and X screen space offset would point the ray far on the (rotated) horizon.

kkostecki:
Old approach it was replaced with the new one:
https://github.com/heremaps/harp.gl/pull/378

PR #369, opened 2019-04-08
HARP-5184: Preserve features' Z-Component.

amsgames94:


@here/harp-lines/lib/Lines.ts
```diff
@@ -148,31 +154,31 @@ export interface VertexAttributeDescriptor {
...
-    { name: "position", itemSize: 2, offset: 0 },
-    { name: "positionLow", itemSize: 2, offset: 2 },
-    { name: "segment", itemSize: 2, offset: 4 },
-    { name: "tangents", itemSize: 4, offset: 6 },
-    { name: "angles", itemSize: 2, offset: 10 },
-    { name: "texcoord", itemSize: 2, offset: 12 }
+    { name: "position", itemSize: 3, offset: 0 },
+    { name: "positionLow", itemSize: 3, offset: 3 },
+    { name: "segment", itemSize: 2, offset: 6 },
+    { name: "tangents", itemSize: 4, offset: 10 },
```
Astrak: 10>8?

@here/harp-utils/lib/Math2D.ts
```diff
@@ -87,15 +87,15 @@ export namespace Math2D {
     /**
      * Computes the squared length of a line.
      *
-     * @param line An array of even length, which form a line via [x,y,x,y,...] pairs.
+     * @param line An array of even length, which form a line via [x,y,z,x,y,z...] tuples.
```
Astrak:
```suggestion
     * @param line An array that forms a line via [x,y,z,x,y,z...] tuples.
```

@here/harp-lines/lib/Lines.ts
```diff
@@ -65,30 +66,34 @@ export function createLineGeometry(
             const L1 = L[i - 1];
             const L2 = L[i];
             if (!highPrecision) {
-                V.push(x, y, L1, L2, T1[0], T1[1], T2[0], T2[1], 0, 0, +1, -1);
-                V.push(x, y, L1, L2, T1[0], T1[1], T2[0], T2[1], 0, 0, +1, +1);
+                V.push(x, y, z, L1, L2, T1[0], T1[1], T2[0], T2[1], 0, 0, +1, -1);
```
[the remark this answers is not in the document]
zbigg: @robertoraggi : While i completly agree, can we make it separate change?

@here/harp-lines/lib/Lines.ts
amsgames94: Good catch!

@here/harp-utils/lib/Math2D.ts
amsgames94: Nice, will apply.

@here/harp-lines/lib/Lines.ts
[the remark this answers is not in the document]
amsgames94: I agree with @zbigg , let's keep this module refactoring for a different ticket.

@here/harp-lines/lib/Lines.ts
```diff
@@ -41,20 +41,21 @@ export function createLineGeometry(
...
         const len = Math.sqrt(dx * dx + dy * dy);
         sum = sum + len;
         L[i + 1] = sum;
         tangents[i] = [dx, dy];
     }
 
     const isClosed =
-        polyline[0] === polyline[polyline.length - 2] &&
-        polyline[1] === polyline[polyline.length - 1];
+        polyline[0] === polyline[polyline.length - 3] &&
```
robertoraggi: An alternative approach that could save same memory for lines projected on 2D flat surfaces (e.g. mercator projected without terrain) would be to pass the dimensions (2 or 3) to createLineGeometry, in a way making it similar to the eartcut triangulation APIs. Of course, this will require creating two different vertex formats for 2D and 3D lines also the line shader should define the `attribute position` using vec2 or vec3 depending on the dimension.

@here/harp-lines/lib/Lines.ts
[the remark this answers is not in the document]
amsgames94: Removed the angles anyways, but I agree that bigger refactoring to ``harp-lines`` module is needed.

@here/harp-lines/lib/Lines.ts
amsgames94: Again, not a bad point, but more in the scope of ``harp-lines`` refactoring ticket...

@here/harp-geojson-datasource/lib/GeoJsonGeometryCreator.ts
```diff
@@ -358,11 +358,15 @@ export class GeoJsonGeometryCreator {
...
+            for (let i = 0; i < polygon.vertices.length; i += 3) {
+                positions.push(
+                    polygon.vertices[i],
+                    polygon.vertices[i + 1],
+                    polygon.vertices[i + 2]
+                );
             }
 
-            const triangles = earcut(polygon.vertices, polygon.holes);
+            const triangles = earcut(polygon.vertices, polygon.holes, 3);
```
robertoraggi: unfortunately, with the current version of earcut, the triangulated geometry obtained from points projected on the globe surface is not very good, and at the least for now it is not usable. In fact, we get better result when triangulating the 2D rings and then reproject the result on the sphere.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -843,25 +846,33 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
 
                 try {
                     // Triangulate the footprint polyline.
-                    const triangles = earcut(vertices, holes);
+                    const triangles = earcut(vertices, holes, 3);
```
robertoraggi: we should double check and re-evaluate this, but it may be better to triangulate before projecting.

@here/harp-lines/lib/Lines.ts
robertoraggi: I agree, we should remove @here/harp-lines and introduce a more generic @here/harp-geometry with all the bits needed to work with geometries, and yes, I am with you, this is definitely not something that we should do in this PR. 

That said, this change is touching createLineGeometry(...) and it is changing its signatures from accepting a sequence of packaged 2D coordinates (something that this function knows how to handle) to a sequence of 3D points. Unfortunately, this new version of createLineGeometry, at least for now, doesn't create correct geometries unless all the input 3D points have the same Z component, and I find this wrong. In my opinion we should change the signature of createLineGeometry to accept 3D points only after we verified that its code can create valid geometries.

amsgames94:
Since this change is bigger than anticipated, we'll proceed by closing this one and follow-up with smaller tasks that lead to a similar output.

PR #371, opened 2019-04-09
MINOR: Fixing malformed theme JSON files

SahinHERE:
* Fixed berlin_tilzen_base.json
 * Fixed berlin_tilezen_day_reduced.json


zbigg:
Duplicate of #370

SahinHERE:
Basically we did the same fix, I don`t care which commit gets in. Do you want me to cancel my commit?

zbigg:
> Basically we did the same fix, I don`t care which commit gets in. Do you want me to cancel my commit?

I will cancel this, as it's farther in pipeline so with high probability we will wait longer fir CI to build it.

Nevertheless, thanks for taking care and ... sorry for problem, because it was me who broke these files.

SahinHERE:
No worries, can happen to the best of us.

PR #374, opened 2019-04-09
Cache eviction happens when last requested

nzjony:
- Cache eviction is now controlled by when the tile was last requested,
and not whether the tile is visible.
- "visibleTiles" renamed because it was a bit confusing as there are
tiles which are not visible in the visibleTiles list because they are
still loading.

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -1897,21 +1897,16 @@ export class MapView extends THREE.EventDispatcher {
 
         const renderList = this.m_visibleTiles.dataSourceTileList;
 
-        renderList.forEach(({ zoomLevel, renderedTiles, visibleTiles, numTilesLoading }) => {
-            // The visible tiles may not actually be rendered, but they have to stay in the cache
-            // for the next frame.
-            visibleTiles.forEach(tile => {
-                tile.frameNumLastVisible = this.m_frameNumber;
-            });
+        renderList.forEach(({ zoomLevel, renderedTiles }) => {
             renderedTiles.forEach(tile => {
                 this.renderTileObjects(tile, zoomLevel);
+                //We know that rendered tiles are visible (in the view frustum), so we update the
+                //frame number, note we don't do this for the visibleTiles because some may still be
+                //loading (and therefore aren't visible in the sense of being seen on the screen).
+                //Note also, this number isn't currently used anywhere so should be considered to be
+                //removed in the future (though could be good for debugging purposes).
                 tile.frameNumLastVisible = this.m_frameNumber;
             });
-        });
-
-        // Update the visibility flag on every tile in the cache, so it does not get evicted.
-        this.forEachCachedTile(tile => {
-            tile.isVisible = tile.frameNumLastVisible === this.m_frameNumber;
         });
 
         if (currentFrameEvent !== undefined) {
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1897,33 +1897,23 @@ export class MapView extends THREE.EventDispatcher {
 
         const renderList = this.m_visibleTiles.dataSourceTileList;
 
-        renderList.forEach(({ zoomLevel, renderedTiles, visibleTiles, numTilesLoading }) => {
-            // The visible tiles may not actually be rendered, but they have to stay in the cache
-            // for the next frame.
-            visibleTiles.forEach(tile => {
-                tile.frameNumLastVisible = this.m_frameNumber;
```
nzjony: This was used to keep the tiles in the cache, but now we use when the given tile was requested as the means of deciding when to evict from the cache.

ninok: So this commit also contains function changes. Please rename the commit message or ideally split into two commits (renaming and functional)

ninok:
> * "visibleTiles" renamed because it was a bit confusing as there are
>   tiles which are not visible in the visibleTiles list because they are
>   still loading.
I find the new name more confusing or quite ambiguous (what was actually computed?). 
Since the class is computing the visible tiles (or tile keys) the name was a good fit.

Finding a proper name for this class is quite hard b/c it is doing a lot of things. That's a good indicator that this class should be split, maybe in VisbileTileSet and RenderableTileSet

@here/harp-mapview/lib/Tile.ts
```diff
@@ -931,9 +927,9 @@ export class Tile implements CachedResource {
         this.clear();
         this.userTextElements.length = 0;
         this.m_disposed = true;
-
-        // Ensure that tile is removable from tile cache.
         this.isVisible = false;
+        // Ensure that tile is removable from tile cache.
+        this.frameNumLastRequested = 0;
```
ninok: I thought dispose is called from the TileCache when the tile is evicted, so it is a bit weird that we have to make sure that the Tile can be evicted.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -405,8 +407,7 @@ export class Tile implements CachedResource {
     numFramesVisible: number = 0;
 
     /**
-     * The visibility status of the `Tile`. It is actually visible or planned to become visible, and
-     * as such it should not be removed from cache.
+     * The visibility status of the [[Tile]]. It is actually visible or planned to become visible.
      */
     isVisible: boolean = false;
```
ninok: Why do we still need it? Can we replace it with a getter method that takes the current frame number:
```suggestion
    isVisibleInFrame(frame:number): boolean {
        return frame === frameNumLastRequested;
    }
```

@here/harp-mapview/lib/Tile.ts
nzjony: Yes, but it is not the only place it is called from, it is called on tiles that have exited the view, search for disposeTile, so in this case, it can also be removed from the tile cache.

@here/harp-mapview/lib/Tile.ts
nzjony: I made it a getter, the frame number can be obtained by the datasource's mapview's framenumber (this should be refactored at some point).

PR #375, opened 2019-04-09
HARP-5371: DataSource exposes getter/setter for storage level offset.

kkostecki:
This allows to update and read this property via MapView and thus
control level offsets in demo app.
WebTileDataSource should use default level offset set to -1 which
corresponds to 512 tile sizes used.


kkostecki:
> Can you also add the usage in the same topic, otherwise you just add unused code.

Hi @ninok, topic created and set on gerrit, I hope it is working the same way as with casual gerrit code - I have also added you to review there.

PR #376, opened 2019-04-09
HARP-5401: Add option to set maximum frame rate

StefanDachwitz:

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -89,6 +89,16 @@ const DEFAULT_MAX_ZOOM_LEVEL = 20;
 const DEFAULT_MIN_CAMERA_HEIGHT = 20;
 
 /**
+ * Amount of framerate values to pick average from
+ */
+const FRAME_RATE_RING_SIZE = 12;
+
+/**
+ * Default starting value for FPS computation.
+ */
+const FALLBACK_FRAME_RATE = 30;
+
+/**
  * The type of `RenderEvent`.
  */
 export interface RenderEvent extends THREE.Event {
@@ -454,6 +464,17 @@ export interface MapViewOptions {
     dynamicPixelRatio?: number;
 
     /**
+     * Set maximum FPS (Frames Per Second). If VSync in enabled, the specified number may not be
+     * reached, but instead the next smaller number than `maxFps` that is equal to the refresh rate
+     * divided by an integer number.
+     *
+     * E.g.: If the monitors refresh rate is set to 60hz, and if `maxFps` is set to a value of `40`
+     * (60hz/1.5), the actual used FPS may be 30 (60hz/2). For displays that have a refresh rate of
+     * 60hz, good values for `maxFps` are 30, 20, 15, 12, 10, 6, 3 and 1. A value of `0` is ignored.
+     */
+    maxFps?: number;
+
+    /**
      * @hidden
      * Disable all fading animations for debugging and performance measurement.
      */
@@ -534,6 +555,8 @@ export class MapView extends THREE.EventDispatcher {
     private m_updatePending: boolean = false;
     private m_renderer: THREE.WebGLRenderer;
     private m_frameNumber = 0;
+    private m_maxFps = 0;
+    private m_detectedFps: number = FALLBACK_FRAME_RATE;
 
     private m_textElementsRenderer?: TextElementsRenderer;
     private m_textRenderStyleCache = new TextRenderStyleCache();
@@ -560,8 +583,13 @@ export class MapView extends THREE.EventDispatcher {
     private m_previousFrameTimeStamp?: number;
     private m_firstFrameRendered = false;
     private m_firstFrameComplete = false;
+    private m_previousRequestAnimationTime?: number;
+    private m_targetRequestAnimationTime?: number;
+    private m_frameTimeIndex: number = 0;
+    private readonly m_frameTimeRing: number[] = [];
 
     private handleRequestAnimationFrame: any;
+    private handlePostponedAnimationFrame: any;
 
     private m_pickHandler: PickHandler;
 
@@ -653,6 +681,10 @@ export class MapView extends THREE.EventDispatcher {
 
         this.m_pixelRatio = options.pixelRatio;
 
+        if (options.maxFps !== undefined) {
+            this.m_maxFps = Math.max(0, options.maxFps);
+        }
+
         this.m_options.enableStatistics = this.m_options.enableStatistics === true;
 
         this.m_languages = this.m_options.languages || MapViewUtils.getBrowserLanguages();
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1759,16 +1784,35 @@ export class MapView extends THREE.EventDispatcher {
 
         this.m_pixelToWorld = undefined;
 
-        const cameraPitch = Math.min(
```
nzjony: I'm guessing this was a mistake? It seems not related to your commit.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1759,16 +1784,35 @@ export class MapView extends THREE.EventDispatcher {
...
     }
 
+    private detectCurrentFps(now: number) {
+        if (this.m_previousRequestAnimationTime === undefined) {
+            this.m_previousRequestAnimationTime = now;
+            return;
+        }
+
+        // Skip the first frames, they are from not originated from requestAnimationFrame()
+        if (this.m_frameNumber > 5 /*&& now - this.m_previousRequestAnimationTime > 5*/) {
```
nzjony: should this comment stay here? Why?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1780,8 +1824,58 @@ export class MapView extends THREE.EventDispatcher {
...
+
+        const vSyncFrameTime = 1000 / (this.m_detectedFps !== undefined ? this.m_detectedFps : 30);
+        const frameInterval = 1000 / this.m_maxFps;
+
+        const startTime =
+            this.m_previousFrameTimeStamp === undefined
+                ? performance.now()
+                : this.m_previousFrameTimeStamp;
+
+        // Magic ingredient to compensate time flux.
```
nzjony: this is 3ms? I think it should be added to the variable name

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1759,16 +1784,35 @@ export class MapView extends THREE.EventDispatcher {
...
-        const cameraPitch = Math.min(
-            MapViewUtils.extractYawPitchRoll(this.camera.quaternion).pitch,
-            Math.PI / 3
-        );
+        const cameraPitch = MapViewUtils.extractYawPitchRoll(this.camera.quaternion).pitch;
 
-        const distance = Math.abs(this.camera.position.z) / Math.cos(cameraPitch);
+        const distance = Math.min(
+            Math.abs(this.camera.position.z) / Math.cos(cameraPitch),
+            this.camera.position.z * 2
```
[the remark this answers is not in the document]
StefanDachwitz: That seems to be a merge issue. Reverting this.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1791,6 +1839,27 @@ export class MapView extends THREE.EventDispatcher {
...
+        }
+
+        // Skip the first frames, they are from not originated from requestAnimationFrame()
+        if (this.m_frameNumber > 5) {
+            const currentFps = 1000 / (now - this.m_previousRequestAnimationTime);
+            this.m_frameTimeRing[this.m_frameTimeIndex] = currentFps;
+            this.m_frameTimeIndex = (this.m_frameTimeIndex + 1) % FRAME_RATE_RING_SIZE;
+
+            let sum: number = 0;
+            for (let i = 0; i < FRAME_RATE_RING_SIZE; i++) {
```
nzjony: minor, you could save some code by using reduce(): https://www.w3schools.com/jsref/jsref_reduce.asp

@here/harp-mapview/lib/MapView.ts
```diff
@@ -952,6 +989,17 @@ export class MapView extends THREE.EventDispatcher {
         this.m_forceCameraAspect = aspect;
     }
 
+    /**
+     * Maximum FPS. If defined (and > 0) it is the maximum FPS that is used.
+     */
+    set maxFps(fps: number | undefined) {
```
nzjony: why can the user set an undefined maxFps?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1791,6 +1839,27 @@ export class MapView extends THREE.EventDispatcher {
...
+        if (this.m_frameNumber > 5) {
+            const currentFps = 1000 / (now - this.m_previousRequestAnimationTime);
+            this.m_frameTimeRing[this.m_frameTimeIndex] = currentFps;
+            this.m_frameTimeIndex = (this.m_frameTimeIndex + 1) % FRAME_RATE_RING_SIZE;
+
+            let sum: number = 0;
+            for (let i = 0; i < FRAME_RATE_RING_SIZE; i++) {
+                sum += this.m_frameTimeRing[i];
+            }
+            this.m_detectedFps = Math.floor(sum / FRAME_RATE_RING_SIZE);
```
nzjony: this is slightly wrong, if the ring is only partially filled, then it should divide by the number of valid entries (?)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1802,8 +1871,61 @@ export class MapView extends THREE.EventDispatcher {
...
+        if (this.m_targetRequestAnimationTime === undefined) {
+            return;
+        }
+
+        if (this.m_animationFrameHandle !== undefined) {
+            cancelAnimationFrame(this.m_animationFrameHandle);
+            this.m_animationFrameHandle = undefined;
+        }
+
+        this.detectCurrentFps(now);
```
nzjony: is this correct? This will detect the fps using a postponed animation frame, which will artificially inflate the FPS

okanishchev: in this case we need detected fps only to measure delay to next rAF callback,
and this delay is computed by browser using previous callbacks duration.
As there is no standard algo for this delay, and no api method, we need to estimate it somehow.

Also, we can't do rAF, then setTimeout (or similar) to next rAF - that won't work as expected, especially on high cpu load.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -965,6 +998,17 @@ export class MapView extends THREE.EventDispatcher {
...
 
+    /**
+     * Maximum FPS. If defined (and > 0) it is the maximum FPS that is used.
+     */
+    set maxFps(fps: number) {
+        this.m_maxFps = Math.max(0, fps);
+    }
+
+    get maxFps(): number {
+        return this.m_maxFps > 0 ? this.m_maxFps : 0;
```
nzjony: Math.max(0, this.m_maxFps) would be equivalent?

okanishchev: Yeah, I think so. As we already using max everywhere, we can also do it here.

PR #379, opened 2019-04-12
HARP-5286: Styling size metric based on screen space pixel sizes.

kkostecki:
Implement screen space sizing (pixelToWorld ratio) based on camera
focal length and camera target distance.
Formula allows to render objects in exact screen space size given in
pixels for the objects close to camera focus point, objects closer or
farther then that are perspective scaled acordingly.
Formula is camera tilt and roll angle independent as also does not require
any raycasting, so works even with camera oriented to sky.

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -515,6 +515,7 @@ export class MapView extends THREE.EventDispatcher {
 
     private m_camera: THREE.PerspectiveCamera;
     private m_focalLength: number;
+    private m_lookAtDistance: number;
     private m_pointOfView?: THREE.PerspectiveCamera;
 
     private m_tempVector3: THREE.Vector3 = new THREE.Vector3();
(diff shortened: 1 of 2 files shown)


kkostecki:
Please check-out demo that shows the functionality. There are two styles are using pixel metric:
- roads marked with pink color,
- country capitals - marked with green.

zbigg:
@heremaps-bot ok to test

kkostecki:
Some explanations for solutioning may be found on jira. 
Most important are relations between focal length and fov:
`
// Focal length vs fov
fov = 2 * atan(canvasSize / 2 * focalLength)
`
and between final relation giving object screen space size:
`
// Object screen size vs its world space size and distance
objScreenSize = (focalLength * objWorldSize) / worldDistance;

// Object world size vs its desired screen size and world distance
objWorldSize = (worldDistance * objScreenSize) / focalLength;
`
So if we choose to have exact pixel metric match at camera focus point (look at) then we may pass look at distance to above equation.

Final equation for pixelToWorld ratio will take a form:
`
// Pixel to world ratio formula
pixelToWorld = (lookAtDistance * 1.0) / focalLength;
`

PR #381, opened 2019-04-15
HARP-4924 Fix switch statement logic

nzjony:
This checks that the switch statements are exhaustive in the TileLoader.

Cancelling a tile which has already loaded, or not loaded, etc has no affect.

### DIFF
--- a/@here/harp-mapview-decoder/lib/TileLoader.ts
+++ b/@here/harp-mapview-decoder/lib/TileLoader.ts
@@ -13,7 +13,7 @@ import {
 } from "@here/harp-datasource-protocol";
 import { TileKey } from "@here/harp-geoutils";
 import { DataSource, TileLoaderState } from "@here/harp-mapview";
-import { LoggerManager } from "@here/harp-utils";
+import { assertNever, LoggerManager } from "@here/harp-utils";
 
 import { DataProvider } from "./DataProvider";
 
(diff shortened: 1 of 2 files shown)


@here/harp-mapview-decoder/lib/TileLoader.ts
```diff
@@ -148,6 +149,17 @@ export class TileLoader {
                     this.requestController = undefined;
                 }
                 break;
+
+            case TileLoaderState.Canceled:
```
zbigg: Tentative.
This just adds bytes that add no value for final application (we have worse boilerplate here and there but here we add more).

`cancel` was written as idempotent method, which is very useful property - you expect only certain effect, without looking at current state.

It's no error to `cancel` loader that for some particular reason is in `loaded` state (or any other), it's should be just  `no-op`.

@here/harp-mapview-decoder/lib/TileLoader.ts
```diff
@@ -111,6 +111,7 @@ export class TileLoader {
                 this.ensureLoadingStarted();
                 return this.donePromise!;
         }
+        assertNever(this.state);
```
zbigg: This is not needed if we trust TypeScript.
Typescript already checks if all paths return value.

Experiment: comment `TileLoaderState.Failed` case and see that typescript doesn't even compile this.

If we don't trust `this.state` then this would be viable check.

nzjony: Thanks for pointing out, yes that is correct. I will remove.

PR #383, opened 2019-04-15
HARP-5185: Add @here/harp-geometry.

robertoraggi:
Introduce the new module @here/harp-geometry for now containing
only OrientedBox3.

### DIFF
--- a/@here/harp-geometry/lib/OrientedBox3.ts
+++ b/@here/harp-geometry/lib/OrientedBox3.ts
@@ -1,0 +1,102 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { OrientedBox3Like } from "@here/harp-geoutils";
+import { Frustum, Matrix4, Plane, Vector3 } from "three";
+
+export class OrientedBox3 implements OrientedBox3Like {
+    /**
+     * The position of the center of this `OrientedBox3`.
+     */
+    readonly position = new Vector3();
+
+    /**
+     * The x-axis of this `OrientedBox3`.
+     */
+    readonly xAxis = new Vector3(1, 0, 0);
+
+    /**
+     * The y-axis of this `OrientedBox3`.
+     */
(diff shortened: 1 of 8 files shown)


@here/harp-geometry/lib/OrientedBox3.ts
```diff
@@ -0,0 +1,102 @@
...
+export class OrientedBox3 implements OrientedBox3Like {
+    /**
+     * The position of the center of this `OrientedBox3`.
+     */
+    readonly position = new Vector3();
+
+    /**
+     * The x-axis of this `OrientedBox3`.
+     */
+    readonly xAxis = new Vector3(1, 0, 0);
```
zbigg: > `xAxis`
> `yAxis`
> `zAxis`

Having threee independent vectors implies we support skewed boxes, so not only box (which is by intuition cuboid ...).
So do we support non-cuboid boxes then ?

Should this be optimized-out to only one rotation vector or is this too complex and not worth effort ?

@here/harp-geometry/lib/OrientedBox3.ts
```diff
@@ -0,0 +1,102 @@
...
+     * @param extents The extents of the `OrientedBox3`.
+     */
+    constructor(position: Vector3, rotationMatrix: Matrix4, extents: Vector3);
+
+    /**
+     * Creates a new `OrientedBox3`.
+     *
+     * @hideconstructor
+     */
+    constructor(position?: Vector3, rotationMatrix?: Matrix4, extents?: Vector3) {
```
zbigg: Minor. My intuitive priority would be `position, extents and rotation` as for many cases last one would be not used at all.

robertoraggi: yeah, that is definitely a possible way of defining the values, but I would prefer if we keep the order consistent with the standard way of composing/decomposing transformations, e.g.
https://threejs.org/docs/#api/en/math/Matrix4.compose
https://threejs.org/docs/#api/en/math/Matrix4.decompose

we could as well use the quaternion, instead of a Matrix, but I decided to go for the matrix4 because what we really need are the basis.

nzjony: especially for the globe, the rotation matrix would always need to be defined.

PR #384, opened 2019-04-16
HARP-5188: 3D Lines buffer generation.

amsgames94:


amsgames94:
GPU tests please

@here/harp-materials/lib/ShaderChunks/LinesChunks.ts
```diff
@@ -5,23 +5,50 @@
...
+        pos += uv.y * lineWidth * dir / cos(angle / 2.0);
+        uv.x = ss + uv.x * lineWidth * uv.y * tan(angle / 2.0);
+    }
+    else {
+        vec3 t = vec3(-dir.y, dir.x, dir.z);
+        pos += uv.y * lineWidth * dir + uv.x * lineWidth * t;
+        uv.x = ss + uv.x * lineWidth;
+    }
+
+    /* if (angle != 0.0) {
```
robertoraggi: I would prefer if you could remove this commented out code.

amsgames94: Adressed in last revision.

nzjony:
I notice there are no tests, I think it would be good to have some IBC tests to check we don't have any regressions in future.

PR #386, opened 2019-04-16
HARP-4924 Tile interface change to accept offset

nzjony:
This change is [Reviewable](https://reviewable.io/reviews/heremaps/harp.gl/386)


nzjony:
Hey guys, this is just the cosmetic changes and doesn't contain any attempt to shared the DecodedTile, this will be part of a separate commit.

nzjony:
> I would expect, that MapView is clamping the longitude of the geoCenter to always be between -180 and 180.

I removed it.

@here/harp-mapview-decoder/test/TileDataSourceTest.ts
```diff
@@ -146,26 +147,23 @@ describe("TileDataSource", function() {
 
         const spyTileSetDecodedTile = sinon.spy(tile, "setDecodedTile");
 
-        // act
-        testedDataSource.updateTile(tile);
```
nzjony: this was removed, because it is already called in getTile

@here/harp-mapview-decoder/test/TileDataSourceTest.ts
```diff
@@ -201,6 +199,167 @@ describe("TileDataSource", function() {
...
+        const tile2 = mockedDataSource.testedDataSource.getTile(
+            TileKey.fromRowColumnLevel(0, 0, 0)
+        )!;
+        assert(tile2);
+        const spyTileSetDecodedTile2 = sinon.spy(tile2, "setDecodedTile");
+        assert.equal(tile1.tileLoader!, tile2.tileLoader);
+
+        // Send two cancel requests (for example, both tiles have left the viewport).
+        tile1.tileLoader!.cancel();
+        tile1.tileLoader!.cancel();
```
nzjony: make tile2

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -152,6 +156,9 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
         });
 
         this.cacheable = true;
+        this.m_tileLoaderCache = new LRUCache<number, TileLoader>(40, tileLoader => {
```
ninok: 40 seems quite big for wrap-around tiles.

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -201,18 +208,74 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
...
-            tileKey,
-            this.m_options.dataProvider,
-            this.decoder,
-            0
-        );
+        const mortonCode = tileKey.mortonCode();
+        const tileLoader = this.m_tileLoaderCache.get(mortonCode);
+        // Filter out disposed TileLoaders.
+        if (tileLoader !== undefined && tileLoader.state !== TileLoaderState.Disposed) {
+            tile.tileLoader = tileLoader;
```
ninok: Shouldn't we remove the disposed ones from the cache immediately?

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -201,18 +208,74 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
...
+            const newTileLoader = new TileLoader(
+                this,
+                tileKey,
+                this.m_options.dataProvider,
+                this.decoder,
+                0
+            );
+            tile.tileLoader = newTileLoader;
+            // We don't cache tiles with level 4 and above, at this level, there are 16 (2^4)
+            // tiles horizontally, given the assumption that the zoom level assumes the tile
```
ninok: This assumption depends on the storageLevelOffset of the datasource, so you should add it to the condition below.

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -201,18 +208,74 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
...
         this.updateTile(tile);
         return tile;
     }
 
+    updateTile(tile: Tile) {
+        const tileLoader = tile.tileLoader;
+        if (tileLoader === undefined) {
+            return;
+        }
+        if (tileLoader.decodedTile !== undefined) {
```
ninok: && this.decodedTileHasGeometry(decodedTile) ?

@here/harp-mapview-decoder/lib/TileDataSource.ts
nzjony: correct, this is the number used in VisibleTileSet for the tile cache, I have made it a bit clearer.

@here/harp-mapview-decoder/lib/TileDataSource.ts
nzjony: Unfortunately it isn't currently possible, the Tile::dispose method calls TileLoader::dispose, however TileLoader doesn't have a reference to the TileDataSource, just to the DataSource (which contains owns the TileLoaders). I could add it, but it would make the code a bit more ugly, because it would involve type checking etc.

Also, I found a bug, currently the TileLoader.dispose method doesn't check correctly if it can actually dispose, or whether there is another Tile waiting on it. I will fix and add test.

Actually... I don't think it makes sense to have a Disposed state, because it anyway just calls cancel, and then sets the disposed state. A disposed TileLoader can't be restarted like a cancelled one, but I don't quite see the use case for this.

@here/harp-mapview-decoder/lib/TileDataSource.ts
nzjony: I will reword.

PR #388, opened 2019-04-23
HARP-5189: Add a simple SubdivisionModifier.

robertoraggi:
The SubdivisionModifier is used to increase the number of
vertices of polygons meant to be rendered on a sphere.

### DIFF
--- a/@here/harp-geometry/lib/SphericalGeometrySubdivisionModifier.ts
+++ b/@here/harp-geometry/lib/SphericalGeometrySubdivisionModifier.ts
@@ -1,0 +1,112 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Face3, Geometry, Vector3 } from "three";
+
+/**
+ * The [[SphericalGeometrySubdivisionModifier]] subdivides triangle mesh geometries positioned
+ * on the surface of a sphere centered at `(0, 0, 0)`.
+ */
+export class SphericalGeometrySubdivisionModifier {
+    /**
+     * Constructs a new [[SphericalGeometrySubdivisionModifier]].
+     *
+     * @param angle The maximum angle in radians between two vertices and the origin.
+     */
+    constructor(readonly angle: number) {}
+
+    /**
+     * Subdivides the faces of the given [[THREE.Geometry]].
+     *
+     * This method modifies (in-place) the vertices and the faces of the geometry.
+     * Please note that only the vertex position attribute is changed
+     * by this modifier. Normals, UVs, and vertex colors are left unmodified.
+     *
+     * @param geometry The [[THREE.Geometry]] to subdivide.
+     */
+    modify(geometry: Geometry): Geometry {
+        const { vertices, faces } = geometry;
+
+        // A cache containing the indices of the vertices added
+        // when subdiving the faces of the geometry.
+        const cache = new Map<string, number>();
+
+        /**
+         * Returns the index of the vertex positioned in the middle
+         * of the given vertices.
+         */
+        function middleVertex(i: number, j: number): number {
+            // build a unique `key` for the pair of indices `(i, j)`.
+            const key = `${Math.min(i, j)}_${Math.max(i, j)}`;
+
+            const h = cache.get(key);
+
+            if (h !== undefined) {
+                // nothing to do, a vertex in the middle of (i, j) was already created.
+                return h;
+            }
+
+            // the position of the new vertex.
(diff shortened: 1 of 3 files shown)


@here/harp-geometry/lib/SubdivisionModifier.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Subdivides the faces of the given `THREE.Geometry`.
+     *
+     * @param geometry The `THREE.Geometry` to subdivide.
+     */
+    modify(geometry: Geometry): Geometry {
+        const { vertices, faces } = geometry;
+
+        const cache = new Map<string, number>();
```
ninok: What is this cache caching?

@here/harp-geometry/lib/SubdivisionModifier.ts
```diff
@@ -0,0 +1,95 @@
...
+import {
+    Face3,
+    Geometry,
+} from "three";
+
+/**
+ * The `SubdivisionModifier` subdivides triangle mesh geometries positioned
+ * on the surface of a sphere centered at `(0, 0, 0)`.
+ */
+export class SubdivisionModifier {
```
ninok: Maybe rename to SphereSubdivisionModifier b/c for terrain we might use a different approach.

@here/harp-geometry/lib/SubdivisionModifier.ts
zbigg: Middle vertices. 
`middleVertexCache` ?

@here/harp-geometry/lib/SubdivisionModifier.ts
```diff
@@ -0,0 +1,95 @@
...
+            } else if (gamma === m) {
+                const d = middleVertex(face.c, face.a);
+                faces.push(new Face3(face.a, face.b, d));
+                faces.push(new Face3(d, face.b, face.c));
+            } else {
+                throw new Error("failed to subdivide the given geometry");
+            }
+        }
+
+        geometry.faces = subdividedFaces;
```
zbigg: Geometry may have some interdependencies between internal data.
Shouldn't we trigger one of these:
```
	elementsNeedUpdate;
	verticesNeedUpdate;
	this.uvsNeedUpdate =...
```
(https://threejs.org/docs/#api/en/core/Geometry)

@here/harp-geometry/lib/SubdivisionModifier.ts
zbigg: I would question necessity to put this in separate class with only one method.

Our guidleines say:
> Use simplest tool for the job: function vs. object/namespace/module vs class.

This is plain "data->data" transformation, with no requirements for state (`angle` is hardly _state_).

My proposal `function subdivideGeometryToMatchSphere` ?

@here/harp-geometry/lib/SubdivisionModifier.ts
```diff
@@ -0,0 +1,95 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+    Face3,
+    Geometry,
+} from "three";
+
+/**
+ * The `SubdivisionModifier` subdivides triangle mesh geometries positioned
```
nzjony: Minor, should be [[SubdivisionModifier]]

@here/harp-geometry/lib/SubdivisionModifier.ts
robertoraggi: I renamed it to `SphericalGeometrySubdivisionModifier`.

@here/harp-geometry/lib/SubdivisionModifier.ts
robertoraggi: I updated the change to set vertices/elementsNeedUpdate. I also I added a paragraph in the documentation explaining that only the position and the faces of the geometry are modified, other common attributes of THREE.Geometry (like normals and UVs) are usually inferred by the vertex position so currently we don't need to preserve them, but if needed we can easily extend the code compute values for them.

@here/harp-geometry/lib/SubdivisionModifier.ts
robertoraggi: the API is following the de-facto standard interface of three.js's modifiers.
https://github.com/mrdoob/three.js/tree/dev/examples/js/modifiers

@here/harp-geometry/lib/SubdivisionModifier.ts
robertoraggi: done

@here/harp-geometry/lib/SubdivisionModifier.ts
robertoraggi: I added some documentation explaining what is that we are caching,

PR #391, opened 2019-04-24
HARP-5184: Preserve Projected Z-Component.

amsgames94:
Continuation of #369

### DIFF
--- a/@here/harp-lines/lib/Lines.ts
+++ b/@here/harp-lines/lib/Lines.ts
@@ -7,13 +7,12 @@
 import * as THREE from "three";
 
 // Preallocate temp variables used during line generation.
-const tmpTangent0_2D = new THREE.Vector2();
-const tmpTangent1_2D = new THREE.Vector2();
-const tmpBitangent_2D = new THREE.Vector2();
-const tmpTangent0_3D = new THREE.Vector3();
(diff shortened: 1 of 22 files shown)


@here/harp-lines/lib/Lines.ts
```diff
@@ -27,92 +27,70 @@ interface VertexAttributeDescriptor {
...
-    constructor(params: { dimensions: 2 | 3 }) {
-        this.dimensions = params.dimensions;
-    }
 }
 
 /**
  * Creates a [[LineGeometry]] object out of a polyline.
  *
  * @param polyline Array of `numbers` describing a polyline.
+ * @param projection [[Projection]] to which project the polyline.
```
ninok: Passing the whole projection is a bit misleading b/c it looks like this method will project the line but it doesn't. The polyline is already projected (as far as I can see).
Passing a boolean flag would make things more clear.
Please also add some comment that polyline is always 3 dimensional now

amsgames94: It's true that currently only the ``getDimensions`` method of the projection is used, but in the future this might not be the case and you might want to do projection-specific computations (i.e. sphere and elliptical projections are not gonna compute the vertex normal in the same way).

Don't mind changing it, but I think it's more future proof this way.

PR #392, opened 2019-04-24
HARP-5488: Improve mesh size (memory cost) approximation.

kkostecki:
MapViewUtils.estimateMeshSize does not count THREE.Points geometry
and throws thus trows warnings. Extend the mesh size estimation for
THREE.Points instances.

Patch-set fixes also some estimation constants, such as estimated cost
per THREE.Object3d and cost of vertex.

### DIFF
--- a/@here/harp-mapview/lib/Utils.ts
+++ b/@here/harp-mapview/lib/Utils.ts
@@ -596,6 +596,54 @@ export namespace MapViewUtils {
         objectSize.gpuSize += attrBytes;
     }
 
+    function estimateGeometrySize(
+        geometry: THREE.Geometry | THREE.BufferGeometry,
+        objectSize: MemoryUsage,
+        visitedObjects: Map<string, boolean>
+    ): void {
+        const isNewObject =
+            geometry.uuid === undefined || visitedObjects.get(geometry.uuid) !== true;
+
+        if (!isNewObject) {
+            return;
+        }
+        visitedObjects.set(geometry.uuid, true);
+
+        let bufferGeometry: THREE.BufferGeometry | undefined;
+
+        if (geometry instanceof THREE.Geometry) {
+            // Each vertex is represented as 3 floats vector (24 bytes).
+            objectSize.heapSize += geometry.vertices.length * 24;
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/Utils.ts
```diff
@@ -596,6 +596,62 @@ export namespace MapViewUtils {
...
+                objectSize.heapSize += geometry.vertices.length * 24;
+                // Face: 3 indices (24 byte), 1 normal (3 floats = 24). Vertex normals and
+                // colors are not counted here.
+                objectSize.heapSize += geometry.faces.length * (24 + 24);
+                // Additionally, the internal _bufferGeometry is also counted:
+                bufferGeometry = (geometry as any)._bufferGeometry;
+            } else if (geometry instanceof THREE.BufferGeometry) {
+                bufferGeometry = geometry;
+            } else {
+                bufferGeometry = undefined;
```
zbigg: You can warn and return here.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -596,6 +596,62 @@ export namespace MapViewUtils {
...
+            }
+
+            if (bufferGeometry !== undefined) {
```
zbigg: and remove this condition because it will be always defined.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -596,6 +596,62 @@ export namespace MapViewUtils {
...
+    function estimateGeometrySize(
+        geometry: THREE.Geometry | THREE.BufferGeometry,
+        objectSize: MemoryUsage,
+        visitedObjects: Map<string, boolean>
+    ): void {
+
+        const isNewObject =
+            geometry.uuid === undefined || visitedObjects.get(geometry.uuid) !== true;
+
+        if (isNewObject) {
```
zbigg: Use early return if you know there is nothing to do.

    if (!isNewObject) { return } 

Less indentation, prettier code

zbigg:
@heremaps-bot ok to test

PR #393, opened 2019-04-25
MINOR: Added example for standard-textured technique

ninok:
Also added texture repeat properties and roughnessMap to StandardTexturedTechnique.


@here/harp-map-theme/resources/berlin_tilezen_textured.json
```diff
@@ -0,0 +1,1169 @@
+{
```
zbigg: I know that it's out-of-scope of this task, but copying whole theme for this example is ... ovekill.

Maybe something like this can work:

1) Ensure all styles in original `berlin...json` are findable by for example unique "description".
2) Load original theme using ThemeLoader - as usual
3) Update specific techniques (like `park) like this:

```
updateTheme(theme, "park", {
    technique: "standard-textured",
    attr: {
        ...
    }
});
```

This also should work

```
function updateStyle(theme: Theme, description: string, newStyle: Style) {
    ThemeVisitor.visit(theme, style => {
         if( style.description === "description") {
              Object.assign(style, newStyle);
        }
    });
}
```

ninok: I was also thinking about this, but since this is an example it would detract the users attention from the really important thing, i.e.:
_You can just use a texture in the stylesheet and that's it. No code changes needed_
If I would do what you propose it looks like you have to load the Theme and modify it to use textures, but that's not the case.

Maybe this discussion just reveals a generic problem we have right now: There are no simple examples for the themes. We only have a full blown theme. So in the future we might want to have really simple theme files that also help to understand our styling capabilities.

ninok: Another idea that just came to my mind was to patch the berlin base theme and generate the textured one at build time (like prepack.ts). Than we would at least have both themes in sync. What do you think?

zbigg: Yes, the problem is big, but on the other hand, source of this example is soooo big, that it's no longer "you can use texture, here is how to do it", it is rather "here is big pile of unreadable JSON with 4 lines of real interest  hidden in 1000 lines of uncommented JSON".

IMHO, Update _here_ (i mean locally, in same source file) as proposed, is much more "local" and discoverable than non-commented JSON blob.

(this is bigger discussion, we can do it later)

@here/harp-examples/src/hello_textured.ts
```diff
@@ -0,0 +1,145 @@
...
+</style>
+<p id="texture-license">Textures by
+<a href="https://opengameart.org/content/wall-grass-rock-stone-wood-and-dirt-480">
+West</a>.</p>`;
+    }
+
+    /**
+     * Modify the theme after loading and replace some area features with the
+     * [[StandardTexturedTechnique]] to render them with textures and lighting.
+     * This could solely be done in the theme json file but for explanatory resons it's done here.
```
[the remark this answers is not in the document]
ninok: StandardTexturedTechnique (and StandardTechnique) implement lighting as opposed to the FillTechnique. I will reword to make it more clear.

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -101,80 +101,88 @@ export function createMaterial(
     material.depthTest = isExtrudedPolygonTechnique(technique) && technique.depthTest !== false;
 
     if (isStandardTexturedTechnique(technique)) {
-        ["map", "normalMap", "displacementMap"].forEach((texturePropertyName: string) => {
-            const textureProperty = (technique as any)[texturePropertyName];
-            if (textureProperty === undefined) {
-                return;
-            }
+        ["map", "normalMap", "displacementMap", "roughnessMap"].forEach(
```
amsgames94: Since the aim of this ticket is to explore three's ``MeshStandardMaterial``, why don't we expose all the other maps that can come into play in the PBR light model?
* Ao
* Metalness
* EnvMap

You can find all of them here: https://threejs.org/docs/index.html#api/en/materials/MeshStandardMaterial

ninok: good point, but at least aoMap, lightMap need additional set of texture coordinates. Implementing that would exceed the scope of this task. Will add the others though.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -1030,6 +1068,9 @@ export interface TextureProperties {
      * Flip texture vertically.
      */
     flipY?: boolean;
+
+    repeatU?: number;
```
amsgames94: Typedoc?

PR #394, opened 2019-04-25
MINOR: Re-enable no-unused-variable tslint option

ninok:


tslint.json
```diff
@@ -34,6 +34,7 @@
             }
         ],
         "no-unused-expression": true,
+        "no-unused-variable": true,
```
haraldF: I'm getting lost in this deprecated / undeprecated jungle :) Is it now undeprecated again? See e.g. https://github.com/palantir/tslint/issues/4100

ninok: It is deprecated and there is a warning but we don't have the proper tsc options in place so I would stick to tslint for now.

PR #387, opened 2019-04-19
HARP-5484: Buildings should be shown at all available storage levels.

kkostecki:
Themes fix, the styles should not filter buildings. Remove the `$level`
condition from all styles relevant for the buildings polygon data.

PR #395, opened 2019-04-25
MINOR: Make NY buildings visible in the hello example.

Astrak:
Broader view when starting the hello example.


@here/harp-map-theme/resources/berlin_tilezen_base.json
```diff
@@ -1,1755 +1,1183 @@
...
-    "startRatio": 0.8
-  },
-  "clearColor": "#87959A",
-  "lights": [
-    {
-      "type": "ambient",
-      "color": "#FFFFFF",
-      "name": "ambientLight",
-      "intensity": 0.9
+    "sky": {
```
zbigg: Why reformatting whole file?
BTW, Krystian isn't this PR dup of #387, which enabled bulidings on all zoom levels?
Is there any functional change hidden here?

ninok: This also happens to me all the time when editing json files in vcode (due to prettier plugin). Ideally the formatting of our json files and the ts/js files should be the same, but that is currently not the case.

Astrak: > Why reformat the whole file

`Ctrl+S`...

> isn't this PR dup of #387, which enabled bulidings on all zoom levels?

Indeed it is. Why is #387 still not merged though?

zbigg: > Ctrl+S...

Don't let tools be more smart than they should be.

Yes, theme formatting is broken because design gave use files formatted with 2 spaces and we usually use 4.

Let's decide finally.
I guess, we have 2 spaces for json written down as  rule somewhere.

zbigg:
@heremaps-bot: retest this please

Retesting with #387 already merged.

PR #414, opened 2019-04-30
HARP-5577 Add missing packages to test-npm-packages.

musculman:
Re-enable the test in Travis


musculman:
I have tested this script with `HEAD` of `master` and also simulating a release scenario.
Just run `lerna version` locally and then run `./scripts/test-npm-packages.sh`
I believe the root cause was just a couple of missing packages

PR #415, opened 2019-04-30
HARP-5603: Fix secondaryRenderOrder style argument.

zbigg:


@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
```diff
@@ -263,6 +263,9 @@ export class StyleSetEvaluator {
             if (currStyle.renderOrderBiasGroup !== undefined) {
                 technique.renderOrderBiasGroup = currStyle.renderOrderBiasGroup;
             }
+            if ((currStyle as any).secondaryRenderOrder !== undefined) {
```
zbigg: This is temporary solution, i don't like it too.

We can either fix it now or think about better solution, so `secondaryRenderOrder` is not alien to rest of styling framework.

PR #416, opened 2019-04-30
HARP-5602: Improved terrain height-based coloring.

amsgames94:


@here/harp-mapview/lib/Tile.ts
```diff
@@ -1285,6 +1285,34 @@ export class Tile implements CachedResource {
                     materials[techniqueIndex] = material;
                 }
 
+                // Modify the standard textured shader to support height-based coloring.
+                if (
+                    isStandardTexturedTechnique(technique) &&
+                    technique.displacementMap !== undefined
```
ninok: What if the displacementMap is set in the stylesheet?
You should either add a bool "useHeightColoring" or maybe it's time to introduce a dedicated TerrainTechnique.

amsgames94: This was a overlook on my behalf, I forgot to add ``&& technique.heightBasedColor === true``. Should be fixed now.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1285,6 +1285,35 @@ export class Tile implements CachedResource {
...
+                            `#include <map_pars_fragment>
+    uniform sampler2D displacementMap;
+    uniform float displacementScale;
+    uniform float displacementBias;`
+                        );
+
+                        shader.fragmentShader = shader.fragmentShader.replace(
+                            "#include <map_fragment>",
+                            `#ifdef USE_MAP
+    float disp = texture2D( displacementMap, vUv ).x / ${EarthConstants.MAX_HEIGHT.toFixed(
```
ninok: I think it would be nicer if we would specify the max( and min) range in the technique instead of using earth max height, b/c this really depends on the source of the displacement map

amsgames94: I don't necessarily disagree, but I don't know if we should do this at this moment in the project (specially since the ``TerrainDataSource`` is a really special kind of ``DataSource``).

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -848,6 +848,30 @@ export interface StandardTexturedTechniqueParams extends BaseStandardTechniquePa
     displacementMapProperties?: TextureProperties;
 }
 
+/**
+ * Technique used to render a terrain geometry with a texture.
+ * When using this technique, the datasource will produce texture coordinates in
+ * local tile space (i.e. [0,0] at south-west and [1,1] at north-east tile corner).
+ */
+export interface StandardTexturedTerrainTechniqueParams extends StandardTexturedTechniqueParams {
```
ninok: Maybe just TerrainTechnique would be enough

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -848,6 +848,30 @@ export interface StandardTexturedTechniqueParams extends BaseStandardTechniquePa
...
+    /**
+     * Colors to be applied at different heights (as a results of a `displacementMap`).
+     */
+    heightBasedColors?: HeightBasedColors;
+
+    /**
+     * If `heightBasedColor` is true, this value defines the interpolation method used to generate
```
ninok: If `heightBasedColor` is _defined_

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -848,6 +848,30 @@ export interface StandardTexturedTechniqueParams extends BaseStandardTechniquePa
...
+     * the height-based gradient texture (defaults to `Discrete`).
+     */
+    heightGradientInterpolation?: "Discrete" | "Linear" | "Cubic";
+
+    /**
+     * If `heightBasedColor` is true, this value defines the width (in pixels) of the generated
```
ninok: If heightBasedColor is _defined_

@here/harp-geoutils/lib/projection/EarthConstants.ts
```diff
@@ -10,4 +10,7 @@ export class EarthConstants {
 
     /** The equatorial radius in meters. */
     static EQUATORIAL_RADIUS: number = 6378137.0;
+
+    /** The highest point on earth (Mt. Everest) in meters. */
+    static MAX_HEIGHT: number = 8848.0;
```
ninok: You also have to map the lowest point(i.e. -433) to 0.0 when creating the texture. If you add the values here, please remove the ones in TerrainElevationRangeSource.ts

@here/harp-mapview/lib/ThemeHelpers.ts
```diff
@@ -6,68 +6,103 @@
...
-        return THREE.LuminanceFormat;
-    } else if (format === "LuminanceAlpha") {
-        return THREE.LuminanceAlphaFormat;
-    } else if (format === "RGBE") {
-        return THREE.RGBEFormat;
-    } else if (format === "Depth") {
-        return THREE.DepthFormat;
-    } else if (format === "DepthStencil") {
-        return THREE.DepthStencilFormat;
+    switch (format) {
```
ninok: Cool. I was doing C++ for too long. I did not even know that you can do a switch with strings.

Can you please add RedFormat as well. When I implemented this RedFormat was not in the type definitions yet.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1332,7 +1362,9 @@ export class Tile implements CachedResource {
 
                 if (
                     !bufferGeometry.getAttribute("normal") &&
-                    (isStandardTechnique(technique) || isStandardTexturedTechnique(technique))
+                    (isStandardTechnique(technique) ||
+                        isStandardTexturedTechnique(technique) ||
+                        isStandardTexturedTerrainTechnique(technique))
```
ninok: For terrain we actually don't need per vertex normals.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
amsgames94: Makes sense. Will change.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
amsgames94: Good catch.

@here/harp-mapview/lib/ThemeHelpers.ts
amsgames94: Will take care of it.

@here/harp-mapview/lib/Tile.ts
amsgames94: True, will remove.

@here/harp-geoutils/lib/projection/EarthConstants.ts
amsgames94: Done.

@here/harp-mapview/lib/Tile.ts
amsgames94: Actually, we do need to compute the normals since they're used in the vertex-shader displacement. Readded it.

ninok: hm, I see. Another reason to have custom shader code for the displacement, b/c the normals will all be (0,0,1).
But maybe having per vertex normals is more future proof when thinking about the globe.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1275,6 +1276,37 @@ export class Tile implements CachedResource {
                     materials[techniqueIndex] = material;
                 }
 
+                // Modify the standard textured shader to support height-based coloring.
+                if (
+                    isTerrainTechnique(technique) &&
+                    technique.heightBasedColors !== undefined &&
+                    technique.displacementMap !== undefined
+                ) {
+                    (material as any).onBeforeCompile = (shader: THREE.Shader) => {
```
ninok: It is really ugly that we have the custom shader code in the (supposed to be generic) Tile. Ideally there should be a onMaterialCreated callback in the technique so that we can move the code there (or eventually the technique should create the material). Anyways this is out of scope for this change. Will create a ticket for it.

PR #418, opened 2019-04-30
HARP-5425: Stylable terrain colors.

amsgames94:


amsgames94:
This changes will be integrated in #416 .

PR #419, opened 2019-04-30
MINOR: Fix error message for level 0.

Astrak:


@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
```diff
@@ -290,13 +290,15 @@ export class AnimatedExtrusionTileHandler {
                     quadTreeSearchDistanceUp > nextLevelDiff &&
                     extrusionStartTime === undefined
                 ) {
-                    const parentTileKey = tileKey.parent();
-                    tileHandler = this.m_animatedExtrusionHandler.find([parentTileKey]);
-                    if (tileHandler !== undefined) {
-                        extrusionStartTime = tileHandler.m_animatedExtrusionStartTime;
-                        break;
+                    if (tileKey.level !== 0) {
```
nzjony: would it not make more sense to put it as part of the while condition above?

PR #424, opened 2019-05-02
HARP-5610: Allow calling addDataSource if the datasource is already added.

Astrak:


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1291,9 +1291,15 @@ export class MapView extends THREE.EventDispatcher {
     addDataSource(dataSource: DataSource): Promise<void> {
         const twinDataSource = this.getDataSourceByName(dataSource.name);
         if (twinDataSource !== undefined) {
-            throw new Error(
-                `A DataSource with the name "${dataSource.name}" already exists in this MapView.`
-            );
+            if (twinDataSource === dataSource) {
+                return new Promise(resolve => resolve());
```
ninok: The behavior of the API is a bit weird with that change:

`
const awesomeDataSource = new MyDataSource();
mapView.addDataSource(awesomeDataSource).then(console.log("datasource is connected"));
mapView.addDataSource(awesomeDataSource).then(console.log("datasource is connected"));
`

The second console.log will be printed even though the datasource is not connected. Not sure though if we have any usecase where we actually use the returned promise. Maybe it would be better to not return the promise at all.

Astrak: I means we need to possibly return `undefined`, which means checking if the answer of `addDataSource` is defined everywhere. This would be a breaking change for typescript users.

ninok: I just mean return void (i.e. never return the Promise). I know this is a breaking API change, but I doubt that the API is used that way.
If we don't want to break the API your change will get more complicated b/c you have to remember the promise when the data source is added the first time, so that you can return the same promise, when the data source is added the second time.

Astrak: So should we break the API with `return void` and modify every use of the promise in the code with a check that the returned value is not undefined? 

Or do we implement the empty Promise that won't trigger the `connect`-related promises?

zbigg: @Astrak, @ninok 
I would remove this promise, in same manner as `DataSource.connect` and other `connect` like methods in other interfaces.

MapView usually doesn't care at all if DS is connected it just wants `Tiles` from it.
DataSource DataProvider should just work or throw exceptions or return empty tiles if doesn't have underlying data yet loaded because it's not yet connected.

Handling connect/disconnect stuff is only burden to core library (MapView and core classes) and in my opinion they are just implementation detail of DataSource.

nzjony: why not reject the promise instead? i.e. `return new Promise((resolve, reject) => reject());`

That way:
- only one "datasource is connected" is shown
- user can `catch` such issues

zbigg: @nzjony The main purpose of this PR was to make `addDataSource` idempotent, so you can just call it twice without worrying if it was added before - which is right.

So, if DS is already added method should be no-op

Problem is that proper, async no-op means that it should resolve only after datasource was successfully added - and that in our case means connected and theme loaded.

In this case it would mean memoizing promise ...

For what reason? Who waits for DS to be connected ? That's why i would argue that `connect` methods should be removed from contract and object should be ready from use from beginning.

Now, we basically have multi-stage initialization which just adds dozens of lines in MapView with very minor gain in usability.
Async initialization should be responsibility of DS.

ninok: @zbigg: So this is what I mentioned in the very beginning: Change the method to return nothing i.e. void.

nzjony: @zbigg, 

> So, if DS is already added method should be no-op
> Problem is that proper, async no-op means that it should resolve only after datasource was successfully added - and that in our case means connected and theme loaded.
> In this case it would mean memoizing promise ...

I don't think it tells us much information by resolving twice, the first call will eventually finish when it is ready. What use case do we need to know that the second one resolved? At least by rejecting, we get immediate information that it failed, and we can react accordingly.

I think it still may make sense to wait for the DS To be connected, say for example you are viewing the maps on a phone with bad reception and the connection fails. It may make sense to show some retry button etc? Otherwise you would need to constantly poll the DataSource to see if it is in some connected state, which is ugly (or I miss something).

zbigg: > wait for the DS To be connected

Most actual datasources don't support this actually? Omv datasource in common use case has `Promise.resolve()` in `connect`.

> connected state

Define connected state of DS. In general way.
I would say - it's not business of renderer.
Renderer can let's say report about completness of scene (number of expected/loaded tiles) and then APP may use it to act on this info, but inheriting this information from so arbitrary - underdefined interface like `DataSource.connect` makes little sense.

BTW, pure offline state is rare. Worst case, is "officially" online, but with such a latency/packet-loss that you can't do anything about but detect this basing on response times. For sure it's not something you want to do in renderer project at this state.

(Although loading higher zoom level tiles in case of poor network conditions may be good option for app and we could/should enable it via our API).

nzjony: hi @zbigg, sorry I didn't respond to this. I actually agree, doing the connect in the MapView is conceptually wrong, they should arrive to the MapView pre connected. But that sounds like a bigger change. There should be a way to stop it, even if it isn't fully perfect yet.

PR #431, opened 2019-05-06
HARP-5630: Road Z-Fighting bugfix.

amsgames94:


@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -140,6 +140,11 @@ export interface SolidLineMaterialParameters extends FadingFeatureParameters {
      */
     depthTest?: boolean;
 
+    /**
+     * Enables/Disable depth write.
+     */
+    depthWrite?: boolean;
```
ninok: Why do we need to expose depthWrite? Does not hurt but it is not needed to fix the issue.

PR #435, opened 2019-05-07
Minor: Public resolveUrls.

amsgames94:


@here/harp-mapview/lib/ThemeLoader.ts
```diff
@@ -38,7 +38,13 @@ export class ThemeLoader {
         return this.resolveUrls(theme);
     }
 
-    private static resolveUrls(theme: Theme): Theme {
+    /**
+     * Resolves all [[Theme]]'s relatives URLs to full URL using the document's base URL
```
zbigg: It resolves relative URLs versus `theme.url` not document's base URL.

amsgames94: True, will change.

@here/harp-mapview/lib/ThemeLoader.ts
```diff
@@ -38,7 +38,13 @@ export class ThemeLoader {
...
     }
 
-    private static resolveUrls(theme: Theme): Theme {
+    /**
+     * Resolves all [[Theme]]'s relatives URLs to full URL using the [[Theme]]'s URL
+     * (see: https://www.w3.org/TR/WD-html40-970917/htmlweb.html#h-5.1.2).
+     *
+     * @param theme The [[Theme]] to resolve.
+     */
+    static resolveUrls(theme: Theme): Theme {
```
ninok: Maybe we should remember in the theme if the urls were resolved already, otherwise if you call resolveUrls after loadAsync the urls will be resolved twice.

PR #436, opened 2019-05-07
HARP-5432 Adding ElevationProvider

nzjony:
Draft pull request


@here/harp-mapview/lib/HeightProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Get the displacement map for a given tile key. If the displacement map for the given tileKey
+     * is not in the cache a lower level tile will be returned.
+     * @param tileKey The tile to get the displacement map for.
+     * @returns Returns the DisplacmentMap for the given tileKey or a lower level tile. Undefined
+     *          if the tile or no parent is in the cache.
+     */
+    getDisplacementMap(tileKey: TileKey): DisplacementMap | undefined;
+
+    getTilingSceme(): TilingScheme;
```
nzjony: I'm wondering why this is needed? Could it be part of the DisplacementMap class instead?

@here/harp-mapview/lib/ElevationWrapperThingy.ts
```diff
@@ -0,0 +1,11 @@
+import { HeightProvider } from "@here/harp-mapview/lib/HeightProvider";
+import { DataSource } from "./DataSource";
+import { ElevationRangeSource } from "./ElevationRangeSource";
+
+export class ElevationWrapperThingy {
```
ninok: I think ElevationSource is good enough as name.

@here/harp-mapview/lib/ElevationWrapperThingy.ts
```diff
@@ -0,0 +1,11 @@
...
+    constructor(
+        readonly terrainDataSource: DataSource,
+        readonly heightProvider: HeightProvider,
```
ninok: Maybe we can combine the height provider and the elevation range source. At least we should be consistent and name everything either Elevation... or Height...

@here/harp-mapview/lib/HeightProvider.ts
ninok: Could be, but if the tiling scheme is not matching you want to early out and don't iterate through the tiles at all.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1752,6 +1755,18 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_fog;
     }
 
+    set elevationWrapperThingy(elevationWrapperThingy: ElevationWrapperThingy | undefined) {
```
ninok: I wonder if we even need the ElevationWrapperThingy. We could also just have multiple parameters.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1752,6 +1755,18 @@ export class MapView extends THREE.EventDispatcher {
...
+        this.m_elevationWrapperThingy = elevationWrapperThingy;
```
ninok: If this.m_elevationWrapperThingy was not undefined before we also have to remove the old terrainDataSource.

@here/harp-mapview/lib/MapView.ts
nzjony: Good point, would be one less class.

@here/harp-mapview/lib/MapView.ts
nzjony: Good point, thanks!

@here/harp-mapview/lib/ElevationWrapperThingy.ts
nzjony: Combining them simplifies the API (one thing to construct as opposed to two), however it would require that some methods use the visible tiles (getHeight / raycast) and some make a separate request (ElevationRangeSource), but it should be possible to make this clear in the documentation / function names.

ninok: Okay, we can merge it later if we still want

@here/harp-mapview/index.ts
```diff
@@ -7,35 +7,37 @@
 export * from "./lib/AnimatedExtrusionHandler";
 export * from "./lib/CameraMovementDetector";
 export * from "./lib/ColorCache";
-export * from "./lib/MapViewPoints";
```
nzjony: I just sorted these by ascending order

@here/harp-mapview/lib/MapView.ts
nzjony: Removed.

@here/harp-mapview/lib/ElevationProvider.ts
```diff
@@ -0,0 +1,37 @@
...
+     */
+    getHeight(geoPoint: GeoCoordinates, level?: number): number | undefined;
+
+    /**
+     * Cast a ray through the given screen position
+     * @param screenPoint screenPoint to cast through
+     * @returns GeoCoordinates of the intersection point with the surface or undefined if the
+     *          surface was not hit by the ray.
+     */
+    rayCast(screenPoint: Vector2): GeoCoordinates | undefined;
```
amsgames94: How could this work without providing the camera position (lat, long, el)?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1752,6 +1757,22 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_fog;
     }
 
+    setElevationSource(
+        elevationSource: DataSource,
+        elevationRangeSource: ElevationRangeSource,
```
amsgames94: Instead of passing ``elevationRangeSource`` and ``elevationProvider``, why don't we just create them inside the function body? It seems to me like these 2 would be really connected to elevationSource.

nzjony: both are unfortunately in the terrain-datasource gerrit repo, so we have to construct them at the application level. Unless there is a way to pass in the constructor? But I don't know if this is so necessary.

@here/harp-mapview/lib/ElevationProvider.ts
nzjony: Opps, I forgot to pass in the camera to the constructor of the TerrainElevationProvider, so it will just ask for the camera position and then construct the ray using its position etc.

nzjony: actually, I will pass in the MapView, otherwise I need to pass in the height / width of the canvas also (needed to convert the screen point to world coordinates).

@here/harp-mapview/lib/MapView.ts
nzjony: another option to simplify it could be to have connect method, similar to the ElevationRangeSource, which would then take the MapView, it would save the application having to pass in the MapView, but then internally I would need to deal with an uninitialized ElevationProvider, so it makes it a bit more code internally.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1752,6 +1757,39 @@ export class MapView extends THREE.EventDispatcher {
...
+     * @param elevationProvider Allows access to the elevation at a given location or a ray
+     *      from the camera.
+     */
+    setElevationSource(
+        elevationSource: DataSource,
+        elevationRangeSource: ElevationRangeSource,
+        elevationProvider: ElevationProvider
+    ) {
+        // Try to remove incase this method was already called, will do nothing if it doesn't exist.
+        this.removeDataSource(elevationSource);
```
zbigg: But this will invalidate cache etc ... Better check if it's already added.
Or make `addDataSource` idempotent (better idea IMHO).

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1752,6 +1757,39 @@ export class MapView extends THREE.EventDispatcher {
...
+     * the [[ElevationRangeSource]] and the [[ElevationProvider]] and access would be granted to
+     * the application when it asks for it, to simplify the API.
+     *
+     * @param elevationSource The datasource containing the terrain tiles.
+     * @param elevationRangeSource Allows access to the elevation min / max per tile.
+     * @param elevationProvider Allows access to the elevation at a given location or a ray
+     *      from the camera.
+     */
+    setElevationSource(
+        elevationSource: DataSource,
```
zbigg: 1. I don't understand why you need three instances of objects here.

2. Looks like they are so interconnected internally, that they should be accessed through unified interface.

3. Isn't `elevationSource` just an implementation detail of how latter (`elevationProvider` and `elevationRangeSource`).

4. Naming. It's really hard to grasp difference between `provider` , `source`.

5. No way to unset these.

@here/harp-mapview/lib/MapView.ts
nzjony: Maybe I will wait for: https://github.com/heremaps/harp.gl/pull/424/ ?

@here/harp-mapview/lib/MapView.ts
nzjony: 1. The three instances are doing separate things
- elevationSource is like a DataSource, it has the same interface and shows the mesh rendering
- elevationRangeSource is similar to a DataSource, however it is special, because it controls which requests we can make to the elevationSource (it saves bandwidth to download the bounding box first before we download the tile).
- elevationProvider accesses the heights based on the visible tiles.

Where is your specific misunderstanding?

2. A unified interface sounds good, but there is a high chance of misusing it. Specifically because some parts are async, and some parts are sync. i.e. the getHeight function is meant to be sync, i.e. it can be used to get the value immediately. Splitting like this makes the responsibilities / functionality clear IMO, unless there is some specific example which shows otherwise?

3. elevationSource is the source of data for the terrain tiles, i.e. the meshes

4. appreciate this, if you have a better idea, let me know!!

5. easy to add later if there is a use case which requires this feature.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1839,6 +1844,32 @@ export class MapView extends THREE.EventDispatcher {
...
```
CraigBerlin: Will we ever have an ElevationProvider that is not associated with a DataSource?
If not you could move the DataSource inside the ElevationProvider interface.
You could also drop the getTilingScheme() method as it should be equal to getDataSource().getTilingScheme().

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1839,6 +1844,32 @@ export class MapView extends THREE.EventDispatcher {
...
+     * If the terrain-datasource is merged with this repository, we could internally construct
+     * the [[ElevationRangeSource]] and the [[ElevationProvider]] and access would be granted to
+     * the application when it asks for it, to simplify the API.
+     *
+     * @param elevationSource The datasource containing the terrain tiles.
+     * @param elevationRangeSource Allows access to the elevation min / max per tile.
+     * @param elevationProvider Allows access to the elevation at a given location or a ray
+     *      from the camera.
+     */
+    setElevationSource(
```
CraigBerlin: For uniformity use setElevationProvider() instead?

@here/harp-mapview/lib/MapView.ts
nzjony: Good question, the ElevationProvider uses the in memory terrain mesh and applies the functions raycast / getHeight to that, so it indirectly needs it, but there isn't a tight coupling (see the TerrainElevationProvider on gerrit). My feeling is that it isn't so necassary to combine, but let's discuss in the office and maybe there is something I'm missing.

@here/harp-mapview/lib/MapView.ts
nzjony: As above, lets discuss in person.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1675,6 +1680,20 @@ export class MapView extends THREE.EventDispatcher {
...
+    /**
+     * Returns a ray caster using the supplied screen positions.
+     *
+     * @param x The X position in css/client coordinates (without applied display ratio).
+     * @param y The Y position in css/client coordinates (without applied display ratio).
+     *
+     * @return Raycaster with origin at the camera and direction based on the supplied x / y screen
+     * points.
+     */
+    raycasterFromScreenPoint(x: number, y: number): THREE.Raycaster {
```
ninok: Why do we need this method in the public interface?

@here/harp-mapview/lib/MapView.ts
ninok: @CraigBerlin: We are not 100% sure how we want to render the terrain if we have vector data. In harp-native we still rendered the terrain mesh and then the vector data on top, but instead we could also just render the elevated vector data (given it covers the whole tile) so the actual terrain data source would not be used.

@here/harp-mapview/lib/MapView.ts
nzjony: It is needed for the current implementation of the TerrainElevationProvider, because it calls raycaster.intersectObject on the TerrainTile mesh, however, when this implementation goes away, I can remove it.

ninok: If this is temporary, please mark it as such (i.e. `@deprecated` or `@alpha`).
Actually we should mark the whole terrain API as `@alpha`. And the rest of the API as `@beta`.
See: https://api-extractor.com/pages/tsdoc/tag_alpha/

PR #437, opened 2019-05-07
MINOR: Fixed wrong geometry for extruded polygons.

ninok:
Signed-off-by: ninok <Nino Kettlitz 1396039+ninok@users.noreply.github.com>

---
This change is [Reviewable](https://reviewable.io/reviews/heremaps/harp.gl/437)

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
+++ b/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
@@ -556,10 +556,7 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
                 isDashedLineTechnique(technique) ||
                 isLineTechnique(technique);
             const isPolygon =
-                isExtrudedPolygonTechnique ||
-                isFillTechnique(technique) ||
-                isStandardTechnique(technique) ||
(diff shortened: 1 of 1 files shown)


@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -710,15 +710,17 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
         featureId: number | undefined,
         env: MapEnv
     ): void {
-        const meshBuffers = this.findOrCreateMeshBuffers(techniqueIndex, GeometryType.Polygon);
+        const isExtruded = isExtrudedPolygonTechnique(technique);
+
+        const geometryType = isExtruded ? GeometryType.ExtrudedPolygon : GeometryType.Polygon;
+        const meshBuffers = this.findOrCreateMeshBuffers(techniqueIndex, geometryType);
```
nzjony: There seems to be a bug at line 541 as well, which should be fixed too.

nzjony: Also, why does the method applyPolygonTechnique extract the techniqueIndex from the technique and pass it in separately.

PR #439, opened 2019-05-08
Javascript friendly bundle

zbigg:
* new module - `@here/harp.gl` which provides crucial APIs under  `window.harp`
* WorkerLoader - support for worker scripts bootstrap protocol
* remove Vector3Like duplicate definition from Theme.ts

Note, see `@here/harp-examples/src/hello_js-bundle.html` for sample usage.

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
+++ b/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
@@ -43,7 +43,8 @@ import {
     StyleSetEvaluator,
     Value
 } from "@here/harp-datasource-protocol/index-decoder";
-import { LineGroup, triangulateLine } from "@here/harp-lines";
+import { LineGroup } from "@here/harp-lines/lib/Lines";
+import { triangulateLine } from "@here/harp-lines/lib/TriangulateLines";
 import { assert, LoggerManager, Math2D } from "@here/harp-utils";
 import earcut from "earcut";
 import * as THREE from "three";
--- a/@here/harp.gl/lib/index.ts
+++ b/@here/harp.gl/lib/index.ts
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { mapBundleMain } from "./BundleMain";
+
+if (!(window as any).THREE) {
+    // tslint:disable-next-line:no-console
+    console.warn(
+        "harp.js: It looks like 'three.js' is not loaded. This script requires 'THREE' object to " +
+            "be defined. See https://github.com/heremaps/harp.gl/@here/harp.gl."
+    );
+}
+
+export {
+    ConcurrentDecoderFacade,
+    ConcurrentTilerFacade,
+    CopyrightElementHandler,
+    DataSource,
+    MapView,
+    MapViewEventNames,
+    MapViewOptions,
+    Tile,
+    ThemeLoader,
+    WorkerLoader
+} from "@here/harp-mapview";
+export * from "@here/harp-omv-datasource";
+export * from "@here/harp-geojson-datasource";
+export * from "@here/harp-webtile-datasource";
+export * from "@here/harp-map-controls/lib/MapControls";
+export * from "@here/harp-map-controls/lib/MapControlsUI";
+export * from "@here/harp-datasource-protocol";
+export * from "@here/harp-geoutils";
+export * from "@here/harp-mapview-decoder";
+
+mapBundleMain();
--- a/@here/harp.gl/package.json
+++ b/@here/harp.gl/package.json
@@ -1,0 +1,53 @@
+{
+    "name": "@here/harp.gl",
+    "version": "0.1.0",
+    "description": "JS Bundle containing all the functionality needed to render a map with harp.gl",
+    "keywords": [
+        "map",
+        "map-renderer",
+        "vector-tiles",
+        "geojson",
+        "typescript",
+        "webgl",
+        "harpgl"
+    ],
+    "main": "dist/harp.js",
+    "scripts": {
+        "prepare": "webpack && NODE_ENV=production webpack",
+        "profile-min": "NODE_ENV=production webpack --profile --json > webpack-stats-production.json",
+        "profile-dev": "webpack --profile --json > webpack-stats-dev.json",
+        "test": "true"
+    },
+    "repository": {
+        "type": "git",
+        "url": "git://github.com:heremaps/harp.gl.git"
+    },
+    "author": {
+        "name": "HERE Europe B.V.",
+        "url": "https://here.com"
+    },
+    "license": "Apache-2.0",
+    "dependencies": {
+        "three": "^0.104.0"
+    },
+    "devDependencies": {
+        "@here/harp-mapview": "^0.8.1",
+        "@here/harp-geoutils": "^0.3.1",
+        "@here/harp-omv-datasource": "^0.3.1",
+        "@here/harp-geojson-datasource": "^0.3.1",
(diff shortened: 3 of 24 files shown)


@here/harp.gl/lib/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    MapView,
+    DataSource,
+    Tile,
+    MapViewEventNames,
+    MapViewOptions,
+    CopyrightElementHandler
+} from "@here/harp-mapview";
+export * from "@here/harp-omv-datasource";
+export * from "@here/harp-geojson-datasource";
+export * from "@here/harp-webtile-datasource";
```
haraldF: This looks quite arbitrary :/ Can we split it a bit more? E.g. have the datasources as plugins that just add to the global `harp` namespace?

Kinda like momentjs.com does with timezones and holidays, e.g. https://momentjs.com/timezone/ and https://www.npmjs.com/package/moment-holiday

zbigg: This is how our API looks like _now_.
Moreover, we have _no_ chance to create pluggable system without turning into SystemJS which is complex and/or modules which are not yet ready for wide audienc.

> have the datasources as plugins 

Which? ATM we have actually one datasource i.e omv-datasource which is our _de_facto_ only one real, supported and featured datasource.
* `WebTileDatasource` is only technology preview.
* `GeoJson...` is being deprecated as OmvDataSource will support geojson.

Part of this story is "make it simple". And simple means 1 or max 2 scripts (including 3JD)
If we will end with
```
<script src="https://unpkg.com/@here/harp.gl-core"></script>
<script src="https://unpkg.com/@here/harp.gl-core"></script>

`moment-holiday.js` works because `moment` by default installs itself in global namespace.
We can have `harp` as root - but what will it contain ?
`MapView` - ok

* `harp.controls` as alias for stuff in `@here/harp-map-controls`
* `harp.omvDatasource` or harp.OmvDatasource` and thus `harp.OmvDatasource.OmvDataSource` ?

Our current module system is and naming is so UMD non-friendly, that it cannot be fixed in scope of such small task. Actually, it's not modular at all.

Comment (zbigg) [reply @here/harp.gl/lib/index.ts]: This is how our API looks like _now_.
Moreover, we have _no_ chance to create pluggable system without turning into SystemJS which is complex and/or modules which are not yet ready for wide audienc.

> have the datasources as plugins 

Which? ATM we have actually one datasource i.e `omv-datasource` which is our _de_facto_ only one real, supported and featured datasource.
* `WebTileDatasource` is only technology preview.
* `GeoJson...` is being deprecated as OmvDataSource will support geojson.

Part of this story is "make it simple". And simple means 1 or max 2 scripts (including 3JS)
If we will end with
```
<script src="https://unpkg.com/@here/harp.gl-core"></script>
<script src="https://unpkg.com/@here/harp.gl-omv-datawhatever"></script>
<script src="https://unpkg.com/@here/harp.gl-map-controls"></script>
```
this is far from being simple.

Going further, 
`moment-holiday.js` works because `moment` is well designed, Javascript first API which is designed to be UMD/IIFE compatible.
It  installs itself in global namespace since always and by default.

Our current module system is and naming is so UMD non-friendly, that it cannot be fixed in scope of such small task. Actually, it's not modular at all.

Comment (haraldF) [reply @here/harp.gl/lib/index.ts]: I'd argue that all 'layers' are plugins:

```
<script src="https://unpkg.com/@here/harp.gl-core"></script>
<script src="https://unpkg.com/@here/harp.gl-omv-datawhatever"></script>
```

This would mean two scripts, and still give us the flexibility to add other layers on top.

Comment (zbigg) [reply @here/harp.gl/lib/index.ts]: > UMD ...

It's all about UMD or not UMD.
Some time ago we dropped  UMD as not supported.
UMD works if we're can import everything efficiently from each-other using `index.ts` or `index-decoder.ts`  etc.

That means, no imports like this:
```
import { LineGroup } from "@here/harp-lines/lib/Lines";
```
Only this:
```
import { LineGroup } from "@here/harp-lines";
```
But then, `decoder` will get be bundled with whole `lines` module including it's deps like materials and TREE.js.

First thing that we have to do is a proper module division, that is built around key concept - don't leak code between UI thread and decoders.

File: @here/harp.gl/package.json [hunk 2]
```diff
@@ -0,0 +1,52 @@
+{
+    "name": "@here/harp.gl",
+    "version": "0.1.0",
+    "description": "Functionality needed to render a map.",
```
musculman: May I suggest using a different description? "JS Bundle containing all the functionality needed to render a map with harp.gl"

zbigg: Thanks for spotting and suggestion
This was copy-pasta.

@here/harp.gl/lib/index.ts
zbigg: > harp.gl-omv-datawhatever

Actually, i'd like to name it `harp.gl-omv-tile-provider`.

Problem with this particular is that `OmvDataSource` and friends which are loaded in main thread is like "5 kilobytes" of pure wrapper over `TileDataSource` (`OmvDataSource is basically `TileDataSource`) and basic `fetch`-based `OmvRestSomethingProvider`.

Same applies to `WebTileDataSource` - 1-2k minified, no more.
So datasources don't weight much in `main bundle` of `harp.js`

The main part that weights is - decoders and specifically, `OmvDecoder` and in future, TerrainDecoder whose loads tons of 3rdparty-libs. 
Ideally, `decoders` script should be split also so user has another bundle/script for all the services in workers:

 * harp-omv-decoder-worker.js for OmvDecoderService
 * harp-tiler-worker.js for TilerService
 * harp-terrain-decoder-worker.js for TerrainDecoderService

Now, each of them requires (usually only math! :/) part of 3JS and tons of common code, so we should avoid creating 4 * eachService workers, because we would end with at least 12 workers in common case (4 for omv, 4 for tiler, 4 for terrain).

So, we wan't to load all the services in each worker (to reuse JS runtimes and memory), ...
... worker scripts should be "UMD/AMD"-like so we can load them using `requirejs`/systemd like semantics.
That leaves us with following "architecture":
 * `three.js`
 * `harp-worker-core.js` (depends on 3js probably)
 * `harp-omv-decoder-worker.js` (depends on 3js, proto,  harp-worker-core.js)
 * `harp-tiler-worker.js` (depends on 3js,  harp-worker-core.js and some custom stuff)
 * `harp-terrain-decoder-worker.js` (depends on 3js, proto,  harp-worker-core.js)

That whole issue is can of worms literally ...

@here/harp.gl/package.json
```diff
@@ -0,0 +1,54 @@
...
+        "type": "git",
+        "url": "git://github.com:heremaps/harp.gl.git"
+    },
+    "author": {
+        "name": "HERE Europe B.V.",
+        "url": "https://here.com"
+    },
+    "license": "Apache-2.0",
+    "dependencies": {
+        "three": "^0.102.0"
```
ninok: 104

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -42,7 +42,8 @@ import {
     StyleSetEvaluator,
     Value
 } from "@here/harp-datasource-protocol/index-decoder";
-import { LineGroup, triangulateLine } from "@here/harp-lines";
+import { LineGroup } from "@here/harp-lines/lib/Lines";
```
ninok: Why not including from index.ts anymore?

zbigg: Because `@here/harp-lines` index.ts depends on whole bunch of stuff required only in main thread.
Without this, all the materials and possibly whole MapView and friends land in decoder bundle.

ninok: I see. This basically boils down to your previous statement to have a module split based on main thread vs. web worker instead of the arbitrary split we have right now.

PR #445, opened 2019-05-10
MINOR: Fixed broken geometry in the extruded-line technique.

ninok:
Also added missing generation of normals.

### DIFF
--- a/@here/harp-lines/lib/TriangulateLines.ts
+++ b/@here/harp-lines/lib/TriangulateLines.ts
@@ -61,7 +61,7 @@ export function numCirclePoints(lineWidth: number): number {
 /**
  * Create a triangle mesh from the given polyline.
  *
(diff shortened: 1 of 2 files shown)


@here/harp-lines/lib/TriangulateLines.ts
```diff
@@ -95,59 +95,60 @@ export function triangulateLine(
...
+    const n = new THREE.Vector3(); // next point
+    const bt = new THREE.Vector3();
+    const averageBt = new THREE.Vector3();
     const p0 = new THREE.Vector3();
     const p1 = new THREE.Vector3();
     const p2 = new THREE.Vector3();
     const p3 = new THREE.Vector3();
 
-    const N = points.length / 2;
+    const N = points.length / 3;
```
ninok: That was the actual bug. All other changes are just renaming to make the code more readable.

PR #446, opened 2019-05-10
HARP-5679: More textures + introduce posteffects.

Astrak:
This change is [Reviewable](https://reviewable.io/reviews/heremaps/harp.gl/446)

### DIFF
--- a/@here/harp-examples/src/hello.ts
+++ b/@here/harp-examples/src/hello.ts
@@ -59,9 +59,9 @@ import { accessToken } from "../config";
  */
 export namespace HelloWorldExample {
     // Create a new MapView for the HTMLCanvasElement of the given id.
-    function initializeMapView(id: string): MapView {
+    function initializeMapView(): MapView {
         // snippet:harp_gl_hello_world_example_0.ts
-        const canvas = document.getElementById(id) as HTMLCanvasElement;
+        const canvas = document.getElementById("mapCanvas") as HTMLCanvasElement;
         // end:harp_gl_hello_world_example_0.ts
 
         // snippet:harp_gl_hello_world_example_1.ts
@@ -100,7 +100,7 @@ export namespace HelloWorldExample {
         return map;
     }
 
-    const mapView = initializeMapView("mapCanvas");
+    const mapView = initializeMapView();
 
     const hereCopyrightInfo: CopyrightInfo = {
         id: "here.com",
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -3,7 +3,7 @@
  * Licensed under Apache 2.0, see full license in LICENSE
  * SPDX-License-Identifier: Apache-2.0
  */
-import { ImageTexture, Light, Sky, Theme } from "@here/harp-datasource-protocol";
+import { ImageTexture, Light, PostEffects, Sky, Theme } from "@here/harp-datasource-protocol";
 import {
     EarthConstants,
     GeoCoordinates,
@@ -501,11 +501,18 @@ export class MapView extends THREE.EventDispatcher {
     dumpNext = false;
 
     /**
+     * Allows discarding the text rendering after the map rendering.
+     */
+    renderLabels: boolean = true;
+
+    /**
      * The instance of [[MapRenderingManager]] managing the rendering of the map. It is a public
      * property to allow access and modification of some parameters of the rendering process at
      * runtime.
      */
     readonly mapRenderingManager: IMapRenderingManager;
+
+    private m_postEffects?: PostEffects;
 
     private m_createdLights?: THREE.Light[];
     private m_skyBackground?: SkyBackground;
@@ -924,6 +931,32 @@ export class MapView extends THREE.EventDispatcher {
     set pointOfView(pointOfView: THREE.PerspectiveCamera | undefined) {
         this.m_pointOfView = pointOfView;
         this.update();
+    }
+
+    /**
+     * Loads a posteffects file.
+     *
+     * @param postEffectsFile File URL describing the post effects.
+     */
+    loadPostEffects(postEffectsFile: string) {
+        fetch(postEffectsFile)
+            .then(response => response.json())
+            .then((postEffects: PostEffects) => {
+                this.m_postEffects = postEffects;
+                this.setPostEffects();
+            });
+    }
+
+    /**
+     * The abstraction of the [[MapRenderingManager]] API for posteffects.
+     */
+    get postEffects(): PostEffects | undefined {
+        return this.m_postEffects;
+    }
+
+    set postEffects(postEffects: PostEffects | undefined) {
+        this.m_postEffects = postEffects;
+        this.setPostEffects();
     }
 
     /**
@@ -1813,6 +1846,30 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_fog;
     }
 
+    private setPostEffects() {
+        // First clear all the effects, then enable them from what is specified.
+        this.mapRenderingManager.bloom.enabled = false;
+        this.mapRenderingManager.outline.enabled = false;
+        this.mapRenderingManager.vignette.enabled = false;
+        this.mapRenderingManager.sepia.enabled = false;
+
+        if (this.m_postEffects !== undefined) {
+            if (this.m_postEffects.bloom !== undefined) {
+                this.mapRenderingManager.bloom = this.m_postEffects.bloom;
+            }
+            if (this.m_postEffects.outline !== undefined) {
+                this.mapRenderingManager.outline.enabled = this.m_postEffects.outline.enabled;
+                this.mapRenderingManager.updateOutline(this.m_postEffects.outline);
+            }
+            if (this.m_postEffects.vignette !== undefined) {
+                this.mapRenderingManager.vignette = this.m_postEffects.vignette;
+            }
+            if (this.m_postEffects.sepia !== undefined) {
+                this.mapRenderingManager.sepia = this.m_postEffects.sepia;
+            }
+        }
+    }
+
     /**
      * Updates the camera and the projections and resets the screen collisions,
      * note, setupCamera must be called before this is called.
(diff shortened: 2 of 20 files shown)


@here/harp-examples/src/effects.ts
```diff
@@ -0,0 +1,209 @@
...
+            offset: 1.0,
+            darkness: 1.0
+        },
+        sepia: {
+            enabled: true,
+            amount: 0.55
+        }
+    };
+
+    mapView.addEventListener(MapViewEventNames.ThemeLoaded, () => {
```
Astrak: Explain that the effects being also possibly defined from the theme, they are reset while setting it, so we first need to wait for this event and apply these defaults after.

@here/harp-examples/src/effects.ts
```diff
@@ -0,0 +1,209 @@
...
+        mapView.mapRenderingManager.outline.enabled = options.outline.enabled;
+        mapView.mapRenderingManager.updateOutline(options.outline);
+        mapView.mapRenderingManager.bloom.enabled = options.bloom.enabled;
+        mapView.mapRenderingManager.bloom = options.bloom;
+        mapView.mapRenderingManager.vignette = options.vignette;
+        mapView.mapRenderingManager.sepia = options.sepia;
+        mapView.update();
+    });
+
+    gui.add(options, "labels").onChange((value: boolean) => {
```
Astrak: What about one function that changes all the effects from the `options` object and is reused everywhere, from the theme-loaded callback to the gui callbacks, it would clean the code.

@here/harp-examples/src/hello.ts
```diff
@@ -124,4 +125,26 @@ export namespace HelloWorldExample {
...
+        theme: {
+            day: "resources/berlin_tilezen_base.json",
+            reducedDay: "resources/berlin_tilezen_day_reduced.json",
+            reducedNight: "resources/berlin_tilezen_night_reduced.json",
+            streets: "resources/berlin_tilezen_effects_streets.json",
+            outlines: "resources/berlin_tilezen_effects_outlines.json",
+            textured: "resources/berlin_tilezen_textured.json"
+        }
+    };
+    gui.add(options, "theme", options.theme).onChange((value: string) => {
```
Astrak: The dropdown list is initially empty. Find how to display the initial theme name as the active one in the dropdown list.

Astrak: Also add the new themes in mapview-demo

@here/harp-mapview/lib/MapView.ts
```diff
@@ -490,6 +490,11 @@ export class MapView extends THREE.EventDispatcher {
      */
     defaultFontCatalog: string = DEFAULT_FONT_CATALOG;
 
+    /**
+     * Allows discarding the text rendering after the map rendering.
+     */
+    renderLabels: boolean = true;
```
Astrak: @zbigg @robertoraggi @amsgames94 I found it a convenient functionnality to only show the map in the effects example, it allows to not switch all the text techniques to `none` or flatout remove them. 

But it introduces a new API :sweat_smile: . Opinions?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -911,6 +916,32 @@ export class MapView extends THREE.EventDispatcher {
         this.m_theme.clearColor = theme.clearColor;
         this.renderer.setClearColor(new THREE.Color(theme.clearColor));
 
+        const effects = theme.effects;
+        if (effects !== undefined) {
+            if (effects.toneMappingExposure !== undefined) {
```
Astrak: Todo: also handle individual effects being undefined, if effects are not undefined.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2125,9 +2158,9 @@ export class MapView extends THREE.EventDispatcher {
                 : Promise.resolve<Theme>(this.m_options.theme);
 
         themePromise.then((theme: Theme) => {
+            this.theme = theme;
```
Astrak: @zbigg @StefanDachwitz In the effects example, I want to use the `mapRenderingManager`'s API to apply effects in the code. But the theme also defines (or does not define) effects, so the API calls need to happen after the theme is loaded, but also after the theme is _applied_, hence this change. It seemed to me that the `ThemeLoaded` event is only used in `datasources/places` and that it does not break, so here is the change.

Thoughts?

@here/harp-mapview/lib/composing/MapRenderingManager.ts
```diff
@@ -46,6 +50,37 @@ export interface IMapAntialiasSettings {
  * to modify some of the rendering processes like the antialiasing behaviour at runtime.
  */
 export interface IMapRenderingManager extends IPassManager {
+    /**
+     * Bloom effect parameters.
+     */
+    bloom: {
+        enabled: boolean;
```
Astrak: Create interface for individual effects and add proper doc about what it does.

@here/harp-mapview/lib/composing/MapRenderingManager.ts
```diff
@@ -93,18 +128,71 @@ export interface IMapRenderingManager extends IPassManager {
...
+        ghostExtrudedPolygons: boolean;
+    }): void;
 }
 
 /**
  * The implementation of [[IMapRenderingManager]] to instantiate in [[MapView]] and manage the map
  * rendering.
  */
 export class MapRenderingManager implements IMapRenderingManager {
+    bloom = {
```
Astrak: Use TS interface for the effects

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -51,6 +51,34 @@ export interface Theme {
      */
     lights?: Light[];
 
+    /**
+     * Post-effects parameters.
+     */
+    effects?: {
+        toneMappingExposure?: number;
+        bloom?: {
```
Astrak: use TS interface for the effects. Make it common with `MapRenderingManager`.

@here/harp-mapview/lib/composing/Pass.ts
```diff
@@ -84,3 +84,78 @@ export class Pass implements IPass {
...
+        scene: THREE.Scene,
+        camera: THREE.Camera,
+        writeBuffer: THREE.WebGLRenderTarget | undefined,
+        readBuffer: THREE.WebGLRenderTarget | undefined
+    ) {
+        renderer.setRenderTarget(this.renderToScreen ? null! : writeBuffer);
+        renderer.render(scene, camera);
+    }
+}
+
```
Astrak: doc

@here/harp-mapview/lib/composing/UnrealBloomPass.ts
```diff
@@ -0,0 +1,358 @@
...
+ */
+
+import { CopyShader, LuminosityHighPassShader } from "@here/harp-materials";
+import * as THREE from "three";
+import { Pass } from "./Pass";
+
+const BlurDirectionX = new THREE.Vector2(1.0, 0.0);
+const BlurDirectionY = new THREE.Vector2(0.0, 1.0);
+
+/**
```
Astrak: no

@here/harp-materials/lib/LuminosityHighPassShader.ts
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as THREE from "three";
+
+export const LuminosityHighPassShader: THREE.Shader = {
```
Astrak: doc

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -199,7 +227,7 @@ export type Style =
     | StandardExtrudedLineStyle
     | ExtrudedPolygonStyle
     | ShaderStyle
-    | TextTechniqueStyle
+    | TextStyle
```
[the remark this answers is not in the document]
ninok: no, this was done by @zbigg when adding the json schema for the styles

@here/harp-examples/src/hello.ts
Astrak: not possible.

@here/harp-map-theme/theme.schema.json
```diff
@@ -0,0 +1,4246 @@
+{
```
zbigg: Why commiting this, it's autogenerated in `datasource-protocol` as for now ?

@here/harp-mapview/lib/MapView.ts
zbigg: Yes and no. Bigger API ;) ?

Btw, You added new theme files, how do they differ from `base` ? Is it only effects ?

If yes, i would propose this solution:
```javascript
const theme = await ThemeLoader.loadAsync("...");
theme.effects = {
    ...
}
const newTheme = new ThemeVisitor(theme).visitStyles(style => {
    if (style.technique === "text") {
        style.technique = "none";
    }
})
const mapView = new MapView({theme: newTheme, ...)
```

Yes, this is code, but this shows clearly -  in example. How to add `effects` to map based on any style.
(Comment valid if we're customizing basic things).

@here/harp-mapview/lib/MapView.ts
zbigg: I think it's actually ok. Previous behavior was kind'a bug, because with event we stated that `theme (was) loaded` but in reality it was not yet applied.

@here/harp-mapview/lib/MapView.ts
Astrak: The new theme files (streets and outlines) should be quite different themes yes.

ninok:
> Why this patch copies and hardcodes and rewrites (selected) sample of post-processing-effects.
> instead of:
> 
> * allowing us to add any post-processing like in #299 or just by supporting "real" `IPass` interface ?
> * and thus reuse effect from rich library delivered by community ?
> 
> Do we really want to maintain this code (bloom, vignette etc ...) ?
> We've just discussed about ever-growing scope of project and size of bundles etc ...

I totally agree with @zbigg here. The rendering manager should be much more generic.

Ideally all the effects should be added by the application but that contradicts with putting the configuration into the theme file.
A bit hybrid aproach would be to still put the configuration of the effects in the theme file for designers convenience, but the effects will only work in applications that add these effects (e.g. your example and maybe map-view demo). If we do this we can also move the effects to a separate module.

Astrak:
@heremaps-bot Retest this please

musculman:
Looking forward to have this in!

PR #450, opened 2019-05-13
HARP-5699 Add simple object demo

nzjony:
Demo which shows how to add a THREE.js object to the scene and position using RTC (relative to center).

---
This change is [Reviewable](https://reviewable.io/reviews/heremaps/harp.gl/450)

### DIFF
--- a/@here/harp-examples/README.md
+++ b/@here/harp-examples/README.md
@@ -23,6 +23,7 @@ We can group the examples by several categories, looking at what kind of feature
 ### Getting Started
 
 1. [The hello world example](src/hello.ts) that displays a map in an HTML element with our default map style and default data source.
+2. [The hello add simple object example](src/hello_add_simple_object.ts) that shows how to add a THREE.js object to the map.
 
 ### Camera
 
--- a/@here/harp-examples/src/threejs_add-simple-object.ts
+++ b/@here/harp-examples/src/threejs_add-simple-object.ts
@@ -1,0 +1,163 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { GeoCoordinates } from "@here/harp-geoutils";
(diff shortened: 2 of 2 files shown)


@here/harp-examples/src/hello_add_simple_object.ts
```diff
@@ -0,0 +1,161 @@
...
+            // end:harp_gl_hello_add_simple_object_2.ts
+
+            // snippet:harp_gl_hello_add_simple_object_3.ts
+            // Add a callback to execute before the items are rendered.
+            map.addEventListener(MapViewEventNames.Render, () => {
+                // Set the cube position relative to the world center. Note, we don't subtract the
+                // [[worldCenter]] from the worldMousePosition, because we need to keep the cubes
+                // world position untouched.
+                cube.position.copy(worldPositionAtMouse);
+                cube.position.sub(map.worldCenter);
```
Astrak: you can chain `copy().sub()`

nzjony: thanks, awesome tip!

musculman:
Why would you call it `hello_add_simple_object`? Wouldn't it be easier to have it as `add_threejs_simple_object`?
I believe that `hello_world` is probably fitting the first and hopefully easiest example we have, but once past that, there is no need to keep the `hello` suffix. Or what was the idea behind?

nzjony:
@musculman, good question. I also discussed this with @ninok, we thought that because it is the equivalent of a hello world example with threejs, that it is probably ok, however true, it might be more appropriate to name it based on the fact that it is showing how to use threejs with the map.

I want to add an example showing raycasting, and we decided to call it threejs_raycast, how about calling this threejs_add_simple_object?

ninok:
> I want to add an example showing raycasting, and we decided to call it threejs_raycast, how about calling this threejs_add_simple_object?
maybe call it threejs_addSimpleObject otherwise you get a lot of "subfolders"

nzjony:
> threejs_addSimpleObject

fail enough regarding the sub folders, the convention, at least with the other example is to use the - to separate the words

PR #453, opened 2019-05-14
Add support for globe projection to MapView.

robertoraggi:

### DIFF
--- a/@here/harp-examples/src/globe.ts
+++ b/@here/harp-examples/src/globe.ts
@@ -1,0 +1,174 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { GeoCoordinates, sphereProjection } from "@here/harp-geoutils";
+import { CopyrightElementHandler, CopyrightInfo, MapView } from "@here/harp-mapview";
+import { APIFormat, OmvDataSource } from "@here/harp-omv-datasource";
+import { accessToken } from "../config";
+
+import * as THREE from "three";
+
+export namespace GlobeExample {
+    class GlobeControls extends THREE.EventDispatcher {
+        private readonly m_mouseDownListener = this.mouseDown.bind(this);
+        private readonly m_mouseWheelListener = this.mouseWheel.bind(this);
+        private readonly m_mouseUpListener = this.mouseUp.bind(this);
+        private readonly m_mouseMoveListener = this.mouseMove.bind(this);
+        private readonly m_cameraMatrix = new THREE.Matrix4();
+        private readonly m_transform = new THREE.Matrix4();
+        private readonly m_quaternion = new THREE.Quaternion();
+
+        private m_enabled = false;
+
+        constructor(readonly mapView: MapView) {
+            super();
+
+            this.enabled = true;
+        }
+
+        get camera() {
+            return this.mapView.camera;
+        }
+
+        get domElement() {
+            return this.mapView.canvas;
+        }
+
+        get projection() {
+            return this.mapView.projection;
+        }
+
+        get enabled() {
+            return this.m_enabled;
+        }
+
+        set enabled(enabled: boolean) {
+            if (this.m_enabled === enabled) {
+                return;
+            }
+
+            this.m_enabled = enabled;
+
+            if (enabled) {
+                this.domElement.addEventListener("mousedown", this.m_mouseDownListener, false);
+                this.domElement.addEventListener("wheel", this.m_mouseWheelListener, false);
+            } else {
+                this.domElement.removeEventListener("mousedown", this.m_mouseDownListener, false);
+                this.domElement.removeEventListener("wheel", this.m_mouseWheelListener, false);
+            }
+        }
+
+        private updateCameraMatrix() {
+            this.camera.updateMatrixWorld(true);
+
+            this.m_cameraMatrix.multiplyMatrices(
+                this.camera.projectionMatrix,
+                this.camera.matrixWorldInverse
+            );
+        }
+
+        private mouseDown(event: MouseEvent) {
+            event.preventDefault();
+            event.stopPropagation();
+            this.domElement.addEventListener("mouseup", this.m_mouseUpListener, false);
+            this.domElement.addEventListener("mousemove", this.m_mouseMoveListener, false);
+            this.mapView.beginAnimation();
+            this.updateCameraMatrix();
+        }
+
+        private mouseUp(event: MouseEvent) {
+            event.preventDefault();
+            event.stopPropagation();
+            this.domElement.removeEventListener("mouseup", this.m_mouseUpListener, false);
+            this.domElement.removeEventListener("mousemove", this.m_mouseMoveListener, false);
+            this.mapView.endAnimation();
+        }
+
+        private mouseMove(event: MouseEvent) {
+            event.preventDefault();
+            event.stopPropagation();
+            const current = this.mapView.getWorldPositionAt(event.clientX, event.clientY);
+            const previous = this.mapView.getWorldPositionAt(
+                event.clientX - event.movementX / window.devicePixelRatio,
+                event.clientY - event.movementY / window.devicePixelRatio
+            );
+            if (current && previous) {
+                this.m_quaternion.setFromUnitVectors(previous.normalize(), current.normalize());
+                this.m_transform.copyPosition(this.camera.matrix);
+                this.m_quaternion.inverse();
+                this.m_transform.makeRotationFromQuaternion(this.m_quaternion);
+                this.camera.applyMatrix(this.m_transform);
+                this.updateCameraMatrix();
+            }
+        }
+
+        private mouseWheel(event: WheelEvent) {
+            event.preventDefault();
+            event.stopPropagation();
+            const surfaceNormal = this.projection.surfaceNormal(
+                this.camera.position,
+                new THREE.Vector3()
+            );
+            const L = this.projection.groundDistance(this.camera.position);
+            this.camera.position.addScaledVector(surfaceNormal, event.deltaY * L * 0.001);
+            this.updateCameraMatrix();
+            this.mapView.update();
+        }
+    }
+
+    // Create a new MapView for the HTMLCanvasElement of the given id.
+    function initializeMapView(id: string): MapView {
+        const canvas = document.getElementById(id) as HTMLCanvasElement;
+
+        const mapView = new MapView({
+            canvas,
+            theme: "resources/berlin_tilezen_base.json",
+            projection: sphereProjection,
+            maxVisibleDataSourceTiles: 400
+        });
+
+        CopyrightElementHandler.install("copyrightNotice", mapView);
+
+        mapView.resize(window.innerWidth, window.innerHeight);
+
+        window.addEventListener("resize", () => {
+            mapView.resize(window.innerWidth, window.innerHeight);
+        });
+
+        return mapView;
+    }
+
+    function main() {
+        const map = initializeMapView("mapCanvas");
+
+        const hereCopyrightInfo: CopyrightInfo = {
+            id: "here.com",
+            year: new Date().getFullYear(),
+            label: "HERE",
+            link: "https://legal.here.com/terms"
+        };
+
+        const copyrights: CopyrightInfo[] = [hereCopyrightInfo];
+
+        const omvDataSource = new OmvDataSource({
+            baseUrl: "https://xyz.api.here.com/tiles/herebase.02",
+            apiFormat: APIFormat.XYZOMV,
+            styleSetName: "tilezen",
+            maxZoomLevel: 17,
+            authenticationCode: accessToken,
+            copyrightInfo: copyrights
+        });
+
+        map.addDataSource(omvDataSource);
+
+        const controls = new GlobeControls(map);
+        controls.enabled = true;
+
+        map.setCameraGeolocationAndZoom(new GeoCoordinates(40.6935, -74.009), 4);
+    }
+
+    main();
+}
--- a/@here/harp-geoutils/lib/projection/Projection.ts
+++ b/@here/harp-geoutils/lib/projection/Projection.ts
@@ -134,6 +134,21 @@ export interface Projection {
     getScaleFactor(worldPoint: Vector3Like): number;
 
     /**
+     * Returns the surface normal at the given world position.
+     *
+     * @param worldPoint The position in world coordinates.
+     */
+    surfaceNormal(worldPoint: Vector3Like): Vector3Like;
+
+    /**
+     * Returns the surface normal at the given world position.
+     *
+     * @param worldPoint The position in world coordinates.
+     * @returns The resulting normal vector.
+     */
+    surfaceNormal<Normal extends Vector3Like>(worldPoint: Vector3Like, result: Normal): Normal;
+
+    /**
      * Returns the signed distance between the given coordinates and
      * the closest point on the surface.
      *
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -141,16 +141,6 @@ const COPYRIGHT_CHANGED_EVENT: RenderEvent = { type: MapViewEventNames.Copyright
 const tmpVector = new THREE.Vector2();
 
 /**
- * Compute visibility for the text elements. Called every frame.
- *
- * @param mapView The current [[MapView]] instance.
- * @param tilt Angle in degrees between the vertical vector (eye to ground) and view vector. The
- * value of `0` is looking straight down. `90` is the "camera on the floor". `>90` is camera looking
- * up.
- */
-export type LabelVisibilityEvaluator = (mapView: MapView, tilt: number) => boolean;
-
-/**
  * Compute far plane distance. May be based on tilt. Is being called every frame.
  *
  * @param mapView The current [[MapView]] instance.
@@ -283,16 +273,6 @@ export interface MapViewOptions {
      * The maximum zoom level. The default is `14`.
      */
     maxZoomLevel?: number;
-
-    /**
-     * Maximum tilt where the labels are visible. Default is `65`.
-     */
-    maxLabelAngle?: number;
-
-    /**
-     * User-defined label distance calculator.
-     */
-    labelVisibilityEvaluator?: LabelVisibilityEvaluator;
 
     /**
      * User-defined far plane distance calculator.
@@ -529,7 +509,7 @@ export class MapView extends THREE.EventDispatcher {
 
     private m_createdLights?: THREE.Light[];
     private m_skyBackground?: SkyBackground;
-    private readonly m_screenProjector = new ScreenProjector();
+    private readonly m_screenProjector: ScreenProjector;
     private readonly m_screenCollisions:
         | ScreenCollisions
         | ScreenCollisionsDebug = new ScreenCollisions();
@@ -761,6 +741,7 @@ export class MapView extends THREE.EventDispatcher {
         this.m_lookAtDistance = 0;
         this.m_focalLength = 0;
         this.m_scene.add(this.m_camera); // ensure the camera is added to the scene.
+        this.m_screenProjector = new ScreenProjector(this.m_camera);
 
         this.setupCamera();
 
@@ -1900,7 +1881,7 @@ export class MapView extends THREE.EventDispatcher {
         this.m_screenCamera.updateProjectionMatrix();
         this.m_screenCamera.updateMatrixWorld(false);
 
-        this.m_screenProjector.update(this.m_camera, this.m_camera.position, width, height);
+        this.m_screenProjector.update(this.camera, width, height);
         this.m_screenCollisions.update(width, height);
 
         this.m_pixelToWorld = undefined;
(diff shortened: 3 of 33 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2227,18 +2282,26 @@ export class MapView extends THREE.EventDispatcher {
     private setupCamera() {
         const { width, height } = this.getCanvasClientSize();
 
-        this.m_scene.add(this.m_camera); // ensure the camera is added to the scene.
+        const defaultGeoCenter = new GeoCoordinates(52.518611, 13.376111, 3000);
+
+        if (this.projection.type === ProjectionType.Spherical) {
```
zbigg: Here, there: 

> // ### TODO: RTC for spherical projection

Is RTC so complex that we can't deliver it right now?
Does lack of RTC introduce same precision issues ? (i assume yes)

In the other hand, for me this condition - here and there should read `if (!this.cameraRelativeToWorldCenter)` because this is more about RTC vs no RTC than spherical vs non-spherical.

robertoraggi: RTC is not difficult implement but there are couple of ways of implementing it. Currently I'm considering to implement it by using an additional internal/private THREE.Camera. I just didn't want to make this change too tricky to review. But I have no problem in waiting to merge until RTC is implemented.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -2072,20 +2061,67 @@ export class Tile implements CachedResource {
...
-            this.center,
+        const color =
             this.dataSource.tileBackgroundColor === undefined
                 ? this.mapView.clearColor
-                : this.dataSource.tileBackgroundColor,
-            this.dataSource.tileBackgroundIsVisible
-        );
+                : this.dataSource.tileBackgroundColor;
+
+        if (this.projection.type === ProjectionType.Spherical) {
```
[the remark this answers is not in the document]
robertoraggi: I agree, and honestly I don't think this "background tile" makes any sense. I just added the support for spherical surfaces to the "background tile" just to make to keep the current change as small as possible.
What I would like to see is an internal BackgroundTileDataSource that the mapview initializes and uses it to provide background to all the data sources. For globe, for example, there's no need to have tessellated quads! a big sphere centered at 0 would be enough and it will also fix the holes at the poles.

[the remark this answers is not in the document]
nzjony: yes, there is a createPlane, there should also be a createCurvedPlace... or similar

@here/harp-mapview/lib/MapView.ts
```diff
@@ -739,8 +748,10 @@ export class MapView extends THREE.EventDispatcher {
             0.1,
             4000000
         );
+        this.camera.up.set(0, 0, 1);
         this.m_lookAtDistance = 0;
         this.m_focalLength = 0;
+        this.m_scene.add(this.m_camera); // ensure the camera is added to the scene.
```
nzjony: shouldn't this be moved into setupCamera?

@here/harp-examples/src/globe.ts
```diff
@@ -0,0 +1,204 @@
...
+        }
+    }
+
+    // Create a new MapView for the HTMLCanvasElement of the given id.
+    function initializeMapView(id: string): MapView {
+        // snippet:harp_gl_globe_example_0.ts
+        const canvas = document.getElementById(id) as HTMLCanvasElement;
+        // end:harp_gl_globe_example_0.ts
+
+        // snippet:harp_gl_globe_example_1.ts
```
nzjony: ditto

@here/harp-examples/src/globe.ts
```diff
@@ -0,0 +1,204 @@
...
+            canvas,
+            theme: "resources/berlin_tilezen_base.json",
+            projection: sphereProjection,
+            maxVisibleDataSourceTiles: 400
+        });
+        // end:harp_gl_globe_example_1.ts
+
+        CopyrightElementHandler.install("copyrightNotice", map);
+
+        // snippet:harp_gl_globe_example_3.ts
```
nzjony: ditto

@here/harp-examples/src/globe.ts
```diff
@@ -0,0 +1,204 @@
...
+                this.m_transform.copyPosition(this.camera.matrix);
+                quaternion.inverse();
+                this.m_transform.makeRotationFromQuaternion(quaternion);
+                this.camera.applyMatrix(this.m_transform);
+                this.updateCameraMatrix();
+            }
+        }
+
+        private mouseWheel(event: WheelEvent) {
+            event.preventDefault();
```
nzjony: we have a harp-map-controls folder, would it make sense to put this logic there?

@here/harp-examples/src/globe.ts
```diff
@@ -0,0 +1,204 @@
...
+            const origin = this.camera.position.clone();
+            const source = new THREE.Vector3(x, y, -1).applyMatrix4(this.m_cameraMatrixInverse);
+            const target = new THREE.Vector3(x, y, +1).applyMatrix4(this.m_cameraMatrixInverse);
+            const direction = target
+                .clone()
+                .sub(source)
+                .normalize();
+
+            const radius = EarthConstants.EQUATORIAL_RADIUS;
+            const a = direction.dot(direction);
```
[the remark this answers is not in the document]
robertoraggi: oops, I reworked this code to use THREE.Ray and THREE.Sphere.

@here/harp-examples/src/globe.ts
```diff
@@ -0,0 +1,204 @@
...
+            const d = new THREE.Vector3();
+            this.camera.getWorldDirection(d);
+            const L = this.camera.position.length() - EarthConstants.EQUATORIAL_RADIUS;
+            this.camera.position.addScaledVector(d, -event.deltaY * L * 0.001);
+            this.updateCameraMatrix();
+            this.map.update();
+        }
+
+        private pointOnSurface(clientX: number, clientY: number): THREE.Vector3 | null {
+            const x = (clientX / this.domElement.clientWidth) * 2 - 1;
```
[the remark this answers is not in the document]
robertoraggi: good point, but `calculateNormalizedDeviceCoordinates()` is in `MapControls` and I cannot use it. Also, GlobeControls is just a temporary gesture manager until we have an improved @here/harp-map-controls with support for spherical projection.

@here/harp-examples/src/globe.ts
robertoraggi: Yes, the code of GlobeControls must be integrated with MapControls.

@here/harp-mapview/lib/MapView.ts
robertoraggi: I'm not sure, but I don't have strong feelings about it. My point is: both the camera and the scene are created in the body of the constructor but then later in some other function (`setupCamera()`) the camera is added to the scene. I find this confusing.

@here/harp-examples/src/globe.ts
```diff
@@ -0,0 +1,204 @@
...
+            const target = new THREE.Vector3(x, y, +1).applyMatrix4(this.m_cameraMatrixInverse);
+            const direction = target
+                .clone()
+                .sub(source)
+                .normalize();
+
+            const radius = EarthConstants.EQUATORIAL_RADIUS;
+            const a = direction.dot(direction);
+            const b = direction.dot(origin);
+            const c = origin.dot(origin) - radius * radius;
```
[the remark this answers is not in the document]
robertoraggi: I reworked this code to use THREE.Sphere.

@here/harp-examples/src/globe.ts
```diff
@@ -0,0 +1,204 @@
...
+            const source = new THREE.Vector3(x, y, -1).applyMatrix4(this.m_cameraMatrixInverse);
+            const target = new THREE.Vector3(x, y, +1).applyMatrix4(this.m_cameraMatrixInverse);
+            const direction = target
+                .clone()
+                .sub(source)
+                .normalize();
+
+            const radius = EarthConstants.EQUATORIAL_RADIUS;
+            const a = direction.dot(direction);
+            const b = direction.dot(origin);
```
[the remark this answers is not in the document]
robertoraggi: This is the standard ray/sphere intersection test, I just re-implemented it using THREE.Sphere.

@here/harp-examples/src/globe.ts
```diff
@@ -0,0 +1,204 @@
...
+                this.domElement.addEventListener("wheel", this.m_mouseWheelListener, false);
+            } else {
+                this.domElement.removeEventListener("mousedown", this.m_mouseDownListener, false);
+                this.domElement.removeEventListener("wheel", this.m_mouseWheelListener, false);
+            }
+        }
+
+        private updateCameraMatrix() {
+            this.map.worldCenter.copy(this.camera.position);
+            mapView.projection.scalePointToSurface(this.map.worldCenter);
```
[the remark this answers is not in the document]
robertoraggi: This is by design. Moving `map.worldCenter` (and changing the camera) is what a `map controller` for `MapView` should do to re-center the scene.

@here/harp-mapview/lib/MapView.ts
nzjony: fair point

@here/harp-examples/src/globe.ts
```diff
@@ -0,0 +1,180 @@
...
+            const L = this.camera.position.length() - EarthConstants.EQUATORIAL_RADIUS;
+            this.camera.position.addScaledVector(d, event.deltaY * L * 0.001);
+            this.updateCameraMatrix();
+            this.map.update();
+        }
+
+        private pointOnSurface(clientX: number, clientY: number): THREE.Vector3 | null {
+            const x = (clientX / this.domElement.clientWidth) * 2 - 1;
+            const y = -(clientY / this.domElement.clientHeight) * 2 + 1;
+            this.m_raycaster.setFromCamera({ x, y }, this.camera);
```
nzjony: very nice

@here/harp-mapview/lib/MapView.ts
```diff
@@ -535,7 +544,8 @@ export class MapView extends THREE.EventDispatcher {
 
     private readonly m_screenCamera = new THREE.OrthographicCamera(-1, 1, 1, -1);
 
-    private m_camera: THREE.PerspectiveCamera;
+    private readonly m_camera: THREE.PerspectiveCamera;
+    private readonly m_rtcCamera: THREE.PerspectiveCamera;
```
nzjony: I think this needs a comment about why we have two perspective camera's now.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -739,8 +749,12 @@ export class MapView extends THREE.EventDispatcher {
             0.1,
             4000000
         );
+        this.m_camera.up.set(0, 0, 1);
         this.m_lookAtDistance = 0;
         this.m_focalLength = 0;
+        this.m_scene.add(this.m_camera); // ensure the camera is added to the scene.
+
+        this.m_rtcCamera = new THREE.PerspectiveCamera(); // this should not be added to the scene.
```
nzjony: can this be constructed when it is declared? similar to m_screenCamera?

@here/harp-mapview/lib/MapView.ts
nzjony: does rtc mean "relative to camera" or "relative to center"?

http://help.agi.com/AGIComponents/html/BlogPrecisionsPrecisions.htm at least it is referred to rte (i wonder if it makes sense to use this terminology instead?)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1818,25 +1870,39 @@ export class MapView extends THREE.EventDispatcher {
         this.m_camera.updateProjectionMatrix();
         this.m_camera.updateMatrixWorld(false);
 
+        this.m_rtcCamera.copy(this.m_camera);
+        this.m_rtcCamera.position.sub(this.m_camera.position);
+        this.m_rtcCamera.updateProjectionMatrix();
```
nzjony: looking at https://github.com/mrdoob/three.js/blob/dev/src/cameras/Camera.js, it seems that the projectionMatrix is copied over from the this.m_camera, so it should not be theoretically necessary? maybe I miss something.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1818,25 +1870,39 @@ export class MapView extends THREE.EventDispatcher {
         this.m_camera.updateProjectionMatrix();
         this.m_camera.updateMatrixWorld(false);
 
+        this.m_rtcCamera.copy(this.m_camera);
+        this.m_rtcCamera.position.sub(this.m_camera.position);
```
nzjony: you are setting the camera to 0,0,0 right? I think setScalar(0) is a bit more clear what you are meaning.

@here/harp-mapview/lib/Tile.ts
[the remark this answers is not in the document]
robertoraggi: I don't think that Tile is the correct place for createPlane/createCurvedPlane. Doing it there means that potentially every data source will get a replacement tile, creating a high number of similar geometries. Also, Tile is not accessible for a worker thread, and usually scaling geometries to the surface is something that will be done at decoding time.
I'm trying to keep this change as small as possible, so I kept the current 'background tile' working with sphere, but I would propose to rework the way we do background tiles completely in a follow up change.

robertoraggi:

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -939,20 +949,37 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
                         }
 
+                        if (isSpherical) {
+                            normal
+                                .set(vertices[i], vertices[i + 1], vertices[i + 2])
+                                .add(this.center)
+                                .normalize();
+                        }
+
+                        const minAxis = normal.clone().multiplyScalar(minHeight * scaleFactor);
```
zbigg: Allocate these in outside of loop and
```
minAxis.copy(normal)...

File: @here/harp-mapview/package.json [hunk 18]
```diff
@@ -32,13 +32,13 @@
...
     "license": "Apache-2.0",
     "dependencies": {
-        "@here/harp-datasource-protocol": "^0.3.3",
-        "@here/harp-geoutils": "^0.3.3",
-        "@here/harp-lines": "^0.2.3",
-        "@here/harp-lrucache": "^0.2.3",
-        "@here/harp-materials": "^0.2.3",
-        "@here/harp-text-canvas": "^0.2.3",
-        "@here/harp-utils": "^0.2.4",
+        "@here/harp-datasource-protocol": "^0.3.2",
```
zbigg: Remove from patch.

@here/harp-mapview/lib/ScreenProjector.ts
```diff
@@ -68,30 +62,10 @@ export class ScreenProjector {
      * @returns {THREE.Vector3} The projected vector (the parameter 'target').
      */
     projectVector(source: THREE.Vector3 | THREE.Vector4, target: THREE.Vector3): THREE.Vector3 {
-        target.x = source.x - this.m_center.x - this.m_cameraPosition.x;
-        target.y = source.y - this.m_center.y - this.m_cameraPosition.y;
-        target.z = source.z - this.m_center.z - this.m_cameraPosition.z;
-        target.applyMatrix4(this.m_projectionViewMatrix);
+        target.set(source.x, source.y, source.z).project(this.m_mapView!.camera);
```
zbigg: Minor ... It looks like now, we need only `camera` and this camera is always same, so maybe just make it `readonly m_camera` in constructor.

Removes also circular dependency between this module and MapView

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1797,19 +1834,51 @@ export class MapView extends THREE.EventDispatcher {
...
-        if (this.m_options.farPlaneEvaluator !== undefined) {
-            this.camera.getWorldDirection(this.m_tempVector3);
-            const angle = THREE.Math.radToDeg(this.m_tempVector3.angleTo(EYE_INVERSE));
-            const nearFarPlane = this.m_options.farPlaneEvaluator(this, angle, nearPlane, farPlane);
-            nearPlane = Math.max(kMinNear, nearFarPlane.near);
-            farPlane = Math.max(nearPlane + kFarOffset, nearFarPlane.far);
+        let nearPlane: number = 0;
+        let farPlane: number = 0;
+
+        if (this.projection.type === ProjectionType.Spherical) {
```
zbigg: Someone commened it some time ago.

constructor is already huuge, maybe move it to `setupNearFarPlanes ?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1797,19 +1834,51 @@ export class MapView extends THREE.EventDispatcher {
...
+        } else {
+            const kMinNear = 0.1;
+            const kMultiplier = 50.0;
+            const kFarOffset = 200.0;
+
+            const groundDistance = this.projection.groundDistance(this.m_camera.position);
+            nearPlane = Math.max(kMinNear, groundDistance * 0.1);
+            farPlane = nearPlane * kMultiplier + kFarOffset;
+
+            if (this.m_options.farPlaneEvaluator !== undefined) {
```
zbigg: Add comment in `MapViewOptions` that `farPlaneEvaluator` is only supported with flat projections.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
amsgames94: Addressed in next revision.

@here/harp-mapview/lib/MapView.ts
robertoraggi: i think farPlaneEvaluator like many of the options of MapViewOptions should go away. They are a bit too low level

@here/harp-mapview/lib/MapView.ts
robertoraggi: this code is not in the constructor, it's in updateCamera.

@here/harp-mapview/lib/ScreenProjector.ts
amsgames94: Fair point. Will be addressed.

@here/harp-mapview/package.json
amsgames94: Will do.

@here/harp-geoutils/lib/projection/Projection.ts
```diff
@@ -133,6 +133,21 @@ export interface Projection {
      */
     getScaleFactor(worldPoint: Vector3Like): number;
 
+    /**
+     * Returns the surface normal at the given world position.
+     *
+     * @param worldPoint The position in world coordinates.
+     */
+    surfaceNormal(worldPoint: Vector3Like): Vector3Like;
```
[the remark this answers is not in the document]
robertoraggi: I'm not sure about that. I think surface normal is the correct name in the context of the projection. Well, if we wanted to be really correct we should have two methods `geocentric surface normal` and `geodetic surface normal` but at least for now we do only sphere flat and we don't model the ellipsoid.

@here/harp-mapview/lib/MapView.ts
robertoraggi: I can try to make the far plane evaluator work with globe as well. I need to refactor a bit the farPlaneEvaluator interface and its usage (e.g. the clamping), I will create a task and make it work in a follow up change.

@here/harp-mapview/lib/ScreenProjector.ts
```diff
@@ -68,38 +67,18 @@ export class ScreenProjector {
      * @returns {THREE.Vector3} The projected vector (the parameter 'target').
      */
     projectVector(source: THREE.Vector3 | THREE.Vector4, target: THREE.Vector3): THREE.Vector3 {
-        target.x = source.x - this.m_center.x - this.m_cameraPosition.x;
-        target.y = source.y - this.m_center.y - this.m_cameraPosition.y;
-        target.z = source.z - this.m_center.z - this.m_cameraPosition.z;
-        target.applyMatrix4(this.m_projectionViewMatrix);
+        target.set(source.x, source.y, source.z).project(this.m_camera);
```
nzjony: 👍

@here/harp-mapview/lib/ScreenProjector.ts
```diff
@@ -52,7 +51,7 @@ export class ScreenProjector {
         target: THREE.Vector2 = new THREE.Vector2()
     ): THREE.Vector2 | undefined {
         const p = this.projectVector(source, ScreenProjector.tempV3);
-        if (p.z > 0 && p.z < 1) {
+        if (p.z > -1 && p.z < 1) {
```
nzjony: 👍

@here/harp-mapview/lib/ScreenProjector.ts
```diff
@@ -68,38 +67,18 @@ export class ScreenProjector {
...
-
     /**
      * Fast test to check if projected point is on screen.
      *
      * @returns {boolean} `true` if point is on screen, `false` otherwise.
      */
     onScreen(source: THREE.Vector3): boolean {
         const p = this.projectVector(source, ScreenProjector.tempV3);
-        if (p.z > 0 && p.z < 1) {
+        if (p.z > -1 && p.z < 1) {
```
nzjony: 👍

PR #456, opened 2019-05-14
HARP-5714: Fix decoder task queue draining with many aborted tasks.

zbigg:
* Fix ConcurrentWorkerSet queue draining in case of many aborted
  tasks.
* Encapsulate ConcurrentWorkerSet even more and adapt tests.


@here/harp-mapview/lib/ConcurrentWorkerSet.ts
```diff
@@ -535,6 +478,61 @@ export class ConcurrentWorkerSet {
         }
     };
 
+    /**
```
zbigg: Moved as is without any changes due to public/private ordering rules.

PR #457, opened 2019-05-15
Fix repository.url in package.json files.

zbigg:
See this conversation: https://github.com/heremaps/harp-font-resources/issues/25#issuecomment-492552436

### DIFF
--- a/@here/generator-harp.gl/package.json
+++ b/@here/generator-harp.gl/package.json
@@ -10,7 +10,7 @@
     },
     "repository": {
         "type": "git",
-        "url": "git://github.com:heremaps/harp.gl.git",
+        "url": "git://github.com/heremaps/harp.gl.git",
         "directory": "@here/generator-harp.gl"
     },
     "author": {
--- a/@here/harp-datasource-protocol/package.json
+++ b/@here/harp-datasource-protocol/package.json
@@ -15,7 +15,7 @@
     },
     "repository": {
         "type": "git",
-        "url": "git://github.com:heremaps/harp.gl.git",
+        "url": "git://github.com/heremaps/harp.gl.git",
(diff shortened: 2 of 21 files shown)


zbigg:
> After further reading, it is not needed. Both formats are accepted.

@musculman 
I'm not sure, https://docs.npmjs.com/files/package.json says

> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification

`git` doesn't like current "url"s:

```
$ git clone git://github.com:heremaps/harp.gl.git
Cloning into 'harp.gl'...
fatal: Unable to look up github.com:heremaps (port 9418) (nodename nor servname provided, or not known)
```

I guess following would be acceptable:
```
"repository": "github:heremaps/harp.gl"
```
but that prevents us from having `directory`.

Nevertheless, proper URL is much better for interoperability so i would commit this

musculman:
If you want to unify the formats, is fine, but we have examples of npm packages already published using the `:` that work fine.
I was wrong on my original comment on the issue in the harp-font-resources catalog, as both are accepted.

zbigg:
@musculman TLDR - closing it

`https://www.npmjs.com/package/@here/harp-map-theme` also says `Unable to find a readme for @here/harp-map-theme@0.2.3`.

So, as we have this problems only with huuge packages (`@here/harp-font-resources`) and `harp-map-theme` i assume this is somehow connected with `npmjs` infra refusing to inspect these and extract README.md files.

Unfortunately, more people have problem like this [1] and seems like there is no resolution.

[1] https://npm.community/search?q=Unable%20to%20find%20a%20readme%20for

musculman:
Thanks for the additional information, from my side I was also not able to find good reasons to change it either way or to investigate further :(

PR #458, opened 2019-05-15
HARP-5719: Update three.js to 0.104

ninok:


@here/harp-mapview/lib/composing/MSAARenderPass.ts
```diff
@@ -187,7 +187,9 @@ export class MSAARenderPass extends Pass {
             // 6. Render the quad on top of the previous renders.
 
             // NOTE: three.js doesn't like undefined as renderTarget, but works with `null`
-            renderer.setRenderTarget(this.renderToScreen ? null! : writeBuffer);
+            renderer.setRenderTarget(
+                this.renderToScreen ? null : writeBuffer !== undefined ? writeBuffer : null
```
CraigBerlin: You could instead define in LowResRenderPass.ts, MSAARenderPass.ts, Pass.ts
 writeBuffer: THREE.WebGLRenderTarget | null,
 readBuffer: THREE.WebGLRenderTarget | null

@here/harp-mapview/lib/MapTileCuller.ts
```diff
@@ -88,15 +88,12 @@ export class MapTileCuller {
...
-        const camera = new THREE.Camera();
-
-        camera.projectionMatrix.copy(this.m_camera.projectionMatrix);
+        const invProjMatrix = (this.m_camera as any).projectionMatrixInverse as THREE.Matrix4;
 
         let cornerIndex = 0;
 
         function addPoint(x: number, y: number, z: number) {
-            frustumCorners[cornerIndex++].set(x, y, z).unproject(camera);
+            frustumCorners[cornerIndex++].set(x, y, z).applyMatrix4(invProjMatrix);
```
okanishchev: You can use `.unproject(this.m_camera)` here, as it takes camera's projectionMatrixInverse now.
See https://github.com/mrdoob/three.js/blob/dev/src/math/Vector3.js#L320,
and PR https://github.com/mrdoob/three.js/pull/15996/files merged in r103

ninok: unproject is also multiplying with the world matrix. This we don't want to do in this case, that's why in the old code we created a temporary camera.

okanishchev: Yep, I've missed that. Maybe some comment for clarity?

@here/harp-mapview/lib/composing/MSAARenderPass.ts
ninok: done

PR #471, opened 2019-05-22
MINOR: Add XYZ Space as APIFormat

ninok:
Add an XYZ Space as an APIFormat to be used within OmvDataSource. This enables a developer to visualize data from an XYZ Space, using the .mvt endpoint from XYZ. Example endpoint:

```
https://xyz.api.here.com/hub/spaces/your-space-id/tile/web/{z}_{x}_{y}.mvt?access_token=your-access-token
```

Example of adding data from XYZ Space to map:
```
const xyzDataSource = new OmvDataSource({
   baseUrl: "https://xyz.api.here.com/hub/spaces/2vDrakce/tile/web",
   apiFormat: APIFormat.XYZSpace,
   styleSetName: "geojson",
   maxZoomLevel: 17,
   authenticationCode: "your-access-token"
});
map.addDataSource(xyzDataSource);
```

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvRestClient.ts
+++ b/@here/harp-omv-datasource/lib/OmvRestClient.ts
@@ -110,7 +110,23 @@ export enum APIFormat {
      *
      * Default authentication method used: [[AuthenticationTypeTomTomV1]].
      */
-    TomtomV1
+    TomtomV1,
+
+    /**
+     * Use the REST API format of XYZ Space Vector Tile API in OMV format.
+     *
+     * Usage:
+     * `<OmvRestClientParams.baseUrl>/hub/spaces/<space-id>/tile/web/<zoom>_<X>_<Y>.mvt?access_token=<OmvRestClientParams.authenticationCode>`
+     *
+     * Format definition:
+     * `http|s://<base-url>/hub/spaces/{spaceId}/tile/web/{z}_{x}_{y}.mvt?access_token={access_token}`
+     *
+     * Sample URL:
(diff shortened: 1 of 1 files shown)


zbigg:
It's like 5th or 6th format in this API.

I wanted to propose it some time ago, but i think we should step back and allow user to add arbitrary type of URL.

Like this:
{
   baseurl: "https://xyz.api.here.com/.../tile/web/{z}_{x}_{y}.mvt?access_token= 
  {accessToken}",
  accessToken: "fooBar"
}

The z/x/y can be documented, others like `accessToken` can be resolved from options in some intelligent fashion.

@ninok 
By no means i want to -1 here, but let's maybe add in next iteration.

ninok:
> I wanted to propose it some time ago, but i think we should step back and allow user to add arbitrary type of URL.

Yes, we should have this. I think it does not even need to be that smart. Maybe it would be good enough if the user could specify a function that get's x,y,z,accessToken,... as input an the user can do whatever he needs to do.

ninok:
> @ninok
> By no means i want to -1 here, but let's maybe add in next iteration.

Yes, we can do this later. For usability reasons I still think we should at least have all the different formats  that HERE offers. The more interesting question is, why do all our services have different APIs. They should only differ in the base url.

dbabbs:
I agree that ideally you should be able to pass an absolute URL instead of having to create a custom APIFormat for each type of endpoint. Would be great to see that as soon as possible.

However, we are giving a harp.gl workshop involving XYZ Spaces in mid-June. Unless we are able to have the arbitrary absolute url as a source merged in by then, can we please merge this PR as a temporary solution?

PR #476, opened 2019-05-22
Simple Map component.

zbigg:
(very draft) 

Proposal wrapper over our APIs.

### DIFF
--- a/@here/harp-examples/src/hello2_js-bundle.html
+++ b/@here/harp-examples/src/hello2_js-bundle.html
@@ -1,0 +1,83 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta
+            name="viewport"
+            content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0"
+        />
+        <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    </head>
+    <style>
+        body {
+            background: #fff;
+            padding: 0;
+            margin: 0;
+            overflow: hidden;
+        }
+    </style>
+    <body>
+        <!--
+            These paths are to be used by clients, when using map bundle from CDN.
+        -->
+        <!--
+            <script src="https://unpkg.com/three/build/three.min.js"></script>
+            <script src="https://unpkg.com/@here/harp.gl/dist/harp.js"></script>
+        -->
+        <!--
+            These paths are used here to be compatible with harp.gl CI.
+        -->
+        <script src="three.min.js"></script>
+        <script src="harp.js"></script>
+
+        <script>
+            const hereCopyrightInfo = {
+                id: "here.com",
+                year: new Date().getFullYear(),
+                label: "HERE",
+                link: "https://legal.here.com/terms"
+            };
+            const xyzAccessToken = "AYlqpxvwl7C8tSVG22lX2lg";
+
+            const map = new harp.Map(document.body, {
+                themeUrl: "resources/berlin_tilezen_base.json",
+                location: [40.6935, -74.009],
+                zoomLevel: 16.9,
+                cameraYaw: 6.3,
+                cameraPitch: 50,
+                dataSources: [
+                    new harp.OmvDataSource({
+                        baseUrl: "https://xyz.api.here.com/tiles/herebase.02",
+                        apiFormat: harp.APIFormat.XYZOMV,
+                        styleSetName: "tilezen",
+                        maxZoomLevel: 17,
+                        authenticationCode: xyzAccessToken,
+                        copyrightInfo: [hereCopyrightInfo]
+                    })
+                ]
+            });
+            const infoElement = document.createElement("div");
+            infoElement.style.position = "absolute";
+            infoElement.style.right = "0";
+            infoElement.style.top = "0";
+            infoElement.style.backgroundColor = "#f0fef1";
+            infoElement.style.zIndex = "400";
+            infoElement.style.padding = "2px";
+            infoElement.style.fontSize = "0.8em";
+            infoElement.style.fontFamily = "sans-serif";
+            document.body.appendChild(infoElement);
+
+            function humanReadable(n) {
+                return n.toPrecision(5).replace(/(\.)?0+$/, "");
+            }
+            map.addEventListener("camera-updated", () => {
+                const location = map.geoCenter;
+                infoElement.innerHTML = `
+                    Lat: ${humanReadable(location.latitude)}<br>
+                    Long: ${humanReadable(location.longitude)}<br>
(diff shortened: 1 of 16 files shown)


@here/harp-map/lib/BuiltinMap.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * This class is an alias for standard Map class that exists in Node and Browser environment.
+ *
+ * Required because as for now thre is no way to reliably alias name from global module from
+ * scope that redefines this name (and Map.ts defines class Map).
+ *
+ * @see https://github.com/Microsoft/TypeScript/issues/983
+ * @see https://stackoverflow.com/questions/51319046
+ */
+export const BuiltinMap = Map;
```
Astrak: Why rewrite the existing `Map` object? How does the user know that he now has to call `BuiltinMap`? Why would we be the only ones to do it that way? Can't we easily find another name than `Map`?

@here/harp-map/lib/Map.ts
```diff
@@ -0,0 +1,611 @@
...
+        const canvas = this.getCanvas(element);
+        const theme =
+            this.params.theme !== undefined
+                ? this.params.theme
+                : this.params.themeUrl !== undefined &&
+                  this.params.themeUrl !== null &&
+                  this.params.themeUrl !== ""
+                ? this.params.themeUrl
+                : Map.getDefaultTheme();
+        this.mapView = new MapView({
```
Astrak: Stupid question (I'm only overlooking the PR): couldn't `Map` simply extend `MapView`?

@here/harp-map/lib/Map.ts
```diff
@@ -0,0 +1,611 @@
...
+            // positioning for container element, while internally we can use 'relative + absolute'
+            // positioning of actual canvas to reliably fill whole parent element.
+
+            const extraParent = document.createElement("div");
+            extraParent.style.position = element === document.body ? "absolute" : "relative";
+            extraParent.style.width = "100%";
+            extraParent.style.height = "100%";
+            extraParent.style.margin = "0";
+            extraParent.style.padding = "0";
+            extraParent.style.border = "none";
```
Astrak: in one go:
```suggestion
            extraParent.style.cssText = `position: ${element === document.body ? "absolute" : "relative"}; width: 100%; height: 100%; margin: 0; padding: 0; border: none;`;
```

@here/harp-map/lib/Map.ts
```diff
@@ -0,0 +1,611 @@
...
+            const canvas = document.createElement("canvas");
+            canvas.style.width = "100%";
+            canvas.style.height = "100%";
+
+            extraParent.appendChild(canvas);
+            return canvas;
+        } else {
+            if (!document) {
+                throw new Error(
+                    "harp.Map: document not yet loaded, Use window.onload to delay script " +
```
Astrak: `addEventListener` for `"DOMContentLoaded"` rather.

@here/harp-examples/src/hello2_js-bundle.html
```diff
@@ -0,0 +1,83 @@
...
+                return n.toPrecision(5).replace(/(\.)?0+$/, "");
+            }
+            map.addEventListener("camera-updated", () => {
+                const location = map.geoCenter;
+                infoElement.innerHTML = `
+                    Lat: ${humanReadable(location.latitude)}<br>
+                    Long: ${humanReadable(location.longitude)}<br>
+                    Zoom: ${humanReadable(map.mapView.zoomLevel)}
+                `;
+            });
```
Astrak: From the GoogleMaps API entry point, the first hello-world example they showcase is even more simple: 
```js
map = new google.maps.Map(document.getElementById('map'), {
      center: {lat: -34.397, lng: 150.644},
      zoom: 8
});
```

For this we would only need to have internal defaults for the datasources and themes parameters:

### DataSources
I wonder if we even need to define the datasources for this top-level component. Likely the base OMV layer is always needed, and a Features layer could be handled internally if we add some custom objects logic. Then we wouldn't need to expose the datasource model.

### Theme URL
Users for a simple level may not be the ones creating or using tailored themes. We could simplify the theme parameter into simply an optional string being `"day"`, `"reducedDay"` or `"night"` and with an internal default.

@here/harp-map/lib/Map.ts
```diff
@@ -0,0 +1,611 @@
...
+     * If `trackResize` option is enabled, then this method is invoked automatically after each
+     * window resize.
+     *
+     * If parent is `document.body`, `Map` ocuppies whole viewport.
+     *
+     * Notes about styling of parent canvas element:
+     *  * If parent is `body` then `body` shouldn't have margin nor padding
+     *  * Other parents can have margin and borders, but padding is not supported.
+     */
+    updateMapSize = () => {
```
Astrak: I'd be curious to know how other webgl-based map libraries handle the resizing.

@here/harp-map/lib/BuiltinMap.ts
zbigg: Use doesn't need to know anything about this atrocity. This was needed only in `Map.ts` to use real "standard" Map, it is leftover which AFAIRis not needed anymore.

Why `Map` ... because, everyone does this ?
* `mapboxgl.Map` - https://docs.mapbox.com/mapbox-gl-js/overview/
 `google.maps.Map` - https://developers.google.com/maps/documentation/javascript/tutorial
...

`Map` works well in pre-ES6 world, where all JS libraries were delivered in some namespace.
Like, our `harp` namepace from `harp.js` bundle.
In Es6 world, when user is expecte to
```
import { Map } from "whatever";
```
This becomes really ugly, because user has to either rename it to say "HarpMap", or import whole module as namespace. Both are suboptimal.

Astrak: Ok. About a solution: would it be possible to rename it to something else for the modular version, and export it `as Map` for the namespaced version?

@here/harp-examples/src/hello2_js-bundle.html
zbigg: > Theme URL

As noted, it's first draft.
Theme URL should be default, indeed but we don't have effective means to locate default theme yet.
This can be`https://unpkg.com/@here/harp-map-theme@x.y.z/...` but that's problematic, because
we hardcode dependencies into bundle...
Forces users to use particular CDN.

Another option, is to bundle harp.js with all the styles, and assets and fonts, but then this package becomes huge and we may expected problems - at least until we use `unpkg.com` as CDN.

...
This shows, how concept of 
1) forced theme - Map/MapView doesn't work without theme now
2) forced universal theme on MapView, even if it's only datasource who uses it
3) OmvDataSource which doesn't work at all without theme because it's only theme, but also set of mandatory rules to process data ... and it's not attached to DS, but  to MapView

It's all ... at best suboptimal ..

PR #479, opened 2019-05-22
HARP-5432 Remove z component from mesh

nzjony:
The z component is required to do ray casting on the CPU, however on the
GPU we use a displacement map to do this, so we zero the z value.

### DIFF
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -1300,6 +1300,14 @@ export class Tile implements CachedResource {
     diffuseColor *= texelColor;
 #endif`
                         );
+
+                        // We remove the displacement map from manipulating the vertices, it is
+                        // however still required for the pixel shader, so it can't be directly
+                        // removed.
+                        shader.vertexShader = shader.vertexShader.replace(
+                            "#include <displacementmap_vertex>",
+                            ""
+                        );
                     };
                 }
 


@here/harp-mapview/lib/Tile.ts
```diff
@@ -1300,6 +1300,12 @@ export class Tile implements CachedResource {
     diffuseColor *= texelColor;
 #endif`
                         );
+
+                        shader.vertexShader = shader.vertexShader.replace(
+                            "#include <displacementmap_vertex>",
+                            `transformed = vec3(position.xy, 0);
```
ninok: It's a bit weird to remove the z component and add it again from the displacement map.
This will also not work for the globe.
Why not just remove the displacement map code from the shader?

nzjony: I think it has to be tested for the globe, if the points are in local space, then it should work, because the vertex normals of the terrain are the surface point at that world location. So, the terrain points z component should not matter (I hope) (because they are displaced using the displacement map in the direction of the surface normal, and rotated according using the model view matrix later).

The displacement map might be worth keeping if we want to do some animation? 
 / Scaling. Otherwise if there is no need for it, I will remove it.

ninok: - Currently tiles are not in local tangent space
- Even if terrain tiles would be in local tangent space, at low zoomlevels the curvature is not negligible
- If we want to do animations with the displacement map, I would rather remove the altitude on the CPU side(before projecting the vertices) and fix the raycasting to take the displacement map into account.

nzjony: Your right, I will remove the displacement.

nzjony:

PR #480, opened 2019-05-23
HARP-5847: Remove SharedArrayBuffer reference.

Astrak:
Geojson and custom features are meeting the `SharedArrayBuffer` check, which is undefined everywhere but in Chrome.

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvUtils.ts
+++ b/@here/harp-omv-datasource/lib/OmvUtils.ts
@@ -8,7 +8,11 @@
  * @hidden
  */
 export function isArrayBufferLike(data: any): data is ArrayBufferLike {
-    return data instanceof ArrayBuffer || data instanceof SharedArrayBuffer;
+    if (typeof SharedArrayBuffer !== "undefined") {
+        return data instanceof ArrayBuffer || data instanceof SharedArrayBuffer;
+    } else {
+        return data instanceof ArrayBuffer;
+    }
 }
 
 /**


@here/harp-omv-datasource/lib/OmvUtils.ts
```diff
@@ -7,8 +7,8 @@
 /**
  * @hidden
  */
-export function isArrayBufferLike(data: any): data is ArrayBufferLike {
-    return data instanceof ArrayBuffer || data instanceof SharedArrayBuffer;
+export function isArrayBuffer(data: any): data is ArrayBufferLike {
```
ninok: should return `data` is ArrayBuffer`

okanishchev:
What if we actually will be using shared buffers? Maybe it worth to do additional check for `typeof SharedArrayBuffer !== 'undefined'` or so.

@here/harp-omv-datasource/lib/OmvUtils.ts
Astrak: `ArrayBufferLike` already only contains `ArrayBuffer`, I don't know why it was implemented this way, I thought it didn't impact the problem so I've let it this way. It could be deleted entirely yes. Should I do it?

ninok: Yes, this is quite confusing. Please remove it.

Astrak: It is actually in the es5 type definition, could not remove. Implemented what Oleksii suggested instead.

PR #481, opened 2019-05-23
Fix test-npm-packages script

zbigg:
* get packages list to be rebuilt using script
* forcefully bump version, to ensure that we test "next" versions of packages, so nothing
  can possibly be retrieved from npm
* re-enable `test-npm-packages` in Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ jobs:
         npx ts-mocha -r tsconfig-paths/register ./@here/*/test/*.ts
         yarn test-browser --headless-firefox
         yarn test-browser --headless-chrome
-        #./scripts/test-npm-packages.sh
-        #git status # just in case test-npm-packages leaves some garbage
+        ./scripts/test-npm-packages.sh
+        git status # just in case test-npm-packages leaves some garbage
     - name: "Build & Deploy"
       script: |
         set -ex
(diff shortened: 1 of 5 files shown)


zbigg:
> sigh - I wanted to prevent the lerna dependency because it pulls in a lot of stuff.

node_modules size 
* without lerna :493M
* with lerna: 472M

So, yes 20mb, but original `node_modules` size is huge already.

Without this it already contains surprises like test coverage of some random package:

```
8,0K	node_modules/unique-filename/test
12K	node_modules/unique-filename/coverage/__root__
64K	node_modules/unique-filename/coverage !!!
12K	node_modules/unique-filename/.nyc_output !!!
104K	node_modules/unique-filename
```

> Is this test really so important? It's a lot of extra code to maintain :(

Well, we test only in artificial environment - `ts-node` / `ts-mocha` and in bundle that is not distributed  - test bundle.

This test, checks if our npm packages actually build and if they really contain what we advertise they contain. (at least at minimal level) - IMHO, it's worth being here.

DO you see other way of checking, that build artifacts (tgz files) are not empty files ?

musculman:
What about this change? The script is still disabled in travis but I guess it would make sense to merge this and re-enable it at some point :)

PR #485, opened 2019-05-24
HARP-5900 Add satellite tiles example

musculman:
Add `ppi` to request only if explicitly initialized with a value, otherwise satellite tiles do not work.
Change valid zoom levels of web tiles from 1 to 20.

### DIFF
--- a/@here/harp-examples/src/datasource_satellitetile.ts
+++ b/@here/harp-examples/src/datasource_satellitetile.ts
@@ -14,20 +14,20 @@ import { appCode, appId } from "../config";
 /**
  * A simple example using the webtile data source. Tiles are retrieved from
  * ```
- * https://1.base.maps.api.here.com/maptile/2.1/maptile/newest/normal.day/${level}/${column}/${row}/512/png8?app_id=${appId}&app_code=${appCode}
+ * https://1.aerial.maps.api.here.com/maptile/2.1/maptile/newest/satellite.day/${level}/${column}/${row}/512/png8?app_id=${appId}&app_code=${appCode}
  * ```
  *
  * A [[WebTileDataSource]] is created with specified applications' appId and appCode passed
  * as [[WebTileDataSourceOptions]]
  * ```typescript
- * [[include:harp_gl_datasource_webtile_1.ts]]
+ * [[include:harp_gl_datasource_satellitetile_1.ts]]
  * ```
  * Then added to the [[MapView]]
  * ```typescript
- * [[include:harp_gl_datasource_webtile_2.ts]]
+ * [[include:harp_gl_datasource_satellitetile_2.ts]]
  * ```
  */
-export namespace WebTileDataSourceExample {
+export namespace SatelliteDataSourceExample {
     // creates a new MapView for the HTMLCanvasElement of the given id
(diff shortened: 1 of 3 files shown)


@here/harp-examples/src/datasource_satellitetile.ts
```diff
@@ -0,0 +1,76 @@
...
+ * as [[WebTileDataSourceOptions]]
+ * ```typescript
+ * [[include:harp_gl_datasource_satellitetile_1.ts]]
+ * ```
+ * Then added to the [[MapView]]
+ * ```typescript
+ * [[include:harp_gl_datasource_satellitetile_2.ts]]
+ * ```
+ */
+export namespace WebTileDataSourceExample {
```
ninok: Please use a different namespace for this example

@here/harp-examples/src/datasource_webtile.ts
```diff
@@ -63,7 +63,8 @@ export namespace WebTileDataSourceExample {
     // snippet:harp_gl_datasource_webtile_1.ts
     const webTileDataSource = new WebTileDataSource({
         appId,
-        appCode
+        appCode,
+        ppi: 320
```
nzjony: this seems to be for a phone, is that the general use case that this example should show?

@here/harp-examples/src/datasource_satellitetile.ts
```diff
@@ -0,0 +1,76 @@
...
+            canvas,
+            theme: "resources/berlin_tilezen_base.json"
+        });
+        map.setCacheSize(100, 100);
+
+        // instantiate the default map controls, allowing the user to pan around freely.
+        const controls = new MapControls(map);
+
+        // Add an UI.
+        const ui = new MapControlsUI(controls);
```
nzjony: should we limit the tilt here as in other examples?

@here/harp-examples/src/datasource_satellitetile.ts
```diff
@@ -0,0 +1,76 @@
...
+export namespace WebTileDataSourceExample {
+    // creates a new MapView for the HTMLCanvasElement of the given id
+    export function initializeMapView(id: string): MapView {
+        const canvas = document.getElementById(id) as HTMLCanvasElement;
+
+        const map = new MapView({
+            canvas,
+            theme: "resources/berlin_tilezen_base.json"
+        });
+        map.setCacheSize(100, 100);
```
nzjony: why is this needed?

@here/harp-examples/src/datasource_webtile.ts
musculman: It is what it was showing before, I think for most of the browsers is the best setting

@here/harp-examples/src/datasource_satellitetile.ts
musculman: They are webtiles, I don't see why is that needed?

@here/harp-examples/src/datasource_satellitetile.ts
musculman: Was already in the previous one, so I did not change it

PR #455, opened 2019-05-14
HARP-4670: Host docs and demo in s3

musculman:
Host docs and demo in s3 following a structure so that there is an archive of releases, and an index file of those releases maintained.
Remove extra script from travis and add it to typedoc generation.
New minisite under /docs with assets that show the available releases.
This minisite is copied as well by typedoc and hosted on s3.
For every release, the minisite is pushed to gh-pages and the docs and examples to s3.
For every master merge, the master documentation on s3 is updated.
This PR depends on #429

PR #490, opened 2019-05-27
HARP-4670: Change link to gh-pages on README

musculman:

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 `harp.gl` is an _experimental and work in progress_ open-source 3D map rendering engine.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 `harp.gl` is an _experimental and work in progress_ open-source 3D map rendering engine.
 
-### [Documentation](https://heremaps.github.io/harp.gl/doc) - [Examples](https://heremaps.github.io/harp.gl/examples)
```
ninok: How to find the examples now?

musculman: With https://github.com/heremaps/harp.gl/pull/455
After we make a release, we will have in gh-pages a new html that will show current master and past releases in order, so it only makes sense to link directly to master gh-pages

PR #494, opened 2019-05-27
HARP-5683: Support Cubemap Sky.

amsgames94:


amsgames94:

@here/harp-examples/src/datasource_satellitetile_globe.ts
```diff
@@ -34,7 +34,7 @@ export namespace SatelliteTileDataSourceGlobeExample {
 
         const map = new MapView({
             canvas,
-            theme: "resources/berlin_tilezen_base.json",
+            theme: "resources/berlin_tilezen_base_globe.json",
```
musculman: Is there no way to use the existing style? Duplicating them for globe sounds like an overkill and a big maintenance overhead

@here/harp-map-theme/resources/berlin_tilezen_base.json
```diff
@@ -1,12 +1,15 @@
 {
     "sky": {
-        "type": "gradient",
-        "colorTop": "#015EBB",
-        "colorBottom": "#DDDCDA",
-        "groundColor": "#87959A",
-        "monomialPower": 1
+        "params": {
```
zbigg: Wouldn't it be simpler without this "params" indirection?
```
"sky": {
     "type": "gradient",
     "colorBottom": "#DDDCDA",
     ...
}
```

amsgames94: The ``params`` indirection is handy at runtime, since it prevents us from checking for multiple attributes that ALL have to be defined for the sky to be generated accordingly.

The other possible ``Sky`` definition would be:
```
export interface Sky {
    type: "gradient" | "cubemap";
    topColor?: string;
    bottomColor?: string;
    groundColor?: string;
    monomialPower?: number;
    faces?: [
        CubemapFaceDefinition,
        CubemapFaceDefinition,
        CubemapFaceDefinition,
        CubemapFaceDefinition,
        CubemapFaceDefinition,
        CubemapFaceDefinition
    ];
}
```

In which only ``type`` would always be defined. Functions like ``isCubemapSky`` and ``isGradientSky`` would have to be used at runtime (which I'm not against if enough people agree this is the way to go).

@here/harp-examples/src/datasource_satellitetile_globe.ts
amsgames94: It's true that for now they only differ in the ``Sky`` property, but decided to added because in the future designer might want to tweak other properties (i.e. text size).

ninok: @musculman: This is what I mentioned in the standup as well. We should support json $ref so we could reuse parts of existing stylesheets. @robertoraggi mentioned that this is also covered by his style refactoring.

@here/harp-map-theme/resources/berlin_tilezen_base.json
ninok: @amsgames94 I guess with this approach we would loose proper json schema. You could write sth. like:
`
"sky" : {
 "type": "cubemap",
 "topColor": "#fff"
`

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -381,29 +381,63 @@ export interface TextStyleDefinition {
...
- *
- * @param enabled Whether the fog is enabled.
- * @param startRatio Distance ratio to far plane at which the linear fogging begins.
+ * Interface that defines the options to configure a cubemap sky.
+ */
+export interface CubemapSkyParams {
+    /** Sky type. */
+    type: "cubemap";
+    /** Array of [[CubemapFaceDefinition]] defining the 6 faces of the cubemap. */
+    faces: [
```
ninok: Can we change this type somehow that it is not possible to define a face twice? I guess this will be a quite common typo that will happen to our users and it would be good if we could catch it with the json schema.

Would be also good  to have this as a dedicated type that we can reuse for the env. map of the standard technique.

amsgames94: Addressed in latest patch set.

@here/harp-map-theme/resources/berlin_tilezen_base.json
zbigg: @amsgames94 
You could equally do this:
```
export type Sky = CubemapSkyParams | GradientSkyParams
``` 
abd have proper typing without artificial `params` node.

Moreover, not "renaming" (`colorTop => topColor`) for sake of renaming would help with backwards compatibility.

Now, we have to clean again (#510, #512) - not including, customers who will have to face with theme change.

amsgames94: @zbigg The renaming didn't happen for the sake of renaming, it happened for the sake of consistency. We had ``colorBottom`` and ``colorTop`` and ``groundColor``, which had different naming approaches when they're all describing similar properties. Took the correct english grammar (adjective before noun) and applied to all members. I tried to fix all the themes, but might have slipped a few. But I still believe than we should try to clean old code instead of not touching it ever again... We can do better ;)

And regarding the ``Sky`` type, it's true that it could be simplified that way. Will look into it and create a minor.

amsgames94: @zbigg Cleaner ``Sky`` definition based on your proposal on review now on #515

@here/harp-mapview/lib/ThemeLoader.ts
```diff
@@ -51,6 +52,14 @@ export class ThemeLoader {
             (childUrl: string) => resolveReferenceUrl(theme.url, childUrl),
             defaultUrlResolver
         );
+        if (theme.sky && theme.sky.params.type === "cubemap") {
```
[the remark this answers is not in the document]
amsgames94: Addressed in #515.

PR #496, opened 2019-05-28
HARP-5883: Interpolation of colors with zoom levels does not work

okanishchev:

### DIFF
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -17,6 +17,9 @@ import {
     isExtrudedLineTechnique,
     isExtrudedPolygonTechnique,
     isFillTechnique,
+    isInterpolatedProperty,
+    isLineTechnique,
+    isSegmentsTechnique,
     isSolidLineTechnique,
     isSquaresTechnique,
     isTerrainTechnique,
@@ -39,6 +42,7 @@ import {
     EdgeMaterialParameters,
     FadingFeature,
     MapMeshBasicMaterial,
+    MapMeshStandardMaterial,
     SolidLineMaterial
 } from "@here/harp-materials";
 import {
@@ -387,7 +391,7 @@ export class TileGeometryCreator {
         const materials: THREE.Material[] = [];
         const mapView = tile.mapView;
         const dataSource = tile.dataSource;
-        const zoomLevel = mapView.zoomLevel;
+        const displayZoomLevel = Math.floor(mapView.zoomLevel);
         const objects = tile.objects;
 
         if (dataSource.addTileBackground) {
@@ -437,7 +441,7 @@ export class TileGeometryCreator {
                     material = createMaterial(
                         {
                             technique,
-                            level: Math.floor(zoomLevel),
+                            level: displayZoomLevel,
                             fog: mapView.scene.fog !== null
                         },
                         onMaterialUpdated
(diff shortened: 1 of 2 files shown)


okanishchev:
As we exclude text, point, and marker techniques, we left with
* **color** and **emissive** for _standard_ technique, and hence _extruded-polygon_;
* **lineColor** for _extruded-polygon_;
* **defaultColor** for _extruded-polygon_ we can't interpolate as it is processed in decoder;
* **color** for _line_, _segments_ and _extruded-line_ techniques;
* **color** and **secondaryColor** for _solid-line_;
* **color** for _dashed-line_ and _fill_ techniques;

Maybe I've missed something?
This request not really depends on #497 while there is no examples here, but may be blocked in future.

okanishchev:
Well, I've forgot color on _line_ and _segments_ techniques.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -387,7 +391,7 @@ export class TileGeometryCreator {
         const materials: THREE.Material[] = [];
         const mapView = tile.mapView;
         const dataSource = tile.dataSource;
-        const zoomLevel = mapView.zoomLevel;
+        const displayZoomLevel = Math.floor(mapView.zoomLevel);
```
amsgames94:
```suggestion
        const zoomLevel = mapView.zoomLevel;
        const displayZoomLevel = Math.floor(zoomLevel);
```
There's no need to remove the ``zoomLevel`` variable here and the patch-set would be smaller that way, since you wouldn't need to update lines from ``zoomLevel`` to ``mapView.zoomLevel``, like...

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -612,9 +639,16 @@ export class TileGeometryCreator {
...
+                                getOptionValue(
+                                    getPropertyValue(technique.color, mapView.zoomLevel),
+                                    "#000000"
+                                )
+                            );
+
                             lineMaterial.lineWidth =
                                 getOptionValue(
-                                    getPropertyValue(technique.lineWidth, zoomLevel),
+                                    getPropertyValue(technique.lineWidth, mapView.zoomLevel),
```
amsgames94: ... here (look at above comment).

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
okanishchev: There are two cases in this method, where zoomLevel is used:
 1. creating objects
 1. pre-render update functions

in first case zoomLevel is used as integer, flooring current value down, passing clear (and persistent) level to properties that will not update further.
in second - we need up-to-date zoomLevel and can't use value from cached variable.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
okanishchev: this is the 2nd case of my previous comment. We need up-to-date value here, and can't use cached variable;

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
amsgames94: You're, right didn't think about this case. Thanks for the explanation!

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
amsgames94: Got it!

PR #497, opened 2019-05-28
HARP-5717: Extend json schema

okanishchev:
Extend json scheme to support editing in monaco editor/map-editor

Fix theme and technique interfaces;
Fix priority interpolation;
Add color format to enable colorpicker in editor;
Export defaultSnippets to json schema;
Cleanup berlin_tilezen theme files;

### DIFF
--- a/@here/harp-datasource-protocol/lib/TechniqueParams.ts
+++ b/@here/harp-datasource-protocol/lib/TechniqueParams.ts
@@ -48,13 +48,13 @@ export interface BaseTechniqueParams {
      * Distance to the camera (0.0 = camera position, 1.0 = farPlane) at which the object start
      * fading out (opacity decreases).
      */
-    fadeNear?: number;
+    fadeNear?: MaybeInterpolatedProperty<number>;
 
     /**
      * Distance to the camera (0.0 = camera position, 1.0 = farPlane) at which the object has zero
      * opacity and stops fading out. A value of <= 0.0 disables fading.
      */
-    fadeFar?: number;
+    fadeFar?: MaybeInterpolatedProperty<number>;
 }
 
 export enum TextureCoordinateType {
@@ -78,8 +78,9 @@ export interface StandardTechniqueParams extends BaseTechniqueParams {
      * Color of the feature in hexadecimal or CSS-style notation, for example: `"#e4e9ec"`,
      * `"#fff"`, `"rgb(255, 0, 0)"`, or `"hsl(35, 11%, 88%)"`.
      * See https://threejs.org/docs/#api/en/materials/MeshStandardMaterial.color.
+     * @format color-hex
      */
-    color?: string;
+    color?: MaybeInterpolatedProperty<string>;
     /**
      * A value of `true` creates a wireframe geometry. (May not be supported with all techniques).
      * See https://threejs.org/docs/#api/en/materials/MeshStandardMaterial.wireframe.
@@ -96,7 +97,7 @@ export interface StandardTechniqueParams extends BaseTechniqueParams {
      * diffuse. Default is `0.5`.
      * See https://threejs.org/docs/#api/en/materials/MeshStandardMaterial.roughness.
      */
-    roughness?: number;
+    roughness?: MaybeInterpolatedProperty<number>;
     /**
      * How much the material is like a metal. Nonmetallic materials such as wood or stone use `0.0`,
      * metallic ones use `1.0`, with nothing (usually) in between. Default is `0.5`. A value between
@@ -104,12 +105,12 @@ export interface StandardTechniqueParams extends BaseTechniqueParams {
      * values are multiplied.
      * See https://threejs.org/docs/#api/en/materials/MeshStandardMaterial.metalness.
      */
-    metalness?: number;
+    metalness?: MaybeInterpolatedProperty<number>;
     /**
      * The material will not be rendered if the opacity is lower than this value.
      * See https://threejs.org/docs/#api/en/materials/Material.alphaTest.
      */
-    alphaTest?: number;
+    alphaTest?: MaybeInterpolatedProperty<number>;
     /**
      * Skip rendering clobbered pixels.
      * See https://threejs.org/docs/#api/en/materials/Material.depthTest.
@@ -126,18 +127,19 @@ export interface StandardTechniqueParams extends BaseTechniqueParams {
      * opaque.
      * See https://threejs.org/docs/#api/en/materials/Material.opacity.
      */
-    opacity?: number;
+    opacity?: MaybeInterpolatedProperty<number>;
     /**
      * Emissive (light) color of the material, essentially a solid color unaffected by other
      * lighting. Default is black.
      * See https://threejs.org/docs/#api/en/materials/MeshStandardMaterial.emissive.
+     * @format color-hex
      */
-    emissive?: string;
+    emissive?: MaybeInterpolatedProperty<string>;
     /**
      * Intensity of the emissive light. Modulates the emissive color. Default is `1`.
      * See https://threejs.org/docs/#api/en/materials/MeshStandardMaterial.emissiveIntensity.
      */
-    emissiveIntensity?: number;
+    emissiveIntensity?: MaybeInterpolatedProperty<number>;
     /**
      * The index of refraction (IOR) of air (approximately 1) divided by the index of refraction of
      * the material. It is used with environment mapping modes `THREE.CubeRefractionMapping` and
@@ -145,7 +147,7 @@ export interface StandardTechniqueParams extends BaseTechniqueParams {
      *  is `0.98`.
      * See https://threejs.org/docs/#api/en/materials/MeshStandardMaterial.refractionRatio.
      */
-    refractionRatio?: number;
+    refractionRatio?: MaybeInterpolatedProperty<number>;
 
     /**
      * Whether and how texture coordinates should be generated. No texture coordinates are
@@ -219,8 +221,9 @@ export interface PointTechniqueParams extends BaseTechniqueParams {
     /**
      * Color of a point in hexadecimal or CSS-style notation, for example: `"#e4e9ec"`, `"#fff"`,
      * `"rgb(255, 0, 0)"`, or `"hsl(35, 11%, 88%)"`.
+     * @format color-hex
      */
-    color?: string;
+    color?: MaybeInterpolatedProperty<string>;
     /**
      * URL of a texture image to be loaded.
      */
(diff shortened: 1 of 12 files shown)


@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -183,6 +184,166 @@ export interface BaseStyle {
...
+ *         "label": "New extruded-polygon",
+ *         "description": "Add a New 'extruded-polygon' Styling Rule",
+ *         "body": {
+ *             "technique": "extruded-polygon",
+ *             "when": "$1",
+ *             "attr": {}
+ *         }
+ *     },
+ *     {
+ *         "label": "New none",
```
zbigg: Do we really need snippets for everything, including unnecessary things like none?

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -183,6 +184,166 @@ export interface BaseStyle {
...
+ *     {
+ *         "label": "New shader",
+ *         "description": "Add a New 'shader' Styling Rule",
+ *         "body": {
+ *             "technique": "shader",
+ *             "when": "$1",
+ *             "attr": {}
+ *         }
+ *     },
+ *     {
```
zbigg: Landmark doesn't exist anyore.

okanishchev: Ok, will remove that.

@here/harp-datasource-protocol/lib/Theme.ts
okanishchev: In fact, there are two cases where `"none"` technique can be useful: 
 * to add when rule or some attrs that will be inherited by nested styles;
 * used together with `"when": "..."` and `"final": true`, to completely disable some features

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -708,12 +708,12 @@ export interface SolidLineTechniqueParams extends BaseTechniqueParams, Polygonal
...
      * opaque.
      */
-    opacity?: number;
+    opacity?: MaybeInterpolatedProperty<number>;
     // TODO: Make pixel units default.
     /**
      * Units in which different size properties are specified. Either `Meter` (default) or `Pixel`.
      */
-    metricUnit?: string;
+    metricUnit?: MaybeInterpolatedProperty<string>;
```
amsgames94: This shouldn't be a MaybeInterpolatedProperty. We cannot interpolate between `Meter` and `Pixel`.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -755,12 +757,12 @@ export interface DashedLineTechniqueParams extends BaseTechniqueParams, Polygona
...
      * opaque.
      */
-    opacity?: number;
+    opacity?: MaybeInterpolatedProperty<number>;
     // TODO: Make pixel units default.
     /**
      * Units in which different size properties are specified. Either `Meter` (default) or `Pixel`.
      */
-    metricUnit?: "Meter" | "Pixel";
+    metricUnit?: MaybeInterpolatedProperty<"Meter" | "Pixel">;
```
amsgames94: Same as above.

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -183,6 +184,157 @@ export interface BaseStyle {
...
+ *         "label": "New dashed-line",
+ *         "description": "Add a New 'dashed-line' Styling Rule",
+ *         "body": {
+ *             "technique": "dashed-line",
+ *             "when": "$1",
+ *             "attr": {
+ *                 "color": "#$2fff",
+ *                 "lineWidth": "^${3:1}"
```
amsgames94: Aren't you supposed to add the ``dashSize`` and ``gapSize`` properties here too?

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -183,6 +184,157 @@ export interface BaseStyle {
...
+ *         }
+ *     },
+ *     {
+ *         "label": "New text",
+ *         "description": "Add a New 'text' Styling Rule",
+ *         "body": {
+ *             "technique": "text",
+ *             "when": "$1",
+ *             "attr": {
+ *                 "scale": "^$2"
```
amsgames94:
```suggestion
 *                 "size": "^$2"
```

@here/harp-datasource-protocol/lib/Theme.ts
amsgames94: I'd say that the snippets are nice, but it would be nice to take a bit more care with them and put some meaningful initial values. Some like the ones for ``standard`` are a bit too empty...

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -386,11 +580,20 @@ export interface TextStyleDefinition {
 export interface GradientSkyParams {
     /** Sky type. */
     type: "gradient";
-    /** Color of the upper part of the gradient. */
+    /**
+     * @format color-hex
```
amsgames94: Be consistent with the typedoc (look at the members below).

@here/harp-geojson-datasource/lib/GeoJsonTile.ts
```diff
@@ -201,7 +202,7 @@ export class GeoJsonTile extends Tile {
             path,
             this.getRenderStyle(technique),
             this.getLayoutStyle(technique),
-            priority,
+            getPropertyValue(priority, this.mapView.zoomLevel),
```
amsgames94: I'm not sure if everything should be ``MaybeInterpolatedProperty``... Some cases like this look a bit weird.

I'd stick to stylable properties being interpolated, and be a bit more strict with ones that modify scene/runtime logic (like this).

Just an opinion though, would be nice to have more input on this.

@here/harp-map-theme/resources/berlin_tilezen_base.json
```diff
@@ -701,26 +694,6 @@
                             2
                         ]
                     },
-                    "secondaryWidth": {
```
amsgames94: Why removing this? The actually help to reuse the same vertex buffers for 2 different objects.

@here/harp-map-theme/resources/berlin_tilezen_day_reduced.json
```diff
@@ -609,26 +589,6 @@
                             2
                         ]
                     },
-                    "secondaryWidth": {
```
amsgames94: Same as above.

@here/harp-map-theme/resources/berlin_tilezen_day_reduced.json
```diff
@@ -1163,19 +1110,9 @@
                     "size": 14.4,
                     "opacity": 0.6,
                     "minZoomLevel": 17,
-                    "roughness": 1.95,
```
amsgames94: Why removing this also?

@here/harp-map-theme/resources/berlin_tilezen_night_reduced.json
```diff
@@ -620,26 +600,6 @@
                             2
                         ]
                     },
-                    "secondaryWidth": {
```
amsgames94: Same as above.

@here/harp-map-theme/resources/berlin_tilezen_night_reduced.json
```diff
@@ -1179,19 +1126,9 @@
                     "size": 14.4,
                     "opacity": 0.6,
                     "minZoomLevel": 17,
-                    "roughness": 1.95,
```
amsgames94: Same as above.

@here/harp-map-theme/resources/berlin_tilezen_day_reduced.json
okanishchev: It's not a valid properties for a text technique. As we would make automatic validation of styles in future, we need to remove all those.

@here/harp-map-theme/resources/berlin_tilezen_night_reduced.json
okanishchev: Also text technique don't have this properties

@here/harp-map-theme/resources/berlin_tilezen_night_reduced.json
okanishchev: This properties are from solid-line technique, not dashed-line, maybe left from copy-paste

@here/harp-map-theme/resources/berlin_tilezen_day_reduced.json
okanishchev: Properties of solid-line technique, not dashed-line.

@here/harp-map-theme/resources/berlin_tilezen_base.json
okanishchev: Yep, this works on solid-line, but not on dashed-line.

@here/harp-geojson-datasource/lib/GeoJsonTile.ts
okanishchev: I'm simply fixing already working feature - we have interpolated priority prop in our styles right now, but not on every technique. And our code supports interpolated properties on almost all techniques. So I think, make all techniques to support priority interpolation would be good decision.

@here/harp-datasource-protocol/lib/Theme.ts
okanishchev: Yay! It's a typo. Thanks

@here/harp-datasource-protocol/lib/Theme.ts
okanishchev: Good catch.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
okanishchev: Why not? We can't do it linearly or in cubic now, but discrete interpolation can be done simply.
For example we can say that some road will be 15 meters wide at level 14 or higher, or 8 pixels on lower levels.

@here/harp-datasource-protocol/lib/Theme.ts
okanishchev: Yep, think I should insert some special properties for every technique.

okanishchev:
I've populated technique snippets with some related properties. Hope they would be more useful now.

@here/harp-map-theme/resources/berlin_tilezen_day_reduced.json
amsgames94: Got it.

@here/harp-map-theme/resources/berlin_tilezen_base.json
amsgames94: You're right, didn't realize.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
okanishchev: We can't because right now string interpolation is restricted to color values only.
So this should be implemented in another ticket.

PR #501, opened 2019-05-30
HARP-6000: Fix mouseUp handling in cross-origin iframes.

zbigg:

### DIFF
--- a/@here/harp-map-controls/lib/MapControls.ts
+++ b/@here/harp-map-controls/lib/MapControls.ts
@@ -802,7 +802,6 @@ export class MapControls extends THREE.EventDispatcher {
             return;
         }
 
-        event.preventDefault();
         event.stopPropagation();
 
         if (this.m_state !== State.NONE) {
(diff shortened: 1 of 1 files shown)


@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -802,7 +802,7 @@ export class MapControls extends THREE.EventDispatcher {
             return;
         }
 
-        event.preventDefault();
+        //event.preventDefault();
```
nzjony: is there some reason for this?

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -827,17 +827,27 @@ export class MapControls extends THREE.EventDispatcher {
...
+            // Registering for parent/top mouse events is not allowed and throws exceptions if we're
+            // in cross-origin frame. We can safely ignore it and don't handle them.
+        }
 
         this.m_cleanupMouseEventListeners = () => {
             window.removeEventListener("mousemove", onMouseMove);
-            window.top.removeEventListener("mousemove", onMouseMove);
             window.removeEventListener("mouseup", onMouseUp);
-            window.top.removeEventListener("mouseup", onMouseUp);
+            // try {
```
nzjony: do you mean to comment this out?

@here/harp-map-controls/lib/MapControls.ts
zbigg: Sorry, i forgot to push cleaned-up patch, back to question. Will do in a moment.

> is there some reason for this?

Actually, i didn't done very thorough research, but without `event.preventDefault` `mouseDown` events properly arrive, even when we drag mouse outside our iframe.

With `event.preventDefault()` we just don't get `mouseUp` that happens outside of our frame.

I don't know why, the only source is enigmatic blog post: https://taye.me/blog/tips/2015/11/16/mouse-drag-outside-iframe/

@here/harp-map-controls/lib/MapControls.ts
zbigg: removed, outdated

@here/harp-map-controls/lib/MapControls.ts
zbigg: So, summarizing it looks like  `window.top` listeners were needed for `iframe` scenario because we called `event.preventDefault`.
Now, without we can safely remove `window.top` event listeners and thus cross-origin issues.

PR #507, opened 2019-06-03
MINOR: batch fill features outline geometries into 1 mesh

okanishchev:
Lines for fill technique outline created as separate mesh for each fill feature, causing performance loss.
Multiple urban fill areas can be viewed around Budapest at 8.00 zoom level.


amsgames94:
Will be addressed in #514 .

PR #513, opened 2019-06-04
HARP-5904: Implement lookAt method.

Astrak:


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1490,7 +1490,28 @@ export class MapView extends THREE.EventDispatcher {
...
+    lookAt(target: GeoCoordinates, distance: number, pitch: number = 0, yaw: number = 0) {
+        const zoomLevel = MapViewUtils.calculateZoomLevelFromDistance(distance, this);
+        const geoLocation = MapViewUtils.getCameraCoordinatesFromTargetCoordinates(
+            target,
+            zoomLevel,
+            yaw,
+            pitch,
+            this
+        );
+        this.setCameraGeolocationAndZoom(geoLocation, zoomLevel, yaw, pitch);
```
ninok: Can we not set the (projected)camera position directly instead of setCameraGeolocationAndZoom?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1490,7 +1490,28 @@ export class MapView extends THREE.EventDispatcher {
...
 
     /**
-     * Moves the camera to specified geocoordinates, sets desired zoom level and adjusts yaw and
+     * The method that sets the camera to the desired angle (`pitch`) and distance (in meters) to
+     * the `target` location.
+     *
+     * @param target The location to look at.
+     * @param distance The distance of the camera to the target in meters.
+     * @param pitch The camera tilt angle in degrees.
+     * @param yaw The camera yaw angle in degrees.
```
ninok: We should align with harp-native, i.e. use azimuth and tilt

Astrak: You mean, replace `yaw` and `pitch` with `azimuth` and `tilt`? We use these values extensively and they are exposed externally. 

Should I do this in this ticket?

@here/harp-mapview/lib/MapView.ts
Astrak: Yes! Thanks! It solved my problem.

@here/harp-mapview/lib/MapView.ts
Astrak: Another thing: I chose the parameter `distance` in order to align with native as well, but the competition prefers a zoom level number. With `setCameraGeoLocationAndZoom` that is also what we did. Do we switch to distance in meters indeed? Setting a global view might be painful.

@here/harp-examples/src/camera_orbit.ts
```diff
@@ -0,0 +1,100 @@
...
+ * ```
+ *
+ * Then we listen to render events to trigger new `lookAt` calls with progressing yaw angle offsets:
+ * ```typescript
+ * [[include:harp_gl_camera_orbit_example_1.ts]]
+ * ```
+ *
+ * Here a GUI is also set up so as to fiddle with the pitch and distance from the page.
+ */
+export namespace HelloWorldExample {
```
nzjony: This conflicts with the hello.ts example name.

@here/harp-mapview/lib/MapView.ts
ninok: If we put zoomlevel here instead of distance, this means we have to define our zoomlevel as distance to the look-at point. Currently this is only true if the tilt is less than 60 degrees and we have a flat map (i.e. no globe and no terrain). 
In mapview Utils.ts we have `calculateDistanceToGroundFromZoomLevel` but that also makes not much sense b/c it is not the distance to ground.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1490,33 +1490,53 @@ export class MapView extends THREE.EventDispatcher {
     }
 
     /**
-     * Moves the camera to specified geocoordinates, sets desired zoom level and adjusts yaw and
+     * The method that sets the camera to the desired angle (`pitch`) and distance (in meters) to
```
nzjony: references to variables should use [[pitchDeg]] and [[distance]]

@here/harp-mapview/lib/Utils.ts
```diff
@@ -129,8 +129,8 @@ export namespace MapViewUtils {
      *
      * @param targetCoordinates Coordinates of the center of the view.
      * @param zoom Zoom level.
-     * @param yaw Camera yaw.
-     * @param pitch Camera pitch.
+     * @param yawDeg Camera yaw in degrees.
+     * @param pitchDeg Camera pitch in degrees.
      * @param mapView Active MapView, needed to get the camera fov and map projection.
      */
     export function getCameraCoordinatesFromTargetCoordinates(
```
ninok: Why is this method taking the zoomlevel and not the distance?

PR #517, opened 2019-06-05
HARP-5657: TechniqueProcessor Definition.

amsgames94:


@here/harp-geometry/lib/TechniqueProcessor.ts
```diff
@@ -0,0 +1,124 @@
...
+     * If not specified in style file, [[StyleSetEvaluator]] will assign monotonically increasing
+     * values according to style position in file.
+     */
+    readonly renderOrder: number;
+}
+
+/**
+ * Base class that handles processing data accordingly to its assigned [[Technique]].
+ */
+export abstract class TechniqueProcessor {
```
zbigg: While i agree, that unification of geometry processing is needed, i will raise key problem from the beginning:

Separation of "decoder" code that lands in worker bundle and "main thread" code.

Class based design binds all this code `createBuffers` and `vertexShader` and finally `createObjects` into one piece of unbreakable code and we result with `MapView` (unnecessarily) and friends linked into decoders ... and vice versa - "createBuffers" linked into main thread bundle.

_If_ this is a concern, then unfortunately, `TechniqueProcessor` has to be more of like a "concept" how to structure code and maybe set of interfaces and implementations, but they cannot be one language entity (class, function, record), so they are really separable from link time perspective.

ninok: I had a quick chat with Andres yesterday and we already identified that the technique processors have to be registered twice:
1. In MapView for the main thread to create objects
2. Somewhere else for the decoders to create buffers.
So you are absolutely right, this could be split in two classes, one for the main thread and one for the web worker.

amsgames94: Splitting this into 2 classes will make it harder to maintain, since it's no longer a single class per technique. ``createBuffers``can be called from both the main thread and the decoding threads (as showcased in one of our examples), and ``createObjects`` should never be called from decoders, since it's aimed to be connected to MapView, as you guys stated.

You're not supposed to instantiate this classes (they're fully static classes, maybe we can change to singletons?), so even if the code is both on the main and decoder bundles, I think we should use these methods wisely (no need to link MapView to the decoders) and keep them in a single coherent place where we can understand the data flow from the fetched tiles to the three.js objects on screen.

zbigg: > Splitting this into 2 classes will make it harder to maintain, since it's no longer a single class per technique. 

Yes, it will be a little harder. Unfortunately, we having decoders makes us "client-server" architecture where client/server implementation shouldn't leak to each other - even if they are very tightly coupled.
We have to work with this "cost".

Now, with `SceneState` we're much better, because at least MapView doesn't leak so much, issue is still there.

What i'm trying to say, is that we should split this "buffer" creation and "3d objects" creation to separate classes language entities, preferably files, so "linker" (webpack/rollup) can link only needed functions in respective bundles.

Specifically, we don't want this in decoder bundle:
 * we don't want shaders in bundles
 * we don't want whole shaders, Object3d and Meshes, PerspectiveCamera and stuff in decoder bundlers

And we don't "buffer creation code"  like `createLines`, `Extuder",  `Outliner" in main bundle.

...
Yes, someone may say that JS code is cheap, but our bundles are already "500k" gzipped/minified or more. They shouldn't grow without a reason.

zbigg: > ~500k

Ok, i was too pessimistic, actual sizes are not so bad:
```
200K	./harp-decoders.min.js
56K	./harp-decoders.min.js.gz
116K	./harp.min.js.gz
460K	./harp.min.js
```

@here/harp-mapview/lib/MapView.ts
```diff
@@ -784,6 +785,7 @@ export class MapView extends THREE.EventDispatcher {
         this.m_camera.up.set(0, 0, 1);
         this.m_lookAtDistance = 0;
         this.m_focalLength = 0;
+        this.m_scene.userData = this;
```
okanishchev: Object3D's userData is subject to JSON.stringify: https://github.com/mrdoob/three.js/blob/dev/src/core/Object3D.js#L749

in case of complex object it may throw circular reference (or similar) error.

amsgames94: This is a left-over from previous solution, it should have been removed. Good catch ;)

@here/harp-geometry/lib/TechniqueProcessor.ts
```diff
@@ -0,0 +1,133 @@
...
+    static readonly uniforms: Map<string, THREE.Uniform>;
+    /**
+     * Vertex shader source code.
+     */
+    static readonly vertexShader: string;
+    /**
+     * Fragment shader source code.
+     */
+    static readonly fragmentShader: string;
+
```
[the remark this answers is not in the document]
amsgames94: For sure we want to conform to three.js default materials/shaders, but that's the easy part. We want to also use these interfaces to create geometries for our custom materials (solid-line, extruded-polygon, etc.) in an abstract way that allow us to ensure the correct processing and runtime display of the data without hard-coding this logic into the decoding pipeline. Decoding will be just an use-case of this interfaces.

@here/harp-geometry/lib/TechniqueProcessor.ts
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Checks if this [[TechniqueProcessor]] can process the specified data.
+     *
+     * @param data Data to check for support.
+     */
+    static supportsData(data: ArrayBufferLike | {}): boolean {
```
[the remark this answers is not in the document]
amsgames94: I agree, but currently cannot think of anything better (and to be honest, is the same types we use in the ``WorkerDecoderProtocol``...). Any suggestion? It's tricky to be flexible and strict at the same time.

@here/harp-geometry/lib/TechniqueProcessor.ts
```diff
@@ -0,0 +1,133 @@
...
+    ): THREE.Object3D | THREE.Object3D[] {
+        return [];
+    }
+
+    protected static updateUniforms(
+        object: THREE.Object3D,
+        params: TechniqueParams,
+        sceneState: SceneState
+    ) {
+        object.onBeforeRender = (renderer, scene, camera, geom, mat, group) => {
```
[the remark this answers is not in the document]
amsgames94: I understand your concerns, but without proper interpolant implementation (which might be on the way, and it's not there), we will do this anyways for almost all materials (look at #496 ). 

We cannot have runtime interpolation right now any other way, and we're not adding any overhead here, just rearranging where the logic is located (which is now hardcoded in ``TileGeometryCreator``).

@here/harp-geometry/lib/TechniqueProcessor.ts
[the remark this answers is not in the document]
robertoraggi: I understand that the code of solid-line and extruded-polygons is sparse a bit all over our code base and we need to find way to consolidate that code. But I do expect something  called TechniqueProcessor to work on all kind of techniques and many of our techniques are based on standard three materials and geometries, in these cases we may look at how to use shader overrides to inject snippets in the built-in three.js shaders. At least to me, an abstraction working on THREE.Uniform and vertex/fragmentShader seems a bit too lowlevel.

@here/harp-geometry/lib/TechniqueProcessor.ts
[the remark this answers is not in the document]
zbigg: If this is a bare "drawing" technique intended at _just_drawing_lines_ i would think that they should accept world coordinates, rather bar buffers.

I think that mere reference to `WorkerDecoderProtocol` shows misunderstanding (possibly mine, but still communication issue) of purpose this task.

Advertised as "refactor-out" bare technique processing, now you're reusing interface of "tile decoder" (`ArrayBufferLike | {}` is basically MVT or MvtJson or GeoJSON) so, it's not layer we would like to interface with technique.

So, on which level technique API should be defined ? Some aspects.
 - should technique know/understand transfer formats? no, it should accept bare polygons/lines/points in some space
 - should technique know/understand projections? no, it should work in eucludian space, like world/tile space, so space doesn't matter as long as it's affine-transformable to world space
 
Ergo -> 
We've already interfaces in `IGeometryProcessor.ts` which IMHO are best suitable for technique input - with one difference, they should use convey world-space compatible coordinates instead of geocoords.

Varia:
_if_ we would lay a line on curved area (terrain or sphere), it's caller responsibility (decoder in this case) to divide geometry and add proper z-offsets

@here/harp-geometry/lib/TechniqueProcessor.ts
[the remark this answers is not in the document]
zbigg: @robertoraggi, @amsgames94 

I think that we cannot do much in scope of this patch, so this one is ok.

Ultimate solution, would need `kind-of` MaterialCache, so we would reuse materials in tiles and pre-render pass over all materials in current scene to update dynamic uniforms dependent on time, pixel2world, and whatnot.

This would save us from tons of `onBeforeRender` calls, at least those that doesn't need toggle stencil buffer.

I must write-out this idea as patch for discussion.

amsgames94:
The ideas and concept in this PR will be taken into account into a future reworking of the current techniques/material system.

PR #523, opened 2019-06-06
HARP-5937: Add APIs to reproject world coordinates.

robertoraggi:
Also, make Projection an abstract class and a default
implementation of reprojectPoint.

### DIFF
--- a/@here/harp-geoutils/lib/projection/Projection.ts
+++ b/@here/harp-geoutils/lib/projection/Projection.ts
@@ -30,11 +30,11 @@ export enum ProjectionType {
  * `Projection` is used to convert positions from geo coordinates to world coordinates and vice
  * versa.
  */
-export interface Projection {
+export abstract class Projection {
     /**
      * The type of this [Projection].
      */
-    readonly type: ProjectionType;
+    abstract get type(): ProjectionType;
 
     /**
      * Returns the world extents in world coordinates.
@@ -43,7 +43,7 @@ export interface Projection {
      * @param maxElevation The maximum elevation in meters.
      * @param result The optional object that will be used to create the resulting bounding box.
      */
-    worldExtent<Bounds extends Box3Like>(
+    abstract worldExtent<Bounds extends Box3Like>(
         minElevation: number,
         maxElevation: number,
         result?: Bounds
@@ -63,7 +63,7 @@ export interface Projection {
      * @param result The optional object used to store the resulting world position, result must
      * implement [[Vector3Like]].
      */
-    projectPoint<WorldCoordinates extends Vector3Like>(
+    abstract projectPoint<WorldCoordinates extends Vector3Like>(
         geoPoint: GeoCoordinatesLike,
         result?: WorldCoordinates
     ): WorldCoordinates;
@@ -80,7 +80,7 @@ export interface Projection {
      *
      * @param worldPoint The position in world coordinates.
      */
-    unprojectPoint(worldPoint: Vector3Like): GeoCoordinates;
+    abstract unprojectPoint(worldPoint: Vector3Like): GeoCoordinates;
 
     /**
      * Projects bounds in geo coordinates to a bounding box in world coordinates.
@@ -93,7 +93,7 @@ export interface Projection {
      *
      * @param geoBox The bounding box in geo coordinates.
      */
-    projectBox(geoBox: GeoBox): Box3Like;
+    abstract projectBox(geoBox: GeoBox): Box3Like;
 
     /**
      * Projects bounds in geo coordinates to a bounding box in world coordinates.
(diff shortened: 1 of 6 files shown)


@here/harp-geoutils/lib/projection/Projection.ts
```diff
@@ -123,43 +123,93 @@ export interface Projection {
...
+
+    /**
+     * Reproject a world position from the given source [[Projection]].
+     *
+     * @param sourceProjection The source projection.
+     * @param worldPos A valid position in the world space defined by the source projection.
+     * @param result The resulting position reprojected using this [[Projection]].
+     * @hidden
+     */
+    reprojectPoint(
```
ninok: Why do we need to declare the method three times? Is this implementation not enough?

robertoraggi: it's for avoiding possible runtime errors when using the `Projection` APIs. In fact, without restricting the signatures the following code will result in a runtime error:
```
const p = projection.reprojectPoint<THREE.Vector3>(sourceProjection, worldP);
p.set(0,0,0); // runtime error THREE.Vector3
```

On the other hand, the restricted APIs works as expected:
```
const p = projection.reprojectPoint(sourceProjection, worldP);
const {x,y,z} = p; // OK, p is a Vector3Like
p.sub(0, 0, 0); // OK, type error, Vector3Like doesn't have methods.

const p = projection.reprojectPoint(sourceProjection, worldP, new THREE.Vector3());
p.set(0, 0, 0); // OK, p is a THREE.Vector3.
const {x,y,z} = p; // OK, THREE.Vector3 is compatible with Vector3Like
```

@here/harp-geoutils/lib/projection/Projection.ts
```diff
@@ -123,43 +123,93 @@ export interface Projection {
...
+        sourceProjection: Projection,
+        worldPos: Vector3Like,
+        result?: Vector3Like
+    ): Vector3Like {
+        if (sourceProjection === this) {
```
nzjony: super edge case: should you check instanceof as well?

@here/harp-geoutils/lib/projection/Projection.ts
ninok: Thanks for the detailed explanation.

@here/harp-geoutils/lib/projection/Projection.ts
```diff
@@ -123,43 +123,93 @@ export interface Projection {
...
+            if (result === undefined) {
+                return { x: worldPos.x, y: worldPos.y, z: worldPos.z };
+            }
+            result.x = worldPos.x;
+            result.y = worldPos.y;
+            result.z = worldPos.z;
+            return result;
+        }
+        return this.projectPoint(sourceProjection.unprojectPoint(worldPos), result);
```
nzjony: you mentioned that there are some short cuts to transform between different projections, i.e. webmercator -> sphere, how do you handle that with this API?

nzjony:
Don't want to block, my comments can be in follow up commit if necessary.

@here/harp-geoutils/lib/projection/Projection.ts
robertoraggi: there shouldn't be any need for checking with `instanceof`. Our projections are implemented as singleton (and the classes are not exported, well MercatorProjection is exported but that is only because WebMercator subclasses it). By the way the only exported symbols are `mercatorProjection, webMercatorProjection, identityProjection, sphereProjection` and so on..). So there should be one and only one instance of each projection so the strict equality check should be enough.

@here/harp-geoutils/lib/projection/Projection.ts
robertoraggi: I was thinking to do something like this:
```
class SphereProjection extends Projection {
    // OK, reprojectPoint is compatible with the runtime method signature of Projection.
    // and since SphereProjection is not exported with don't have to care about wrong
    // usages of Generics.
    reprojectPoint(sourceProjection: Projection,
        worldPos: Vector3Like, result?: Vector3Like): Vector3Like {
        if (sourceProjection === mercatorProjection) {
            ...
        }
        ...
    }
```

robertoraggi: If needed we could also maintain a global list of `converters` so users could inject their own reprojection and we could call it in the default implementation of reprojectPoint. I'm not sure it is need, but the following should work:
```
export abstract class Projection {
   ...
    static registerReprojection(...): void {
        reprojections.push(...)
    }

    // ... reprojectPoint restricted signatures...

    // runtime reprojectPoint method
    reprojectPoint(sourceProjection: Projection, ...) {
        ...
        const reprojection = Projection.findReprojection(this, sourceProjection);
        if (reprojection) ... else ...
    }
```

PR #528, opened 2019-06-11
Fix decoder url detection for tiler in harp.js bundle.

zbigg:


zbigg:
There is bug, don't merge!

PR #529, opened 2019-06-11
HARP-6042: StyleCacheId bugfix.

amsgames94:


@here/harp-mapview/lib/text/TextStyleCache.ts
```diff
@@ -35,7 +35,7 @@ export function computeStyleCacheId(
     technique: Technique,
     zoomLevel: number
 ): string {
-    return `${datasourceName}_${technique.renderOrder}_${zoomLevel}`;
+    return `${datasourceName}_${(technique as any)._index}_${zoomLevel}`;
```
zbigg: This is hard issue (interpolated text or POI properties1!), so it's still workaround, but ..

Why not `Math.floor(zoomLevel)` here, so patch is smaller ?

amsgames94: Interpolated text and poi properties need a more complex setup (since their runtime evaluation is way different from all the other objects, look at ``TextElementsRenderer.renderTextElements()``), so until we don't rework the text placement system, we will not allow interpolation of these properties.

And the patch is minimal, and looks more correct to explicitly floor the input to this function to do it behind the curtains (inside the function body).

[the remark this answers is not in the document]
amsgames94: Applied this suggestion in last patch. Thanks!

@here/harp-mapview/lib/ThemeLoader.ts
```diff
@@ -82,6 +82,7 @@ export class ThemeLoader {
         }
 
         if (theme.styles) {
+            let globalThemeIndex = 0;
```
zbigg: remove ?

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -178,6 +178,12 @@ export interface IndexedTechniqueParams {
      * @hidden
      */
     _index: number;
+
+    /**
+     * Optimization: Index into [[StyleSet]] in [[StyleSetEvaluator]].
```
zbigg: Maybe this is more informative:

> unique, deterministic index of `Technique` within source [[StyleSet]].

@here/harp-mapview/lib/ThemeLoader.ts
amsgames94: Thanks for catching this one! Was a leftover.

@here/harp-datasource-protocol/lib/Techniques.ts
amsgames94: Good suggestion, will apply.

PR #533, opened 2019-06-11
HARP-6101 Adding clear method to DataSource

nzjony:

### DIFF
--- a/@here/harp-mapview-decoder/lib/TileDataSource.ts
+++ b/@here/harp-mapview-decoder/lib/TileDataSource.ts
@@ -164,6 +164,10 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
         this.decoder.dispose();
     }
 
+    clear() {
+        this.m_tileLoaderCache.clear();
+    }
+
     ready(): boolean {
         return this.m_isReady && this.m_options.dataProvider.ready();
     }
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/DataSource.ts
```diff
@@ -151,6 +151,14 @@ export abstract class DataSource extends THREE.EventDispatcher {
         // to be overloaded by subclasses
     }
 
+    /**
+     * Clears this `DataSource`. Clearing as opposed to [[dispose]] just resets any internal state,
+     * as opposed to actually destroying allocated objects.
+     */
+    clear() {
```
zbigg: Continuing internal discussion.

> it acceptable to add a clear method to a DataSource

For sure it's acceptable, but naming could be better like `resetState`, `clearCaches` ...

On the other hand, the responsibility for this cache is in `TileDataSource` so this class should know about this and just clear any needed cache in it's `.setStyleSet` implementation and we don't need `DataSource#clear` (and discussion about it's name and semantics).

nzjony: Agreed.

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -164,6 +164,10 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
         this.decoder.dispose();
     }
 
+    clear() {
```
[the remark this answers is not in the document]
nzjony: @robertoraggi, I like @zbigg 's idea of just having the cache clearing in the `setStyleSet` method.

In the case you mentioned, the `clear()` method has a clear (forgive the pun) correlation to the `add()` method, however with the TileDataSource, it isn't apparent when `clear()` should be called, because the user of the class can't (directly) add anything to it.

PR #535, opened 2019-06-12
MINOR: Fix azimuth in examples.

Astrak:
Azimuth screwed up in examples.

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -1538,14 +1538,19 @@ export class MapView extends THREE.EventDispatcher {
      * @param target The location to look at.
      * @param distance The distance of the camera to the target in meters.
      * @param tiltDeg The camera tilt angle in degrees (0 is vertical).
-     * @param azimuthDeg The camera azimuth angle in degrees relative to the target (0 is north).
+     * @param azimuthDeg The camera azimuth angle in degrees (0 means looking north).
      */
-    lookAt(target: GeoCoordinates, distance: number, tiltDeg: number = 0, azimuthDeg: number = 0) {
-        MapViewUtils.setRotation(this, 180 - azimuthDeg, tiltDeg);
+    lookAt(
+        target: GeoCoordinates,
+        distance: number,
+        tiltDeg: number = 0,
+        azimuthDeg: number = 0
+    ): void {
+        MapViewUtils.setRotation(this, azimuthDeg, tiltDeg);
         this.geoCenter = MapViewUtils.getCameraCoordinatesFromTargetCoordinates(
             target,
             distance,
-            180 - azimuthDeg,
+            azimuthDeg,
             tiltDeg,
             this
         );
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1538,9 +1538,15 @@ export class MapView extends THREE.EventDispatcher {
      * @param target The location to look at.
      * @param distance The distance of the camera to the target in meters.
      * @param tiltDeg The camera tilt angle in degrees (0 is vertical).
-     * @param azimuthDeg The camera azimuth angle in degrees relative to the target (0 is north).
-     */
-    lookAt(target: GeoCoordinates, distance: number, tiltDeg: number = 0, azimuthDeg: number = 0) {
+     * @param azimuthDeg The camera azimuth angle in degrees relative to the target (180 means that
```
zbigg: Missed chance to review original submission, but it's never too late.

TLDR - azimuth 0 should be north.

From internal discussion, it appears that `lookAt` parameters are relative to "target" point which i find heavily misleading  esp. w.r.t to `azimuth` parameter which now means 
1) looking at north = 180
2) looking at south = 0

So, you really have to wrap your head around concept that you setup eye  in relative to target, when "intuition" and default is that usually you setup direction of eye - camera.

This is additionally counter-intuitive as everywhere in culture `azimuth` is defined as more-or-less as direction  of "someone looking at something" from "north" [1], [2]. So, north is 0.
On compass, north is 0.

Note, "someone looking at something" in our case, it's "camera looking at something", so it matches `lookAt` name perfectly.

So, i think that in general, `lookAt` parameters should be relative to camera it's not `target` looking at `camera` but `camera` looking at target.

@ninok, @Astrak 

Btw, It would be nice to document direction also (clockwise/counter-clockwise) so, users understand without experimenting if `east`/`west` is 90 or -90.

[1] https://en.wikipedia.org/wiki/Azimuth
[2] http://www.differencebetween.net/science/difference-between-azimuth-and-bearing/

Astrak: I think it depends how we expect this method to read:
1. look at this point with this camera orientation
1. look at this point from this angle

I believe that we are free to pick either of the above, as long as we make it clear. If we decide that the origin of the system is the lookAt point, then to look north the camera needs to be at the south of the target, azimuth 180, and it complies with these references. 

I initially implemented it as you suggest in the PR when Nino asked for this method, but it also made sense for me in the end.

robertoraggi: I wonder if we could split the APIs in two methods. One method could be used to set up the camera lookAt direction, and an additional method e.g. `rotateCameraAround()` could be used to rotate the camera around the target or around its position.

ninok: @zbigg, @Astrak: To be honest I was not reviewing this that closely but my assumption was as well, that an azimuth of 0 means looking north. Also I agree that this method should read as look-at target from camera. This is a more natural interpretation than the other way around.

PR #537, opened 2019-06-12
HARP-6096: Add fast conversions between the WEB Mercator projections.

robertoraggi:
Also, this change places the "WEB Mercator"-projection variants
currently supported by @here/harp-geoutils in the same file. This is
needed to avoid possible circular dependencies (e.g reprojectPoint()
needs to access both webMercatorProjection and mercatorProjection).

Placing both variants in the same file has the nice side-effect to
remove the need to export the class MercatorProjection.


@here/harp-geoutils/lib/projection/MercatorProjection.ts
```diff
@@ -183,9 +183,131 @@ export class MercatorProjection extends Projection {
...
+        worldPos: Vector3Like,
+        result?: Vector3Like
+    ): Vector3Like {
+        // this implementation of [[reprojectPoint]] supports both
+        // [[WebMercatorProjection]] and [[MercatorProjection]]. The only
+        // difference betweeen these two variants of WEB Mercator
+        // is in the orientation of the Y axis, so we just flip Y coordinates
+        // when reprojecting between them.
+        if (
+            sourceProjection !== this &&
```
ninok: if sourceProjection === this we should return the input point but instead you call super.reprojectPoint ?!

@here/harp-geoutils/lib/projection/MercatorProjection.ts
```diff
@@ -183,9 +183,131 @@ export class MercatorProjection extends Projection {
...
+            (sourceProjection === webMercatorProjection || sourceProjection === mercatorProjection)
```
ninok: If source is mercator (and this is also mercator) we should not flip the y axis

@here/harp-geoutils/lib/projection/MercatorProjection.ts
robertoraggi: when sourceProjection === this we call the base implementation that already does what you expect.

@here/harp-geoutils/lib/projection/MercatorProjection.ts
robertoraggi: yes, and that is captured by "sourceProjection !== this"

PR #538, opened 2019-06-13
HARP-6160: Improve the classification of outer/inner rings.

robertoraggi:
Use the winding of the 1st ring of a polygon feature
to distinguish between inner and outer rings.

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
+++ b/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
@@ -837,9 +837,13 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
                 }
 
                 // Repeat the process for all the inner rings (holes).
+                const outerRingWinding = ring.winding;
                 const holes: number[] = [];
-                for (; ringIndex < rings.length && rings[ringIndex].isInnerRing; ++ringIndex) {
-                    const current = rings[ringIndex];
+
+                // Iterate over the inner rings. The inner rings have the oppositve winding
+                // of the outer rings.
+                while (ringIndex < rings.length && rings[ringIndex].winding !== outerRingWinding) {
+                    const current = rings[ringIndex++];
                     holes.push(vertices.length / stride);
 
                     // As we are predicting the indexes before the vertices are added,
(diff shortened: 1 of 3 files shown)


@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -837,9 +837,13 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
 
                 // Repeat the process for all the inner rings (holes).
+                const outerRingWinding = ring.winding;
                 const holes: number[] = [];
-                for (; ringIndex < rings.length && rings[ringIndex].isInnerRing; ++ringIndex) {
-                    const current = rings[ringIndex];
+
+                // Iterate over the inner rings. The inner rings have the oppositve winding
+                // of the outer rings.
+                while (ringIndex < rings.length && rings[ringIndex].winding !== outerRingWinding) {
```
zbigg: I'm worried, that this loop stops on first "subring" that has same winding as ultimate parent ring, whereas old loop processed all rings. Is this intended ?

@here/harp-omv-datasource/lib/OmvDecoder.ts
```diff
@@ -70,11 +70,7 @@ export class Ring {
         readonly contourOutlines?: boolean[],
         readonly points: number[] = contour
     ) {
-        this.isOuterRing = this.area() < 0;
-    }
-
-    get isInnerRing(): boolean {
-        return !this.isOuterRing;
+        this.winding = Math.sign(this.area());
```
ninok: What about 0? The winding will be 0 as well. Not sure if the code will handle it properly.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
zbigg: Nevermind, outerloop takes care of processing remaining "outer" rings

@here/harp-omv-datasource/lib/OmvDecoder.ts
robertoraggi: that should not really happen, unless the data is broken.

robertoraggi: i changed the condition to cover this case

PR #541, opened 2019-06-13
HARP-6151: Refactored the OMV decoder to work in webmercator space.

robertoraggi:
Changed the OMV decoder and its helper classes to work in webmercator
space and use `Projection.reprojectPoint` to fast reproject between
projected coordinates. This change should avoid a huge number of
conversions from WEB Mercator to geo space and then back to web mercator
or globe.

### DIFF
--- a/@here/harp-omv-datasource/lib/IGeometryProcessor.ts
+++ b/@here/harp-omv-datasource/lib/IGeometryProcessor.ts
@@ -5,7 +5,7 @@
  */
 
 import { MapEnv } from "@here/harp-datasource-protocol/index-decoder";
-import { GeoCoordinates } from "@here/harp-geoutils";
+import { Vector3 } from "three";
 
 /**
 + *  An interface to represent inner and outer rings of a [[IPolygonGeometry]].
@@ -14,13 +14,13 @@ export interface IRing {
     /**
      * The [[GeoCoordinates]] of this [[IRing]].
      */
-    coordinates: GeoCoordinates[];
+    readonly positions: Vector3[];
 
     /**
      * Boolean array representing if the edges of this [[IRing]] should be considered as potential
      * outlines.
      */
-    outlines?: boolean[];
+    readonly outlines?: boolean[];
 }
 
 /**
@@ -30,7 +30,7 @@ export interface IPolygonGeometry {
     /**
      * The rings of this polygon.
      */
-    rings: IRing[];
+    readonly rings: IRing[];
 }
 
 /**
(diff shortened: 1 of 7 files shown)


@here/harp-omv-datasource/lib/IGeometryProcessor.ts
```diff
@@ -30,17 +30,17 @@ export interface IPolygonGeometry {
...
+    readonly rings: IRing[];
 }
 
 /**
  * An interface to represent line geometries.
  */
 export interface ILineGeometry {
     /**
-     * The [[GeoCoordinates]] of the line.
+     * The positions of this [[IlineGeometry]] projected in world space.
```
ninok: What world space? The one of the tiling scheme or the one of the map view? Please also add comments to the Vector3 above.
Maybe we should also consider to add a new type that adds additional information about the projection to Vector3 to make debugging and runtime checking easier (we would not even need to pass the source projection to the reproject method)

robertoraggi: good point, it's the world space of the tiling scheme. I will update the documentation.
In theory the processor (e.g. the decoders) that get these data have access to the TilingScheme. So I'm not sure we need more than a paragraph in the documentation mentioning that that the geo coordinates can be obtained by using `tilingScheme.projection`.

ninok: For this change it's fine to just mention it.
I was just thinking that it might come in handy in the future if the world coordinates know what projection was used to produce them so the reprojectPoint would not need an additional sourceProjection parameter and we could also add some runtime checks in some places if the world coordinates are in the "wrong" world.

@here/harp-omv-datasource/lib/OmvData.ts
```diff
@@ -185,15 +185,13 @@ export class GeometryCommands {
...
     ) {
-        const longitudeScale = geoBox.longitudeSpan / extent;
-
         const geometryCount = geometry.length;
 
         const { north, west } = geoBox;
-
         const N = Math.log2(extent);
-
+        const scale = Math.pow(2, tileKey.level + N);
```
ninok: Shouldn't that code be in the tiling scheme?

@here/harp-omv-datasource/lib/OmvData.ts
```diff
@@ -221,9 +219,15 @@ export class GeometryCommands {
                         yCoords.push(currY);
                     }
 
-                    const longitude = west + currX * longitudeScale;
-                    const latitude = tile2lat(top + currY, tileKey.level + N);
-                    commands.push({ kind, latitude, longitude });
+                    const R = EarthConstants.EQUATORIAL_CIRCUMFERENCE;
```
ninok: Same as above!?

@here/harp-omv-datasource/lib/OmvData.ts
robertoraggi: yes but unfortunately our TileKey/TilingScheme doesn't handle tiles with level greater than 23. So we won't be able to represent anything with level 15+

@here/harp-omv-datasource/lib/OmvData.ts
robertoraggi: I guess if we really want we could export some more function from TileKey/SubdivisionScheme/TilingScheme to deal with this but I'm not sure it is worth doing it. Also, it will create a bit of an additional APIs. I don't have strong feelings about it, so if we want I can export helpers to deal with the normalized mercator space and higher subdivisions.

PR #542, opened 2019-06-13
MINOR: Updated yeoman generator to use latest packages and look at ca…

ninok:
…mera

Signed-off-by: Nino Kettlitz <1396039+ninok@users.noreply.github.com>

### DIFF
--- a/@here/generator-harp.gl/generators/app/templates/index.ts
+++ b/@here/generator-harp.gl/generators/app/templates/index.ts
@@ -22,5 +22,7 @@ window.addEventListener("resize", () => {
 });
 
 // center the camera to New York
-mapView.camera.position.set(0, 0, 1600);
-mapView.geoCenter = new GeoCoordinates(40.70398928, -74.01319808, 0);
+mapView.lookAt(new GeoCoordinates(40.70398928, -74.01319808), 1500, 40, 190);
+
+// make sure the map is rendered
+mapView.update();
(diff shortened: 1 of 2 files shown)


@here/generator-harp.gl/generators/app/templates/index.ts
```diff
@@ -22,5 +22,7 @@ window.addEventListener("resize", () => {
 });
 
 // center the camera to New York
-mapView.camera.position.set(0, 0, 1600);
-mapView.geoCenter = new GeoCoordinates(40.70398928, -74.01319808, 0);
+mapView.lookAt(new GeoCoordinates(40.70398928, -74.01319808), 1500, 40, 190);
```
zbigg: After last #535, azimuth should be 10 ... or -10 ;) ?

robertoraggi: > Now, whtn #535 is merged, azimuth should be 10 ... or -10 ;) ?

does it mean that we need publish new packages before to get the new behavior of `lookAt` befor merging this change? if that is the case, we update the generate package.json with the correct patch version for harp-mapview.

zbigg: No, this package will be published together with the rest, we just have to remember to bump deps in `package.json` of generated package too and AFAIR, we don't have automagic mechanism for this.

ninok: Yes, but the yeoman-generator is always one version behind. So what we should do is, increase the minor version of the mapview module (b/c of #535) when we do next release (yeoman will still use the old one) and after the release we have to update yeoman generator again for the next after next release.
I totally don't like this process right now. We should automatically update the yeoman generator whenever we release, but this is how it is right now.

PR #543, opened 2019-06-13
HARP-6087: SolidLines vertex color support.

amsgames94:


@here/harp-geojson-datasource/lib/GeoJsonGeometryCreator.ts
```diff
@@ -548,12 +550,12 @@ export class GeoJsonGeometryCreator {
                 (this.isOnTileBorder(contour[i + 1], tileExtents) &&
                     this.isOnTileBorder(contour[i + 4], tileExtents))
             ) {
-                lines.add([...outline]);
+                lines.add(lineCenter, [...outline]);
```
zbigg: `[...outline]`
I know it's old code, but is that copy really necessary ?
`outline` is forgotten two lines later and both `lines.add` and `buffer.push` use it as `readonly` array.

(same few lines below)

PR #544, opened 2019-06-13
Makefile based build orchestrator for development.

zbigg:
This is draft idea, how we can ease pre-checks before stuff is pushed to review.

Our development pipeline is getting bigger and bigger there are like 8 tasks executed for each build in CI, which are hard-coded in `.travis.yml` and thus not available directly for developer before he pushes to review.

This makefile, does the same with
 - simple usage: `make -j4`
 - hides output of successful steps (outputs is dumped only in case of failure)
 - parallel build support
 - basic dependency resolution - will not rebuild examples, if you play with tests and vice versa

This "old-fashioned" can run CI" tasks in on developer machine in around 1 minute, thanks to fact that is is parallelizable (yes, `make -j4` works).

Yes, `make` is not available on windows - but there is no proper build orchestrator that can do proper dependency resolution for out-of-box parallel build support.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,60 @@
+#
+# Usage
+#    make <script from package.json>
+#
+# Runs all tasks from package.json in correct order and deps, so this can be parallelized.
+# (Good results observed up to -j4).
+#
+# By default, all output of successfull command is skipped, to enable it set VERBOSE env variable
+# like this:
+#
+#    VERBOSE=1 make
+
+tasks = \
+	tslint prettier \
+	test \
+	build-examples \
+	build-bundle \
+	build-tests \
+	test-browser-chrome test-browser-firefox \
+	typedoc
+
+default: $(tasks)
+
+# custom targets and
+build: build-examples build-bundle
+test: test-browser
+test-browser: test-browser-chrome test-browser-firefox
+
+app_sources_ts=$(shell find @here/*/src -name "*.ts")
+lib_sources_ts=$(shell find @here/*/lib -name "*.ts" | egrep -v ".d.ts") $(shell find @here/*/index*.ts)
+test_sources_ts=$(shell find test @here/*/test -name "*.ts")
+
+all_sources=$(lib_sources_ts) $(test_sources_ts) $(app_sources_ts)
+
+# dependencies definitions
+.tasks/build-examples: $(test_sources_ts) .tasks/build-bundle
+.tasks/build-bundle: $(lib_sources_ts)
+
+.tasks/tslint: $(all_sources)
+.tasks/typedoc: $(lib_sources_ts)
+.tasks/prettier: $(all_sources)
+
+.tasks/test: $(lib_sources_ts) $(test_sources_ts)
+.tasks/build-tests: $(lib_sources_ts) $(test_sources_ts)
+
+# custom tasks, should be added to package.json actually
+.tasks/test-browser-chrome: build-tests
+	@./scripts/ci-task.sh $@ yarn test-browser --headless-firefox
+
+.tasks/test-browser-firefox: build-tests
+	@./scripts/ci-task.sh $@ yarn test-browser --headless-chrome
+
+# workaround, firefox depend on chrome tests, so we don't run in parallel and we have problem with
+# EADDRINUSE
+.tasks/test-browser-firefox: .tasks/test-browser-chrome
+
+.tasks/%: package.json
+	@./scripts/ci-task.sh $@ yarn run $(@F)
+
+$(tasks): %: .tasks/%
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@ node_modules/
 dist/
 doc/
 test-npm-packages-app/
+.tasks/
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,60 @@
...
+	typedoc
+
+default: $(tasks)
+
+# custom targets and
+build: build-examples build-bundle
+test: test-browser
+test-browser: test-browser-chrome test-browser-firefox
+
+app_sources_ts=$(shell find @here/*/src -name "*.ts")
```
okanishchev: Windows users start to hate you down from this line

zbigg: Yes, but windows users have MSYS2, Cygwin and WSL.
More, they have choice to not use this script and manually use `yarn whatever` in proper order.

`gradle`, `basel`, `cmake`, `ninja`, `Jake`, `grunt`, `gulp` based solutions are warmly welcomed but i seriously doubt that they can beat this one w.r.t robustness and simplicity.

Makefile
```diff
@@ -0,0 +1,60 @@
...
+tasks = \
+	tslint prettier \
+	test \
+	build-examples \
+	build-bundle \
+	build-tests \
+	test-browser-chrome test-browser-firefox \
+	typedoc
+
+default: $(tasks)
```
okanishchev: is there any use cases of "default" rule? make without arguments runs first rule in Makefile by default

zbigg: No particular use case, other that readability.

> make without arguments runs first rule in Makefile
As you see, this rule is first for exactly this reason, you just need some name for it ... and `default` is actually good one

Makefile
okanishchev: Yep, windows users have a lot of options not to use windows.

But seriously, make runs pretty good on plain windows, but there are some hard limitations:
 * no bash
 * no find

All other programs (like ```grep```, ```sed```, whatever) can be simply installed and used from default windows console.
But for ```bash``` and ```find``` you need to replace console: ```bash``` for obvious reasons, and ```find``` because windows has own builtin ```find``` command that can't be replaced.

So if we avoid this two conditions, we can run ```make``` on plain windows, if someone will need this feature of course.

Makefile
okanishchev: Oh, I'm firstly took tasks declaration for the rule, didn't see the =

Makefile
musculman: While it is fine to make it work in windows, if we can, our software does not work on plain windows already today. Some of our release scripts are already linux only. I would not take a too hard stand here if we have no alternative.

haraldF:
I'm giving a big -1 to this - `make` and shell are two more technologies that come with their own baggage of pain and that we would have to learn, understand and maintain.

Yes, the current system is not optimal, but I don't see enough pain to warrant introduction of `make`.

Makefile
haraldF: If it doesn't work on Windows we should fix it. Introducing technology that we know doesn't work well on Windows is a no-no. I would like Windows developers to build our stack and also locally run the pre-submit checks so they can reproduce CI errors locally.

zbigg: > If it doesn't work on Windows

There are no major obstacles for it to run in proper environment on windows, just use MSYS2 - it should work. In GIT bash shell, it should work too.

@haraldF 
The point of this patch is: Improve developers experience by _enabling_ them to pre-check code in same way as in CI. CI works on Linux, more, it works only in Docker.

Counter question - can we run `.travis-yml` on Windows, Linux, OSX? No (at least not directly*), and yet it's here for it's purpose.

Let's not discard tools that help majority (majority work on Linux/Mac) given that minority has workarounds.

* [yes](https://andy-carter.com/blog/setting-up-travis-locally-with-docker-to-test-continuous-integration), we can but it's also, but we can run Makefile stuff in Docker too, it's just question of writing instruction

haraldF: @zbigg "no major obstacles" - I quite disagree :) It should work out of the box on a normal Windows terminal.

Plan B - what about having a simple task runner directly in node? So we can write `npm run ci:tests` and don't have any external tool dependencies (and no `sh` or `make` dependency)

zbigg: > Plan B - what about having a simple task runner directly in node? 

I tried hard to be pragmatic by not doing it by myself (DYI is my second name) so, namely using right tool for right job. As discussed internally, i still believe that `make` + `sh` - despite all of its limitations - has several qualities that qualify them. but ... ok.

Will rethink again proper task runner in bare TS (no gulp or other frameworks) for this, maybe it will "offspring" as proper "make" for JS projects.

okanishchev: Maybe we can just write down some additional webpack rule (or yarn script) to make everything?

nzjony:
Or though as the discussion above shows that this isn't a perfect solution, this seems like a good starting point to have developers be able to run the CI locally at least in mac / linux. I would like to very much be able to use this to run locally, because it is annoying to have to wait for the CI to give feedback, when I can do it locally (if I so choose to).

I think we should submit this and wait until the community cries loud enough to introduce windows support.

PR #550, opened 2019-06-14
HARP-5949: Default language support.

amsgames94:
Avoid calling ``MapViewUtils.getBrowserLanguages()`` implicitly when no language settings are set, since ``navigator.languages`` and ``navigator.language`` return unconsistent values across browsers (i.e. chrome treats your configured languages as an array and doesn't care about which one you've actually set).


zellm:
Note: This change will also allow to display labels in the default language as defined in the data.

PR #551, opened 2019-06-14
HARP-5430 Adding correct panning with terrain.

nzjony:


@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -1110,8 +1110,27 @@ export class MapControls extends THREE.EventDispatcher {
...
+        let fromWorld: THREE.Vector3 | null;
+        if (this.mapView.elevationProvider === undefined) {
+            fromWorld = MapViewUtils.rayCastWorldCoordinates(this.mapView, from.x, from.y);
+            toWorld = MapViewUtils.rayCastWorldCoordinates(this.mapView, to.x, to.y);
+        } else {
+            fromWorld = this.mapView.elevationProvider.rayCast(fromX, fromY);
+            if (fromWorld === null) {
+                return;
+            }
+            const geoCoordFromWorld = this.mapView.projection.unprojectPoint(fromWorld);
```
ninok: You only need the altitude and that is a no-op for globe.
So maybe we should add a method to the projection that returns the altitude for a world coordinate

nzjony: You mean a no-op for planar projections?
Good idea to add a method to projection that returns the altitude. I will add it now and we can discuss the best name.
Renamed.

@here/harp-geoutils/lib/projection/MercatorProjection.ts
```diff
@@ -267,6 +271,10 @@ class WebMercatorProjection extends MercatorProjection {
         return new GeoCoordinates(latitude, longitude, worldPoint.z);
     }
 
+    unprojectAltitude(worldPoint: Vector3Like): number {
```
CraigBerlin: This class extends MercatorProjection, can omit this

@here/harp-mapview/lib/Utils.ts
```diff
@@ -190,8 +194,11 @@ export namespace MapViewUtils {
         rayCaster.set(cameraZPosition, pointInCameraSpace.normalize());
 
         const vec3 = new THREE.Vector3();
+        if (elevation !== undefined) {
+            groundPlane.constant = -elevation;
```
CraigBerlin: Would it be safer in this case just to create a new Vector3 for the groundPlane?
Setting and then clearing the global could cause undefined behavior if this function is ever called simultaneously.

nzjony: My understanding is that because of the event loop, we can only execute one part of our code at once, so the function can't be shared.

In the case of Web Workers, then they have their own copy, so nothing is shared. @zbigg, @robertoraggi, @StefanDachwitz  is that correct?

@here/harp-geoutils/lib/projection/MercatorProjection.ts
```diff
@@ -105,6 +105,10 @@ class MercatorProjection extends Projection {
         return geoPoint;
     }
 
+    unprojectAltitude(worldPoint: Vector3Like): number {
+        return worldPoint.z;
```
ninok: for mercator worldPoint.z is not in meters. You need to apply the mercator scale

@here/harp-mapview/lib/Utils.ts
ninok: Since rayCaster.ray.intersectPlane is not async so this should be fine.

PR #553, opened 2019-06-15
Create new website for harp.gl

dbabbs:
- [x] HERE branding
- [x] mobile friendly
- [x] selector for different versions of releases
- [x] looks dope


docs/index.html
```diff
@@ -1,79 +1,112 @@
...
+      <title>harp.gl</title>
+      <link href="https://fonts.googleapis.com/css?family=Fira+Sans:400,500,600&display=swap" rel="stylesheet">
+      <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+      <meta name="viewport" content="width=device-width,initial-scale=1">
+      <link rel="stylesheet" type="text/css" href="css/index.css">
+      <link rel="shortcut icon" href="img/favicon.ico">
+   </head>
+   <body>
+      <header>
+         <iframe src="https://dbabbs.github.io/harp-temp-site/dist/"></iframe>
```
musculman: I guess that the src of this iframe should not be that?

docs/js/index.js
```diff
@@ -0,0 +1,47 @@
...
+}).catch(() => {
+   
+   //In case network request to build information fails, add master link
+   const option = document.createElement('option');
+   option.innerText = 'master';
+   dropdown.appendChild(option);
+})
+
+setTimeout(() => {
+   document.querySelector('header').style.backgroundImage = `url('img/background.png')`;
```
musculman: The image we have for background right now is under img/bgmap.png. I guess you are not using it, can you remove it as part of this PR so that there are no leftovers from previous version?

docs/index.html
dbabbs: I had thought about this too, but where else should we host? Within this repository inside of `docs`?

musculman: I think we can host it directly in our `gh-pages`. Same way we host right now the `releases.json` resource that is only needed for this site.

docs/js/index.js
dbabbs: Deleted

musculman:
Successfully tested at https://github.com/musculman/harp.gl
Deployment here: https://musculman.github.io/harp.gl

nzjony:
Thanks @musculman for the links. I notice there is a lot of tiles flickering in / out in the above, might pay to reduce the zoom level (or FOV), I have created a ticket for this: HARP-6192.

docs/index.html
```diff
@@ -1,79 +1,112 @@
...
-                            <a href="https://github.com/heremaps/harp.gl" class="dp-button dp-button-transparent" target="_blank">BROWSE ON GITHUB</a>
-                    </div>
-                </div>
+         <div class="container header-container">
+            <nav>
+               <div class="logo"></div>
+               <div>
+                  <a id="docs-nav-mobile" href="http://harp.gl/docs" class="nav-item">Docs</a>
+                  <a id="docs-nav" href="http://harp.gl/docs" class="nav-item">Documentation</a>
+                  <a href="http://harp.gl/examples" class="nav-item">Examples</a>
```
nzjony: This goes to a different page, when compared with the examples button lower in the page, is this intended, my gut feeling is that this is not desired.

nzjony:
Thanks @musculman for the links. I notice there is a lot of tiles flickering in / out in the above, might pay to reduce the zoom level (or FOV), I have created a ticket for this: HARP-6192.

docs/index.html
dbabbs: That is by design. Currently this link (http://harp.gl/docs) doesn't work because the site hasn't been published. However, if you view @musculman's forked published version, you can see the short url working:

https://musculman.github.io/harp.gl/docs
https://musculman.github.io/harp.gl/examples

During jsconf asia, we identified the need for short urls to the examples and docs, rather than the long s3 bucket urls.

nzjony: Ok, thanks for clarifying.

dbabbs:
Logo and standard footer now removed. This is what the page now looks like: 
screencapture-localhost-8002-dist-2019-06-19-20_52_55

PR #554, opened 2019-06-17
HARP-5700: Fix texture coordinate computations for globe.

robertoraggi:


@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -578,11 +584,11 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
-                          const { x: u, y: v } = projection
-                              .reprojectPoint(webMercatorProjection, pos, texCoord)
-                              .sub(tileBounds.min)
-                              .divide(tileSize);
-                          return { u, v };
+                          const { x: u, y: v } = pos
+                              .clone()
+                              .sub(tileDataBounds.min)
+                              .divide(tileDataSize);
+                          return { u, v: 1 - v };
```
robertoraggi: Computes the UVs in the OMV tile space (Y points down). The previous code was computing the UVs in the target projection space (webmercator with Y up or spherical globe). Projecting the UVs in spherical globe doesn't really work with the subdivision modifier, so it's better to keep these attributes in their original space.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -628,6 +634,9 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
                         if (computeTexCoords !== undefined) {
                             const { u, v } = computeTexCoords(pos);
                             ringContour.push(u, v);
+                            if (ringPoints !== undefined) {
+                                ringPoints.push(u, v);
+                            }
```
robertoraggi: Rings, contours, ringPoints and stride are a bit messy. In a follow up change I will try to rework the way we work with multi-polygons.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -531,6 +532,15 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
     ): void {
         this.processFeatureCommon(env);
 
+        const { projection, center, tileKey } = this.m_decodeInfo;
+
+        // get the bounds, and the size of the OMV Tile in its webmercator space.
+        const tileDataBounds = new THREE.Box3();
+        webMercatorTilingScheme.getWorldBox(tileKey, tileDataBounds);
```
amsgames94: Cannot this be passed as part of ``OmvDecoder.DecodeInfo``? Otherwise, we're recalculating this for all features in a tile (when it's the same for all of them).

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -578,11 +584,11 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
             const computeTexCoords =
                 texCoordType === TextureCoordinateType.TileSpace
                     ? (pos: THREE.Vector3) => {
-                          const { x: u, y: v } = projection
-                              .reprojectPoint(webMercatorProjection, pos, texCoord)
-                              .sub(tileBounds.min)
-                              .divide(tileSize);
-                          return { u, v };
+                          const { x: u, y: v } = pos
+                              .clone()
```
amsgames94: Maybe copy into a temporary object?

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
amsgames94: Agree, we need to clean this one up.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
robertoraggi: done

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
robertoraggi: We could but it doesn't really make any difference, also it is decoder-specific if the bounds are needed or not.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -531,6 +532,15 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
     ): void {
         this.processFeatureCommon(env);
 
+        const { projection, center, tileKey } = this.m_decodeInfo;
+
+        // get the bounds, and the size of the OMV Tile in its webmercator space.
+        const tileDataBounds = new THREE.Box3();
```
amsgames94: The names ``tileDataBounds`` and ``tileDataSize`` (specially this last one) are a bit confusing, since it seems like you're taking about array ranges and sizes (i.e. bytes), while you're talking about meassurements. Why renamig to this?

PR #559, opened 2019-06-18
HARP-6057: Tweaked MapView default cache values (globe).

amsgames94:


@here/harp-examples/src/datasource_satellitetile_globe.ts
```diff
@@ -35,8 +35,7 @@ export namespace SatelliteTileDataSourceGlobeExample {
         const map = new MapView({
             canvas,
             theme: "resources/berlin_tilezen_base_globe.json",
-            projection: sphereProjection,
-            tileCacheSize: 400
+            projection: sphereProjection
         });
         map.setCacheSize(100, 100);
```
robertoraggi: what about the call to `setCacheSize`? can we find a way to remove it?

amsgames94: Didn't realize, thanks for pointing it out!

PR #561, opened 2019-06-18
HARP-5693: View is changing depending on aspect ratio

SergiiKulakovskyi:

### DIFF
--- a/@here/harp-examples/src/hello.ts
+++ b/@here/harp-examples/src/hello.ts
@@ -73,7 +73,8 @@ export namespace HelloWorldExample {
         // snippet:harp_gl_hello_world_example_1.ts
         const map = new MapView({
             canvas,
-            theme: "resources/berlin_tilezen_base.json"
+            theme: "resources/berlin_tilezen_base.json",
+            fovCalculation: { type: "fixed", fov: 40 }
         });
         // end:harp_gl_hello_world_example_1.ts
 


@here/harp-examples/src/hello.ts
```diff
@@ -73,7 +73,8 @@ export namespace HelloWorldExample {
         // snippet:harp_gl_hello_world_example_1.ts
         const map = new MapView({
             canvas,
-            theme: "resources/berlin_tilezen_base.json"
+            theme: "resources/berlin_tilezen_base.json",
+            fovCalculation: { type: "fixed", fov: 40 }
```
nzjony: @zellm, is this acceptable? Do we want to have a fixed fov, as opposed to the dynamic?

The distortion seen is due to the extreme edge case being tested. The small window mentioned in the ticket has a FOV of 40, and then when stretched, the fov is increased.

If the window needs to start out small, and will be stretched, it would make more sense to choose either smaller FOV, or fix it as in the case above, but I don't think this is necessarily the use case that the hello example should support?

zbigg: @nzjony I want to say only that i expected this kind of patch from someone not familiar with current default - fixed focal length.
Current behavior of map is _really weird_, especially when you have 3D objects near by that are easily distorted.

Users really don't expect `fov` to change, when resizing window. Not sure, what they expect - probably `cameraHeight` should just change, but `zoomLevel` and `fov` should be maintained because this is what (IMO) our brain expects when broadening/limiting viewport.
Maybe `zoomLevel` should actually depend on "meters per pixel", not cameraHeight+fov.

BTW, competitors maintain FOV and change camera height.

nzjony: > Users really don't expect fov to change, when resizing window. Not sure, what they expect

I guess it depends on the use case.

>  probably cameraHeight should just change, but zoomLevel and fov should be maintained because this is what (IMO) our brain expects when broadening/limiting viewport.

Would have to be checked with the use case, the initial focal length could be saved, but you risk then having a lower zoom level, so a map with a stretch canvas will have lower resolution tiles when compared to a same size canvas. Maybe undesirable.

> Maybe zoomLevel should actually depend on "meters per pixel", not cameraHeight+fov.

Meters per pixel is computed based on the focal length.

I'm all for improving it to make it as simple as possible for developers to understand, maybe it should be solutioned to find the best way to go forward, if this is a headache for them.

zellm: The user expectation and what most (all?) competition does at least for 2d maps is the following. When resizing the map, you only see a larger or smaller area of the map, the size of the objects  in screenspace or any distance between the objects stays the same, basically you just change the view port size. This is the expectation for harp.gl and as well as for native. The question is how to technically realize this. In google maps you see two different approaches, for the normal map zoomed in, if you change the >height< of the view port you will notice that the all the map features stick to their position on the left or right edges of the view port which indicates that map is NOT zoomed. This is the behavior we implemented with the dynamic fov/focal lengths. In satellite you will notice that changing the height of the viewpoints makes the map actually zooming in and out. That is the approach with fixed fov. I believe we need to have a longer discussion to agree which approach is more appropriate for a given use case but HARP-5693 probably not even related to this. It seems there is a bug with keeping the center point.

zellm: Adrien pointed out that the initial issue is probably already fixed by https://github.com/heremaps/harp.gl/pull/513 . We just took a look and indeed, everything seems to work as expected.

nzjony: @SergiiKulakovskyi , @zbigg , I will improve the documentation to make the trade offs clear. They are not documented well enough.

zbigg: @zellm Sorry, i've checked again behavior our map and i think that visually, dynamic fov based on constant focal length is ok, but we have wrong defaults for default focal length - maybe they should be calculated basing on, say initial viewport size

Compare our version [1] with f4 [2]
When starting with maxed window (2k), both apps behave more or less the same, but when starting with small window things go weird.

- our app is showing "big panorama", but "scaled down"
- f4 shows only "viewport" over original panorama, and when window is maxed in, shows big panorama - same as you would open it in maxed window
- our map, when maximized shows "fisheye view", which is unexpected

[1] us - http://harp.gl.s3-website-us-east-1.amazonaws.com/docs/master/examples/hello.html
[2] f4 for more or less ame view: https://demo.f4map.com/#lat=40.7050809&lon=-74.0083066&zoom=16&camera.theta=44.141&camera.phi=2.005

nzjony: > but we have wrong defaults for default focal length - maybe they should be calculated basing on, say initial viewport size

The focal length is computed based on the fov and the initial viewport size, see: `private calculateFocalLength(height: number)`

The reason for the big panarama is that we are using a fixed distance, as opposed to a fixed zoom level. If we want the same look regardless of the height then a fixed zoom level needs to be used.

SergiiKulakovskyi: @zellm, with the #513 initial positioning issue is fixed. Noted in the ticket.

PR #563, opened 2019-06-18
HARP-6075: Expose all MapView exports in harp.js bundle

zbigg:


sasha240100:
@zbigg Could you please add harp.gl/dist to S3 artifacts, so I am able to quickly download build and test it?

PR #566, opened 2019-06-18
HARP-6217: Introduce resize event for external use cases

sasha240100:


zbigg:
Not sure why it's needed.

The task says:

> mapview.resize(width, height); is a public method that sometimes can be used externally. The only way > to subscribe for a resize event currently is to wrap resize method (...)

`MapView` doesn't resize automatically by itself, only if user calls resize, so actually there is no `event`.
So, if map is resized, it's someone who called `resize` in response and he already know that resize happened.


BTW, We may "auto-resize" (and rationale for `resize`) event add it in future because users expect auto-resize out of box.

okanishchev:
This can be used by standalone plugin to mapview that expects mapview in the constructor, and then listens to mapview events. So the user hasn't need to resize plugin.

Besides, the event can be emitted only if actual size change is made - if we provide some checks.

PR #569, opened 2019-06-19
HARP-6060 Introducing scale to convert from unit to world coordinates

nzjony:

### DIFF
--- a/@here/harp-datasource-protocol/lib/DecodedTile.ts
+++ b/@here/harp-datasource-protocol/lib/DecodedTile.ts
@@ -5,7 +5,9 @@
  */
 
 import {
+    equirectangularProjection,
     mercatorProjection,
+    normalizedEquirectangularProjection,
     Projection,
     sphereProjection,
     webMercatorProjection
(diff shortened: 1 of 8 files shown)


@here/harp-datasource-protocol/lib/DecodedTile.ts
```diff
@@ -217,6 +223,10 @@ export function getProjectionName(projection: Projection): string | never {
         return "webMercator";
     } else if (projection === sphereProjection) {
         return "sphere";
+    } else if (projection === equirectangularProjection) {
+        return "equirectangular";
+    } else if (projection === equirectangularWorldProjection) {
+        return "equirectangularworld";
```
robertoraggi: we use `camel case` for the projection names, so please rename it `equirectangularWorld`. Actually, about the name `equirectangularWorldProjection` it would be nice if we could fine a less ambiguous name.

@here/harp-datasource-protocol/lib/DecodedTile.ts
```diff
@@ -200,6 +202,10 @@ export function getProjection(projectionName: string): Projection | never {
             return webMercatorProjection;
         case "sphere":
             return sphereProjection;
+        case "normalizedequirectangular":
+            return normalizedEquirectangularProjection;
```
robertoraggi: please use `camel case` for the name `normalizedEquirectangular`

PR #570, opened 2019-06-19
HARP-6211: Tweaked tile visibility evaluation.

amsgames94:


amsgames94:
> Please show an example where this makes a difference?
> 
> Otherwise we are just making a change without really understanding the root cause.

Added comment to the change to better understand why this makes a difference.

PR #577, opened 2019-06-19
HARP-5727: Integrate GlobeControls in MapControls.

Astrak:
First step to integrate globe controls: Merge existing logic in MapControls.


musculman:
I know that @dbabbs will like this PR :)

dbabbs:
This is excellent. Was just working on moving the JSConf workshop to developer.here.com and had to remove the globe projection because of our `extras.js` nonsense hack we had to make...

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -463,21 +463,37 @@ export class MapControls extends THREE.EventDispatcher {
...
+
+            const surfaceNormal = this.mapView.projection.surfaceNormal(
+                this.camera.position,
+                new THREE.Vector3()
+            );
+
+            this.camera.position.addScaledVector(
+                surfaceNormal,
+                (this.zoomLevelTargetted - zoomLevel) * L * 0.1
+            );
```
robertoraggi: there's a typo `zoomLevelTargetted`

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -463,21 +463,37 @@ export class MapControls extends THREE.EventDispatcher {
...
+
+            if (this.mapView.projection.groundDistance(this.camera.position) < 500) {
```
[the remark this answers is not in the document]
Astrak: Don't know... copied from the GlobeControls, didn't do functional change. @robertoraggi ? Was that maybe a remnant of an early version of the globe?

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -463,21 +463,37 @@ export class MapControls extends THREE.EventDispatcher {
...
+            const L = this.mapView.projection.groundDistance(this.camera.position);
+
+            const surfaceNormal = this.mapView.projection.surfaceNormal(
+                this.camera.position,
+                new THREE.Vector3()
+            );
+
+            this.camera.position.addScaledVector(
+                surfaceNormal,
+                (this.zoomLevelTargetted - zoomLevel) * L * 0.1
```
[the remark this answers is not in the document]
Astrak: GlobeControls had an arbitrary 0.001 but it was much faster. Here the function gets a different number, that I need to re-divide to get the original mouse wheel figure. But, if I then let 0.001, the movement is not perceptible, I don't know why. In addition, this method's purpose is to target a specific zoom level, which this code is not doing whatever the number. It should be modified in the feature to align zoom level calculations with camera movement in this globe view.

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -463,21 +463,37 @@ export class MapControls extends THREE.EventDispatcher {
...
-
         // Assign the new animation start time.
         this.m_zoomAnimationStartTime = performance.now();
 
-        this.handleZoom();
+        if (this.mapView.projection.type === geoUtils.ProjectionType.Planar) {
+            this.m_zoomTargetNormalizedCoordinates.set(screenTarget.x, screenTarget.y);
+            this.handleZoom();
+        } else {
+            const L = this.mapView.projection.groundDistance(this.camera.position);
```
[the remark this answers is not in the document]
Astrak: Can you do it? I'm not familiar with that code, its concepts and stakes.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -171,7 +173,7 @@ export namespace MapViewUtils {
         mapView: MapView,
         pointOnScreenXinNDC: number,
         pointOnScreenYinNDC: number
-    ): THREE.Vector3 | null {
+    ): THREE.Vector3 {
```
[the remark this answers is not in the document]
Astrak: Seemed to come from an old type definition, the result is always a Vector3 currently. Doc updated.

@here/harp-map-controls/lib/MapControls.ts
Astrak: See https://github.com/heremaps/harp.gl/pull/577#discussion_r295361895

@here/harp-map-controls/lib/MapControls.ts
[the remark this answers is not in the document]
robertoraggi: I guess the values of the min distance to the surface and the max distance from the surface should be configurable because they depend from the data (e.g. the max storage level of the data source), the height provider, and so on. 500 is the number that look ok to avoid precision issues when rendering data at storage level 17.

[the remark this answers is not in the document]
nzjony: I will add a comment that this limitation should be configurable.

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -363,6 +372,9 @@ export class MapControls extends THREE.EventDispatcher {
      * @param deltaAltitude Delta altitude in degrees.
      */
     orbitFocusPoint(deltaAzimuth: number, deltaAltitude: number) {
+        if (this.mapView.projection.type === geoUtils.ProjectionType.Spherical) {
+            return;
```
ninok: Why early return here? Please add a comment to the typedoc that this method will do nothing for globe.

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -463,21 +475,43 @@ export class MapControls extends THREE.EventDispatcher {
...
+        if (this.mapView.projection.type === geoUtils.ProjectionType.Planar) {
+            this.m_zoomTargetNormalizedCoordinates.set(screenTarget.x, screenTarget.y);
+            this.handleZoom();
+        } else {
+            const surfaceNormal = this.mapView.projection.surfaceNormal(
+                this.camera.position,
+                new THREE.Vector3()
+            );
+
+            // TODO: HARP-5341 Use the elevation provider to find the ground distance
```
ninok: Typo: HARP-5431

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -463,21 +475,43 @@ export class MapControls extends THREE.EventDispatcher {
...
-
         // Assign the new animation start time.
         this.m_zoomAnimationStartTime = performance.now();
 
-        this.handleZoom();
+        if (this.mapView.projection.type === geoUtils.ProjectionType.Planar) {
+            this.m_zoomTargetNormalizedCoordinates.set(screenTarget.x, screenTarget.y);
+            this.handleZoom();
+        } else {
+            const surfaceNormal = this.mapView.projection.surfaceNormal(
```
ninok: Shouldn't all this code go to handleZoom? What do we do differently when zooming in on the globe? Do we have a zoom animation for globe?

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -1113,19 +1147,25 @@ export class MapControls extends THREE.EventDispatcher {
...
+            this.m_panAnimationStartTime = performance.now();
 
-        // Assign the new animation start time.
-        this.m_panAnimationStartTime = performance.now();
+            this.m_panDistanceFrameDelta = fromWorld.sub(toWorld);
 
-        this.pan();
+            this.pan();
+        } else {
+            quaternion.setFromUnitVectors(fromWorld.normalize(), toWorld.normalize()).inverse();
```
ninok: This should go to MapControls.pan. Will there be inertia for globe panning(spinning) ?

@here/harp-map-controls/lib/MapControls.ts
Astrak: It should do something, it's only that it is not supported yet. It should be implemented in further tickets.

@here/harp-map-controls/lib/MapControls.ts
Astrak: This PR only integrates the existing globe controls. I want to create further tickets to improve their handling in MapView and have a consistent approach, but at the moment it cannot easily be. With the planar version we zoom by specifying a zoom level in this method, not yet in globecontrols, I have to figure that out. In the planar version the panning is actually a rotation. The orbitting does not have the same reference and up axis. All that calls for different implementation that I want to break down in tickets. The progress will be clearer, more visible for the team, and easier to review.

@here/harp-map-controls/lib/MapControls.ts
Astrak: Yes. Problem is that the pan function receives a vector in world units, which doesn't work for the globe, at best we want a delta in geolocations for a common approach, which requires some change in the structure of MapControls. I want to separate these questions into tickets and tackle them separately, so that the controls are an abstraction of the movement that works for all projections and where inertia handle both the same way too, yes.

PR #579, opened 2019-06-21
HARP-6194 Adding run time change to projection

nzjony:


nzjony:
Please also see gerrit patchset.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1234,6 +1234,13 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_visibleTileSetOptions.projection;
     }
 
+    set projection(projection: Projection) {
+        this.m_visibleTileSetOptions.projection = projection;
```
zbigg: I think we should maintain proper `geoCenter` by stg like:

```
const x = this.mapView.geoCenter;
// change projection
this.mapView.geoCenter = x;
```

User shouldn't be forced to whole internal known state, if `MapView` knows enough to maintain proper view.

Same applies to camera, with such a basic transformation i'm worried that camera will go crazy when you change from mercator to globe.

nzjony: You're absolutely right. I will fix.

sasha240100:
@nzjony It looks like `SkyBackground` class also depends on `projectionType`. I think you have to add some code into `MapView#updateSkyBackground` (which is triggered by `set theme`) that should set new `skyBackground.m_skyTexture` if it is `SkyGradientTexture` with the new `projectionType`. I see the flow as following:

 1. `set theme`
 2. `MapView#updateSkyBackground`
 3. `MapView#updateSkyBackgroundColors`
 4. -> `this.m_skyBackground.updateTexture(sky, projectionType?);` **projectionType** should be optional and it should set local m_projectionType when is provided.

nzjony:
@sasha240100, thanks a lot for your feedback and good spotting. I have made a fix.
@amsgames94, is this correct?

@here/harp-mapview/lib/SkyBackground.ts
```diff
@@ -71,14 +71,15 @@ export class SkyBackground {
      *
      * @param params New sky configuration parameters.
      */
-    updateTexture(params: GradientSky | CubemapSky) {
-        const isSameSkyType = this.m_sky.type === params.type;
+    updateTexture(params: GradientSky | CubemapSky, projectionType: ProjectionType) {
```
sasha240100: Should have a line:
```js
this.m_projectionType = projectionType;
```

Because we've updated projectionType to a new one.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2555,12 +2575,16 @@ export class MapView extends THREE.EventDispatcher {
         }
     }
 
-    private updateSkyBackgroundColors(sky: Sky, clearColor: string | undefined) {
+    private updateSkyBackgroundColors(
+        sky: Sky,
+        clearColor: string | undefined,
+        projType: ProjectionType
```
sasha240100: `projType` is not used.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2528,7 +2548,7 @@ export class MapView extends THREE.EventDispatcher {
         const theme = this.m_theme;
         if (this.m_skyBackground instanceof SkyBackground && theme.sky !== undefined) {
             // there is a sky in the view and there is a sky option in the theme. Update the colors
-            this.updateSkyBackgroundColors(theme.sky, theme.clearColor);
+            this.updateSkyBackgroundColors(theme.sky, theme.clearColor, this.projection.type);
```
sasha240100: 3rd parameter is not needed here as it accessible in `updateSkyBackgroundColors`

@here/harp-mapview/lib/SkyBackground.ts
```diff
@@ -71,14 +71,15 @@ export class SkyBackground {
...
+        const isSameSkyType =
+            this.m_sky.type === params.type && this.m_projectionType === projectionType;
         switch (params.type) {
             case "gradient":
                 if (isSameSkyType) {
                     (this.m_skyTexture! as SkyGradientTexture).updateTexture(params);
                 } else {
-                    this.m_skyTexture = new SkyGradientTexture(params, this.m_projectionType);
+                    this.m_skyTexture = new SkyGradientTexture(params, projectionType);
```
amsgames94:
```suggestion
                    this.m_projectionType = projectionType;
                    this.m_skyTexture = new SkyGradientTexture(params, projectionType);
```

@here/harp-mapview/lib/MapView.ts
nzjony: opps, done.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1234,6 +1234,26 @@ export class MapView extends THREE.EventDispatcher {
...
+        // camera's position is based on the projected geo center.
+        const geoCenter = this.geoCenter;
+        this.m_visibleTileSetOptions.projection = projection;
+        this.clearTileCache();
+        // We reset the theme, this has the affect of ensuring all caches are cleared.
+        this.theme = this.theme;
+        this.geoCenter = geoCenter;
+        // Necessary for the sphereProjection, however this also resets the camera position, so it
+        // should be fixed.
+        //this.setupCamera();
```
sasha240100: Remove unnecessary comment

@here/harp-mapview/lib/SkyBackground.ts
```diff
@@ -70,9 +70,13 @@ export class SkyBackground {
...
      * @param params New sky configuration parameters.
+     * @param projectionType Which projection is used, this may also change (in which case the
+     * textures should be recreated).
      */
-    updateTexture(params: GradientSky | CubemapSky) {
-        const isSameSkyType = this.m_sky.type === params.type;
+    updateTexture(params: GradientSky | CubemapSky, projectionType: ProjectionType) {
+        const isSameSkyType =
+            this.m_sky.type === params.type && this.m_projectionType === projectionType;
+        this.m_projectionType = projectionType;
```
sasha240100: I'd move this line to the end of `updateTexture` (to be next to `this.m_sky`)

@here/harp-mapview/lib/MapView.ts
nzjony: I want to keep it, because it is necassary for the sphere projection (see the above TODO as well), I could create a ticket and put it here instead?

@here/harp-mapview/lib/SkyBackground.ts
nzjony: I could, but then I would need to change line 85, and it would make it slightly more complicated.
Maybe I could instead move this.m_sky to be closer to this (I don't see the need to have it at the bottom).

sasha240100: @nzjony One reason I could tell it should be in the bottom is that if for example you provide a wrong projectionType, or any other thing will go wrong, then it will throw an error and exit a function, and changes won't be applied (`this.m_sky` & `this.m_projectionType` will be the same as before) and it should be normal behavior.

@here/harp-mapview/lib/MapView.ts
sasha240100: @nzjony Got it, it's fine then

@here/harp-mapview/lib/SkyBackground.ts
nzjony: Done

PR #580, opened 2019-06-22
MINOR: update readme and getting started guide to include simple bundle

dbabbs:
Now that harp.gl is getting some publicity (conferences, workshops, [blog posts](https://developer.here.com/blog/introducing-harp.gl-3d-vector-maps-for-the-web), [tutorials](https://developer.here.com/tutorials/harpgl)), the readme should prioritize information about how to quickly get started with the api.

- [x] add section to top-level README.md about simple bundle and package generator
- [x] add section to Getting Started Guide about simple bundle

### DIFF
--- a/docs/GettingStartedGuide.md
+++ b/docs/GettingStartedGuide.md
@@ -1,19 +1,57 @@
 # Getting Started Guide
 
-To begin with `harp.gl`, we provide few starting points
+To begin with `harp.gl`, we provide a few starting points:
 
--   [Create simple app](#yeoman) using Yeoman
--   [Integrate `harp.gl` into your existing Webpack based project](#integrate)
--   [Look at examples](#examples)
+- [Import harp.gl with simple bundle](#simple)
+- [Create simple app](#yeoman) using Yeoman
+- [Integrate `harp.gl` into your existing Webpack based project](#integrate)
+- [Look at examples](#examples)
+
+## <a name="simple"></a> Import harp.gl with simple bundle
+
+Add `three.js` and `harp.gl` to your html and create a canvas with an id `map`:
+```html
+<html>
+   <head>
+      <style>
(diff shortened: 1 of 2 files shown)


docs/GettingStartedGuide.md
```diff
@@ -1,19 +1,57 @@
...
 
--   [Create simple app](#yeoman) using Yeoman
--   [Integrate `harp.gl` into your existing Webpack based project](#integrate)
--   [Look at examples](#examples)
+- [Import harp.gl with simple bundle](#simple)
+- [Create simple app](#yeoman) using Yeoman
+- [Integrate `harp.gl` into your existing Webpack based project](#integrate)
+- [Look at examples](#examples)
+
+## <a name="simple"></a> Import harp.gl with simple bundle
```
nzjony: Can this instead not link back to the README.md examples? Having two versions increases the change that they differentiate and / or maintenance cost.

dbabbs: This examples link is from a prior commit. However, I don’t understand your question. Where would you like this to link to? And where is the duplication?

nzjony: Sorry I wasn't clear, I mean the html and javascript below are exactly the same in both places.

dbabbs: Ah, I understand now. I do think it's crucial to have basic getting started information on the repo's top-level readme file (hence this pull request) in order to make it as easy as possible for new users to get started with harp.gl. But yes, this information should also be included in the getting started guide page for obvious reasons.

I would prioritize having this getting started information in two visible locations (top-level readme and getting started readme) rather than getting started only, even if that means having two versions with the risk of them possibly getting out of sync. At some point, having a single source of code samples will be unavoidable as the project grows. 

What does the rest of the team think?

nzjony: Valid points!

PR #581, opened 2019-06-24
Harp styling

zbigg:
Support for definitions and inheritance in Theme

* `Theme` supports global `definitions`, so repeated values can be
defined once.

* `Theme` supports `extends` clause which inherits content of
other theme or file pointed by URL.

See new `hello_custom_theme` example and `ThemeLoaderTest` for use-cases.


@here/harp-mapview/test/ThemeLoaderTest.ts
```diff
@@ -0,0 +1,196 @@
...
+                        technique: "solid-line",
+                        attr: {
+                            lineColor: { $ref: "roadColor" }
+                        }
+                    }
+                },
+                styles: {
+                    tilezen: [
+                        {
+                            when: "kind == 'road", // or here, or both with override ?
```
zbigg: @SahinHERE, @robertoraggi  Which place for `when` is better, here or in `definitions`.

I guess, that it should be here but with possibility to define base, default when in `definitions`

SahinHERE: Hey @zbigg,

I would leave out the 'when' condition in the 'roadStyle' defined in 'definitions'.
The idea would be to define styles and conditions in 'definitions' and use them potentially for different datasources. 

So in that way, we would be able to define a set of styles and conditions and could overwrite the given conditions depending on the data source. This would also allow to use the same styling for different data sources.

Let me give you an example to illustrate the idea better.
Also lets imagine that we would have another datasource where the kind attribute for a road is defined as 'roadClass' but is semantically the same as 'kind == road'

```js
 const baseTheme: Theme = {
                definitions: {
                    roadColor: { type: "color", value: "#f00" },
                    roadCondition: 'when: "kind == 'road"',
                },
                styles: {
                    tilezen: [
                        {
                            description: "roads",
                            when: { $ref : roadCondition }
                            technique: "solid-line",
                            attr: {
                                lineColor: { $ref: "roadColor" }
                            }
                        }
                    ]
                }
            };

 const otherDataSourceTheme: Theme = {
                extends: baseTheme,
                definitions: {
                    roadCondition: 'when: "kind == 'roadClass"',
                }
            };
```


One more thing to test would be to define a condition in 'definitions' and refer to the condition in a given style in a stlyset.

zbigg: @SahinHERE New patch contains support for
```
roadCondition: { type: "selector", value: "kind == 'road'"}
```

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1027,6 +1027,17 @@ export class MapView extends THREE.EventDispatcher {
      * Changes the `Theme` used by this `MapView` to style map elements.
      */
     set theme(theme: Theme) {
+        if (!ThemeLoader.isThemeLoaded(theme)) {
```
zbigg: Theme with `extends` may need asynchronous loading.

Poll: 
1) should we throw here, saying that one should apply `ThemeLoader.load` on inherited theme
1) should we do as in patch, just set theme asynchronously

2 seems more user-friendly (inherited from URL theme is valid theme too) and `set theme` is already heavily asynchronous without sane way to tell when it actually realized, so don't change grand scheme of things.

SahinHERE: Looking at it, I question the existence of the theme setter function. Since setting a theme is asynchronous by nature and could involve resolving of external resource like fonts , poi atlas etc, I would propose to remove the setter function and only provide a async theme resolve/load function for the public API.

We have to differentiate between resolving theme inheritance through resolving the "extends" and loading the actual theme which involves resolving resources related to the theme. 

Also by doing so, we would add a sane way to detect when a theme is full ready to be used

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -531,6 +737,28 @@ export interface NoneStyle extends BaseStyle {
     };
 }
 
+export interface ReferenceStyle {
+    $ref: string;
+
+    /**
+     * TODO, not needed - only to satisfy typedoc
```
StefanDachwitz: Did we not decide to close a comment with a period?

@here/harp-examples/src/hello_custom_theme.ts
```diff
@@ -0,0 +1,129 @@
...
+import { accessToken } from "../config";
+
+/**
+ * This example shows how to customize buildin `Berlin` theme using [Theme] `definition` mechanism.
+ *
+ * > NOTE:
+ * > This example focuses on [[Theme]] customization. Please refer to [[HelloWorldExample]] for
+ * > introduction how to use [[MapView]]
+ *
+ * [Theme]s can inherit from another theme by using `extens: URL` property:
```
StefanDachwitz: extends
("Code Spell Checker" extension from streetsidesoftware.code-spell-checker is what I am using)

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: I would rather not see an implicit/uncontrollable) asynchronous behavior here. Why not return a promise if it is a theme that needs async loading, which would, of course, require a function not a setter. If a setter is required/desired, I could live with an exception here.

@here/harp-utils/lib/ContextLogger.ts
```diff
@@ -0,0 +1,72 @@
...
+    pushAttr(name: string): void;
+    pushIndex(index: string | number): void;
+    pop(): void;
+}
+
+export class ContextLogger implements IContextLogger {
+    readonly context: string[] = [];
+    private m_headerLogged = false;
+    constructor(readonly m_logger: IChannel, readonly headerMessage: string) {}
+    pushAttr(name: string) {
```
StefanDachwitz: did you run prettier?

@here/harp-utils/lib/ContextLogger.ts
```diff
@@ -0,0 +1,72 @@
...
+    info(message?: any, ...optionalParams: any[]): void;
+    warn(message?: any, ...optionalParams: any[]): void;
+    error(message?: any, ...optionalParams: any[]): void;
+
+    pushAttr(name: string): void;
+    pushIndex(index: string | number): void;
+    pop(): void;
+}
+
+export class ContextLogger implements IContextLogger {
```
StefanDachwitz: No docs?

@here/harp-utils/lib/ContextLogger.ts
```diff
@@ -0,0 +1,72 @@
...
+        this.context.push(`[${index}]`);
+    }
+    pop() {
+        this.context.pop();
+    }
+
+    // They, are public member functions it's just tslint who doesn't understand this.
+
+    // tslint:disable:member-ordering
+    warn = this.createLogMethod("warn");
```
StefanDachwitz: I can live with this JS style, but I would prefer TS, statically typed, tslintable code. Don't you?

@here/harp-utils/lib/ObjectUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Deep clone of object.
+ *
+ * Like `JSON.parse(JSON.stringify(obj))`, but supports basic javascript types (string, number,
+ * object), `Date` and `RegExp`s and cycles.
+ *
+ * @param obj
+ */
+export function cloneDeep<T>(obj: T): T {
```
StefanDachwitz: Are there any known limitations, especially those that would make it fail horribly? If no, great, if yes, please add comments and maybe test(s).
This may be important if the theme is no longer fully loaded from JSON, but may have been modified by users.

@here/harp-mapview/lib/MapView.ts
zbigg: > I would rather not see an implicit/uncontrollable

But it's already uncontrollable. You never know when this theme (and it's resources) are actually loaded. How `extends` differs from `poiTable[*].url` ?

Everything is uncontrollable now. `theme`, `addDataSource`
`FrameComplete` event is barely stable.

If we want to fix uncontrollable aspect of loading our stuff, we should tackle it as separate task.

Comments noted.

SahinHERE: I see the point and agree that this should be tackled on a separate task.
Otherwise I would also go for option 2.

@here/harp-utils/lib/ObjectUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+                for (let i = 0; i < result.length; ++i) {
+                    result[i] = cloneInternal(src[i]);
+                }
+                return result;
+            } else if (src.constructor === Date) {
+                const result = new Date(src.getTime());
+                cache.set(src, result);
+                return result;
+            } else if (src.constructor === RegExp) {
+                const result = new Date(src.source, src.flags);
```
robertoraggi: new RegEx.. maybe it's better if you check with `instanceof` instead of comparing the prototype

PR #583, opened 2019-06-24
HARP-6118: Implement whole GeoJson management in FeaturesDataSource.

Astrak:


Astrak:

@here/harp-examples/src/geojson.ts
```diff
@@ -0,0 +1,261 @@
...
+        const reader = new FileReader();
+        reader.onload = () => {
+            (document.querySelector(
+                "#editor textarea"
+            ) as HTMLTextAreaElement).value = reader.result as string;
+            featuresDataSource.setFromGeojson(JSON.parse(reader.result as string));
+            map.update();
+        };
+        if (file.type === "application/geo+json" || file.type === "application/json") {
+            reader.readAsBinaryString(file);
```
zbigg: https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsBinaryString

> At that time, the `result` attribute contains the *raw binary* data from the file.

Are you sure you want to use this one ?

Shouldn't you use this:
https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsText ???

@here/harp-features-datasource/lib/FeaturesDataSource.ts
```diff
@@ -38,6 +43,35 @@ export class FeaturesDataSource extends OmvDataSource {
...
+     * the data source, one should loop through it to create [[MapViewFeature]] and add them with
+     * the `add` method.
+     *
+     * @param geojson A javascript object matching the GeoJSON specification.
+     */
+    setFromGeojson(geojson: FeatureCollection | GeometryCollection | Feature) {
+        if (geojson.type === "FeatureCollection") {
+            this.m_featureCollection = geojson;
+        } else if (geojson.type === "Feature") {
+            this.m_featureCollection = DEFAULT_GEOJSON;
```
zbigg: You're adding objects to `DEFAULT_GEOJSON` instance.

~You probably wanted
```
{ ...DEFAULT_GEOJSON }
```
But this will always write to same instance of `DEFAULT_GEOJSON.features` ...
So, this `DEFAULT_GEOJSON` doesn't work here at all, you just to call `clear`.

@here/harp-examples/src/geojson.ts
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    function processGeoJsonFile(file: File) {
+        const reader = new FileReader();
+        reader.onload = () => {
+            (document.querySelector(
+                "#editor textarea"
+            ) as HTMLTextAreaElement).value = reader.result as string;
+            featuresDataSource.setFromGeojson(JSON.parse(reader.result as string));
+            map.update();
```
zbigg: > map.update();

It shouldn't be needed. `setFromGeojson` should be enough to properly update map until new stuff is visible.

If you need this, then you should fix `FeaturesDataSource`.

@here/harp-features-datasource/lib/FeaturesDataSource.ts
```diff
@@ -38,6 +39,35 @@ export class FeaturesDataSource extends OmvDataSource {
         this.addTileBackground = false;
     }
 
+    /**
+     * This method allows to directly add a GeoJSON wihout using [[MapViewFeature]] instances. It
```
StefanDachwitz: typo: without

PR #585, opened 2019-06-24
HARP-6270, HARP-5431 Improve zoom level & near / far plane calculation for terrain.

nzjony:

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2109,11 +2109,14 @@ export class MapView extends THREE.EventDispatcher {
             // want to see far in the distance, hence we use a sufficiently large offset to ensure
             // the user can see enough, anyway we never have much detail close up, so it doesn't
             // make sense to max out the accuracy here.
-            const kFarOffset = 20000.0;
+            const kFarOffset = 200.0;
 
             const heightAboveGround = this.getHeightAboveTerrain();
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2108,6 +2112,19 @@ export class MapView extends THREE.EventDispatcher {
         }
     }
 
+    private getHeightAboveTerrain() {
+        let height = Math.abs(this.projection.groundDistance(this.m_camera.position));
+
+        if (this.elevationProvider !== undefined) {
+            const heightAboveTerrain = this.elevationProvider.getHeight(this.geoCenter);
+            if (heightAboveTerrain !== undefined) {
+                height =
```
sasha240100: The reason why I don't like using "let" (in general) is that for example in this case, let's say you have:
- `this.elevationProvider !== undefined`
- and `heightAboveTerrain !== undefined`

You do:

1. `let height = Math.abs(this.projection.groundDistance(this.m_camera.position));`
2. overwrite `height` with `this.projection.unprojectAltitude(this.m_camera.position) - heightAboveTerrain;`

you do unnecessary computation of `this.projection.groundDistance` (see 1st step) and don't use it later elsewhere.

Please use `if-else` construction instead.

nzjony: Done.

nzjony:
@sasha240100, you are still requesting changes, can you please re-review?

PR #586, opened 2019-06-25
HARP-5430 Panning now checks for terrain intersections.

nzjony:

### DIFF
--- a/@here/harp-mapview/lib/Utils.ts
+++ b/@here/harp-mapview/lib/Utils.ts
@@ -25,6 +25,8 @@ const MINIMUM_ATTRIBUTE_SIZE_ESTIMATION = 56;
 const groundPlane = new THREE.Plane(new THREE.Vector3(0, 0, 1));
 const groundSphere = new THREE.Sphere(undefined, EarthConstants.EQUATORIAL_RADIUS);
 const cameraZPosition = new THREE.Vector3(0, 0, 0);
+// Careful that these aren't both used at the same time!
+const surfaceNormal = cameraZPosition;
 const rotationMatrix = new THREE.Matrix4();
 const unprojectionMatrix = new THREE.Matrix4();
 const rayCaster = new THREE.Raycaster();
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/Utils.ts
```diff
@@ -233,10 +233,31 @@ export namespace MapViewUtils {
...
+    export function pan(
+        mapView: MapView,
+        offsetX: number,
+        offsetY: number,
+        radius: number = 100
+    ): void {
+        const cameraPosition = mapView.camera.position;
+        cameraPosition.x += offsetX;
+        cameraPosition.y += offsetY;
+        const cameraHeight = mapView.projection.unprojectAltitude(cameraPosition);
```
Astrak: shouldn't this line belong to the `if` block?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -233,10 +233,31 @@ export namespace MapViewUtils {
...
+        if (mapView.elevationProvider !== undefined) {
+            const geoPosition = mapView.projection.unprojectPoint(cameraPosition);
+            const terrainHeight = mapView.elevationProvider.getHeight(geoPosition);
+            const surfaceNormal = mapView.projection.surfaceNormal(
+                cameraPosition,
+                new THREE.Vector3()
```
Astrak: Is it possible to use one of the cached vectors constants at the top of the file instead of creating a new one?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -233,10 +233,31 @@ export namespace MapViewUtils {
...
+        offsetY: number,
+        radius: number = 100
+    ): void {
+        const cameraPosition = mapView.camera.position;
+        cameraPosition.x += offsetX;
+        cameraPosition.y += offsetY;
+        const cameraHeight = mapView.projection.unprojectAltitude(cameraPosition);
+        if (mapView.elevationProvider !== undefined) {
+            const geoPosition = mapView.projection.unprojectPoint(cameraPosition);
+            const terrainHeight = mapView.elevationProvider.getHeight(geoPosition);
```
CraigBerlin: If terrainHeight is undefined can we skip the rest? Also do we need to compute surfaceNormal is we know height check will not pass?

PR #589, opened 2019-06-26
HARP-5432 Simplified the interface to make it easier to show internal data

nzjony:


@here/harp-mapview/lib/ElevationProvider.ts
```diff
@@ -39,7 +39,8 @@ export interface ElevationProvider {
     getDisplacementMap(tileKey: TileKey): DisplacementMap | undefined;
 
     /**
-     * @returns the TilingScheme used for the DisplacementMaps returned by [[getDisplacementMap]].
+     * @returns the TilingScheme used for the DisplacementMaps returned by [[getDisplacementMap]] or
+     * null if there is no elevation [[DataSource]] attached to the [[MapView]].
      */
-    getTilingSceme(): TilingScheme;
+    getTilingSceme(): TilingScheme | null;
```
zbigg: `| undefined` maybe, we use it everywhere to mark unavailable/optional/not-always-relevant things.

While justyfying, i've also stumbled across this - https://github.com/basarat/typescript-book/blob/master/docs/javascript/null-undefined.md ... and this https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined, while we're not TypeScript i think we effectively follow their most of rules from beginning of project.
`null` is

nzjony: I liked the idea that null means "Something is currently unavailable".

In this case, if there is no ElevationProvider, then I think null is a reasonable thing to return.

But I am no Typescript expert, so if we come up with some guideline, I will follow it.

nzjony: Done

PR #593, opened 2019-06-26
HARP-3891: Display generic (open-sourced by Here) road-shields.

kkostecki:
Simple solution for generic road-shields display (same for every world
region). Solves problem with different road names (name length).


zbigg:
@heremaps-bot ok to test

kkostecki:
For now I have used mapbox icons, generated with their tool (requires license check):
https://labs.mapbox.com/map-icon-editor/editor/

Later on we may change it to custom icons created by graphics.

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -58,7 +58,7 @@ type Literal = "boolean" | "number" | "string";
 /**
  * @hidden
  */
-type ExprKind = "var" | "in" | Literal | UnaryOp | RelationalOp | EqualityOp | LogicalOp;
+type ExprKind = "var" | "in" | "length" | Literal | UnaryOp | RelationalOp | EqualityOp | LogicalOp;
```
ninok: What happens if there is an attribute "length"?

kkostecki: It will still work, please note double brackets '(...)', and function call semantics. I have checked it by changing 'length' function name to 'ref', so we could even use such conditions:
"when": "$layer == 'roads' && ... && has(ref) && (ref(ref) == 3)"

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -458,6 +411,85 @@ export class PoiManager {
...
+        tile: Tile,
+        poiGeometry: PoiGeometry,
+        technique: PoiTechnique,
+        techniqueIdx: number,
+        positions: THREE.BufferAttribute
+    ) {
+        const poiTechnique = technique as PoiTechnique;
+        let imageTextureName = poiTechnique.imageTexture;
+        if (typeof poiTechnique.imageTexturePrefix === "string") {
+            imageTextureName = poiTechnique.imageTexturePrefix + imageTextureName;
```
sasha240100: My opinion is that this duplicating "add prefix/postfix" code should be separated in a function. Something like `getImageTextureName(technique)`.

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -458,6 +411,85 @@ export class PoiManager {
...
+        }
+        if (typeof poiTechnique.imageTexturePostfix === "string") {
+            imageTextureName = imageTextureName + poiTechnique.imageTexturePostfix;
+        }
+
+        if (poiGeometry.stringCatalog === undefined) {
+            return;
```
sasha240100: This condition should be at the start of the function. Otherwise, we execute unnecessary prefix/postfix code

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -458,6 +411,85 @@ export class PoiManager {
...
+        const poiTableName = poiTechnique.poiTable;
+        let poiName = poiTechnique.poiName;
+
+        for (let i = 0; i < positions.count; ++i) {
+            const x = positions.getX(i);
+            const y = positions.getY(i);
+            const z = positions.getZ(i);
+
+            assert(poiGeometry.texts.length > i);
+            let text = poiGeometry.stringCatalog[poiGeometry.texts[i]];
```
sasha240100: Why not just?
`const text = poiGeometry.stringCatalog[poiGeometry.texts[i]] || "";`

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -458,6 +411,85 @@ export class PoiManager {
...
+            if (text === undefined) {
+                text = "";
+            }
+
+            if (poiGeometry.stringCatalog !== undefined) {
```
sasha240100: Could `poiGeometry.stringCatalog` be `undefined` here if on line 434 we exit the function if it is?

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -458,6 +411,85 @@ export class PoiManager {
...
+        }
+        if (typeof poiTechnique.imageTexturePostfix === "string") {
+            imageTextureName = imageTextureName + poiTechnique.imageTexturePostfix;
+        }
+
+        if (poiGeometry.stringCatalog === undefined) {
+            return;
+        }
+        const poiTableName = poiTechnique.poiTable;
+        let poiName = poiTechnique.poiName;
```
sasha240100: `const poiName = poiTechnique.poiName === undefined ? imageTextureName : poiTechnique.poiName`;

@here/harp-mapview/lib/poi/PoiManager.ts
kkostecki: Don't ask me ... not my code :).

@here/harp-mapview/lib/poi/PoiManager.ts
kkostecki: Again not my code, but ok, will fix.

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -205,6 +205,36 @@ class ContainsExpr extends Expr {
...
+ * For boolean data types always returns 1.
+ */
+class LengthExpr extends Expr {
+    constructor(readonly attribute: string) {
+        super("length");
+    }
+
+    evaluate(env: Env): Value | never {
+        const attrib = env.lookup(this.attribute);
+        if (typeof attrib === "boolean") {
```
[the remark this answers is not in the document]
kkostecki: Hmm.. may look better....

@here/harp-mapview/lib/poi/PoiManager.ts
kkostecki: Will try something.

@here/harp-mapview/lib/poi/PoiManager.ts
kkostecki: Again I just moved some stuff to separate function (refactor). Will check it.

@here/harp-datasource-protocol/lib/Expr.ts
ninok: okay, but I think readability of this is bad, but hopefully this will never happen.
For the length function I would prefer to rename it to sth. like strLength b/c the probability that there will be a length attribute in the data is quite high (i.e. we have it already for truck restrictions).

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -136,75 +136,21 @@ export class PoiManager {
             );
 
             if (isLineMarkerTechnique(technique) && positions.count > 0) {
-                this.addLineMarker(tile, poiGeometry, technique, techniqueIndex, positions);
+                this.addLineMarker(
+                    tile,
+                    poiGeometry,
+                    technique as LineMarkerTechnique,
```
ninok: Minor: the as LineMarkerTechnique is not needed, b/c typescipt can deduce the type from the condition above.

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -136,75 +136,21 @@ export class PoiManager {
...
+                    techniqueIndex,
+                    positions
+                );
             } else {
```
ninok: `else if (isPoiTechnique(...))` ?

@here/harp-mapview/lib/poi/PoiManager.ts
zbigg: @sasha240100, @kkostecki 

While i agree that `|| ""` is more readable and correct in this case, please note that we have very strict rule disallowing any truthy/falsy statements an this one is exactly false`y for `""` in array albeit correct ("" yields "", so no harm) .

That's why `||` and `&&` on non-booleans is avoided like a plague even if we loose some compactness of the code.

@here/harp-map-theme/resources/berlin_tilezen_base.json
```diff
@@ -86,19 +91,125 @@
...
+                    "priority": 64,
+                    "minDistance": 200,
+                    "vAlignment": "Center",
+                    "hAlignment": "Center",
+                    "textIsOptional": false,
+                    "iconIsOptional": false,
+                    "textMayOverlap": false,
+                    "iconMayOverlap": false,
+                    "iconReserveSpace": true,
+                    "renderTextDuringMovements": true,
```
zbigg: > renderTextDuringMovements

Is this possible, that this one is responsible for severe flickering during mouse interactions ?

@here/harp-map-theme/resources/berlin_tilezen_base.json
```diff
@@ -86,19 +91,125 @@
...
+                    "color": "#000000"
+                },
+                "renderOrder": 14.3
```
zbigg: @Astrak, @
`renderOrder` is not needed for techniques drawn in "2d" space

@here/harp-mapview/lib/poi/PoiManager.ts
kkostecki: We may add this, but it is impossible to have something else anyway, see condition above:
`
if (!isLineMarkerTechnique(technique) && !isPoiTechnique(technique)) {
   continue;
}
`

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -205,6 +205,39 @@ class ContainsExpr extends Expr {
...
+            return 1;
+        }
+        else if(typeof attrib === "number") {
```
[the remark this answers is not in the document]
kkostecki: Hmm... we can do it like that, if it still makes sense, will add this.

@here/harp-map-theme/resources/berlin_tilezen_base.json
kkostecki: I guess that rather, occlusion test:
 "iconMayOverlap": false,
causes the flicker, because changing it on POIs causes the same issue.

@here/harp-map-theme/resources/berlin_tilezen_base.json
kkostecki: Ok, then I will remove it.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -214,17 +215,11 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
-                    const imageTextureValue = env.lookup(poiTechnique.imageTextureField);
-                    imageTexture = imageTextureValue as string;
-                    if (typeof poiTechnique.imageTexturePrefix === "string") {
-                        imageTexture = poiTechnique.imageTexturePrefix + imageTexture;
-                    }
-                    if (typeof poiTechnique.imageTexturePostfix === "string") {
-                        imageTexture = imageTexture + poiTechnique.imageTexturePostfix;
-                    }
+                    const imageTextureValue = env.lookup(poiTechnique.imageTextureField) as string;
+                    imageTexture = PoiManager.composeTextureName(imageTextureValue, poiTechnique);
```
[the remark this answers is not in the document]
kkostecki: Ok.

kkostecki:
@heremaps-bot ok to test

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -366,3 +366,25 @@ export function addBuffersToTransferList(technique: Technique, transferList: Arr
         }
     }
 }
+
+/**
+ * Compose full texture name for given image name with technique specified.
+ * Some techniques allows to add prefix/posix to icons names specified, this
```
okanishchev:
```suggestion
 * Some techniques allows to add prefix/postfix to icons names specified, this
```

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -450,10 +394,75 @@ export class PoiManager {
...
+        for (let i = 0; i < positions.count; ++i) {
+            const x = positions.getX(i);
+            const y = positions.getY(i);
+            const z = positions.getZ(i);
+
+            assert(poiGeometry.texts.length > i);
+            const text: string = poiGeometry.stringCatalog[poiGeometry.texts[i]] || "";
+            if (poiGeometry.imageTextures !== undefined && poiGeometry.imageTextures[i] >= 0) {
+                assert(poiGeometry.imageTextures.length > i);
+                imageTextureName = poiGeometry.stringCatalog[poiGeometry.imageTextures[i]];
```
okanishchev: imageTextureName defined out of loop above gets overwritten and will never be restored. So if next loop iteration will fail this ```if``` condition it will get imageTextureName from previous loop iteration, and not default defined above.

If this is desired behavior, there should be some description comment. And if not, I suggest to reset this variable to default value at the beginning of each loop iteration.

@here/harp-datasource-protocol/lib/Techniques.ts
kkostecki: Ok will fix.

@here/harp-mapview/lib/poi/PoiManager.ts
kkostecki: Good finding, will fix!

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -450,10 +394,78 @@ export class PoiManager {
...
+        const poiTableName = poiTechnique.poiTable;
+        let poiName = poiTechnique.poiName;
+
+        for (let i = 0; i < positions.count; ++i) {
+            const x = positions.getX(i);
+            const y = positions.getY(i);
+            const z = positions.getZ(i);
+
+            assert(poiGeometry.texts.length > i);
+            let imageTextureName;
```
okanishchev: You even can do
```suggestion
            let imageTextureName = techniqueTextureName;
```

kkostecki: Yeah, I was even struggling, but old way won this time :+1:

PR #597, opened 2019-06-28
Update Readme.md and GetstartedGuide.md

chfw:
- </cavas> is add
- authentication code is noted.

-----

Note, every developer need to trade in birthday to get authentication code. Is it a common practice?


ninok:
@heremaps-bot ok to test

musculman:
Unfortunately, we have an issue right now when we mix in content from an external fork.
I am overriding the failure to merge it.

Thanks for your contribution @chfw 

As for the birth date that is requested, this is not required of course to use harp.gl, but only to use HERE Data. You could use different data if you would like to.

dbabbs:
Thanks for pointing out this `</canvas>` tag issue.

PR #601, opened 2019-07-02
Create style specification guide

dbabbs:
Twitter user is asking for information about the styling syntax for the map theme. Adding this document as a basic starting point. To be expanded in the future.


docs/StyleSpecificationGuide.md
```diff
@@ -0,0 +1,217 @@
...
+   baseUrl: "https://xyz.api.here.com/tiles/herebase.02",
+   apiFormat: harp.APIFormat.XYZOMV,
+   styleSetName: "tilezen", // <- style set name goes here
+   authenticationCode: 'YOUR-XYZ-TOKEN HERE', 
+});
+```
+
+You are free to have as many style set names as you would like.
+
+Within each style set name is where you can define style properties for layers. An individual style rule object has at least four required properties:
```
ninok: The user of "layer" in this section is not really correct.
One style refers to a subset of  the data (defined by the condition). So a style could be a subset of a layer but could also work across layers (e.g. you could have one style for all points in the datasource)

musculman:
> There was also some work done by Oleksii to generate the style out of the json schema. I think there is some overlap.
> https://github.com/heremaps/harp.gl/blob/master/%40here/harp-datasource-protocol/README.md

Definitively, it should be merged or combined somehow

PR #604, opened 2019-07-05
HARP-6181: Zombie Level [0...3] Tiles

okanishchev:

Prevent text elements to be rendered or compared multiple times;

Performance of different ways of getting unique element list:
https://jsperf.com/set-vs-array-addition


okanishchev:
The complete description is in the HARP-6181 ticket.
Shortly, clearing tile cache on zoom levels 0 to 3 causes some text labels to disappear.
Inspecting the issue, I've found that text elements getting pushed to same array twice, and then make themselves to fail collision test.

PR #606, opened 2019-07-08
HARP-6426: Non-text features missing after cache clear on lower level tiles

CraigBerlin:


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -515,14 +520,14 @@ export class TileGeometryCreator {
                 const technique = decodedTile.techniques[techniqueIndex];
 
                 if (
-                    group.created === true ||
+                    group.createdOffsets!.some(offset => offset === tile.offset) ||
```
okanishchev: As tile offset is a plain number, we can use indexOf
```suggestion
                    group.createdOffsets!.indexOf(tile.offset) !== -1 ||
```

CraigBerlin: Thanks changed, agree code is easier to read.

PR #608, opened 2019-07-09
HARP-6418: Fix label filtering

StefanDachwitz:
* Labels can now be hidden and disabled
 * Added option to disable phased loading

* [contribution guidelines](../CONTRIBUTING.md)
* [coding style](../CODINGSTYLE.md)

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -31,7 +31,7 @@ import { DataSource } from "./DataSource";
 import { ElevationProvider } from "./ElevationProvider";
 import { ElevationRangeSource } from "./ElevationRangeSource";
 import { PhasedTileGeometryManager } from "./geometry/PhasedTileGeometryManager";
-import { TileGeometryManager } from "./geometry/TileGeometryManager";
+import { SimpleTileGeometryManager, TileGeometryManager } from "./geometry/TileGeometryManager";
 import { MapViewImageCache } from "./image/MapViewImageCache";
 import { MapViewFog } from "./MapViewFog";
 import { PickHandler, PickResult } from "./PickHandler";
@@ -497,6 +497,14 @@ export interface MapViewOptions {
     maxFps?: number;
 
     /**
+     * Enable phased loading. If `false`, the geometry on a [[Tile]] is always being created in a
+     * single step, instead of (potentially) over multiple frames to smoothen animations.
+     *
+     * @default `true`
+     */
+    enablePhasedLoading?: boolean;
+
+    /**
      * @hidden
      * Disable all fading animations for debugging and performance measurement.
      */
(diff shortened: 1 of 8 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -496,6 +496,14 @@ export interface MapViewOptions {
      */
     maxFps?: number;
 
+    /**
+     * Disable phased loading. If `true`, the geometry on a [[Tile]] is always being created in a
+     * single step, instead of (potentially) over multiple frames to smoothen animations.
+     *
+     * @default `undefined`
+     */
+    disablePhasedLoading?: boolean;
```
okanishchev: It is usually better to have "enabled"-kind flags (defaulted to true in this case), it's simply to read and understand.
```suggestion
    enablePhasedLoading?: boolean;
```

StefanDachwitz: That is true, changing the wording. But your "suggested change" would not be sufficient since more changes are required if the logic/default value is not supposed to change.

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
```diff
@@ -240,7 +240,10 @@ export abstract class TileGeometryManagerBase implements TileGeometryManager {
                 if (geometryKind !== undefined) {
                     let isVisible = true;
```
sasha240100:
```js
const isVisible = !geometryKind.some(kind => this.hiddenKinds.has(kind));
```

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1324,6 +1324,10 @@ export class TextElementsRenderer {
         const tempMeasurementParams: MeasurementParameters = {};
         const tempBufferAdditionParams: TextBufferAdditionParameters = {};
 
+        const tileGeometryManager = this.m_mapView.tileGeometryManager;
+        const hiddenKinds =
```
sasha240100:
```js
const hiddenKinds = tileGeometryManager && tileGeometryManager.hiddenGeometryKinds;
```

Will return `tileGeometryManager.hiddenGeometryKinds` if `tileGeometryManager !== undefined`;
Otherwise returns `undefined`;

Unless there is no specific need in typescript check 🤔

@here/harp-mapview/lib/MapView.ts
okanishchev: Yes, it's not sufficient. I just don't know how to suggest a couple of changes around files.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
StefanDachwitz: Please read our coding style: Avoid truthies

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: You do not need to add suggestions here, the author will handle that. If the motivation for the requested change is substantiated and well explained (as is is here), your job is done! ;)

okanishchev: Ok. Thanks for the advice!

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
StefanDachwitz: I am always careful to call functions, since my experience tells me that they are slower, so I had to do the test (https://jsperf.com/use-for-loop-instead-of-anonymous-function/1).
And voilá, it save a couple of code lines along with a WHOPPING 30% of execution time.
So, applying the suggested change!

PR #610, opened 2019-07-09
HARP-5398: Stable labels across zoom levels

pit-rpg:

### DIFF
--- a/@here/harp-mapview/lib/text/TextElement.ts
+++ b/@here/harp-mapview/lib/text/TextElement.ts
@@ -293,12 +293,30 @@ export class RenderState {
             this.state === FadingState.FadingOut;
         return visible;
     }
+
+    /**
+     * Creates a clone of this object, uid is ignored.
+     */
+    clone() {
+        return new RenderState(
+            this.state,
+            this.value,
+            this.startTime,
+            this.opacity,
+            this.lastFrameNumber,
+            this.fadingTime
+        );
+    }
 }
 
 /**
  * `TextElement` is used to create 2D text elements (for example, labels).
  */
 export class TextElement {
+    /**
+     * [[TextElement]] unique ID counter
+     */
+    private static uidCounter = 0;
     /**
      * Determines visibility. If set to `false`, it will not be rendered.
      */
@@ -397,26 +415,6 @@ export class TextElement {
 
     /**
      * @hidden
-     * Used during rendering.
-     */
-    iconRenderState?: RenderState;
-
-    /**
-     * @hidden
-     * Used during rendering.
-     */
-    textRenderState?: RenderState;
-
-    /**
-     * @hidden
-     * Used during rendering. Used for line markers only, which have a points array and multiple
-     * icon positions to render. Since line markers use the same renderState for text part and icon,
-     * there is no separate array of [[RenderState]]s for the text parts of the line markers.
-     */
-    iconRenderStates?: RenderState[];
-
-    /**
-     * @hidden
      * Text rendering style.
      */
     renderStyle?: TextRenderStyle;
@@ -462,7 +460,13 @@ export class TextElement {
      */
     dbgPathTooSmall?: boolean;
 
+    /**
+     * A unique id for each text element
+     */
+    readonly uid = TextElement.uidCounter++;
+
     private m_poiInfo?: PoiInfo;
+    private m_identityKey?: string;
 
     /**
      * Creates a new `TextElement`.
@@ -616,5 +620,22 @@ export class TextElement {
                 );
             }
         }
+    }
+
+    /**
+     * Generates a string from text object properties that can be used to find similar elements.
+     * The text element position and uid are ignored
+     */
+    getWeakLabelHash(): string {
+        if (this.m_identityKey === undefined) {
+            this.m_identityKey =
+                this.text +
+                (this.poiInfo === undefined ? "-" : this.poiInfo.imageTextureName) +
+                (this.featureId || "_") +
+                (this.isLineMarker ? "+" : "-") +
+                (this.path ? this.path.length : "-");
+        }
+
+        return this.m_identityKey;
     }
 }
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -71,6 +71,30 @@ export interface TextElementStyle {
     poiRenderer?: PoiRenderer;
 }
 
+interface RenderStateLabelData {
+    renderStateText: RenderState;
+    renderStateIcon: RenderState;
+    priority: number;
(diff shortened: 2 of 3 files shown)


@here/harp-mapview/lib/text/TextElement.ts
```diff
@@ -609,4 +612,21 @@ export class TextElement {
             }
         }
     }
+
+    /**
+     * Something like a unique hash for each text element, the text element position, and uid are
+     * ignored
+     */
+    getIdentityKey(): string {
```
ninok: As far as I can see the unique id is mainly based on the text. What if the label has no text (I guess this can happen for POIs) or even worse if the text is the same (this will happen for sure for POIs like gas stations).

pit-rpg: I can't identify labels between zoom levels because their position changes at each level.
This function generates a key by which I can place similar labels render states in one array, and then if needed we can find and copy a state for a new label at another zoom level.
I'll try to rename this method

@here/harp-mapview/lib/text/TextElement.ts
```diff
@@ -609,4 +612,20 @@ export class TextElement {
             }
         }
     }
+
+    /**
+     * Something like a hash for similar text elements, the text element position, and uid ignored
```
haraldF: "Something like" is not good documentation style. Can you reword please?

@here/harp-mapview/lib/text/TextElement.ts
```diff
@@ -291,8 +291,25 @@ export class RenderState {
...
+            this.startTime,
+            this.opacity,
+            this.lastFrameNumber,
+            this.fadingTime
+        );
+    }
 }
 
+// Unique ID counter
+let TEXT_ELEMENT_UID = 0;
```
okanishchev: It would be better to make it as private static property of TextElement class:

```js
class TextElement {
    private static counter: number = 0;
    readonly uid: number = ++TextElementCounter;
}
```

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -280,13 +306,49 @@ export class TextElementsRenderer {
...
+        // if level do not loaded or old data cleaned up do nothing
+        if (!zoomLevelCompletelyLoaded || this.m_renderStatesCleanedUp) {
+            return;
+        }
+
+        // if level completely loaded clean up unnecessary states
+        for (let [key, textElementStates] of Object.entries(this.m_renderStateStore)) {
+            // states that older then 120 frames are removed
+            textElementStates = textElementStates.filter(
+                state => state.renderStateText.lastFrameNumber + 120 > frameNumber
```
okanishchev: Why 120 is used? Saving this as constant with a descriptive name somewhere at the top and referencing here will be more readable.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -858,17 +906,15 @@ export class TextElementsRenderer {
...
         // than once per element. Also, make sorting stable by taking the index into the array into
         // account, this is required to get repeatable results for testing.
-        for (let i = 0; i < textElements.length; i++) {
-            const textElement = textElements[i];
-
+        for (const textElement of textElements) {
             textElement.sortPriority =
                 textElement.priority +
-                i * indexPriorityFactor +
+                textElement.priority * uidPriorityFactor +
```
okanishchev: I believe this is a mistake
```suggestion
                textElement.uid * uidPriorityFactor +
```

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -2232,32 +2242,27 @@ export class TextElementsRenderer {
...
-            if (renderState.isFading()) {
-                this.updateFading(renderState, time);
-                fadeAnimationStarted = true;
-            }
-
-            renderState.lastFrameNumber = frameNumber;
+    // if not fading in fade in, if fading out reverse
+    private ensureFadeIn(renderState: RenderState, frameNumber: number, time: number) {
+        renderState.lastFrameNumber = frameNumber;
+        if (renderState.state === FadingState.FadingIn) {
```
okanishchev: It's a perfect case to use ```switch```

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -2232,32 +2242,27 @@ export class TextElementsRenderer {
...
-            }
-
-            if (renderState.isFading()) {
-                this.updateFading(renderState, time);
-                fadeAnimationStarted = true;
-            }
-
-            renderState.lastFrameNumber = frameNumber;
+    // if not fading in fade in, if fading out reverse
+    private ensureFadeIn(renderState: RenderState, frameNumber: number, time: number) {
```
okanishchev: The function named ensureFadeIn returns false if render state is FadedIn and true otherwise. And that is completely non-intuitive.
I would prefer to preserve that variable ```fadeAnimationStarted``` to keep awareness on what's going on.

@here/harp-mapview/lib/text/TextElement.ts
```diff
@@ -609,4 +613,21 @@ export class TextElement {
             }
         }
     }
+
+    /**
+     * Generates a string from text object properties that can be used to find similar elements.
+     * The text element position and uid ignored
```
musculman: Perhaps it is missing "are ignored"? instead of just "ignored"

@here/harp-mapview/lib/text/TextElementsRenderer.ts
musculman: Yes please, say no to magic numbers :)

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -858,17 +906,15 @@ export class TextElementsRenderer {
 
     private sortTextElements(textElements: TextElement[], maxViewDistance: number) {
         const distancePriorityFactor = 0.1;
-        const indexPriorityFactor = 0.01 * (1 / textElements.length);
+        const uidPriorityFactor = 0.0000000001;
```
musculman: Why this change from previous formula to hard-coded value?

pit-rpg: The previous one calculates the priority used the array length, but, the length is changing and this affects the priority across frames. To make the priority value more stable I've used the text element UID. So if the visible tileset is changed during camera movement this does not affect the displayed label's priority.
This value just a very small but static number for reducing the UID weight on text element priority.

pit-rpg:
Decoders tests please

DimitryRd:
Decoders tests please

PR #611, opened 2019-07-10
HARP-6181: Zombie Level [0...3] Tiles

okanishchev:

Simply clearing tile loader cache when mapview tile cache is being cleared.

### DIFF
--- a/@here/harp-mapview/lib/DataSource.ts
+++ b/@here/harp-mapview/lib/DataSource.ts
@@ -152,6 +152,13 @@ export abstract class DataSource extends THREE.EventDispatcher {
     }
 
     /**
+     * Purges all the caching done by this `DataSource`
+     */
+    clearCache() {
+        // to be overloaded by subclasses
+    }
+
+    /**
      * Returns `true` if this `DataSource` is ready and the [[MapView]] can invoke `getTile()` to
      * start requesting data.
      */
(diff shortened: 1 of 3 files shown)


Astrak:
Thanks! This is the issue that was blocking my ticket for a week :)

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -151,6 +151,10 @@ export abstract class DataSource extends THREE.EventDispatcher {
         // to be overloaded by subclasses
     }
 
+    clearCache() {
```
CraigBerlin: Please add comments describing the purpose of the method

okanishchev: Done. But it should be obvious from the method name.

CraigBerlin: True but this is public API in a core base class that will be implemented by users. For these classes we should have documentation, even if it is trivial.

okanishchev:

nzjony:
hey @okanishchev, I don't quite understand how this changes make a difference. Why does clearing  the cache make a difference to the loading of the text. I tried a revert locally and I couldn't see any difference, I will make a revert and see if I can prove it using the CI

okanishchev:
@nzjony, decoded tiles for [0..3] zoomlevels are cached in tile loaders, and when you clear tile cache, new tiles for that levels are retrieved from tile loader cache, and not from decoder.
This matters when you change map theme - new styles are not applied to already decoded tiles, only to newly decoded.

nzjony:
> This matters when you change map theme - new styles are not applied to already decoded tiles, only to newly decoded.

`setStyleSet` already calls `this.m_tileLoaderCache.clear();`

PR #582, opened 2019-06-24
HARP-5574: StringEncodedNumeral support.

amsgames94:
* Added support for ``StringEncodedNumeral``, which are number values which can be specifed as strings (i.e. colors, distances in pixel, etc.).
* Improved the ``TechniqueParams`` typings to be more strict with the new ``color`` and ``length`` types.
* Implemented mix interpolation between:
    - Pixels and meter values (``metricUnit`` is now deprecated).
    - Hex, rgb and hsl colors (interpolation is performed internally in HSL space for better results).
* Refactored the ``getPropertyValue`` API to always return ``number`` (and also handle ``length`` and ``color`` interpolation properly).

PR #614, opened 2019-07-11
HARP-6436: POI and transport icons are missing for all data souces

okanishchev:

### DIFF
--- a/@here/harp-mapview/lib/poi/PoiManager.ts
+++ b/@here/harp-mapview/lib/poi/PoiManager.ts
@@ -491,11 +491,11 @@ export class PoiManager {
             const fadeNear =
                 technique.fadeNear !== undefined
                     ? getPropertyValue(technique.fadeNear, displayZoomLevel)
-                    : 0;
+                    : technique.fadeNear;
             const fadeFar =
                 technique.fadeFar !== undefined
                     ? getPropertyValue(technique.fadeFar, displayZoomLevel)
-                    : 0;
+                    : technique.fadeFar;
 
             const positions = Array.isArray(x)
                 ? (x as THREE.Vector3[])
--- a/@here/harp-datasource-protocol/lib/TechniqueParams.ts
+++ b/@here/harp-datasource-protocol/lib/TechniqueParams.ts
@@ -194,7 +194,7 @@ export interface BaseTechniqueParams {
 
     /**
      * Distance to the camera (0.0 = camera position, 1.0 = farPlane) at which the object has zero
-     * opacity and stops fading out. A value of <= 0.0 disables fading.
+     * opacity and stops fading out. An undefined value disables fading.
      */
     fadeFar?: MaybeInterpolatedProperty<number>;
 


@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -491,11 +491,11 @@ export class PoiManager {
             const fadeNear =
                 technique.fadeNear !== undefined
                     ? getPropertyValue(technique.fadeNear, displayZoomLevel)
-                    : 0;
+                    : technique.fadeNear;
```
kkostecki: I think it should be rather 0 or even better defaultRenderParams.fadeNear, otherwise you will assign undefined here (intentionally?).

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -491,11 +491,11 @@ export class PoiManager {
...
             const fadeFar =
                 technique.fadeFar !== undefined
                     ? getPropertyValue(technique.fadeFar, displayZoomLevel)
-                    : 0;
+                    : technique.fadeFar;
```
kkostecki: Similarly to above, but by default I would assign 1.0 or defaultRenderParams.fadeFar.

okanishchev:
Yep, that undefineds are intentional - it indicates that no clipping should occur, because 0 is valid value both for fadeNear and fadeFar. And getPropertyValue was completely reworked in #582 and as I assume, not accept undefineds anymore.

kkostecki:
> Yep, that undefineds are intentional - it indicates that no clipping should occur, because 0 is valid value both for fadeNear and fadeFar. And getPropertyValue was completely reworked in #582 and as I assume, not accept undefineds anymore.

Ok, I see it now... (TextElementsRenderer.ts, getDistanceFadingFactor) seems that fadeNear === undefined may be interpreted as 0 (zero) while fadeFar === undefined disables fading... hmmm... quite intuitive :1st_place_medal:  isn't it?
Please update at least documentation (mentioned above) cause it is misleading.

kkostecki:
> > Yep, that undefineds are intentional - it indicates that no clipping should occur, because 0 is valid value both for fadeNear and fadeFar. And getPropertyValue was completely reworked in #582 and as I assume, not accept undefineds anymore.
> 
> Ok, I see it now... (TextElementsRenderer.ts, getDistanceFadingFactor) seems that fadeNear === undefined may be interpreted as 0 (zero) while fadeFar === undefined disables fading... hmmm... quite intuitive 1st_place_medal isn't it?
> Please update at least documentation (mentioned above) cause it is misleading.

Ok, I have spoke with Zyszek and we are quite sure that the problems comes from different source, maybe it is better just to support undefined in InterpolatedProperty.getPropertyValue(... | undefined) which will obviously result in undefined and reduce the amount of boilerplate code.

okanishchev:
Updated docs.

okanishchev:
> Ok, I have spoke with @zbigg and we are quite sure that the problem comes from different source, maybe it is better just to support undefined in InterpolatedProperty.getPropertyValue(... | undefined) which will obviously result in undefined and reduce the amount of boilerplate code.

Well, actual implementation allows passing undefined to getPropertyValue.

PR #619, opened 2019-07-13
Improve examples CSS.

Astrak:
This PR implements the following:

1.  **Categories for the examples** > an example named A goes into category "Miscellaneous", those named A_B_... go into category "A" (and create it if missing). The categories can be optionally ordered.
2. **CSS in the HTML** > no more broken CSS during bundle parsing.
3. **Standardized mobile version**.
4. **Colors consistency** > Background color consistent with MapControlsUI, "harp.gl" gradient consistent with landing page. 
5. **Links** to doc + repo are included.

preview


dbabbs:
The categorization is a great improvement. Easier to find exactly what you want.

Why did you choose that color scheme and font? We should try to comply with the HERE design system.

In the future, I think next steps should be to combine the harp.gl landing page and this examples page instead a single unified site (consistent header and url structure). Happy to work on that.

Astrak:
> The categorization is a great improvement. Easier to find exactly what you want.

It was an idea from Nino I think. A next step is then to rename the examples to use them more accurately. 

> Why did you choose that color scheme and font? 

I reused the hue you picked for the controls but varied the luminosity values. The color for "harp" is the complementary, with more saturation and luminosity as well. The font for is the default Arial font, for "harp" it has a narrowed letter-spacing. 

> We should try to comply with the HERE design system.

When I originally designed the controls' UI it was HERE branded, I was then told not to do it, because harp.gl is not HERE branded?

Astrak:

@here/harp-examples/index.html
```diff
@@ -1,65 +1,367 @@
...
+            #magnifier-placeholder {
+                position: absolute;
+                top: 22px;
+                left: 16px;
+                pointer-events: none;
+                font-size: 25px;
+                transform: rotateZ(-45deg);
+                color: #4e5a6e;
+            }
+
```
okanishchev: You can also hide icon when search field gains focus
```suggestion

#filterInput:focus + #magnifier-placeholder {
    display: none;
}
```

@here/harp-examples/example-browser.ts
```diff
@@ -51,21 +50,57 @@ function exampleBrowser(exampleDefinitions: ExampleDefinitions) {
...
+                let linkSubMenu = linkElements.length > 1 ? linkElements[0] : "miscellaneous";
+                linkSubMenu = linkSubMenu.charAt(0).toUpperCase() + linkSubMenu.slice(1);
+                if (menuElements[linkSubMenu] === undefined) {
+                    menuElements[linkSubMenu] = [];
+                }
+                const visibleText =
+                    linkElements.length === 1
+                        ? linkElements[0]
+                        : linkElements.slice(1).join(" (") +
+                          (linkElements.slice(1).length > 1 ? ")" : "");
```
okanishchev:
```suggestion
                          (linkElements.length > 2 ? ")" : "");
```

@here/harp-examples/example-browser.ts
```diff
@@ -51,21 +50,57 @@ function exampleBrowser(exampleDefinitions: ExampleDefinitions) {
...
+                const linkElements = linkName.split(" / ");
+                let linkSubMenu = linkElements.length > 1 ? linkElements[0] : "miscellaneous";
+                linkSubMenu = linkSubMenu.charAt(0).toUpperCase() + linkSubMenu.slice(1);
+                if (menuElements[linkSubMenu] === undefined) {
+                    menuElements[linkSubMenu] = [];
+                }
+                const visibleText =
+                    linkElements.length === 1
+                        ? linkElements[0]
+                        : linkElements.slice(1).join(" (") +
```
okanishchev: No closing paren for inner elements.
with array ```["datasource", "geojson", "property", "styling"]``` resulting string would be ```geojson (property (styling)```

ninok:
One generic comment: We have to find a way to sort the categories by "importance). I.e. the hello category should come first.
Maybe we can have as scheme like this:
`1_hello_hello`
Where the 1 is used to sort the categories but not part of the category. If the number is missing you could fallback to sorting by name.

musculman:
> When I originally designed the controls' UI it was HERE branded, I was then told not to do it, because harp.gl is not HERE branded?

Agreed, harp.gl is not in general HERE branded.
The www.harp.gl site is using the HERE colour scheme but not HERE branded, it does not have the HERE logo for example.
But, I would argue that yes, it would make sense to align the www.harp.gl minisite and the examples so that ideally the would feel like a single unified page, like in many other projects.
In any case this PR is a big improvement of what we currently have and I would like to have it.

Astrak:
> One generic comment: We have to find a way to sort the categories by "importance). I.e. the hello category should come first.
Maybe we can have as scheme like this:
1_hello_hello
Where the 1 is used to sort the categories but not part of the category. If the number is missing you could fallback to sorting by name.

Yes agreed, I remember this idea! However I haven't found a way to implement it:
- first we can't begin a variable name with a number... but we could put that optional number in the second position.
- the problem is then, if it's defined in the variable, that one may put a wrong number one day and a small category might disappear without us noticing, or it requires other assumptions in the script to work around it.

I think the most stable way to implement a custom category order is to hardcode it...?

Astrak:
I've added the "Fira Sans" font, used in the landing page, looks even better.

devallois:
This is looking much better. 
Question: Is there a reason why the category title is repeated for every item in the list beneath it?

okanishchev:
@devallois, when you filter the list with a search term, category titles are gone.

devallois:
> @devallois, when you filter the list with a search term, category titles are gone.

I see. So this means people will primarily look at examples by filtering for something specific rather than looking through the list, and so we want to make sure they see what category the example belongs to - is that right?

Astrak:
About repeating category names in the example link, received comments fall as follows so far, please edit if necessary:

PROS: 
- it allows to filter keywords comprised into categories across other categories ("geojson", "globe"...) or filter a category quicker ("xyz")

CONS:
- it is repetitive and takes useless place, clutters the interface, creates line breaks. It would be leaner without them.
- the example fonts could be bigger, more readable and more comfortable for people with lower visual accuracy.

Astrak:
Implemented the solution found by @okanishchev for the category names: only enable them during search.

@here/harp-examples/example-browser.ts
```diff
@@ -50,22 +50,76 @@ function exampleBrowser(exampleDefinitions: ExampleDefinitions) {
...
+        });
     }
 
-    function installFilter(filterInput: HTMLInputElement, fileteredElements: HTMLElement[]) {
+    function installFilter(filterInput: HTMLInputElement) {
         filterInput.addEventListener("input", e => {
             const filterValue = (filterInput.value || "").trim();
 
-            for (const element of fileteredElements) {
+            if (filterValue) {
```
okanishchev: The whole if-else block should be called only when actual condition changes, in this case ```filterValue.length > 0```, and not on every keypress. Also remember about truthy values.

@here/harp-examples/example-browser.ts
```diff
@@ -50,22 +50,76 @@ function exampleBrowser(exampleDefinitions: ExampleDefinitions) {
         Object.keys(exampleDefinitions)
             .sort()
             .forEach(pageUrl => {
-                const linkName = getName(pageUrl);
+                const linkName = getName(pageUrl).replace(new RegExp("-", "g"), " ");
```
okanishchev: Why not
```suggestion
                const linkName = getName(pageUrl).replace(/-/g, " ");
```
?

@here/harp-examples/example-browser.ts
```diff
@@ -86,12 +152,16 @@ function exampleBrowser(exampleDefinitions: ExampleDefinitions) {
      * @param pageUrl example page url, must exist in `examples` map
      */
     function showExample(pageUrl: string) {
+        const expandButton = document.getElementById("hamburgerMenu") as HTMLAnchorElement;
+        if (navPanel.classList.contains("collapsed")) {
+            expandButton.classList.remove("expanded");
+        } else {
+            expandButton.classList.add("expanded");
+        }
```
okanishchev: You can use ```.classList.toggle()``` here

Astrak: As far as I understand, the if condition prevents using toggle, as we want to remove or add depending on it, while we wouldn't control it with toggle.

@here/harp-examples/example-browser.ts
Astrak: Correct! However it is the only remaining change and I'd like not to retrigger the CI for it and merge the topic if you're ok with it.

@here/harp-examples/example-browser.ts
okanishchev: You can pass second argument to toggle
```js
expandButton.classList.toggle("expanded", !navPanel.classList.contains("collapsed"));
```

PR #623, opened 2019-07-15
Harp 5722 background data source

atomicsulfate:
There are 3 commits in this PR:

1. Move creation of background area geometry to a separate data source: BackgroundDataSource.
2. Fix a bug found in TileTreeTraverse.subTiles() where it returns invalid tiles for halfquad tile subdivision scheme.
3. Add tile intersection caching so that intersection tiles are reused for data sources with same tiling scheme. Otherwise, with the addition of BackgroundDataSource we would be calculating the same tile intersections at least twice.

### DIFF
--- a/@here/harp-geoutils/lib/tiling/SubTiles.ts
+++ b/@here/harp-geoutils/lib/tiling/SubTiles.ts
@@ -6,7 +6,7 @@
 
 import { TileKey } from "./TileKey";
 
-export class SubTiles {
+export class SubTiles implements Iterable<TileKey> {
     private m_tileKey: TileKey;
     private m_level: number;
     private m_count: number;
@@ -23,9 +23,6 @@ export class SubTiles {
         // tslint:enable:no-bitwise
     }
 
-    get length(): number {
-        return this.m_count;
-    }
     get level(): number {
         return this.m_level;
     }
@@ -33,8 +30,12 @@ export class SubTiles {
         return this.m_tileKey;
     }
 
+    [Symbol.iterator](): SubTiles.SubTileIterator {
+        return new SubTiles.SubTileIterator(this, this.m_count);
+    }
+
     iterator() {
-        return new SubTiles.Iterator(this);
+        return this[Symbol.iterator]();
     }
 
     skip(index: number): number {
@@ -50,13 +51,15 @@ export class SubTiles {
 }
 
 export namespace SubTiles {
-    export class Iterator {
+    export class SubTileIterator implements Iterator<TileKey> {
         private m_parent: SubTiles;
         private m_index: number;
+        private m_totalSubTileCount: number;
 
-        constructor(parent: SubTiles, index: number = 0) {
+        constructor(parent: SubTiles, totalSubTileCount: number, index: number = 0) {
             this.m_parent = parent;
             this.m_index = parent.skip(index);
+            this.m_totalSubTileCount = totalSubTileCount;
         }
 
         get value() {
@@ -73,7 +76,9 @@ export namespace SubTiles {
         }
 
         next() {
+            const current = { value: this.value, done: this.m_index >= this.m_totalSubTileCount };
             this.m_index = this.m_parent.skip(++this.m_index);
+            return current;
         }
     }
 }
--- a/@here/harp-mapview/lib/BackgroundDataSource.ts
+++ b/@here/harp-mapview/lib/BackgroundDataSource.ts
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { TileKey, TilingScheme, webMercatorTilingScheme } from "@here/harp-geoutils";
+import { DataSource } from "./DataSource";
+import { TileGeometryCreator } from "./geometry/TileGeometryCreator";
+import { Tile } from "./Tile";
+
+/**
+ * Provides background geometry for all tiles.
+ */
+export class BackgroundDataSource extends DataSource {
+    private static readonly DEFAULT_TILING_SCHEME = webMercatorTilingScheme;
+    private m_referenceDataSource: DataSource | undefined;
+    private m_tilingScheme: TilingScheme;
+
+    constructor() {
+        super("background", undefined, 1, 20);
+        this.cacheable = true;
+        this.m_referenceDataSource = undefined;
+        this.m_tilingScheme = BackgroundDataSource.DEFAULT_TILING_SCHEME;
+    }
+
+    updateTilingScheme() {
+        if (
+            this.m_referenceDataSource &&
+            this.mapView.isDataSourceEnabled(this.m_referenceDataSource)
+        ) {
+            return;
+        }
+
+        const newReferenceDataSource = this.mapView.dataSources.find(
+            ds => ds !== this && this.mapView.isDataSourceEnabled(ds)
+        );
+
+        const tilingSchemeChanged = this.setReferenceDataSource(newReferenceDataSource);
+
+        if (tilingSchemeChanged) {
+            this.mapView.clearTileCache(this.name);
+        }
+    }
+    getTilingScheme(): TilingScheme {
+        return this.m_tilingScheme;
+    }
+
+    getTile(tileKey: TileKey): Tile | undefined {
+        const tile = new Tile(this, tileKey);
+        tile.forceHasGeometry(true);
+        tile.removeDecodedTile(); // Skip geometry loading.
+        TileGeometryCreator.instance.addGroundPlane(tile);
+
+        return tile;
+    }
+
+    // Returns true if the tiling scheme changed, false otherwise.
+    private setReferenceDataSource(referenceDataSource: DataSource | undefined): boolean {
+        this.m_referenceDataSource = referenceDataSource;
+        let newTilingScheme = BackgroundDataSource.DEFAULT_TILING_SCHEME;
+
+        if (this.m_referenceDataSource !== undefined) {
+            newTilingScheme = this.m_referenceDataSource.getTilingScheme();
+            this.storageLevelOffset = this.m_referenceDataSource.storageLevelOffset;
+        }
+        const tilingSchemeChanged = this.m_tilingScheme !== newTilingScheme;
+
+        this.m_tilingScheme = newTilingScheme;
+
+        return tilingSchemeChanged;
+    }
+}
--- a/@here/harp-mapview/lib/FrustumIntersection.ts
+++ b/@here/harp-mapview/lib/FrustumIntersection.ts
@@ -1,0 +1,307 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { OrientedBox3 } from "@here/harp-geometry";
+import { MathUtils, Projection, ProjectionType, TileKey, TilingScheme } from "@here/harp-geoutils";
+import { assert } from "chai";
+import * as THREE from "three";
+import { CalculationStatus, ElevationRangeSource } from "./ElevationRangeSource";
+import { MapTileCuller } from "./MapTileCuller";
+import { MapViewUtils, TileOffsetUtils } from "./Utils";
+
+/**
+ * Represents a unique TileKey and the area it takes up on screen.
+ *
+ * Note, in certain tiling projections, it is possible to have an offset, which represents a tile
+ * which has fully wrapped around, hence this defaults to 0 to simplify usage for projections which
+ * don't require it.
+ */
+export class TileKeyEntry {
+    constructor(public tileKey: TileKey, public area: number, public offset: number = 0) {}
+}
+
+function getGeoBox(tilingScheme: TilingScheme, childTileKey: TileKey, offset: number) {
+    const geoBox = tilingScheme.getGeoBox(childTileKey);
+    const longitudeOffset = 360.0 * offset;
+    geoBox.northEast.longitude += longitudeOffset;
+    geoBox.southWest.longitude += longitudeOffset;
+    return geoBox;
+}
+
+namespace FrustumIntersection {
+    export interface Result {
+        /**
+         * Tiles intersected by the frustum. Keys are a combination of morton code and tile offset,
+         * see [[TileOffsetUtils.getKeyForTileKeyAndOffset]].
+         */
+        readonly tileKeyEntries: Map<number, TileKeyEntry>;
+        /**
+         * True if the intersection was calculated using precise elevation data, false if it's an
+         * approximation.
+         */
+        calculationFinal: boolean;
+    }
+}
+
+/**
+ * Computes the tiles intersected by the frustum defined by the current camera setup.
+ */
+export class FrustumIntersection {
+    private readonly m_frustum: THREE.Frustum = new THREE.Frustum();
+    // used to project global coordinates into camera local coordinates
+    private readonly m_viewProjectionMatrix = new THREE.Matrix4();
+    private readonly m_mapTileCuller: MapTileCuller;
+    private m_rootTileKeys: TileKeyEntry[] = [];
+    private readonly m_tileKeyEntries: Map<number, TileKeyEntry> = new Map();
+
+    constructor(
+        private readonly m_camera: THREE.PerspectiveCamera,
+        private readonly m_projection: Projection,
+        private readonly m_extendedFrustumCulling: boolean
+    ) {
+        this.m_mapTileCuller = new MapTileCuller(m_camera);
+    }
+
+    /**
+     * Updates the frustum to match the current camera setup.
+     */
+    updateFrustum() {
+        this.m_viewProjectionMatrix.multiplyMatrices(
+            this.m_camera.projectionMatrix,
+            this.m_camera.matrixWorldInverse
+        );
+
+        this.m_frustum.setFromMatrix(this.m_viewProjectionMatrix);
+
+        if (this.m_extendedFrustumCulling) {
+            this.m_mapTileCuller.setup();
+        }
+        this.computeRequiredInitialRootTileKeys(this.m_camera.position);
+    }
+
+    /**
+     * Computes the tiles intersected by the updated frustum, see [[updateFrustum]].
+     *
+     * @param tilingScheme The tiling scheme used to generate the tiles.
+     * @param maxTileLevel The maximum tile level that will be checked for intersections.
+     * @param elevationRangeSource Source of elevation range data if any.
+     * @returns The computation result, see [[FrustumIntersection.Result]].
+     */
+    compute(
+        tilingScheme: TilingScheme,
+        maxTileLevel: number,
+        elevationRangeSource: ElevationRangeSource | undefined
+    ): FrustumIntersection.Result {
+        this.m_tileKeyEntries.clear();
+        let calculationFinal = true;
+
+        for (const item of this.m_rootTileKeys) {
+            this.m_tileKeyEntries.set(
+                TileOffsetUtils.getKeyForTileKeyAndOffset(item.tileKey, item.offset),
+                new TileKeyEntry(item.tileKey, Infinity, item.offset)
+            );
+        }
+
+        const useElevationRangeSource: boolean =
+            elevationRangeSource !== undefined &&
+            elevationRangeSource.getTilingScheme() === tilingScheme;
+
+        const tileBounds = new THREE.Box3();
+        const workList = [...this.m_rootTileKeys];
+
+        while (workList.length > 0) {
+            const tileEntry = workList.pop();
+
+            if (tileEntry === undefined) {
+                continue;
+            }
+
+            const tileKey = tileEntry.tileKey;
+            const uniqueKey = TileOffsetUtils.getKeyForTileKeyAndOffset(tileKey, tileEntry.offset);
+            const cachedTileEntry = this.m_tileKeyEntries.get(uniqueKey);
+
+            assert(cachedTileEntry !== undefined);
+            assert(cachedTileEntry!.area > 0);
+
+            if (tileKey.level > maxTileLevel) {
+                continue;
+            }
+
+            tilingScheme.getSubTileKeys(tileKey).forEach(childTileKey => {
+                const offset = tileEntry.offset;
+                const tileKeyAndOffset = TileOffsetUtils.getKeyForTileKeyAndOffset(
+                    childTileKey,
+                    offset
+                );
+
+                assert(this.m_tileKeyEntries.get(tileKeyAndOffset) === undefined);
+
+                const geoBox = getGeoBox(tilingScheme, childTileKey, offset);
+
+                if (useElevationRangeSource) {
+                    const range = elevationRangeSource!.getElevationRange(childTileKey);
+                    geoBox.southWest.altitude = range.minElevation;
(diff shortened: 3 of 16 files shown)


@here/harp-geoutils/lib/tiling/SubTiles.ts
```diff
@@ -73,7 +78,9 @@ export namespace SubTiles {
         }
 
         next() {
+            const current = { value: this.value, done: this.m_index >= this.m_parent.length };
```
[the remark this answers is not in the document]
atomicsulfate: Yes, the skip() function sets the index to this.m_parent.length when next() is called after the last masked subtile was returned.
The test in SubTilesTest is checking that.

@here/harp-geoutils/lib/tiling/SubTiles.ts
```diff
@@ -23,6 +23,7 @@ export class SubTiles {
         // tslint:enable:no-bitwise
     }
 
+    // Returns the total number of sub tiles disregarding the mask.
```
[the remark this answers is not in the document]
atomicsulfate: Yes, even though i added a comment to clarify what that length is, it can be misused. I've removed the getter.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -168,6 +169,63 @@ export interface DataSourceTileList {
...
 }
 
+/**
+ * Properties of a frustum intersection cached in [[FrustumIntersectionCache]].
+ */
+interface FrustumIntersection {
+    /**
+     * Tiles intersected by the frustum.
+     */
+    tiles: Map<number, TileKeyEntry>;
```
zbigg: `tileKeyEntries`

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -831,107 +886,141 @@ export class VisibleTileSet {
...
     }
 
-    // Computes the visible tiles for each supplied datasource.
-    private getVisibleTilesForDataSources(
-        worldCenter: THREE.Vector3,
+    private getVisibleTilesForDataSource(
+        dataSource: DataSource,
         zoomLevel: number,
-        dataSources: DataSource[],
+        rootTiles: TileKeyEntry[],
```
zbigg: `rootTileKeyEntries`

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -831,107 +886,141 @@ export class VisibleTileSet {
...
-            const displayZoomLevel = dataSource.getDisplayZoomLevel(zoomLevel);
-
-            const tilingScheme = dataSource.getTilingScheme();
-            const useElevationRangeSource: boolean =
-                elevationRangeSource !== undefined &&
-                elevationRangeSource.getTilingScheme() === tilingScheme;
-
-            const tileBounds = new THREE.Box3();
-            const workList: TileKeyEntry[] = this.getRequiredInitialRootTileKeys(worldCenter);
+    ): { visibleTiles: TileKeyEntry[]; bboxesFinal: boolean } {
```
zbigg: My general comment (with one about caches) is that this function should be idempotent and just return brand new `FrustumIntersection` basing on `tilingScheme`, `rootTiles`, `elevationRangeSource` and `maxZoomLevel`.

So, it should be named `getFrustumIntersection`.

This would make it perfectly compatible with `cache` concept and would make `FrustumInterfsectionCache` class almost redundant as it would be just plain `Map<TilingScheme, FrustumIntersection>` local to `getVisibleTilesForDataSources`.

Note on `maxZoomLevel` in next comment.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -831,107 +886,141 @@ export class VisibleTileSet {
...
+            const uniqueKey = TileOffsetUtils.getKeyForTileKeyAndOffset(tileKey, tileEntry.offset);
+            const cachedTileEntry = frustumIntersection.tiles.get(uniqueKey);
 
-            const tileFrustumIntersectionCache = this.createIntersectionCache(workList);
+            assert(cachedTileEntry !== undefined);
+            assert(cachedTileEntry!.area > 0);
 
-            while (workList.length > 0) {
-                const tileEntry = workList.pop();
+            if (tileKey.level > displayZoomLevel) {
```
zbigg: This is broken in very obscure way.
`displayZoomLevel` is actually bounded by `maxZoomLevel` of first datasource that happens be processes by this method.

If for example, this first  datasource allows only data up to zoomLevel 15, all other datasources (that have data up to, 20) will also draw tiles up to 15.

So, one we should calculate like this:
```
maxDisplayZoomLevel = enabledDataSources.map((ds) => ds.getDisplayZoomLevel(zoomLevel)).reduce(max, 0);
```
and it should be done in `getVisibleTilesForDataSources` where we know all the DSs to be processed.

@here/harp-mapview/lib/BackgroundDataSource.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    attach(mapView: MapView): void {
+        super.attach(mapView);
+        mapView.addEventListener(MapViewEventNames.ThemeLoaded, () => {
+            this.m_themeLoaded = true;
+        });
+    }
+
+    getTilingScheme(): TilingScheme {
+        return webMercatorTilingScheme;
```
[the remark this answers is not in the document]
zbigg: Proposal to resolve this:
```
getTilingScheme() {
     const other =    this.mapView.enabledDataSources.find(ds => ds !== this);
     return other && other.tilingScheme || webMercator;
}
```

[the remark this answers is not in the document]
atomicsulfate: makes sense, will do.

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: done.

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: done.

@here/harp-mapview/lib/BackgroundDataSource.ts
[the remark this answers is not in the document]
zbigg: @ninok , @atomicsulfate Agree, that care should be taken to maintain runtime stability. My proposal was just pseudo-code (should've mention this).

We should memoize if detected tiling scheme changed since previous frame and if, then `this.mapView.clearTileCache(this)`.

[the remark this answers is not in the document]
atomicsulfate: @zbig @ninok The frustum intersection cache is no reused between frames, it's just used to avoid recalculating the intersection in the same frame multiple times.

Anyway, I'm working on a refactor now that will remove this cache since it's actually not needed. It's enough to order the data sources by tiling scheme, calculate the intersection, and reuse it.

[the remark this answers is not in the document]
zbigg: @atomicsulfate 
> The frustum intersection cache is no reused between frames

I already know this but we discuss other cache here.
If `BacgroundDataSource.getTilingScheme` would be implemented in naive way like in my proposition, it would be possible for tiling scheme to change from frame to frame, but tiles in tile cache are indexed by tiles morton code, so this cache would be invalid.

That's why "dynamic" tiling scheme in background data source requires that we detect tiling scheme change and invalidate tile cache.

> It's enough to order the data sources by tiling scheme, calculate the intersection, and reuse it.

 Good point +1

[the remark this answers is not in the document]
atomicsulfate: done

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -831,107 +886,141 @@ export class VisibleTileSet {
...
+    } {
+        const tiles = [];
+        let allBoundingBoxesFinal: boolean = true;
+        const rootTileKeys: TileKeyEntry[] = this.getRequiredInitialRootTileKeys(worldCenter);
+        const frustumIntersectionCache = new FrustumIntersectionCache(rootTileKeys);
+        let frustumIntersectionCount = 0;
 
-                    if (subTileArea > 0) {
-                        workList.push(new TileKeyEntry(childTileKey, subTileArea, offset));
+        for (const dataSource of dataSources) {
```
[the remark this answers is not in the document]
atomicsulfate: done

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: done.

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: done.

atomicsulfate:

@here/harp-mapview/test/MapViewTest.ts
```diff
@@ -306,4 +281,46 @@ describe("MapView", function() {
...
+                    callback(0);
+                };
+                if (customMapView) {
+                    callbackWrapper();
+                } else {
+                    setupFrameCallback = callbackWrapper;
+                }
+                return 0;
+            };
+            global.cancelAnimationFrame = () => {};
```
zbigg: remove this in in `afterEach`

@here/harp-mapview/test/MapViewTest.ts
```diff
@@ -306,4 +281,46 @@ describe("MapView", function() {
...
+
+    it("updates background tiling scheme", async function() {
+        this.timeout(500);
+        let setupFrameCallback: FrameRequestCallback = () => {};
+
+        let customMapView: MapView;
+
+        if (inNodeContext) {
+            global.requestAnimationFrame = (callback: FrameRequestCallback) => {
+                const callbackWrapper = () => {
```
zbigg: I would wrap all of this in `setTimeout(..., 0)` or `process.nextTick` so it is really async as real `requestAnimationFrame. is.

@here/harp-mapview/test/MapViewTest.ts
```diff
@@ -306,4 +281,46 @@ describe("MapView", function() {
             }
         }
     });
+
+    it("updates background tiling scheme", async function() {
+        this.timeout(500);
+        let setupFrameCallback: FrameRequestCallback = () => {};
+
+        let customMapView: MapView;
```
zbigg: Please reuse suite `mapView` variable, as it ensures `mapView` is destroyed regardless of test outcome.

Test case in this state will leak live `MapView` instance which can do some async stuff after the test case and fail random test cases later.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -831,107 +845,161 @@ export class VisibleTileSet {
...
-                        }
-                    } else {
-                        this.options.projection.projectBox(geoBox, tileBounds);
-                        subTileArea = this.computeSubTileArea(tileBounds);
-                    }
+        if (dataSources.length === 0) {
+            return { tiles, allBoundingBoxesFinal, frustumIntersectionCount };
+        }
+        // Sort data sources with same tiling scheme into the same bucket.
+        dataSources.sort((lhs, rhs) => {
```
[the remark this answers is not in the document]
atomicsulfate: Forgot that code there from a previous implementation. done.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -831,107 +845,161 @@ export class VisibleTileSet {
...
+            return lhs.getTilingScheme().id - rhs.getTilingScheme().id;
+        });
 
-                    tileFrustumIntersectionCache.set(tileKeyAndOffset, subTileArea);
+        const dataSourceBuckets = new Map<number, DataSource[]>();
+        dataSources.forEach(dataSource => {
+            const tilingSchemeId = dataSource.getTilingScheme().id;
```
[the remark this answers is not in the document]
atomicsulfate: done.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -831,107 +845,161 @@ export class VisibleTileSet {
...
+    // Computes the visible tiles for each supplied datasource.
+    private getVisibleTilesForDataSources(
+        worldCenter: THREE.Vector3,
+        zoomLevel: number,
+        dataSources: DataSource[],
+        elevationRangeSource: ElevationRangeSource | undefined
+    ): {
+        tiles: Array<{ dataSource: DataSource; visibleTiles: TileKeyEntry[] }>;
+        allBoundingBoxesFinal: boolean;
+        frustumIntersectionCount: number;
```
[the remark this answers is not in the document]
atomicsulfate: yes, true, it's only for testing purposes.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1521,6 +1536,11 @@ export class MapView extends THREE.EventDispatcher {
                 });
 
                 this.update();
+
+                if (addBackgroundDataSource) {
```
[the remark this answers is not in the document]
atomicsulfate: not sure, what would be the point of that? i can't think of any use case.

@here/harp-geoutils/lib/tiling/TilingScheme.ts
```diff
@@ -80,4 +84,13 @@ export class TilingScheme {
     getWorldBox(tileKey: TileKey, result?: Box3Like): Box3Like {
         return this.boundingBoxGenerator.getWorldBox(tileKey, result);
     }
+
+    /**
+     * Returns the id of the tiling scheme instance, which can be used for tile scheme comparisons.
+     *
+     * @returns The tiling scheme instance id.
+     */
+    get id(): number {
```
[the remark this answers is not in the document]
atomicsulfate: done.

@here/harp-mapview/test/MapViewTest.ts
atomicsulfate: done.

@here/harp-mapview/test/MapViewTest.ts
atomicsulfate: done.

@here/harp-mapview/test/MapViewTest.ts
atomicsulfate: done.

@here/harp-mapview/test/VisibleTileSetTest.ts
```diff
@@ -193,4 +204,31 @@ describe("VisibleTileSet", function() {
...
+            expect(result.tileList[0].visibleTiles).to.eql(result.tileList[1].visibleTiles);
+        }
+
+        // If the second data source has a different tile scheme, the intersection must be
+        // calculated twice, once for each tiling scheme.
+        sinon.replace(secondDataSource, "getTilingScheme", sinon.stub().returns(hereTilingScheme));
+
+        {
+            const result = updateRenderList(zoomLevel, storageLevel);
+            assert.equal(result.intersectionCount, 2);
```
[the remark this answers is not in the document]
atomicsulfate: I agree it's not a nice approach, but the main argument against it is that it exposes a useless piece of information in the return from VisibleTileSet. Overhead is probably not even measurable (we are talking about incrementing a counter).

The approach you suggest is too fragile, because it depends on implementation details (the number of calls to getSubTileKeys) to deduce what's been done (how many frustum intersections). It can be easily broken by implementation changes because the relation between there's no strong relation between those two pieces of information. If someone needs to call getSubTileKeys once more for any reason, they'll get a broken unit test and will have to figure out how to reimplement the test.

The only better way i can find of testing this is to actually refactor the whole frustum intersection calculation into a separate class and expose there the needed information. But I didn't think it was worth the effort.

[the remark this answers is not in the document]
zbigg: Fragility depends on what you're comparing to.
If you compare to hardcoded values, they i agree it may be very fragile.
If you compare to previous "runs", then it may be very stable.

Imagine scenario

```javascript
updateRenderList(..., [onlyOmvDataSource])
const originalCallsCount = getSubTileKeysStub.count();
getSubTileKeysStub.reset();
updateRenderList(..., [onlyOmvDataSource, backgrounDataSource])
assert.equals(getSubTileKeysStub.count(), originalCallsCount);
```
And next one:

```javascript
getSubTileKeysStub.reset();
updateRenderList(..., [onlyOmvDataSource, backgrounDataSource, someOtherDataSourceWithOtherTilingScheme])
assert.greaterThan(getSubTileKeysStub.count(), originalCallsCount);
```
This should give us pretty good insight if w reuse frustum calculations basing on the only observable behaviour - calls to `TilingScheme` interface.

I don't push though, i consider this as hint.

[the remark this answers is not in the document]
zbigg: Actually, i would count most expensive calls - my guess `getGeoBox & getWorldBox` whose perform lot of or math internally and probably have biggest impact on real performance.

@here/harp-mapview/lib/MapView.ts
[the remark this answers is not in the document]
zbigg: Hmm, i'm not very sure, i'm just wondering ...
If background is independent, then why wait at all ?

Other thing - probably very uncommon scenario, is a map that at this point of time,  has some `mapAnchors` defined (i.e custom three js objects), so they should be drawn on some background.

@here/harp-mapview/test/VisibleTileSetTest.ts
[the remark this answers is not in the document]
atomicsulfate: getSubTileKeys (as well as getGeoBox or getWorldBox) may be called anywhere else outside of getVisibleTilesForDataSources(). For instance, fillMissingTilesFromCache() is calling getSubTileKeys a non-obvious number of times and that gets executed in the unit test.
This forces to relax the assertions in the unit tests, as you show in your code snippet where you do
`assert.greaterThan(getSubTileKeysStub.count(), originalCallsCount);`

when you should be able to say:

`assert.equals(getSubTileKeysStub.count(), originalCallsCount+1);`

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -831,107 +845,156 @@ export class VisibleTileSet {
...
+                    workList.push(subTileEntry);
+                }
+            });
+        }
+        return frustumIntersection;
+    }
 
-                    let subTileArea = 0;
+    // Computes the visible tiles for each supplied datasource.
+    private getVisibleTilesForDataSources(
```
nzjony: Should be renamed, the visible tile keys are returned, and not the visible tiles.

@here/harp-mapview/test/VisibleTileSetTest.ts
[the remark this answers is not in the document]
nzjony: Having also worked on this method, I wonder if it makes sense to split the VisibleTileSet into two classes, (or maybe one class and one free function), the computation of the TileKey's visible for the given DataSources doesn't need to be in the same class as the caching functionality etc. I think this discussion is showing that the current design is someone lacking.

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: done

@here/harp-mapview/lib/MapView.ts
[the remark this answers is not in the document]
atomicsulfate: ok, I moved the BackgroundDataSource construction to MapView's constructor.

@here/harp-mapview/test/VisibleTileSetTest.ts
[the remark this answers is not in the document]
atomicsulfate: @nzjony @zbigg I think it makes sense to move computeFrustumIntersection() method. I can move it the the FrustumIntersection interface and make a full blown class. 
That way we can test directly frustum intersection code without having to go through the VIsibleTileSet class. Is that ok?

atomicsulfate:

@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -0,0 +1,306 @@
...
+    geoBox.southWest.longitude += longitudeOffset;
+    return geoBox;
+}
+
+namespace FrustumIntersection {
+    export interface Result {
+        /**
+         * Tiles intersected by the frustum.
+         */
+        readonly tileKeyEntries: Map<number, TileKeyEntry>;
```
nzjony: It isn't apparent what the key is, I guess morton code?

@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -0,0 +1,306 @@
...
+        if (this.m_extendedFrustumCulling) {
+            this.m_mapTileCuller.setup();
+        }
+        this.computeRequiredInitialRootTileKeys(this.m_camera.position);
+    }
+
+    /**
+     * Computes the tiles intersected by the updated frustum, see [[updateFrustum]].
+     *
+     * @param tilingScheme The camera storage level, see [[MapView.storageLevel]].
```
nzjony: Is this comment correct?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -831,143 +671,64 @@ export class VisibleTileSet {
...
+        });
 
-                    if (subTileArea > 0) {
-                        workList.push(new TileKeyEntry(childTileKey, subTileArea, offset));
+        this.m_frustumIntersection.updateFrustum();
+
+        // For each bucket of data sources with same tiling scheme, calculate frustum intersection
+        // once using the maximum display level.
+        for (const [tilingScheme, bucket] of dataSourceBuckets) {
+            const maxDisplayLevel = Math.max(
```
nzjony: This looks like we could generate tiles which the DataSource has no data for, say for example you have in the map call below, 10 & 12, then maxDisplayLevel would be 12, and it would generate tileKeyEntries at 12, however the datasource which has max zoom level 10 would not show any data(?)

I guess then you need to differentiate the DataSources based on the max zoom levels? Or did I miss something?

@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -0,0 +1,306 @@
...
+     * @param maxTileLevel The maximum tile level that will be checked for intersections.
+     * @param elevationRangeSource Source of elevation range data if any.
+     * @returns The computation result, see [[FrustumIntersection.Result]].
+     */
+    compute(
```
nzjony: Is there a need to have update and compute separate methods? In what use case would I call updateFrustum and not compute (and vice versa).

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: We check this in the next loop where we call dataSource.shouldRender(), which returns false if the tile belongs to a level higher than the data source maximum level.

@here/harp-mapview/lib/FrustumIntersection.ts
atomicsulfate: I'll add a comment.

@here/harp-mapview/lib/FrustumIntersection.ts
atomicsulfate: clearly not. I'll update.

@here/harp-mapview/lib/FrustumIntersection.ts
atomicsulfate: updateFrustum needs to be done once per frame. computation is done as many times as different tile schemes.

PR #624, opened 2019-07-15
HARP-6492: Move unit tests setup code to 'before' functions.

atomicsulfate:


@here/harp-geoutils/test/ProjectionTest.ts
```diff
@@ -148,7 +148,7 @@ describe("Mercator", function() {
         assert.equal(coords.altitude, unprojected.altitude);
     });
 
-    (function() {
+    before(function() {
```
zbigg: Not sure if that one works though.

Note, that deep in this anon-function, we define test-cases by it.
I am worried (but not sure), that this one should be called in `describe` phase, not in `before`.

atomicsulfate: will fix that.

PR #640, opened 2019-07-22
HARP-6503: Fix MapView.dispose

zbigg:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/ConcurrentWorkerSet.ts
+++ b/@here/harp-mapview/lib/ConcurrentWorkerSet.ts
@@ -31,7 +31,7 @@ interface ReadyPromise {
 
 interface RequestEntry {
     promise: Promise<any>;
-    resolver: (error?: object, response?: object) => void;
+    resolver: (error?: Error, response?: object) => void;
 }
 
 export interface ConcurrentWorkerSetOptions {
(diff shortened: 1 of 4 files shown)


@here/harp-mapview/lib/ConcurrentWorkerSet.ts
```diff
@@ -216,9 +229,10 @@ export class ConcurrentWorkerSet {
 
         // respond with all pending request
         this.m_requests.forEach(entry => {
-            entry.resolver(new Error("worker destroyed"));
+            entry.resolver("worker destroyed");
```
sasha240100: It looks like there is inconsistency with some other parts of the code that it use resolver() call with `new Error()` object. (See lines 534, 470)

zbigg: Good catch. Fixed

@here/harp-mapview/lib/ConcurrentWorkerSet.ts
```diff
@@ -453,7 +467,11 @@ export class ConcurrentWorkerSet {
             } else {
                 logger.error(`[${this.m_options.scriptUrl}]: onWorkerMessage: invalid workerId`);
             }
-            entry.resolver(response.error, response.response);
+            if (response.error !== undefined) {
+                entry.resolver(new Error(response.error.toString()));
```
sasha240100: I think `new Error(response.error.toString())` is not needed anymore as you have **instanceof check**

zbigg: It's needed because this `response.error` is `any` as defined in `WorkerServiceDefs.ts`. This decoder boundary, so we can't have "real" `Error` instances, only serializable forms which as for now are just results of `toString` applied on `Error` instances in workers.

PR #644, opened 2019-07-22
HARP-5404: Add builtin operators

robertoraggi:


@here/harp-datasource-protocol/lib/ExprEvaluator.ts
```diff
@@ -69,64 +77,406 @@ export class ExprEvaluator implements ExprVisitor<Value, Env> {
...
-            case "||":
-                return value || this.evaluate(expr.right, env);
-            case "&&":
-                return value && this.evaluate(expr.right, env);
         } // switch
-        throw new Error(`invalid logical op '${expr.op}'`);
     }
 }
+
+ExprEvaluator.registerBuiltin("length", {
```
okanishchev: Maybe we should store all this builtin tokens in some enum, and address them like ```ExprBuiltin.Length``` for example?

robertoraggi: The code was reorganized a bit, now separate files expose the builtin operators and a typed string enumeration (e.g. `CastOperators.ts` exports the `CastOperators` builtins and the `CastOperatorNames` enumeration).

PR #647, opened 2019-07-23
MINOR: Improve getting started documentation and add tutorial to website

dbabbs:
- Minor code improvements to getting started code in `README.md`
- Window resize and starting location added to `GettingStartedGuide.md`
- Add tutorial to website navigation

### DIFF
--- a/docs/GettingStartedGuide.md
+++ b/docs/GettingStartedGuide.md
@@ -14,7 +14,7 @@ Add `three.js` and `harp.gl` to your html and create a canvas with an id `map`:
 <html>
    <head>
       <style>
-         body, html { border: 0; margin: 0; padding: 0 }
+         body, html { border: 0; margin: 0; padding: 0; }
          #map { height: 100vh; width: 100vw; }
       </style>
       <script src="https://unpkg.com/three/build/three.min.js"></script>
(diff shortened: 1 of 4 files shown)


docs/GettingStartedGuide.md
```diff
@@ -29,15 +29,23 @@ Add `three.js` and `harp.gl` to your html and create a canvas with an id `map`:
 Initialize the map:
 ```javascript
 const map = new harp.MapView({
-   canvas: document.getElementById('map'),
+   canvas: document.getElementById("map"),
    theme: "https://unpkg.com/@here/harp-map-theme@latest/resources/berlin_tilezen_night_reduced.json",
 });
-const mapControls = new harp.MapControls(map);
+const controls = new harp.MapControls(map);
+
```
robertoraggi: Please use typescript and yo in the getting started guide, because that should be the preferred way to use Harp.gl

dbabbs: There are two sections in the getting started guide. Simple api and yo generator. Yo generator is the second section on this page.

robertoraggi: Most of the code snippets using the `harp.` prefix will not work with harp.gl packages, e.g.

```
// MapView and MapControls are in two different packages, so to the make the
// code snippet work with the npm packages, users have to write something like the following code:

// import {MapView} from "@here/harp-mapview";
// import {MapControls} from "@here/harp-map-controls";
// const harp = {MapView, MapControls};  // :(

const map = new harp.MapView(...);
const controls = new harp.MapControls(map);
```

I would like to avoid the situation where snippets in our documentation differ only for the `harp.` prefix, it will create work for us to maintain two sets of snippets and it can potentially confuse the users.

We could try to reduce the duplication by having specific sections for `harp.js` and npm, e.g.
```
// for bundled 'harp.js'
const {MapView, MapControls} = harp;
```

```
// for packaged harp.gl
import {MapView } from "@here/harp-mapview";
import {MapControls } from "@here/harp-map-controls";
```

common code describing how to set up a map and so on...
```
// works with both the bundled `harp.js` and the npm packaged harp.gl (with noth javascript and typescript)
const map = new MapView(...);
const controls = new MapControls(map);
```

We should also consider to publish the bundled `harp.js` in jsm, then we can use the browser loader, and the "bits" specific for harp.js in jsm for the browser will be:
```
import {MapView, MapControls } from "./harp.js"; // harp.js in jsm

const map = new MapView(...);
const controls = new MapControls(map);
```

dbabbs: I agree that documentation should always prioritize the modules. However, for examples and guides (just like this getting started guide), I don't think it is harmful having multiple sections depending on how you'd like to consume the API. Sure, it may slightly be more work for us to maintain, but offering a quality developer experience that caters to multiple preferences is definitely worth it.

I do like the javascript module approach, however. I haven't look recently so I'm not fully aware if  javascript modules are supported out of the box (no webpack) with all the major browsers.

For now, I am going to merge since this commit is simply fixing some characters and adding the tutorial to the website. The getting started guide with both simple api and yo generator has already been live for a few weeks now.

PR #653, opened 2019-07-25
HARP-6098: Add replacement geometries for the poles

okanishchev:
The new kind of datasource is used to replace missing pole geometries in globe projection. It uses tiling scheme with regular quadtree subdivision and the transverse Mercator projection (https://en.wikipedia.org/wiki/Transverse_Mercator_projection) to effectively use the pole space. This projection is just regular Mercator projection, but with the projection cylinder is rotated 90 degrees, so it also has "poles" cut out but at the (0, -90) and (0, +90). It also has the specificity of non-linear bounding boxes projection, so the culling is not super effective.

The tiles geometries are regular quads, cutted by the pole line if the intersection occurs. They can be styled by the theme as the datasource has it's own `styleSetName` (`"polar"` by default) and the objects get two special **kind**s: `"north_pole"` and `"south_pole"`.

The datasource has a usual `storageLevelOffset` property that controls level of displayed tiles for this datasource, and by default is reduced to -2 to show as less as possible tiles (currently a maximum of 6 pole tiles can be visible). Also it has a special property `geometryLevelOffset` that controls cutting at the pole line and needs to be set equal to regular tile datasource's (omv or satellite etc) `storageLevelOffset` property to match tile bounds.

The datasource is heavy using `shouldSubdivide` to avoid frustum checks on the tiles that are completely out of pole radius. It also skips geometry creation if no styles for the poles are set.

Also this datasource is added by MapView and by default enabled when spherical projection set. There are also controls can be passed to MapView constructor to configure to polar datasource (`enablePolarDataSource`, `polarStyleSetName` and `polarGeometryLevelOffset`).


@here/harp-mapview/lib/PolarTileDataSource.ts
```diff
@@ -0,0 +1,167 @@
...
+} from "@here/harp-geoutils";
+
+import { DataSource } from "./DataSource";
+import { createMaterial } from "./DecodedTileHelpers";
+import { Tile } from "./Tile";
+
+/**
+ * [[DataSource]] providing geometry for poles
+ */
+export class PolarTileDataSource extends DataSource {
```
ninok: I like the idea of encapsulating the pole caps into a dedicated datasource but I think we cannot do it because:
1. We will have a lot of draw calls b/c there will be a lot of tiny tiles at the pole (compared to one draw call if you generate dedicated geometry)
2. We force the VisibleTileSet to traverse the tile tree twice, once for webMercatorTilingScheme and once for the polarTilingScheme (btw. we also call it HereTilingScheme b/c this is what we used in the past for our proprietary map formats).
3. The geometry will not match the webMercator tiles

okanishchev: Thanks for reivew. Yes, there are issues to this solution, it's just the best one I've found so far.
 1. This is kind of a tradeoff because from one side there are cpu and memory complexity with dedicated geometry (tons of subdivisions at higher levels - 2^20 actually), and gpu complexity from the other with tons of quads. But I think this solution has a bit more scalability.
 2. Yep, this is a bad one. But the amount of tiles can be reduced significantly if we could disable subdivision by latitude. I have tried to develop such subdivision scheme, but the SubTiles iterator seems to be not able to create tiles with no subdivision (e.g. with getSubdivisionY() => 1) properly.
 3. Yes, it will not match, and there are no tiling scheme that can both address latitudes >85 and math mercator tiles. But as long as webMercator tiling has same horizontal subdivision scheme, they will match at least horizontally.

okanishchev: Just checked, zooming in directly to poles makes things really bad.

ninok: If we encounter that the triangles for a dedicated geometry are to many we could do some smart culling that renders only sub-ranges of that geometry (i.e. a segment of the circle instead of the whole circle)
I think we should limit the tile level close to the poles somehow. Already rendering the "real" data at level 20 close to the pole will kill us with to many draw calls
Since the raw data is at level 14 the deeper levels are just clipped duplicates of the same data we could make the max storage level that is used depending on the latitude.

Astrak: Note how google maps does not give a hang about this problem :sob: https://www.google.fr/maps/@-85,149.3853301,5.69z

ninok: @Astrak interesting. And when you zoom in closer than 10 you cannot see the pole anymore. They force the camera to stay inside the valid webmercator range.

@here/harp-geoutils/lib/projection/TransverseMercatorProjection.ts
```diff
@@ -0,0 +1,309 @@
...
+import { EarthConstants } from "./EarthConstants";
+import { Projection, ProjectionType } from "./Projection";
+
+/**
+ *
+ * https://en.wikipedia.org/wiki/Transverse_Mercator_projection
+ * http://mathworld.wolfram.com/MercatorProjection.html
+ *
+ */
+export class TransverseMercatorProjection extends Projection {
```
ninok: Please add tests for this class

@here/harp-mapview/lib/BackgroundDataSource.ts
```diff
@@ -55,19 +52,23 @@ export class BackgroundDataSource extends DataSource {
         return tile;
     }
 
-    // Returns true if the tiling scheme changed, false otherwise.
-    private setReferenceDataSource(referenceDataSource: DataSource | undefined): boolean {
-        this.m_referenceDataSource = referenceDataSource;
-        let newTilingScheme = BackgroundDataSource.DEFAULT_TILING_SCHEME;
+    private setReferenceDataSource(referenceDataSource: DataSource | undefined): void {
```
ninok: This complexity is actually not needed.
Just hardcode webMercatorTilingScheme for the background data source.
If it is really needed that the background data source needs to have a different tiling scheme we could make it explicit in the MapView options

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -37,6 +37,12 @@ export abstract class DataSource extends THREE.EventDispatcher {
      */
     useGeometryLoader: boolean = false;
 
+    /**
+     * Set to `true` for the [[BackgroundDataSource]] can use this `DataSource`
+     * as reference.
+     */
+    useForBackground: boolean = false;
```
ninok: Same as above.

@here/harp-mapview/lib/PolarTileDataSource.ts
```diff
@@ -0,0 +1,317 @@
...
+
+        const { north, south } = this.m_tilingScheme.getGeoBox(tileKey);
+
+        const shouldRender = north > this.m_maxLatitude || south < -this.m_maxLatitude;
+
+        return shouldRender;
+    }
+
+    shouldSubdivide(zoomLevel: number, tileKey: TileKey): boolean {
+        if (zoomLevel < tileKey.level) {
```
ninok: <= ?

@here/harp-geoutils/lib/projection/MercatorProjection.ts
```diff
@@ -14,7 +14,7 @@ import { Vector3Like } from "../math/Vector3Like";
 import { EarthConstants } from "./EarthConstants";
 import { Projection, ProjectionType } from "./Projection";
 
-class MercatorProjection extends Projection {
+export class MercatorProjection extends Projection {
```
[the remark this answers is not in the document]
okanishchev: Because it contains static member MAXIMUM_LATITUDE, needed to define pole radius

@here/harp-mapview/lib/PolarTileDataSource.ts
okanishchev: nope, if the datasource has not shouldRender redefined, the tile will render at 
```zoomLevel === tileKey.level```.
So, why check sub-tiles if parent tile will be renderd already?

File: @here/harp-mapview/lib/BackgroundDataSource.ts (hunk 3 above)
Comment (okanishchev) [reply @here/harp-mapview/lib/BackgroundDataSource.ts]: Yep, we had discussion with Andres about that. Will do as a quick fix, but for future I think the whole BackgroundDataSource should be removed and it's functionality integrated into OmvDataSource or TileDataSource.

File: @here/harp-geoutils/lib/projection/TransverseMercatorProjection.ts (hunk 2 above)
Comment (okanishchev) [reply @here/harp-geoutils/lib/projection/TransverseMercatorProjection.ts]: Good point!

File: @here/harp-mapview/lib/PolarTileDataSource.ts (hunk 5 above)
Comment (ninok) [reply @here/harp-mapview/lib/PolarTileDataSource.ts]: that's why shouldSubdive should also return false when `zoomLevel === tileKey.level`

Comment (okanishchev) [reply @here/harp-mapview/lib/PolarTileDataSource.ts]: Wow, you're right. I was thinking about initial method in DataSource where it does return true, and here it returns false. Good catch, thanks!

**Comment by okanishchev** [comment]
The tests are coming...

File: @here/harp-geoutils/lib/projection/TransverseMercatorProjection.ts [hunk 7]
```diff
@@ -0,0 +1,348 @@
...
```
ninok: No need to export this class

@here/harp-geoutils/lib/projection/MercatorProjection.ts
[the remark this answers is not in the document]
ninok: I would rather move the MAXIMUM_LATITUDE out of this class and only export the constant

@here/harp-examples/src/getting-started_globe-projection.ts
```diff
@@ -53,7 +59,12 @@ export namespace GlobeExample {
             copyrightInfo: copyrights
         });
 
+        const polarDataSource = new PolarTileDataSource({
```
ninok: I think it would be better to handle this internally and not put the burden on the user to add this data source

okanishchev: Will think how it can be done automatically.

@here/harp-geoutils/lib/projection/TransverseMercatorProjection.ts
okanishchev: Ok

okanishchev: Actually, how it would be best to do with the case when i need some method special to transverse Mercator?

https://github.com/heremaps/harp.gl/pull/653/files#diff-1b0ff12045051a049d60bf7d1c58e4a3R186

Here i cast to TransverseMercatorProjection explicitly just to use this method. If i don't export that class, i will need to cast to any.

ninok: as far as I can see you are using alignLongitude but this method could actually be a free function in this file or you could also move it to the PolarTileDataSource

okanishchev:
Well, I have some numbers. On my i7-8850H @ 2.6GHz:
```
      tiles  gettile avgframe  cull  cull/a  collect collect/a total frame
       4.57    0.07    0.3     0.51    1.06    0.35     3.77   0.86  27.21
       5.28    0.05    0.25    0.66    2.27    0.28     9.53   0.94  32.54
edge   5.34    0.05    0.28    0.68    3.53    0.32    15.8    1.0   31.86
       5.57    0.05    0.28    0.79    4.94    0.32    22.55   1.11  35.13
       6.27    0.06    0.35    0.74    6.25    0.39    29.36   1.13  33.2
--------------------------------------------------------------------------
       6.61    0.05    0.34    0.67    7.05    0.38    34.17   1.05  11.9
       7.49    0.06    0.46    0.77    7.9     0.49    38.94   1.26   9.84
pole   7.09    0.06    0.44    0.75    8.72    0.46    43.8    1.21   9.38
       6.62    0.04    0.28    0.67    9.48    0.32    48.96   0.99   9.74
       7.53    0.04    0.34    0.81   10.38    0.37    54.93   1.18  11.11
--------------------------------------------------------------------------
       0.13    0.07    0.01    0.17   11.26    0.01    67.67   0.18  20.5
       0.22    0.08    0.02    0.18   11.69    0.03    74.44   0.21  18.46
world  0.4     0.08    0.03    0.21   12.09    0.05    81.76   0.26  19.41
       1.54    0.11    0.16    0.22   12.48    0.18    90.11   0.4   23.21
       0.56    0.05    0.03    0.16   12.82    0.04    97.91   0.2   15.77
```
Each line is average measurement of 1000 frames. All the numbers is in milliseconds, except `tiles` - in pieces. Benchmarking is done by zooming in and out at different regions: 
 * `edge` - pole edge to get cutting;
 * `pole` - pole itself w/o cuts;
 * `world` - random location far away from pole;

Fields:
 * `tiles` - number of `PolarTileDataSource.getTile` called per frame;
 * `gettile` - average `PolarTileDataSource.getTile` execution time;
 * `avgframe` - average of all `PolarTileDataSource.getTile` calls execution time per frame;
 * `cull` - all the code executed for `PolarTileDataSource` in `getVisibleTileKeysForDataSources`;
 * `cull/a` - all the `VisibleTileSet.getVisibleTileKeysForDataSources` execution time;
 * `collect` - all the code executed  for `PolarTileDataSource` in `VisibleTileSet.updateRenderList`;
 * `collect/a` - all the code executed for all datasources in `VisibleTileSet.updateRenderList`;
 * `total` - combined time spent on `PolarTileDataSource` in `VisibleTileSet`;
 * `frame` - `MapView.render` execution time;

Note. `collect` and `collect/a` doesn't include the `cull` and `cull/a`, but includes `avgframe`

@here/harp-examples/src/getting-started_globe-projection.ts
```diff
@@ -18,6 +18,7 @@ export namespace GlobeExample {
         const mapView = new MapView({
             canvas,
             theme: "resources/berlin_tilezen_base_globe.json",
+            maxVisibleDataSourceTiles: 360,
```
ninok: Please remove. We should not hide the fact that we are showing to many tiles close to the poles. There is a ticket for it as well and we should address it properly.

okanishchev: Done. Thanks.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -546,6 +552,22 @@ export interface MapViewOptions {
      */
     backgroundTilingScheme?: TilingScheme;
 
+    /**
+     * Should be the [[PolarTileDataSource]] used on spherical projection.
```
ninok: What is the default?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -808,6 +832,10 @@ export class MapView extends THREE.EventDispatcher {
                 options.quadTreeSearchDistanceDown;
         }
 
+        if (options.enablePolarDataSource !== undefined) {
+            this.m_enablePolarDataSource = options.enablePolarDataSource;
```
ninok: I think if it is not explicitly turned of I would enable it by default if projection is spherical

@here/harp-mapview/lib/MapView.ts
```diff
@@ -546,6 +552,22 @@ export interface MapViewOptions {
...
 
+    /**
+     * Should be the [[PolarTileDataSource]] used on spherical projection.
+     */
+    enablePolarDataSource?: boolean;
+
+    /**
+     * The name of the [[StyleSet]] used by [[PolarTileDataSource]] to evaluate for the decoding.
+     */
+    polarStyleSetName?: string;
```
ninok: what's the default?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -930,6 +958,20 @@ export class MapView extends THREE.EventDispatcher {
         this.m_backgroundDataSource = new BackgroundDataSource();
         this.addDataSource(this.m_backgroundDataSource);
 
+        if (this.m_enablePolarDataSource) {
+            const styleSetName =
```
ninok: I think we should still check for the projection here, otherwise you would also add the polar datasource for flat projections

ninok: Forget this comment. I saw you do this in updatePolarDataSource

@here/harp-mapview/lib/MapView.ts
okanishchev: there is `const DEFAULT_POLAR_STYLE_SET_NAME = "polar";` above

@here/harp-mapview/lib/MapView.ts
okanishchev: there is `const DEFAULT_POLAR_STYLE_SET_NAME = "polar";` above

@here/harp-mapview/lib/MapView.ts
okanishchev: the default is enabled, so it works this way right now.

@here/harp-mapview/lib/MapView.ts
okanishchev: think, I should left a comment here

@here/harp-mapview/lib/MapView.ts
okanishchev: added comment

@here/harp-mapview/lib/MapView.ts
okanishchev: added comment

PR #655, opened 2019-07-26
HARP-6293: Plug zoom level handling to globe projection.

Astrak:


@here/harp-mapview/lib/Utils.ts
```diff
@@ -240,6 +258,23 @@ export namespace MapViewUtils {
...
     }
 
+    /**
+     * The function doing a pan when [[MapView]]'s active [[ProjectionType]] is spherical.
+     *
+     * @param mapView MapView instance.
+     * @param fromWorld Start vector representing the scene position of a geolocation.
+     * @param toWorld End vector representing the scene position of a geolocation.
+     */
+    export function rotateGlobe(
```
nzjony: rotateGlobe is actually rotating the camera about the globe, to make it look like the globe is rotating (as far as I can tell)

Why not just rotate the globe and keep the camera still?

nzjony: Another comment, rotating the camera is also fine, I would just rename the function to mention that the camera is moved, as opposed to the globe, something like rotateCameraAroundGlobe... or equivalent

Astrak: Ah indeed. All the camera logic was not implemented by me so I don't know the ins and outs of the decision, but I will rename to this for clarity.

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -355,7 +345,10 @@ export class MapControls extends THREE.EventDispatcher {
             this.stopZoom();
         }
 
-        const yawPitchRoll = MapViewUtils.extractYawPitchRoll(this.camera.quaternion);
+        const yawPitchRoll = MapViewUtils.extractYawPitchRoll(
```
ninok: It is a bit weird, that you need the projection to get yaw, pitch and roll from the camera.
Maybe we have to work on our naming. Yaw, pitch and roll are just the angles of the camera rotation in world space.
If you are referring to the angles relative to the surface or local tangent space we should call them differently, e.g.:
azimuth, tilt and surfaceRoll (unfortunately there is no distinct name for roll).

@here/harp-mapview/lib/Utils.ts
```diff
@@ -240,6 +258,23 @@ export namespace MapViewUtils {
...
+        mapView: MapView,
+        fromWorld: THREE.Vector3,
+        toWorld: THREE.Vector3
+    ) {
+        tmpQuaternion.setFromUnitVectors(fromWorld.normalize(), toWorld.normalize()).inverse();
+        tmpMatrix.copyPosition(mapView.camera.matrix).makeRotationFromQuaternion(tmpQuaternion);
```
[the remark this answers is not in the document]
Astrak: `tmpMatrix` first needs to be set to the `mapView.camera.matrix`?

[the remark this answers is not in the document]
nzjony: I mean *can* be removed, not is removed.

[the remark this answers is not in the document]
Astrak: Indeed, thanks.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -298,7 +350,8 @@ export namespace MapViewUtils {
      * the raycast is above the horizon.
      *
      * @param mapView Instance of MapView.
-     * @param pointOnScreenNDC  Point in NDC space.
+     * @param pointOnScreenXNDC  Abscissa in NDC space.
```
nzjony: 🚀

PR #656, opened 2019-07-26
HARP-5144: Add picking for omv-datasource

pit-rpg:
HARP-5144, HARP-6565, Picking implemented for OMV data source,
fixed issues with road and label picking.

### DIFF
--- a/@here/harp-datasource-protocol/lib/TileInfo.ts
+++ b/@here/harp-datasource-protocol/lib/TileInfo.ts
@@ -20,6 +20,15 @@ export interface TileInfo {
 }
 
 /**
+ * Represents a feature group type for tile info.
+ */
+export enum FeatureGroupType {
+    Point,
+    Line,
+    Polygon
+}
+
+/**
  * Minimum estimated size of a JS object.
  */
 const MINIMUM_OBJECT_SIZE_ESTIMATION = 100;
@@ -143,7 +152,7 @@ export class LineFeatureGroup extends FeatureGroup {
      * An array of object defined by the user. Certain elements may be `undefined` (if this line
      * feature is not a road, or if the object for that feature is undefined).
      */
-    userData?: Array<{} | undefined> = [];
+    userData: Array<{} | undefined> = [];
 
     getNumBytes(): number {
         return (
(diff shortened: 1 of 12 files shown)


@here/harp-datasource-protocol/lib/TileInfo.ts
```diff
@@ -19,6 +19,15 @@ export interface TileInfo {
     readonly numBytes: number;
 }
 
+/**
+ * Represents a feature group type for tile info.
+ */
+export enum FeatureGropeType {
```
StefanDachwitz: What is a "Grope" ?

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -137,8 +134,8 @@ export class PickHandler {
      */
     intersectMapObjects(x: number, y: number): PickResult[] {
         const worldPos = this.mapView.getNormalizedScreenCoordinates(x, y);
-        const rayCaster = this.m_rayCaster;
 
+        const rayCaster = this.mapView.raycasterFromScreenPoint(x, y);
```
StefanDachwitz: Why is it necessary to create a new raycaster on every call?

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -102,7 +101,6 @@ export interface PickResult {
  * Handles the picking of scene geometry and roads.
  */
 export class PickHandler {
-    private readonly m_rayCaster: PickingRaycaster;
     private readonly m_plane = new THREE.Plane(new THREE.Vector3(0, 0, 1));
```
StefanDachwitz: If PickingRayCaster is no longer required, please remove file.

pit-rpg: The PickingRaycaster is used for MapViewPoints that used for Circles and Squares classes. They are almost unused in the code but I don't really sure that I can remove them.

@here/harp-mapview/lib/PickHandler.ts
pit-rpg: here the raycasterFromScreenPoint method returns only a link to the MapView.m_raycaster object

PR #657, opened 2019-07-29
HARP-6556: Generalize dynamic attributes for techniques.

zbigg:
Work-in-progress, 

Support for general dynamic attributes in techniques (fade, interpolation, time) based on `Expr`
* any attribute can be `Expr` in JSON
* most of them are resolved to constitute unique `Technique`, but
* some attributes are `DynamicAttributes` which can vary accross
 -  features (like `height`, `color`), those affect geometry creation
 - time - time/frame/zoom/camera dependent material attributes


zbigg:
Splitted into #695, #697

PR #659, opened 2019-07-29
HARP-5431 Zoomlevel now based on distance from... low resolution terrain (level 4 tiles).

nzjony:


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2158,7 +2162,7 @@ export class MapView extends THREE.EventDispatcher {
             );
         } else {
             const cameraPitch = MapViewUtils.extractYawPitchRoll(this.m_camera.quaternion).pitch;
-            const cameraPosZ = Math.abs(this.projection.groundDistance(this.m_camera.position));
+            const cameraPosZ = this.getHeightAboveTerrain(4);
```
CraigBerlin: Would be clearer is we declared a constant for the low resolution terrain zoom level.

nzjony: I also renamed the method, I think it should be clearer now.

nzjony:

PR #660, opened 2019-07-29
HARP-6277: Globe rotation inertia.

Astrak:


@here/harp-mapview/lib/Utils.ts
```diff
@@ -21,37 +21,19 @@ const MINIMUM_OBJECT3D_SIZE_ESTIMATION = 1000;
...
-    /**
-     * Cached quaternion for intermediate maths to avoid creating new instances.
-     */
-    const tmpQuaternion = new THREE.Quaternion();
-
-    /**
-     * Cached matrix for intermediate maths to avoid creating new instances.
-     */
-    const tmpMatrix = new THREE.Matrix4();
+//Caching those for performance reasons.
```
kkostecki: Maybe stupid question, but why you have extracted helpers outside the namespace? Are they used somewhere else?

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -1312,25 +1341,22 @@ export class MapControls extends THREE.EventDispatcher {
...
-            // Assign the new animation start time.
-            this.m_panAnimationStartTime = performance.now();
             this.m_panDistanceFrameDelta.subVectors(fromWorld, toWorld);
-            this.pan();
         } else if (this.mapView.projection.type === geoUtils.ProjectionType.Spherical) {
-            // TODO: HARP-6277: Implement inertia in globe.
-            MapViewUtils.rotateCameraAroundGlobe(this.mapView, fromWorld, toWorld);
+            this.m_lastRotateGlobeFromVector.copy(fromWorld);
+            this.m_lastRotateGlobeAxis.crossVectors(fromWorld, toWorld).normalize();
+            this.m_lastRotateGlobeAngle = Math.acos(fromWorld.normalize().dot(toWorld.normalize()));
```
kkostecki: If you are limited to 90 degrees (to angles gives same sinus value between 0-PI), angle may be retrieved directly from cross product of normalized vectors, then divide cross by sines and normalization of cross is done.
Maybe not important, but gives small optimization.

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -1184,7 +1213,7 @@ export class MapControls extends THREE.EventDispatcher {
                 const pinchDistance = this.calculatePinchDistanceInWorldSpace();
                 this.moveAlongTheViewDirection(pinchDistance);
             } else if (this.mapView.projection.type === geoUtils.ProjectionType.Spherical) {
-                // TODO: HARP-6597: Implement rotation for globe.
+                // TODO: HARP-6597: Implement yaw rotation for globe.
                 const pinchDistance = this.calculatePinchDistanceInWorldSpace();
                 this.setZoomLevel(this.currentZoom + pinchDistance * 0.01);
```
kkostecki: Would love to have this as some static const definition, maybe PINCH_ZOOM_LEVEL_RATIO ?

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -777,41 +789,73 @@ export class MapControls extends THREE.EventDispatcher {
...
+            );
+        } else if (this.mapView.projection.type === geoUtils.ProjectionType.Spherical) {
+            MapViewUtils.rotateCameraAroundGlobe(
+                this.mapView,
+                this.m_lastRotateGlobeFromVector,
+                this.m_tmpVector3
+                    .copy(this.m_lastRotateGlobeFromVector)
+                    .applyQuaternion(this.m_rotateGlobeQuaternion)
+            );
+            this.mapView.camera.updateMatrixWorld(true);
```
kkostecki: If its really needed (no update from outside), I guess it should be encapsulated in MapViewUtils.rotateCameraAroundGlobe() method implementation, it already has mapView reference to access.

@here/harp-mapview/lib/Utils.ts
Astrak: Yes that was a request from @robertoraggi, because they should not be in the API of the Utils namespace, they are just there to avoid creating and dumping objects every frame.

@here/harp-map-controls/lib/MapControls.ts
Astrak: Right, however this is only for the in progress work on the controls. Later the use of `moveAlongTheViewDirection` will be implemented in lieu of `setZoomLevel` when zooming with touch, as in flat version above.

PR #661, opened 2019-07-30
HARP-6379: Camera change interception/hooks

SergiiKulakovskyi:

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -105,7 +105,9 @@ export enum MapViewEventNames {
     /** Called when the WebGL context is lost. */
     ContextLost = "webglcontext-lost",
     /** Called when the WebGL context is restored. */
-    ContextRestored = "webglcontext-restored"
+    ContextRestored = "webglcontext-restored",
+    /** Called when camera position has been changed. */
+    CameraPositionChanged = "camera-changed"
 }
 
 const logger = LoggerManager.instance.create("MapView");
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2392,6 +2395,23 @@ export class MapView extends THREE.EventDispatcher {
 
         this.m_movementDetector.checkCameraMoved(this, time);
 
+        if (this.m_movementDetector.cameraMovedLastFrame) {
+            const { yaw, pitch, roll } = MapViewUtils.extractYawPitchRoll(this.camera.quaternion);
+            const { latitude, longitude, altitude } = this.projection.unprojectPoint(
```
ninok: TLDR: Please pass the camera position instead of the look-at point to the callback.

{latitude, longitude, altitude} represents the look at point, not the camera position. There are several concerns with this:
1. the event is called CameraPositionChanged so I would expect that it returns the camera position
2. the rayCast method does not always return a result, especially when the camera is looking up or when using sphere projection
3. even if the ray is hitting the plane you might get a world point that is not representable in geo space. I guess the unproject method will just clamp it, but this might not be what the user wants

So if the user really needs the look-at point instead of the position he/she has to compute it. Maybe you can provide a helper method in MapViewUtils `getCameraTargetCoordinatesFromCameraCoordinates`. We already have the inverse method there.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2392,6 +2395,23 @@ export class MapView extends THREE.EventDispatcher {
 
         this.m_movementDetector.checkCameraMoved(this, time);
```
ninok: This is already returning true/false, so just move it into the if condition

SergiiKulakovskyi: Thanks for noticing

@here/harp-mapview/lib/MapView.ts
SergiiKulakovskyi: Totally agree

PR #662, opened 2019-07-30
HARP-6378: Synchronous rendering

SergiiKulakovskyi:

### DIFF
--- a/@here/harp-examples/src/rendering_synchronous.ts
+++ b/@here/harp-examples/src/rendering_synchronous.ts
@@ -1,0 +1,200 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { GeoCoordinates } from "@here/harp-geoutils";
+import { CopyrightElementHandler, MapView, MapViewEventNames } from "@here/harp-mapview";
+import { APIFormat, OmvDataSource } from "@here/harp-omv-datasource";
+import THREE = require("three");
+import { accessToken } from "../config";
+
+/**
+ * The example shows how to render map synchronously within your own render loop.
+ * By default, when map is updated, changes will be rendered in the next animation frame.
+ *
+ * Setting `synchronousRendering` to `true` allows to control rendering process
+ * and by calling `mapView.renderSync()`.
+ *
+ * ```typescript
+ * [[include:harp_gl_rendering_synchronous_1.ts]]
+ * ```
+ *
+ * `MapViewEventNames.Update` event fired when [[MapView]] requests for a redraw.
+ * E.g.: Tiles asynchronously decoded and ready for rendering, labels animation, etc...
+ *
+ * Subscribe to this event, and call your `update` method.
+ *
+ * ```typescript
+ * [[include:harp_gl_rendering_synchronous_2.ts]]
+ * ```
+ *
+ * Implement your own render loop like in the example below.
+ * With `mapView.renderSync()` you will immediately redraw the map scene.
+ *
+ * Make checks to avoid multiple redrawing at one frame.
+ *
+ * ```typescript
+ * [[include:harp_gl_rendering_synchronous_3.ts]]
+ * ```
+ *
+ */
+export namespace SynchronousRendering {
+    // Creates a new MapView for the HTMLCanvasElement of the given id
+    function initializeMapView(id: string): MapView {
+        const canvas = document.getElementById(id) as HTMLCanvasElement;
+
+        // snippet:harp_gl_rendering_synchronous_1.ts
+        const map = new MapView({
+            canvas,
+            theme: "resources/berlin_tilezen_base.json",
+            synchronousRendering: true
+        });
+        // end:harp_gl_rendering_synchronous_1.ts
+        map.renderLabels = false;
+
+        CopyrightElementHandler.install("copyrightNotice", map);
+
+        // Resize the mapView to maximum
+        map.resize(window.innerWidth, window.innerHeight);
+
+        // React on resize events
+        window.addEventListener("resize", () => {
+            map.resize(window.innerWidth, window.innerHeight);
+        });
+
+        const omvDataSource = new OmvDataSource({
+            baseUrl: "https://xyz.api.here.com/tiles/herebase.02",
+            apiFormat: APIFormat.XYZOMV,
+            styleSetName: "tilezen",
+            authenticationCode: accessToken
+        });
+        map.addDataSource(omvDataSource);
+
+        return map;
+    }
+
+    class Popup {
+        private canvas: HTMLCanvasElement;
+        private context: CanvasRenderingContext2D;
+
+        constructor(text: string, private coordinates: GeoCoordinates) {
+            this.addHTMLElements(text);
+
+            this.canvas = document.getElementById("popupLine") as HTMLCanvasElement;
+            this.context = this.canvas.getContext("2d") as CanvasRenderingContext2D;
+
+            this.canvas.width = window.innerWidth;
+            this.canvas.height = window.innerHeight;
+            window.addEventListener("resize", () => {
+                this.canvas.width = window.innerWidth;
+                this.canvas.height = window.innerHeight;
+            });
+        }
+
+        addHTMLElements(text: string) {
+            document.body.innerHTML += `
+                <style>
+                    #popupLine {
+                        position: absolute;
+                        border: 0;
+                        left: 0;
+                        width: 100%;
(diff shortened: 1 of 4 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1624,9 +1624,22 @@ export class MapView extends THREE.EventDispatcher {
...
+                zoomLevel
+            );
+            const pitchRad = THREE.Math.degToRad(pitchDeg);
+            const projectedDistanceOnTheGround = distanceToGround * Math.tan(pitchRad);
+
+            const yawRad = THREE.Math.degToRad(yawDeg);
+            const worldTargetCoordinates = this.projection.projectPoint(geoPos);
+
+            this.m_camera.position.set(
+                worldTargetCoordinates.x + Math.sin(yawRad) * projectedDistanceOnTheGround,
```
ninok: Why do you move the camera position in x and y direction? This API should just set the position of the camera using the projected geoPos and adjust the height (z for planar projection) based on the provided zoomlevel.
If you want to set the target/look-at point of the camera you would not use this API.
Looking at the code below for the sphere projection, I think it would also work for the planar projection, so just remove the if part and keep the else part.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2030,6 +2051,13 @@ export class MapView extends THREE.EventDispatcher {
         this.m_elevationProvider = elevationProvider;
     }
 
+    /**
+     * Draw a new frame synchronously.
+     */
+    redraw() {
```
zbigg: `renderSync` ?

If this is public, then maybe remove this bool attribute from `update`.
Update is meant mainly for other APIs to inform MAP, that update is needed, not to really re-render.

You should also add `if (!this.m_updatePending) { return}` to `private render()` so, if we rendered stuff asynchronously, nobody will try to redraw it again. Or similar mechanism.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1624,9 +1624,22 @@ export class MapView extends THREE.EventDispatcher {
         pitchDeg: number = 0
     ): void {
         if (this.projection.type === ProjectionType.Planar) {
+            const distanceToGround = MapViewUtils.calculateDistanceToGroundFromZoomLevel(
```
zbigg: Please add comment what are you trying to achieve.
This doesn't look like it has anything to do with synchronous rendering,

@here/harp-mapview/lib/MapView.ts
nzjony: There is already a method renderFunc where you can pass in the time, is that not good enough?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1921,8 +1934,16 @@ export class MapView extends THREE.EventDispatcher {
...
     /**
      * Requests a redraw of the scene.
+     *
+     * @param forceRedraw If `true`, draws scene immediately, without waiting for the next
+     * animation frame
      */
-    update() {
+    update(forceRedraw?: true) {
+        if (forceRedraw === true) {
+            this.redraw();
```
nzjony: in `drawFrame` (on MapView), there is some code to cancel an active requestAnimationFrame, is this necessary here also?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1921,8 +1934,16 @@ export class MapView extends THREE.EventDispatcher {
 
     /**
      * Requests a redraw of the scene.
```
nzjony: This might be a bit confusing, there is now a public method update which requests a redraw and a method redraw (which I guess does the same)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1921,8 +1934,16 @@ export class MapView extends THREE.EventDispatcher {
...
+     *
+     * @param forceRedraw If `true`, draws scene immediately, without waiting for the next
```
nzjony: Maybe add a clarifying comment that this also overrides the internal maxFps setting?

@here/harp-mapview/lib/MapView.ts
zbigg: Following other comments, like 

>  there is some code to cancel an active requestAnimationFrame, is this necessary here also?

So, looks like `renderFunc` is intended to be called by `rAF`, so it assumes that frame was canceled - no need to cancel.

This, one on the other hand can cancel requested rAF as we know it won't be needed (unless new update is scheduled), so proposed

> `if (!this.m_updatePending) { return} to private render()`

is not needed.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1939,8 +1939,17 @@ export class MapView extends THREE.EventDispatcher {
 
     /**
      * Requests a redraw of the scene.
+     *
+     * @param forceRedraw If `true`, updates scene immediately, without waiting for the next
+     * animation frame. Note: Internal `maxFps` will be overriden as this inteded to be called
+     * from external render loop.
      */
-    update() {
+    update(forceUpdate?: boolean) {
```
ninok: forceUpdate or forceRedraw?
I'm also wondering if having a parameter here is enough. If an external application is managing the rneder loop and requesting animation frames MapView should never request animation frames. Currently MapView will still request animation frames in a lot of situations (e.h. whenever data comes in, building animation is ongoing, ...).

So I think this needs to be a MapView option to completely switch to synchronous rendering.

SergiiKulakovskyi: I think we can't rely on the user to deal with all `updates` that we have.
And this `forceUpdate` option is an easy solution for not to wait for the next animation frame and just re-render.

`forceUpdate` sounds better as we have `update` method.

@here/harp-mapview/lib/MapView.ts
SergiiKulakovskyi: I removed this code.
And replaced `camera.matrixWorldNeedsUpdate = true` with `camera.updateMatrixWorld()` in `zoomOnTargetPosition` method

SergiiKulakovskyi: ...and camera moves smoothly with sync rendering.

@here/harp-mapview/lib/MapView.ts
ninok: Why not? If we set synchronous rendering to true in the MapViewOptions, we could just do nothing in MapView.update, except triggering an event that an update was requested.
The application could register to that event and request an animation frame. In the animation frame the application would call MapView.redraw().
I don't think that the current solution will work if the application is drawing some additional stuff in the animation frame. If the animation frame is coming from map view that additional stuff would be missing, i.e. leading to some flickering.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -511,6 +512,13 @@ export interface MapViewOptions {
      * Disable all fading animations for debugging and performance measurement.
      */
     disableFading?: boolean;
+
+    /**
+     * Set to `true` to allow rendering scene synchronously.
```
ninok: I think we need more documentation here what are the impacts if this is set to true (e.g. update will do nothing except triggering an event)
Would be also good to have a dedicated example for that.

@here/harp-examples/src/rendering_synchronous.ts
```diff
@@ -0,0 +1,167 @@
...
+        }
+        updatePending = true;
+
+        currentState.yaw += 0.05;
+        requestAnimationFrame(draw);
+    }
+
+    function draw() {
+        // Avoids multiple redrawing
+        if (updating) {
```
ninok: I would call it drawing to avoid confusion

@here/harp-examples/src/rendering_synchronous.ts
```diff
@@ -0,0 +1,167 @@
...
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { GeoCoordinates } from "@here/harp-geoutils";
+import { CopyrightElementHandler, MapView, MapViewEventNames } from "@here/harp-mapview";
+import { APIFormat, OmvDataSource } from "@here/harp-omv-datasource";
+import THREE = require("three");
+import { accessToken } from "../config";
+
+export namespace SynchronousRendering {
```
ninok: Please add proper documentation what this example is showing and how it is achieved using typedoc snippets, like it is done for the hello world example:
https://github.com/heremaps/harp.gl/blob/master/%40here/harp-examples/src/getting-started_hello-world_npm.ts
https://www.harp.gl/docs/master/doc/modules/harp_examples.helloworldexample.html
The documentation is generated with:
`yarn typedoc`

PR #664, opened 2019-07-30
HARP-6638: Continuous redraws in MapView

okanishchev:
Phased loader can't load generated tiles, and reports they are not finished, producing continuous redraws.

### DIFF
--- a/@here/harp-mapview-decoder/lib/TileDataSource.ts
+++ b/@here/harp-mapview-decoder/lib/TileDataSource.ts
@@ -156,6 +156,7 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
             storageLevelOffset: this.m_options.storageLevelOffset
         });
 
+        this.useGeometryLoader = true;
         this.cacheable = true;
         this.m_tileLoaderCache = new LRUCache<number, TileLoader>(this.getCacheCount());
     }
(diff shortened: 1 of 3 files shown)


@here/harp-debug-datasource/lib/DebugTileDataSource.ts
```diff
@@ -113,6 +113,8 @@ export class DebugTile extends Tile {
 }
 
 export class DebugTileDataSource extends DataSource {
+    readonly loadable = false;
```
StefanDachwitz: This is kinda hacky, it overrides the non-writable value in the base class. It works, but is bad style. Better:
* make "loadable" protected in base class
* set this.loadable in the constructors to false
* please rename "loadable" to "useGeometryLoader"

okanishchev: Good point, thanks!

okanishchev:
@ninok, I also think so, but there are some misconception with the name TileDataSource.
For example, BackgroundDataSource actually provides tiles, so should it inherit TileDataSource? Perhaps DebugTileDataSource?

It provides functionality for loading and decoding tiles, so it would be more convinient to name it something like LoadableDataSource.

ninok:
> @ninok, I also think so, but there are some misconception with the name TileDataSource.
> For example, BackgroundDataSource actually provides tiles, so should it inherit TileDataSource? Perhaps DebugTileDataSource?
> 
> It provides functionality for loading and decoding tiles, so it would be more convinient to name it something like LoadableDataSource.

Yeah, the name is confusing and LoadableDataSource might be a better name.
My point was just that the implementer of a data-source should not need to care if a geometry loader is used or not. You either choose to derive from DataSource(when creating three.js objects) or TileDataSource (when creating decoded tiles).

PR #672, opened 2019-07-31
HARP-6663: Improve performance by reducing tile subdivisions

okanishchev:


@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -130,6 +133,11 @@ export class FrustumIntersection {
                 continue;
             }
 
+            const proceed = dataSources.some((ds, i) => ds.shouldSubdivide(zoomLevels[i], tileKey));
```
kkostecki: Oleksii I guess we could move this section, as also condition above just before uniqueKey getter, this would be even faster, wright? Otherwise looks good!

okanishchev:
@kkostecki, the good thing here, is that storageLevelOffset is used to get right zoomlevel by dataSource.getDisplayZoomLevel, so it's already computed. That's why there are special zoomlevel for every datasource in array.

@here/harp-mapview/lib/FrustumIntersection.ts
okanishchev: Yes, we can, as well as the section with maxTileLevel above. But I think this unique key is something like sanity check needed for development and beleive it can be completely removed for production build somehow. Anyway we can ping @atomicsulfate for more information about that.

okanishchev:
Don't see any issues in mapview demo, but I'd prefer to receive more feedback from people.

@here/harp-mapview/lib/FrustumIntersection.ts
ninok: The uniqueKey is for wrap around at the antimeridian(date border) and basically just tilekey and offset packed into one number.
It is needed to distinguish the wraped tiles from the "original" tiles.
So please move your code up.

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -329,6 +329,19 @@ export abstract class DataSource extends THREE.EventDispatcher {
...
+    /**
+     * Returns `true` if [[MapView]] should traverse tiles further with given [[TileKey]] and
+     * zoom level.
+     *
+     * @param zoomLevel The zoom level of the [[MapView]].
+     * @param tileKey The unique identifier for a map tile.
+     * @returns `true` if the subtiles of the given [[TileKey]] should be
+     * checked for collisions.
+     */
+    shouldSubdivide(zoomLevel: number, tileKey: TileKey): boolean {
```
nzjony: I don't think this method is necessary in DataSource, it is fine enough to have it in the FrustumIntersection

@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -120,16 +123,21 @@ export class FrustumIntersection {
             }
 
             const tileKey = tileEntry.tileKey;
+            if (tileKey.level > maxTileLevel) {
```
nzjony: Why not just have a condition, i
`if (tileKey.level === maxTileLevel) {
continue;
}`
I don't understand the need for iterating over the datasources.

nzjony: @atomicsulfate , you also worked on this? Is this acceptable?

okanishchev: @nzjony, this enables custom culling on datasources, defined not only by zoomlevel.
For example, in poles datasource you don't ever want to check tiles outside of pole radius,
https://github.com/heremaps/harp.gl/pull/653/files#diff-1b0ff12045051a049d60bf7d1c58e4a3R140

okanishchev: besides, only this change reduces overall subdivisions per frame. I didn't check why, but it is

@here/harp-mapview/lib/DataSource.ts
okanishchev: The idea here is to allow datasources to skip tile subdivision, defined by datasource's own rules

@here/harp-mapview/lib/FrustumIntersection.ts
nzjony: Shouldn't this then be a property of the TilingScheme and not the DataSource? I guess the poles have some custom TilingScheme, i think it makes more sense to have it there no?

okanishchev: No.
For polar datasource, the transverse Mercator projection is used along with regular quadtree. With that scheme, all the spherical space except two regions at (0, -90), (0, +90) is defined and can be addressed. And for pole datasource, it is vital only to have north and south pole addressed (+90, 0), (-90, 0) with a radius of ~5 degrees.
So, although it sounds like a good idea, this can't be defined in terms of TilingScheme

nzjony: Just add the method to TilingScheme, and have a PolarTilingScheme inherit from the TransverseMercatorProjection?
Would make more sense that having the DataSource knowing if it should subdivide or not.

okanishchev: It's a bad idea because we can start to spam tiling schemas here and there, while we actually want to reduce them due to high traverse complexity.
If in other way you'd have two equal schemas traversed once, with this approach you'll have two different schemas traversed twice.

nzjony: I don't quite follow, you have a polarTilingScheme:

```
export const polarTilingScheme = new TilingScheme(
    quadTreeSubdivisionScheme,
    transverseMercatorProjection
);
```

This will be treated separately in the VisibleTileSet, so you anyway have to traverse it separately from the other TilingSchemes, why would specializing this further cause a double traversal?

okanishchev: You can add another datasource with same tiling scheme, but different clipping rules, and traversing would occur only once. But I see your point, if we skip some tiles completely, wouldn't it be different tiling scheme? May be. It's a topic for discussion and testing

ninok: @nzjony Having the shouldSubdivide in the DataSource is much more flexible as having it in the tiling scheme. Imagine a usecase where you depending on the data in a tile dynamically decide that you don't want to subdivide any further. This could not be done if the shouldSubdivide is in the TilingScheme

nzjony: All good, I anyway approved these changes on the 5th of August, so I don't know why it wasn't submitted.

okanishchev:
Done in #653

PR #677, opened 2019-08-01
HARP-6654: Move creation of credential.js to build-bundle step

ninok:


package.json
```diff
@@ -35,7 +35,7 @@
         "start": "webpack-dev-server -d --config @here/harp-examples/webpack.config.js",
         "build": "npm run build-bundle && npm run build-examples",
         "build-examples": "webpack -d --config @here/harp-examples/webpack.config.js",
-        "build-bundle": "cd @here/harp.gl && webpack",
+        "build-bundle": "ts-node ./scripts/credentials.ts && cd @here/harp.gl && webpack",
```
zbigg: This should be actually, part of `build-examples`.
This `credentials.js` file is by no means connected with bundle, but is a dependency of generated examples.

ninok: But it is only used in the example that uses the bundle and when just building the examples the bundle is not build.
Nevertheless it doesn't matter. I can move it to the example build.

ninok: This was still not fixing the issue. I guess there is also sth. wrong in our lovely webpack config.

PR #678, opened 2019-08-01
HARP-6605: Added map-repeat param to MapView.

bohdanhorpynchuk:
Signed-off-by: bohdanhorpynchuk <ext-bogdan.gorpynchuk@here.com>

### DIFF
--- a/.ort.yml
+++ b/.ort.yml
@@ -2,5 +2,5 @@ excludes:
   scopes:
   - name: "devDependencies"
     reason: "BUILD_TOOL_OF"
-    comment: "These are dependencies only used for development. They are not distributed in the context of this product."
+    comment: "These are dependencies only used for development."
 


@here/harp-mapview/lib/MapView.ts
```diff
@@ -506,6 +506,13 @@ export interface MapViewOptions {
      */
     enablePhasedLoading?: boolean;
 
+    /**
+     * Enable map repeat. If `false`, map will be repeated on X axis continuously.
+     *
+     * @default `true`
+     */
+    mapRepeat?: boolean;
```
ninok: mapRepeat is not very descriptive name. Maybe stick to tileWrappingEnabled

@here/harp-mapview/lib/MapView.ts
```diff
@@ -506,6 +506,13 @@ export interface MapViewOptions {
      */
     enablePhasedLoading?: boolean;
 
+    /**
+     * Enable map repeat. If `false`, map will be repeated on X axis continuously.
```
ninok: 1. If true, the map will repeat
2. only for planar projections
3. instead of X axis I would reword it in longitudal direction (i.e. there is no X axis in geo space)
4. also mention where it will repeat/wrap around (i.e. data-border or antimeridian)

@here/harp-mapview/test/VisibleTileSetTest.ts
```diff
@@ -44,7 +44,8 @@ class Fixture {
         this.frustumIntersection = new FrustumIntersection(
             this.camera,
             MapViewDefaults.projection,
-            MapViewDefaults.extendedFrustumCulling
+            MapViewDefaults.extendedFrustumCulling,
+            true
```
nzjony: add a test for false please

PR #681, opened 2019-08-01
MINOR: fix SubTiles for custom subdivision schemes

okanishchev:
Do we need that SubTiles at all?

### DIFF
--- a/@here/harp-geoutils/lib/tiling/TileTreeTraverse.ts
+++ b/@here/harp-geoutils/lib/tiling/TileTreeTraverse.ts
@@ -5,7 +5,6 @@
  */
 
 import { SubdivisionScheme } from "./SubdivisionScheme";
-import { SubTiles } from "./SubTiles";
 import { TileKey } from "./TileKey";
 
 export class TileTreeTraverse {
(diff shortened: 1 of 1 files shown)


@here/harp-geoutils/lib/tiling/TileTreeTraverse.ts
```diff
@@ -16,18 +15,22 @@ export class TileTreeTraverse {
...
-            this.m_subdivisionScheme.getSubdivisionX(tileKey.level) *
-            this.m_subdivisionScheme.getSubdivisionY(tileKey.level);
+        const level = tileKey.level;
+        const divX = this.m_subdivisionScheme.getSubdivisionX(level);
+        const divY = this.m_subdivisionScheme.getSubdivisionY(level);
 
-        // tslint:disable-next-line:no-bitwise
-        const subTileMask = ~(~0 << subTileCount);
-
-        const subTiles = new SubTiles(tileKey, 1, subTileMask);
```
ninok: I guess SubTiles was added for testability, but I think we could also add a test that checks TileTreeTraverse.subTiles directly. So for me it's fine to remove SubTiles but maybe also double check with Andres(Mandado) b/c he created that class.

@here/harp-geoutils/lib/tiling/TileTreeTraverse.ts
```diff
@@ -16,18 +15,22 @@ export class TileTreeTraverse {
...
         const result = new Array<TileKey>();
 
-        for (const subTile of subTiles) {
-            result.push(subTile);
+        for (let y = 0; y < divY; y++) {
+            for (let x = 0; x < divX; x++) {
+                result.push(
```
ninok: Would be more performant to change the interface this method to take a callback that is called for every subtile instead of creating a temporary array and calling foreach on it.

@here/harp-geoutils/lib/tiling/TileTreeTraverse.ts
okanishchev: Fortunately, I've added test cases for TileTreeTraverse that fails with current SubTiles implementation.
https://github.com/heremaps/harp.gl/pull/683/files

okanishchev:
Close in favor of #708

PR #686, opened 2019-08-05
HARP-6690: Add APIs to get the local tangent space of a geopoint

robertoraggi:


@here/harp-geoutils/lib/math/TransformLike.ts
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Vector3Like } from "./Vector3Like";
+
+/**
+ * The interface [[TransformLike]] is used to represent a 3D transform.
```
ninok: It actually represents an affine 3d transformation(i.e. rotation + translation)

@here/harp-geoutils/lib/math/TransformLike.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export interface TransformLike {
```
ninok: Maybe AffineTransformLike would be a more precise name

PR #689, opened 2019-08-05
MINOR: Added @types/tween.js as dependency for typescript users of harp

ninok:
Signed-off-by: Nino Kettlitz <1396039+ninok@users.noreply.github.com>

### DIFF
--- a/@here/harp-map-controls/package.json
+++ b/@here/harp-map-controls/package.json
@@ -26,14 +26,14 @@
         "@here/harp-geoutils": "^0.4.3",
         "@here/harp-mapview": "^0.9.4",
         "@here/harp-utils": "^0.2.7",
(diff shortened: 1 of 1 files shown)


@here/harp-map-controls/package.json
```diff
@@ -26,14 +26,14 @@
         "@here/harp-geoutils": "^0.4.3",
         "@here/harp-mapview": "^0.9.4",
         "@here/harp-utils": "^0.2.7",
-        "@tweenjs/tween.js": "^17.4.0"
+        "@tweenjs/tween.js": "^17.4.0",
+        "@types/tween.js": "^17.2.0"
```
robertoraggi: why do we need both?

robertoraggi: oops, yep, we tween.js doesn't bundle the typings.

PR #694, opened 2019-08-06
Harp 5728 area overlay

atomicsulfate:
Overlay of areas over terrain using vertex displacement maps.

Add DisplacementFeature to MapMeshBasicMaterial, and include the displacement map vertex
shader chunk to its shader.
Change FillTechnique to generate vertex normals and texture coordinates.

### DIFF
--- a/@here/harp-datasource-protocol/lib/Techniques.ts
+++ b/@here/harp-datasource-protocol/lib/Techniques.ts
@@ -326,6 +326,7 @@ export function isShaderTechnique(technique: Technique): technique is ShaderTech
  */
 export function needsVertexNormals(technique: Technique): boolean {
     return (
+        isFillTechnique(technique) ||
         isStandardTechnique(technique) ||
         isTerrainTechnique(technique) ||
         isStandardExtrudedLineTechnique(technique)
--- a/@here/harp-mapview-decoder/lib/TileDataSource.ts
+++ b/@here/harp-mapview-decoder/lib/TileDataSource.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { DecodedTile, ITileDecoder, StyleSet, TileInfo } from "@here/harp-datasource-protocol";
+import { ITileDecoder, StyleSet, TileInfo } from "@here/harp-datasource-protocol";
 import { TileKey, TilingScheme } from "@here/harp-geoutils";
 import {
     ConcurrentDecoderFacade,
@@ -15,7 +15,7 @@ import {
 } from "@here/harp-mapview";
 
 import { LRUCache } from "@here/harp-lrucache";
-import { assert, LoggerManager } from "@here/harp-utils";
+import { LoggerManager } from "@here/harp-utils";
 import { DataProvider } from "./DataProvider";
 import { TileInfoLoader, TileLoader } from "./TileLoader";
 
@@ -226,6 +226,8 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
                 0
             );
             tile.tileLoader = newTileLoader;
+            tile.copyrightInfo = this.m_options.copyrightInfo;
+
             // We don't cache tiles with level 4 and above, at this level, there are 16 (2^4) tiles
             // horizontally, given the assumption that the zoom level assumes the tile should be 256
             // pixels wide (see function [[calculateZoomLevelFromDistance]]), and the current
@@ -240,34 +242,12 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
             }
         }
 
-        this.updateTile(tile);
-        return tile;
-    }
-
-    updateTile(tile: Tile) {
-        const tileLoader = tile.tileLoader;
-        if (tileLoader === undefined) {
-            return;
-        }
-        if (tileLoader.decodedTile !== undefined) {
-            this.setDecodedTileOnTile(tileLoader.decodedTile, tile);
+        if (tile.tileLoader.decodedTile !== undefined) {
+            tile.decodedTile = tile.tileLoader.decodedTile;
         } else {
-            tileLoader
-                .loadAndDecode()
-                .then(tileLoaderState => {
-                    assert(tileLoaderState === TileLoaderState.Ready);
-                    const decodedTile = tileLoader.decodedTile;
-                    this.setDecodedTileOnTile(decodedTile, tile);
-                })
-                .catch(tileLoaderState => {
-                    if (
-                        tileLoaderState !== TileLoaderState.Canceled &&
-                        tileLoaderState !== TileLoaderState.Failed
-                    ) {
-                        this.logger.error("Unknown error" + tileLoaderState);
-                    }
-                });
+            tile.load();
         }
+        return tile;
     }
 
     /**
@@ -300,37 +280,10 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
         return promise;
     }
 
-    decodedTileHasGeometry(decodedTile: DecodedTile) {
-        return (
-            decodedTile.geometries.length ||
-            (decodedTile.poiGeometries !== undefined && decodedTile.poiGeometries.length) ||
-            (decodedTile.textGeometries !== undefined && decodedTile.textGeometries.length) ||
-            (decodedTile.textPathGeometries !== undefined && decodedTile.textPathGeometries.length)
-        );
-    }
-
     private getCacheCount(): number {
         // We support up to [[maxLevelTileLoaderCache]] levels, this equates to roughly
         // 2^maxLevelTileLoaderCache^2 tiles in total (at level maxLevelTileLoaderCache), we don't
         // generally see that many, so we add a factor of 2 to try to get the worst case.
         return Math.pow(2, maxLevelTileLoaderCache) * 2;
-    }
-
-    // Applies the decoded tile to the tile.
-    // If the geometry is empty, then the tile's forceHasGeometry flag is set.
-    // Map is updated.
-    private setDecodedTileOnTile(decodedTile: DecodedTile | undefined, tile: Tile) {
-        if (decodedTile && this.decodedTileHasGeometry(decodedTile)) {
-            tile.copyrightInfo =
-                decodedTile.copyrightHolderIds !== undefined
-                    ? decodedTile.copyrightHolderIds.map(id => ({ id }))
-                    : this.m_options.copyrightInfo;
-
-            tile.setDecodedTile(decodedTile);
-        } else {
-            // empty tiles are traditionally ignored and don't need decode
-            tile.forceHasGeometry(true);
-        }
-        this.requestUpdate();
     }
 }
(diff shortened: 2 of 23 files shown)


@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -249,25 +253,23 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
...
-                    if (
-                        tileLoaderState !== TileLoaderState.Canceled &&
-                        tileLoaderState !== TileLoaderState.Failed
-                    ) {
-                        this.logger.error("Unknown error" + tileLoaderState);
-                    }
-                });
-        }
+
+        tileLoader
```
nzjony: this changes the implementation of the method, should `updateTile` be renamed?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -648,17 +648,21 @@ export class VisibleTileSet {
...
-            retainedTiles.add(tileCode);
-            tile.reload();
-        });
-        renderListEntry.renderedTiles.forEach(tile => {
+
+        function markTileDirty(tile: Tile, tileGeometryManager: TileGeometryManager) {
             const tileCode = TileOffsetUtils.getKeyForTileKeyAndOffset(tile.tileKey, tile.offset);
             if (!retainedTiles.has(tileCode)) {
                 retainedTiles.add(tileCode);
+                tileGeometryManager.initTile(tile);
```
nzjony: Would it be worth making a new method to `reinit` a tile, this allocates a new loader, however the internal state should just need to be cleared, @StefanDachwitz, do you have an opinion?

StefanDachwitz: I guess I would implement a reinit(), to clear() the tile of any old geometry, and to make sure the old loader is not adding anything, is cancelled and destroyed properly.

@here/harp-mapview-decoder/lib/TileDataSource.ts
atomicsulfate: I'm not sure why you think the name needs to be changed, now the tile is always updated if you call the method.
Maybe you mean it should better be `loadTile()`?
I've realized there's no reason to go through TileDataSource to do what we're doing in here. I think we should move this method to `Tile.load()`.
The same with `setDecodedTileOnTile()`, which can be moved to `Tile.setDecodedTile()` `decodedTileHasGeometry()`, which should be moved to `DecodedTile.hasGeometry()`.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2087,6 +2088,26 @@ export class MapView extends THREE.EventDispatcher {
...
+        this.m_tileGeometryManager.setTileUpdateCallback((tile: Tile) => {
+            overlayOnElevation(tile);
+        });
+        this.clearTileCache();
+    }
+
+    /**
+     * Clears any elevation sources and provider previously set.
+     */
+    clearElevationSource() {
```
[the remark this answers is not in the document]
atomicsulfate: done.

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -342,6 +343,8 @@ export function textureCoordinateType(technique: Technique): TextureCoordinateTy
         return technique.textureCoordinateType;
     } else if (isTerrainTechnique(technique)) {
         return technique.textureCoordinateType;
+    } else if (isFillTechnique(technique)) {
+        return TextureCoordinateType.TileSpace;
```
nzjony: Is there a reason this is different from the others, i.e. all the others mention `technique.textureCoordinateType`

@here/harp-mapview-decoder/lib/TileDataSource.ts
nzjony: `loadTile` to me gives a clearer picture of what is happening behind the scenes, `update` to me seems like it implies there is data that is added and is less precise (but I understand than an update can of course be implemented in terms of a complete re loading of the data).
The changes you made to not go through the `TileDataSource` also make it much cleaner!

nzjony: at least the performance implications of `loadTile` are a bit clearer also than `updateTile`

@here/harp-omv-datasource/lib/OmvDataSource.ts
```diff
@@ -290,4 +296,15 @@ export class OmvDataSource extends TileDataSource<OmvTile> {
             storageLevelOffset: this.storageLevelOffset
         });
     }
+
+    setEnableElevationOverlay(enable: boolean) {
+        if (this.m_decoderOptions.enableElevationOverlay !== enable) {
+            this.m_tileLoaderCache.clear();
```
nzjony: part of this code seems similar to that in setStyleSet, would it make sense to create a method to bundle this together, something like `reconfigureDecoders` (for want of a better name).

@here/harp-omv-datasource/test/OmvDecodedTileEmitterTest.ts
```diff
@@ -65,7 +65,13 @@ describe("OmvDecodedTileEmitter", function() {
 
         const styleSetEvaluator = new StyleSetEvaluator(styleSet);
 
-        const tileEmmiter = new OmvDecodedTileEmitter(decodeInfo, styleSetEvaluator, false, false);
+        const tileEmmiter = new OmvDecodedTileEmitter(
```
nzjony: Emmiter -> Emitter

atomicsulfate:
https://github.com/heremaps/harp.gl/pull/694#discussion_r312380509

The difference with the other techniques listed there is that FillTechnique does not expose any texture parameter, so it doesn't make sense to have a textureCoordinateType parameter either.
However we do need to generate texture coordinates for the elevation displacement map.

Anyway, I'll remove the line you commented, it's not needed because we take care of generating texture coordinates for the fill technique in OmvDecodedTileEmitter.

nzjony:
Also, general comment, why do you need to

nzjony:
Also, general comment, why do you need to

nzjony:
Sorry!!!! Didn't mean to close, clicked by accident. Wanted to add a general comment, but will make it a bit more specific.

@here/harp-datasource-protocol/lib/Techniques.ts
atomicsulfate: The difference with the other techniques listed there is that FillTechnique does not expose any texture parameter, so it doesn't make sense to have a textureCoordinateType parameter either.
However we do need to generate texture coordinates for the elevation displacement map.

Anyway, I'll remove the line you commented, it's not needed because we take care of generating texture coordinates for the fill technique in OmvDecodedTileEmitter.

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
```diff
@@ -98,6 +100,15 @@ export interface TileGeometryManager {
...
     getAvailableKinds(tiles: Tile[]): GeometryKindSet;
+
+    /**
+     * Sets a callback that will be called for every updated tile on [[updateTiles]].
+     *
+     * @param {TileUpdateCallback} callback The callback that will be called after a tile has been
+     * updated, passing the updated tile as argument. If `undefined`, a previously set callback will
+     * be cleared.
+     */
+    setTileUpdateCallback(callback?: TileUpdateCallback): void;
```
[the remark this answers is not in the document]
atomicsulfate: It's call "update" because that's the verb the methods TileGeometryManager.updateTiles() and TileGeometryLoader.update() use. 

>  because the tile is loaded from disk
TileGeometryLoader takes a decoded tile, creates objects from its geometry and adds them to the tile. I don't see it loads anything from disk.

Maybe you thought this code was related to the **TileLoader** ?

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
```diff
@@ -98,6 +100,15 @@ export interface TileGeometryManager {
      * @returns {GeometryKindSet}
      */
     getAvailableKinds(tiles: Tile[]): GeometryKindSet;
+
+    /**
+     * Sets a callback that will be called for every updated tile on [[updateTiles]].
+     *
+     * @param {TileUpdateCallback} callback The callback that will be called after a tile has been
```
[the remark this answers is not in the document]
atomicsulfate: It's the type of the parameter in JSDoc. I'm following the convention I see in this file documentation, the other methods are documenting the parameter types.

@here/harp-omv-datasource/lib/OmvDataSource.ts
atomicsulfate: done.

@here/harp-omv-datasource/test/OmvDecodedTileEmitterTest.ts
atomicsulfate: done.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2087,6 +2088,28 @@ export class MapView extends THREE.EventDispatcher {
         this.m_elevationRangeSource = elevationRangeSource;
         this.m_elevationRangeSource.connect();
         this.m_elevationProvider = elevationProvider;
+        this.dataSources.forEach(dataSource => {
+            dataSource.setEnableElevationOverlay(true);
+        });
+        this.m_tileGeometryManager.setTileUpdateCallback((tile: Tile) => {
+            overlayOnElevation(tile);
+        });
+        this.clearTileCache();
```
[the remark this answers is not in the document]
atomicsulfate: The decoder configuration is also updated to generate the texture coordinates needed to sample the displacement map texture.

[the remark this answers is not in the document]
nzjony: Of course, ok makes sense.

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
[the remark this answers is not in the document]
nzjony: Now I understand, I thought this was called when the tile was finished loading, but now it makes sense.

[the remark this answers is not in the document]
nzjony: Yes `TileGeometryLoader` takes a decoder tile, but clearing the cache means that the Tile must be re-requested, i.e. the the `TileLoader` is recreated, which will pull the data from the `DataProvider`, hence reloading from disk (which is however likely incorrect, sorry for the potential confusion there, i guess it is in memory).

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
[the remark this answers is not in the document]
nzjony: Ok, I wasn't aware of this. It seems it isn't necessary, as typedoc seems to pull out the types automatically (at least when I run it locally), but it doesn't hurt to have it in, so you can leave it for consistency.

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { GeometryKind } from "@here/harp-datasource-protocol";
+import { DisplacementFeature, MapMeshBasicMaterial } from "@here/harp-materials";
+import THREE = require("three");
```
nzjony:
```suggestion
import * as THREE from "three";
```

atomicsulfate: done.

PR #702, opened 2019-08-08
HARP-6038: Always treat three.module.js dependencies as external

CraigBerlin:


@here/harp.gl/webpack.config.js
```diff
@@ -55,19 +55,29 @@ const mapComponentConfig = merge(commonConfig, {
...
     },
-    externals: {
-        three: "THREE"
-    }
+    externals: [
+        {
+            three: "THREE"
+        },
+        function(context, request, callback) {
+            return /three\.module\.js$/.test(request) ? callback(null, "THREE") : callback();
```
zbigg: My eyes bleed ...

Frankly i would prefer a coding rule mandating that we only import THREE as a whole (`import * as three from 'three'`) than forcing people to use magic like this in their guides and in their code.

Only because we can't force ourselves to be consistent and pragmatic in such a simple things like imports

CraigBerlin: Agreed this is ugly but the problem is in the threejs jsm loader code itself.
See [GLTFLoader.js](https://github.com/mrdoob/three.js/blob/master/examples/jsm/loaders/GLTFLoader.js):
`import ... from "../../../build/three.module.js"`.

If we do not force the external with webpack we end up with two copies of threejs at runtime, which breaks BufferGeometry and probably other things.
On the plus side this should prevent similar problems where threejs is not imported as you suggest.

robertoraggi: @CraigBerlin  good catch!

@zbigg The harp.gl webpack configurations mark the references to `"three"` as references to the global variable `THREE`. Unfortunately this breaks when using 3d party code that imports `"three.module.js"` directly (e.g. the loaders, the controls, and the helper code in `three/examples/jsm/**/*`). In this case webpack will bundle the imported code (in Craig's case GLTLoader.js) and a copy of three.module.js so the final app will end up using two different copies of three.js (one externally provided using THREE and one bundled using es2015 modules) causing all kind of issues. An alternative way to fix this problem is to avoid importing code from `three/external/jsm/` and use `three/external/js/ instead, *but* unfortunately only the `jsm` examples have correct typings.

zbigg: @CraigBerlin, @robertoraggi When third party code imports `three.module.js` they're screwed too, because they need to add this into _theirs_ `webpack.config.js`.

AFAIR, "core" `harp.gl` doesn't use `three.module.js` (yet) so this patch is not needed in this repo (yet ..., how long?).

This hack is needed only in apps which use both `THREE` and `three.module.js`.

Nevertheless, i would still skip that "technology" until it's stable and usable without workarounds like this, even if we gain a little on type safety.

Contrary to patch in this form i would more warmly welcome, a a note in `README` about possible problems with integrations with rest of 3js would be welcomed with this workaround documented.

robertoraggi: @zbigg in this repository we're not including three/examples/jsm's code but are including three/example/js resulting in a lot of untyped code. So as part of this change we could replace these imports to use their jsm equivalent module and get the benefit of having typed code.

```
@here/harp-examples/src/getting-started_free-camera.ts:import "three/examples/js/controls/TrackballControls";
@here/harp-examples/src/getting-started_free-camera.ts:import "three/examples/js/controls/TransformControls";
@here/harp-examples/src/threejs_animation.ts:import "three/examples/js/loaders/FBXLoader";
```

PR #703, opened 2019-08-08
MINOR: labels picking fixed.

pit-rpg:


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -484,6 +484,8 @@ export class TileGeometryCreator {
                 textElement.mayOverlap = technique.mayOverlap === true;
                 textElement.reserveSpace = technique.reserveSpace !== false;
                 textElement.kind = technique.kind;
+                // get the userData for text element picking
```
StefanDachwitz: Please start a sentence with uppercase, and end with a "." Thanks! Same next change below.

pit-rpg: Done.

PR #708, opened 2019-08-09
HARP-6763: SubTiles generate bad tile keys on custom subdivision schemes

okanishchev:

Fixed SubTiles to support custom subdivision schemes;
Force SubTiles to return an iterator and as result avoid using IIFE while iterating over SubTiles.

### DIFF
--- a/@here/harp-geoutils/lib/tiling/SubTiles.ts
+++ b/@here/harp-geoutils/lib/tiling/SubTiles.ts
@@ -7,78 +7,41 @@
 import { TileKey } from "./TileKey";
 
 export class SubTiles implements Iterable<TileKey> {
-    private m_tileKey: TileKey;
-    private m_level: number;
-    private m_count: number;
-    private m_mask: number;
-    private m_shift: number;
+    constructor(public tileKey: TileKey, public sizeX: number, public sizeY: number) {}
 
-    constructor(tileKey: TileKey, level: number, mask: number) {
-        this.m_tileKey = tileKey;
-        this.m_level = level;
-        // tslint:disable:no-bitwise
-        this.m_count = 1 << (level << 1);
-        this.m_mask = mask;
-        this.m_shift = level > 2 ? (level - 2) << 1 : 0;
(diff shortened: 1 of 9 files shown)


@here/harp-geoutils/lib/tiling/SubTiles.ts
```diff
@@ -7,78 +7,45 @@
...
-                // tslint:enableno-bitwise
-            );
+            return TileKey.fromRowColumnLevel(tk.row * dy + py, tk.column * dx + px, tk.level + 1);
         }
 
-        next() {
-            const current = { value: this.value, done: this.m_index >= this.m_totalSubTileCount };
-            this.m_index = this.m_parent.skip(++this.m_index);
-            return current;
+        next(): IteratorResult<TileKey> {
```
robertoraggi: do you think you can refactor this code to use `yield`?

@here/harp-geoutils/lib/tiling/SubTiles.ts
```diff
@@ -7,78 +7,45 @@
...
-        // tslint:enable:no-bitwise
-    }
-
-    get level(): number {
-        return this.m_level;
-    }
-    get tileKey(): TileKey {
-        return this.m_tileKey;
-    }
+    constructor(public tileKey: TileKey, public sizeX: number, public sizeY: number) {}
```
robertoraggi: SubTiles was really working only on a quad-tree, so yeah it is not working. Still, it would be nice if we could keep using for quad trees.

@here/harp-geoutils/lib/tiling/SubTiles.ts
okanishchev: thanks for the tip! the code is really simpler now.

@here/harp-geoutils/lib/tiling/SubTiles.ts
okanishchev: done. and i think, z-curve method would be a bit cheaper, as it uses just one loop and no multiplications.

PR #709, opened 2019-08-09
HARP-6597: Camera rotations in globe.

Astrak:


@here/harp-mapview/lib/Utils.ts
```diff
@@ -26,29 +26,30 @@ const groundNormalPlanarProj = new THREE.Vector3(0, 0, 1);
...
-    const pitchAxis = new THREE.Vector3(1, 0, 0);
+/**
+ * Cached ThreeJS instances for realtime maths.
+ */
+const space = {
+    x: new THREE.Vector3(),
+    y: new THREE.Vector3(),
+    z: new THREE.Vector3()
+};
+const tangentSpace = {
```
amsXYZ: Doesn't it make more sense to use [`Matrix3`](https://threejs.org/docs/#api/en/math/Matrix3) here?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -26,29 +26,30 @@ const groundNormalPlanarProj = new THREE.Vector3(0, 0, 1);
...
-
-    /**
-     * The pitch axis which we use to rotate around when we change the pitch.
-     * The axis is fix and is the +X axis `(1,0,0)`.
-     */
-    const pitchAxis = new THREE.Vector3(1, 0, 0);
+/**
+ * Cached ThreeJS instances for realtime maths.
+ */
+const space = {
```
amsXYZ: What does this actually describes/stores? It's really tricky to just say by its name...

@here/harp-mapview/lib/Utils.ts
```diff
@@ -146,22 +147,35 @@ export namespace MapViewUtils {
...
+        mapView.projection.projectPoint(targetCoordinates, vector);
+        const groundDistance = distance * Math.sin(pitchRad);
+        if (mapView.projection.type === geoUtils.ProjectionType.Planar) {
+            vector.set(
+                vector.x + Math.sin(yawRad) * groundDistance,
+                vector.y - Math.cos(yawRad) * groundDistance,
+                0
+            );
+        } else if (mapView.projection.type === geoUtils.ProjectionType.Spherical) {
+            // In globe yaw and pitch are understood to be in tangent space. The approach below is
```
amsXYZ: Would appreciate if this was broken down into line-by-line comments for clearer comprehension.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -146,22 +147,35 @@ export namespace MapViewUtils {
...
+            // to find the Z and Y tangent space axes, then rotate Y around Z by the given yaw, and
+            // set its new length (groundDistance). Finally the up vector's length is set to the
+            // camera height and added to the transformed Y above.
+            tangentSpace.z.copy(vector).normalize();
+            const obj = new THREE.Object3D();
```
amsXYZ: Why is an `Object3D` needed here? You can use a regular `Matrix4` to do the `lookAt` and the world direction can be easily extracted from that matrix like this https://github.com/mrdoob/three.js/blob/master/src/core/Object3D.js#L563

@here/harp-mapview/lib/Utils.ts
```diff
@@ -248,87 +258,114 @@ export namespace MapViewUtils {
...
      * @param pitch Pitch in degrees.
      */
     export function setRotation(mapView: MapView, yaw: number, pitch: number) {
-        yawQuaternion.setFromAxisAngle(yawAxis, geoUtils.MathUtils.degToRad(yaw));
-        pitchQuaternion.setFromAxisAngle(pitchAxis, geoUtils.MathUtils.degToRad(pitch));
-
-        yawQuaternion.multiply(pitchQuaternion);
-        mapView.camera.quaternion.copy(yawQuaternion);
-        mapView.camera.matrixWorldNeedsUpdate = true;
+        const transform = {
```
amsXYZ: Why is not `space` a `Matrix4`?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -248,87 +258,114 @@ export namespace MapViewUtils {
...
+            yAxis: space.y,
+            zAxis: space.z,
+            position: vector
+        };
+        mapView.projection.localTangentSpace(mapView.geoCenter, transform);
+
+        // `matrix2` thereafter is the transform matrix, `matrix1` the rotation matrix.
+        matrix2.makeBasis(transform.xAxis, transform.yAxis, transform.zAxis);
+        matrix2.setPosition(transform.position.x, transform.position.y, transform.position.z);
+        quaternion1.setFromAxisAngle(vector.set(0, 0, 1), THREE.Math.degToRad(yaw));
```
amsXYZ: `matrix1` can be set up as the multiplication of two simple rotation matrices (no need to create quaternion, multiply, and go back to rotation):

https://en.wikipedia.org/wiki/Rotation_matrix#Basic_rotations

@here/harp-mapview/lib/MapView.ts
```diff
@@ -180,7 +180,8 @@ const CONTEXT_LOST_EVENT: RenderEvent = { type: MapViewEventNames.ContextLost }
 const CONTEXT_RESTORED_EVENT: RenderEvent = { type: MapViewEventNames.ContextRestored } as any;
 const COPYRIGHT_CHANGED_EVENT: RenderEvent = { type: MapViewEventNames.CopyrightChanged } as any;
 
-const tmpVector = new THREE.Vector2();
+const vector2 = new THREE.Vector2();
```
amsXYZ: Why renaming? Also, `vector2` is quite generic and doesn't provide info about this being a temporary object.

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -403,45 +376,84 @@ export class MapControls extends THREE.EventDispatcher {
...
-        const currentAzimuthAltitude = utils.directionToAzimuthAltitude(
-            this.m_currentViewDirection
-        );
+        const pitchRad = MapViewUtils.extractYawPitchRoll(this.camera, this.mapView.projection.type)
+            .pitch;
+
+        // In sphere view, from afar, when the horizon is below the screen center (NDC of 0
+        // vertically), find the uppermost NDC so that the interaction does not break.
+        let verticalNDCToRaycast = 0;
+        if (this.mapView.projection.type === geoUtils.ProjectionType.Spherical) {
```
amsXYZ: Not sure if this is the most correct behaviour from the interaction point of view... 

Since earth at far distances don't cover the entire vFov, tilting from far away will not keep the tilt point in the center of the screen, and when it does, it might mean that that point fall into earth's horizon and is no longer visible.

This also gets worse whenever we introduce any of the acceleration-based algorithms into play, since the rotation in which earth will end up is not deterministic and can translate into inconsistent behaviour (specially when you use the `3D` mapUi button).

I'd stick to limiting the tilt based on the distance to the earth's surface, but that's just my opinionated view on the topic :)

@here/harp-mapview/lib/Utils.ts
Astrak: `space` and `tangentSpace` below are used to do some space transformation math when a tangent space is involved, computed through a new method `localTangentSpace` implemented a bit earlier in another PR. Unfortunately the module it belongs to cannot depend on ThreeJS yet, so it has to implement `Vector3Like` for the math used in it. We were not keen on having a `Matrix3Like`, since legal might agree on the ThreeJS dependency soon. It might not be necessary to explain these internal and momentary legal issues there, and let people with a definition of a space as 3 axes, in the `localTangentSpace` method. This is also straightforward to use in the associated code, I believe?

@here/harp-mapview/lib/Utils.ts
Astrak: Agree, though see other comment.

@here/harp-mapview/lib/Utils.ts
Astrak: True, unfortunately see other comment.

@here/harp-map-controls/lib/MapControls.ts
Astrak: If there is a target point in the center of the screen, then this method will not get the earth outside the center of the screen. This math is rather to bring it back.

This is to address the problem of very small zoom levels. The tilt uses to be enabled at up to 45 by default. Problem, if we tilt with the middle mouse button (camera tilt instead of target tilt), then the earth goes out of sight inevitably. Bringing it back with the middle mouse button is a possibility, but this code also allows to bring it back with right mouse button and the "3D" button, and also with the 3 fingers interaction on mobile. Since it's already there, I think it is a strong benefit?

I don't understand the part about acceleration-based algorithms?

@here/harp-mapview/lib/MapView.ts
Astrak: A `vector3` variable was also created just below, some others like `quaternion1`, `quaternion2`, `matrix1`, `matrix2`... in MapViewUtils in this PR...

Not sure that this is a worthwhile topic to debate, but here are my motivations if need be:
- "tmp" is an abbreviation, we use to avoid abbreviations, it should be "temporary",
- but "temporary" is very long,
- then it should actually be "cached", because these are _cached_, constant variables that we create to _avoid temporary objects_,
- but then I find it personally a bit of a pleonasm, their use is obviously the one of cached variables created to avoid temporary ones, when we see how they are created out of class scopes. There is also a comment telling it in all places except here, unfortunately.
- especially when one joins ordinal numbers to them.

In short I found it more readable, but I frankly don't care much. I'm open to drawbacks I didn't see.

kkostecki:
Adrien I have improved near plane distance calculus a little bit, so you won't see the sphere parts clipped out. Please rebase your commit on top of mine when you are close to finish or cherry-pick to test:
https://github.com/heremaps/harp.gl/pull/764/commits/5f5d58a75a0587eaaddad5beb39a21d0e8ad6524

@here/harp-map-controls/lib/MapControls.ts
Astrak: Update: after discussions with @ninok, the controls should prevent that the center of the screen points in the void (in sphere projection) at all.

PR #710, opened 2019-08-09
HARP-6551: Examples fixing and refinements.

Astrak:
This PR fixes a bunch of examples and makes the navigation panel always collapsible so that the example is fullscreen.


devallois:
For the 'heatmap' example: This is actually not a heatmap, but a choropleth. Please change the example name to "choropleth" Thank you!

devallois:
Regarding the choropleth example - Could you also add a legend like in the polygons example? Needs some context for color and extrusion attributes.

Astrak:

PR #711, opened 2019-08-12
HARP-6781: Add support for null object literals.

robertoraggi:


@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -137,7 +140,7 @@ export type BinaryOp = RelationalOp | EqualityOp;
 /**
  * @hidden
  */
-export type Value = undefined | boolean | number | string;
+export type Value = undefined | null | boolean | number | string;
```
amsgames94:
```suggestion
export type Value = null | boolean | number | string;
```
To have a bit stricter types, it would be nice to remove ``undefined`` from this union type and instead use ``Value | undefined`` wherever we need.

robertoraggi: Yes, we need to remove `undefined` from `Value`, but currently `Value` is used to represent both the the result of an evaluation and also to configure techniques. So a lot of code in the OMV and JSON decoder needs to change for us to remove undefined from Value.

PR #715, opened 2019-08-13
HARP-6781: Restrict the values manipilated by style expressions.

robertoraggi:
This change removes 'undefined' from the possible types that
the evaluation of expression can generate. That is, 'undefined'
values when evaluating expressions should be treated as errors.


@here/harp-datasource-protocol/lib/ExprEvaluator.ts
```diff
@@ -48,7 +48,10 @@ export class ExprEvaluatorContext {
     ) {}
 
     evaluate(expr: Expr | undefined) {
```
kkostecki: Maybe not related, but would be good to add return type (Result).

PR #720, opened 2019-08-14
HARP-6811: Improve map view fog calculation.

kkostecki:
MapViewFog is now tilt angle dependent and should work with any kind of
projection type. Moreover it allows to define both vertical (looking top-
down) and horizontal density. This values may be moved in future to the
Theme fog definition.
This change is required in order to support dynamic frustum (near/far
planes) changes, because old implementation is only far plane related
and does not support precise far plane setup just below the ground level.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
+++ b/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
@@ -10,6 +10,30 @@ import * as THREE from "three";
 
 const epsilon = 0.000001;
 
+/**
+ * Structure that holds near, far planes distances and maximum visibility range.
+ */
+export interface ViewRanges {
+    /**
+     * Distance from camera to near clipping plane along camera eye vector.
+     * @note This value is always positive and in camera space, should be bigger then zero.
+     */
+    near: number;
+    /**
+     * Far clipping plane distance in camera space, along its eye vector.
+     * @note Should be always positive and bigger then [[near]] plane distance.
+     */
+    far: number;
+    /**
+     * Maximum possible visibility range in camera space, regardless of camera orientation.
+     * @note Holds the maximum value that may be returned for [[far]] plane distance at
+     * certain camera position and orientation. You may use this value to calculate
+     * fog or other depth effects that are related to frustum (visibility), but should not
+     * change as dynamically as near/far plane distances on certain evaluators.
+     */
+    maximum: number;
+}
+
 export interface ClipPlanesEvaluator {
     /**
      * Compute near and far clipping planes distance.
@@ -24,7 +48,7 @@ export interface ClipPlanesEvaluator {
      * @note Camera clipping planes aren't automatically updated via #evaluateClipPlanes()
      * call, user should do it manually if needed.
      */
-    evaluateClipPlanes(camera: THREE.Camera, projection: Projection): { near: number; far: number };
+    evaluateClipPlanes(camera: THREE.Camera, projection: Projection): ViewRanges;
 }
 
 /**
@@ -63,10 +87,7 @@ export class InterpolatedClipPlanesEvaluator implements ClipPlanesEvaluator {
         this.farMin = nearMin * nearFarMultiplier + farOffset;
     }
 
-    evaluateClipPlanes(
-        camera: THREE.Camera,
-        projection: Projection
-    ): { near: number; far: number } {
+    evaluateClipPlanes(camera: THREE.Camera, projection: Projection): ViewRanges {
         let nearPlane: number = this.nearMin;
         let farPlane: number = this.farMin;
         if (projection.type === ProjectionType.Spherical) {
@@ -93,13 +114,17 @@ export class InterpolatedClipPlanesEvaluator implements ClipPlanesEvaluator {
             farPlane = p.sub(camera.position).dot(fwd);
             const bias = 2000; // TODO: generalize.
             nearPlane = Math.max(this.nearMin, projection.groundDistance(camera.position) - bias);
-        } else {
+        } else if (projection.type === ProjectionType.Planar) {
             const groundDistance = projection.groundDistance(camera.position);
             nearPlane = Math.max(this.nearMin, groundDistance * this.nearMultiplier);
             // Will be already clamped to minFar due to clamping above.
             farPlane = nearPlane * this.nearFarMultiplier + this.farOffset;
+        } else {
+            assert(false, "Unsuported projection type");
         }
-        return { near: nearPlane, far: farPlane };
+
+        const viewRanges: ViewRanges = { near: nearPlane, far: farPlane, maximum: farPlane };
+        return viewRanges;
     }
 }
 
@@ -133,6 +158,7 @@ export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
     protected m_tmpQuaternion: THREE.Quaternion = new THREE.Quaternion();
     private m_maxElevation: number;
     private m_minElevation: number;
+    private m_minimumViewRange: ViewRanges;
 
     /**
      * Allows to setup near/far offsets (margins), rendered geometry elevation relative to sea
@@ -169,7 +195,7 @@ export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
         minElevation: number = -50,
         readonly nearMin: number = 0.1,
         readonly nearFarMargin: number = 1.0,
-        readonly farMaxRatio = 8.0
+        readonly farMaxRatio = 1.8
     ) {
         assert(maxElevation >= minElevation);
         assert(nearMin > 0);
@@ -177,6 +203,11 @@ export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
         assert(farMaxRatio > 1.0);
         this.m_minElevation = minElevation;
         this.m_maxElevation = maxElevation;
+        this.m_minimumViewRange = {
+            near: nearMin,
+            far: nearMin + nearFarMargin,
+            maximum: Math.max(nearMin * farMaxRatio, nearMin + nearFarMargin)
+        };
     }
 
     /**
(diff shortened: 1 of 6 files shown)


@here/harp-mapview/lib/Utils.ts
```diff
@@ -330,6 +333,53 @@ export namespace MapViewUtils {
...
+     * @param position The position above the ground for which normal is calculated.
+     * @param projection The projection used to transform geo coordinates to world space.
+     */
+    function getGroundNormalInv(
+        position: THREE.Vector3,
+        projection: geoUtils.Projection
+    ): THREE.Vector3 {
+        // Position on the ground for surface normal calculation does not matter for
+        // any planar projections, so we may simplify calculus by returning const vector:
+        if (projection.type !== geoUtils.ProjectionType.Spherical) {
```
ninok: == Planar is safer if new projection types are added later

@here/harp-mapview/lib/Utils.ts
```diff
@@ -330,6 +333,53 @@ export namespace MapViewUtils {
...
+     * @returns angle in radians.
+     */
+    export function getCameraTiltAngle(
+        camera: THREE.Camera,
+        projection: geoUtils.Projection
+    ): number {
+        // Note using different temporary vector then internal function used.
+        const lookAt: THREE.Vector3 = tmpVectors[0];
+        camera.getWorldDirection(lookAt);
+        return lookAt.angleTo(getGroundNormalInv(camera.position, projection));
```
Astrak: Many examples do not work because `lookAt` is undefined

@here/harp-mapview/lib/Utils.ts
Astrak: Yes please, use if planar, if spherical, or a switch statement if typescript complains that not all code paths return a Vector3

@here/harp-mapview/lib/Utils.ts
```diff
@@ -330,6 +333,53 @@ export namespace MapViewUtils {
...
+    /**
+     * Return normal to the ground surface directly above camera position.
+     *
+     * @note Function is not exposed cause it returns internal temporary vectors that
+     * can not be modified. Use with care and clone the result if required.
+     *
+     * @param position The position above the ground for which normal is calculated.
+     * @param projection The projection used to transform geo coordinates to world space.
+     */
+    function getGroundNormalInv(
```
Astrak: Does this need a function? The normal is 
- the Z axis in planar, 
- the normalized position vector in sphere.

@here/harp-mapview/lib/Utils.ts
kkostecki: Yep, good point! Will change it.

@here/harp-mapview/lib/MapViewFog.ts
```diff
@@ -5,25 +5,16 @@
...
-    private m_fogIsDefined: boolean = false;
-    private m_cachedTheme: Theme = { styles: {} };
-
-    /**
-     * Constructs a `MapViewFog` instance.
-     *
-     * @param m_scene The scene used in [[MapView]] that contains the map objects.
-     */
-    constructor(private m_scene: THREE.Scene) {}
-
```
okanishchev: Why moving `enabled` getter/setter before plain properties and even before constructor?

kkostecki: Auto formatter has done this ;(

@here/harp-mapview/lib/Utils.ts
kkostecki: Changed to === Planar, otherwise take Projection.surfaceNormal implementation, which may be more computationally expensive, but always correct - regardless of projection type (even assuming that new projection types may be implemented).

@here/harp-mapview/lib/Utils.ts
kkostecki: True, but this gives us some level of abstraction and allows to be compatible with new projection types if added in the future.

@here/harp-mapview/lib/MapViewFog.ts
kkostecki: Will fix this.

@here/harp-mapview/lib/Utils.ts
kkostecki: Fixed.

kkostecki:
> Fog not visible in the TryMe anymore?

Fixed now.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -32,6 +34,7 @@ const yawQuaternion = new THREE.Quaternion();
 const pitchQuaternion = new THREE.Quaternion();
 const tmpQuaternion = new THREE.Quaternion();
 const tmpMatrix = new THREE.Matrix4();
+const tmpVectors = new Array<THREE.Vector3>(3).fill(new THREE.Vector3());
```
okanishchev: It's wrong, you're filled whole array with one vector instance

kkostecki: Hmm... looks so, but why it's still working, that is kind of mystery... especially that I am using two array elements at the same time :).
Will fix it though...

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -271,16 +306,21 @@ export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
...
+        farPlane = Math.min(farPlane, farMax);
         farPlane = Math.max(farPlane, nearPlane + this.nearFarMargin);
 
-        return { near: nearPlane, far: farPlane };
+        const viewRanges: ViewRanges = {
+            near: nearPlane,
+            far: farPlane,
+            maximum: Math.max(farMax, farPlane)
+        };
+        return viewRanges;
```
okanishchev: and here

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -372,7 +412,8 @@ export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
         farPlane = Math.max(farPlane, cameraAltitude - this.minElevation);
         farPlane = Math.max(farPlane, nearPlane + this.nearFarMargin);
 
-        return { near: nearPlane, far: farPlane };
+        const viewRanges: ViewRanges = { near: nearPlane, far: farPlane, maximum: farPlane };
+        return viewRanges;
```
okanishchev: and here

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -709,13 +759,15 @@ export class FixedClipPlanesEvaluator implements ClipPlanesEvaluator {
...
+    evaluateClipPlanes(camera: THREE.Camera, projection: Projection): ViewRanges {
         // We do not need to perform actual evaluation cause results are precomputed and
         // kept stable until somebody changes the properties.
-        return { near: this.m_nearPlane, far: this.m_farPlane };
+        const viewRanges: ViewRanges = {
+            near: this.m_nearPlane,
+            far: this.m_farPlane,
+            maximum: this.m_farPlane
+        };
+        return viewRanges;
```
okanishchev: and here

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -93,13 +114,17 @@ export class InterpolatedClipPlanesEvaluator implements ClipPlanesEvaluator {
...
             const groundDistance = projection.groundDistance(camera.position);
             nearPlane = Math.max(this.nearMin, groundDistance * this.nearMultiplier);
             // Will be already clamped to minFar due to clamping above.
             farPlane = nearPlane * this.nearFarMultiplier + this.farOffset;
+        } else {
+            assert(false, "Unsuported projection type");
         }
-        return { near: nearPlane, far: farPlane };
+
+        const viewRanges: ViewRanges = { near: nearPlane, far: farPlane, maximum: farPlane };
```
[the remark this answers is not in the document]
kkostecki: Of course, but I prefer clarity over shortening, if it hurts you so much I can change it, but please take a look how many dependent commits will need to be rebased for that.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: We may do it in some other commit, as above it's not so crucial but will make a real rebasing mess (many dependent PR & commits).

kkostecki:

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: Will do it in next commit.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: Will do in the next commit.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -330,6 +333,53 @@ export namespace MapViewUtils {
...
+     *
+     * Tilt is the angle between camera __look at__ (forward) vector and ground normal.
+     *
+     * @note It may depend on the projection type.
+     *
+     * @param camera The camera used for tilt angle calculation.
+     * @param projection The projection used to convert geo coordinated to world coordinates.
+     * @returns angle in radians.
+     */
+    export function getCameraTiltAngle(
```
ninok: Why not use MapViewUtils.extractYawPitchRoll?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -330,6 +333,53 @@ export namespace MapViewUtils {
...
+     * can not be modified. Use with care and clone the result if required.
+     *
+     * @param position The position above the ground for which normal is calculated.
+     * @param projection The projection used to transform geo coordinates to world space.
+     */
+    function getGroundNormalInv(
+        position: THREE.Vector3,
+        projection: geoUtils.Projection
+    ): THREE.Vector3 {
+        // Position on the ground for surface normal calculation does not matter for
```
ninok: This statement is not true. For example is the normal of the mercatorProjection (0,0,1) whereas the normal of the webMercatorProjection is (0,0,-1). Despite the fact that this is quite confusing I would rather always take the code path below.

kkostecki: Ok, good to know!

kkostecki: Will be fixed in follow up commit.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -10,6 +10,30 @@ import * as THREE from "three";
...
+     * @note This value is always positive and in camera space, should be bigger then zero.
+     */
+    near: number;
+    /**
+     * Far clipping plane distance in camera space, along its eye vector.
+     * @note Should be always positive and bigger then [[near]] plane distance.
+     */
+    far: number;
+    /**
+     * Maximum possible visibility range in camera space, regardless of camera orientation.
```
nzjony: I'm still not understanding what maximum means based on this description.

`Holds the maximum value that may be returned for [[far]] plane distance at certain camera position and orientation`

Am I correct in saying that this value acts to `override` the far plane in certain circumstances?

Why does this have to be part of the ClipPlanesEvaluator? To me this is unnecessarily complicating this interface, by forcing users to return some value which they have to experiment with so that they don't break the fog. Couldn't the fog calculation been smart enough to take the tilt into consideration, and thus not complicate the interface?

Or am I missing some detail?

PR #722, opened 2019-08-14
HARP-6605: Added MapView map wrapping param support.

bohdanhorpynchuk:
Signed-off-by: bohdanhorpynchuk <ext-bogdan.gorpynchuk@here.com>


ninok:
@heremaps-bot ok to test

PR #724, opened 2019-08-15
HARP-6636: MapView.FrameComplete event is broken

StefanDachwitz:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2570,8 +2570,11 @@ export class MapView extends THREE.EventDispatcher {
                 this.m_tileDataSources.length &&
             !this.m_updatePending &&
             !this.animating &&
+            !this.cameraIsMoving &&
+            !this.m_animatedExtrusionHandler.isAnimating &&
             this.m_textElementsRenderer !== undefined &&
-            !this.m_textElementsRenderer.loading
+            !this.m_textElementsRenderer.loading &&
+            this.m_poiTableManager.finishedLoading
         ) {
             this.m_firstFrameComplete = true;
 
@@ -2661,7 +2664,7 @@ export class MapView extends THREE.EventDispatcher {
     private finishRenderTextElements() {
         const canRenderTextElements = this.m_pointOfView === undefined;
 
-        if (canRenderTextElements && this.m_textElementsRenderer) {
+        if (canRenderTextElements && this.m_textElementsRenderer !== undefined) {
             // copy far value from scene camera, as the distance to the POIs matter now.
             this.m_screenCamera.far = this.m_camera.far;
             this.m_textElementsRenderer.renderText(this.m_screenCamera);
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -759,12 +759,12 @@ export class Tile implements CachedResource {
     }
 
     /**
-     * Overrides the default value for [[hasGeometry]].
+     * Overrides the default value for [[hasGeometry]] if value is not `undefined`.
      *
      * @param value A new value for the [[hasGeometry]] flag.
      */
-    forceHasGeometry(value: boolean) {
-        if (value) {
+    forceHasGeometry(value: boolean | undefined) {
+        if (value === true) {
             this.setDecodedTile(defaultEmptyDecodedTile);
         }
         this.m_forceHasGeometry = value;
--- a/@here/harp-mapview/lib/geometry/PhasedTileGeometryLoader.ts
+++ b/@here/harp-mapview/lib/geometry/PhasedTileGeometryLoader.ts
@@ -151,7 +151,7 @@ export class PhasedTileGeometryLoader implements TileGeometryLoader {
      *      enabled.
      * @param {(GeometryKindSet | undefined)} disabledKinds The [[GeometryKind]]s that should be
      *      disabled.
-     * @returns {boolean} `true` if actual geometry has been created.
+     * @returns {boolean} `true` if `updateToPhase` was successful.
      */
     updateToPhase(
         toPhase: number,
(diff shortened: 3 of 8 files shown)


@here/harp-mapview/lib/geometry/PhasedTileGeometryManager.ts
```diff
@@ -52,15 +52,13 @@ export class PhasedTileGeometryManager extends TileGeometryManagerBase {
...
-        }
+        let needUpdate = this.mapView.isDynamicFrame
+            ? this.updateSomeTiles(tiles)
+            : this.updateAllTilesTogether(tiles);
 
-        this.updateTileObjectVisibility(tiles);
+        needUpdate = needUpdate || this.updateTileObjectVisibility(tiles);
 
-        if (!this.checkTilesFinished(tiles)) {
+        if (needUpdate || !this.checkTilesFinished(tiles)) {
```
okanishchev:
```suggestion
        needUpdate = needUpdate || !this.checkTilesFinished(tiles);
        
        if (needUpdate) {
```

StefanDachwitz: Unfortunately, 
needUpdate = needUpdate || !this.checkTilesFinished(tiles);
is just as wrong as my line
if (needUpdate || !this.checkTilesFinished(tiles)) {
since checkTilesFinished() has to be called anyways.

Your suggestion makes sense, but is is a bit of a matter of personal taste. I am sticking with the shorter version, since there are only a few expressions, and readability is "good enough". It also saves a store operation (... which may be optimized away by the compiler, I know...)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2554,8 +2554,12 @@ export class MapView extends THREE.EventDispatcher {
                 this.m_tileDataSources.length &&
             !this.m_updatePending &&
             !this.animating &&
+            !this.cameraIsMoving &&
+            !this.m_animatedExtrusionHandler.isAnimating &&
             this.m_textElementsRenderer !== undefined &&
```
ninok: If ` this.m_textElementsRenderer === undefined` the event will never be triggered. Is this expected behavior (same for poiTableManager)?

@here/harp-mapview/lib/Tile.ts
```diff
@@ -759,12 +759,12 @@ export class Tile implements CachedResource {
...
 
     /**
-     * Overrides the default value for [[hasGeometry]].
+     * Overrides the default value for [[hasGeometry]] if value is not `undefined`.
      *
      * @param value A new value for the [[hasGeometry]] flag.
      */
-    forceHasGeometry(value: boolean) {
-        if (value) {
+    forceHasGeometry(value: boolean | undefined) {
```
ninok: a call to `Tile.forceHasGeometry()` will not force the tile to have geometry. This is a bit weird. Shouldn't undefined be handled like true?

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: m_poiTableManager is created properly in the constructor, so be check for undefined is superfluous (fixing that).
m_textElementsRenderer may be undefined, but _will_always_ be created during rendering. The test for "defined" is to make the compiler happy.

@here/harp-mapview/lib/geometry/PhasedTileGeometryManager.ts
okanishchev: so, it would be
```suggestion
        needUpdate = needUpdate | !this.checkTilesFinished(tiles);
        
        if (needUpdate) {
```

note the |

@here/harp-mapview/lib/Tile.ts
StefanDachwitz: Somebody made forceHasGeometry() pretty complex, such that it uses some empty DecodedTile, please go look and see Mr. Architect ;)
I added undefined to make it possible to properly reset the value to "look into the actual geometry (tile.objects[]) and see if there is something there", this is new and does no harm.

@here/harp-mapview/lib/geometry/PhasedTileGeometryManager.ts
StefanDachwitz: In the "typeless" world of javascript, this makes sense, and would work fine. Except that, for a C++ developer, it smells like a hack, since boolean !== number, and binary operators are not defined for booleans, or are they? 
"Long live static types!", and "No more truthies"!  8^)

@here/harp-mapview/lib/geometry/PhasedTileGeometryLoader.ts
```diff
@@ -161,10 +161,10 @@ export class PhasedTileGeometryLoader implements TileGeometryLoader {
         let didUpdate = false;
         toPhase = Math.min(toPhase, this.numberOfPhases);
         while (this.currentPhase < toPhase) {
-            didUpdate = this.update(enabledKinds, disabledKinds);
-            if (!didUpdate) {
+            if (!this.update(enabledKinds, disabledKinds)) {
                 break;
             }
+            didUpdate = true;
```
CraigBerlin: Please update the api comment that this will now return true if only one phase update was successful, we may not have actually reached "toPhase".

@here/harp-mapview/lib/geometry/PhasedTileGeometryManager.ts
okanishchev: Another try
```suggestion
        if (!this.checkTilesFinished(tiles)) {
                needUpdate = true;
        }
        
        if (needUpdate) {
```

okanishchev: If you look on booleans like on 1-bit clamped integers, then all integer operations work just fine. But ok, no more truthies.

PR #726, opened 2019-08-15
HARP-5728: Move displacement map texture creation to TileGeometryCreator.

atomicsulfate:

### DIFF
--- a/@here/harp-mapview/lib/DisplacementMap.ts
+++ b/@here/harp-mapview/lib/DisplacementMap.ts
@@ -6,14 +6,14 @@
 
 import { TileKey } from "@here/harp-geoutils";
 
-export interface DisplacementMapTexture {
-    texture: Float32Array;
-    width: number;
-    height: number;
+export interface DisplacementMap {
+    xCountVertices: number;
+    yCountVertices: number;
+    buffer: Float32Array;
 }
 
-export interface DisplacementMap {
+export interface TileDisplacementMap {
     tileKey: TileKey;
-
-    texture: DisplacementMapTexture;
+    texture: THREE.DataTexture;
+    displacementMap: DisplacementMap;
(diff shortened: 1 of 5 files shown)


atomicsulfate:

@here/harp-mapview/lib/DisplacementMap.ts
```diff
@@ -6,14 +6,14 @@
...
 import { TileKey } from "@here/harp-geoutils";
 
-export interface DisplacementMapTexture {
-    texture: Float32Array;
-    width: number;
-    height: number;
+export interface DisplacementMap {
+    xCountVertices: number;
+    yCountVertices: number;
+    buffer: Float32Array;
```
ninok: Why do we need this buffer? Couldn't we use the buffer from the DataTexture?

atomicsulfate: That DisplacementMap  is passed from the decoder and used in the TileGeometryCreator to create the DataTexture. It looks like we can't create the DataTexture object directly in the decoder because the structured cloning of the objects passed back from the web workers can only handle plain structures.

TerrainElevationProvider does keep this DisplacementMap in the returned TileDisplacementMap because the original array is not stored in the DataTexture.
DataTexture only keeps the underlying raw buffer corresponding to the Float32Array here in
texture.image.data.buffer, which is of type ArrayBuffer. The only way to access to the original floats is by recreating a Float32Array using that buffer:
new Float32Array(texture.image.data.buffer).

So there's two options: either will store the original array with the terrain tiles or we recreate the array object everytime someone calls getDisplacementMap(). I went for the first option to avoid recreating these array objects. The additional memory overhead should be small because the underlying buffer is not copied, TypedArrays are just views of ArrayBuffers.

PR #732, opened 2019-08-19
HARP-6718: Overlay lines on terrain.

atomicsulfate:
- Add texture coordinates and normals to line geometry.
- Change line shader to use displacement map.

### DIFF
--- a/@here/harp-lines/lib/Lines.ts
+++ b/@here/harp-lines/lib/Lines.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { assert } from "@here/harp-utils";
 import * as THREE from "three";
 
 // Preallocate temp variables used during line generation.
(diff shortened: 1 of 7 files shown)


@here/harp-lines/lib/Lines.ts
```diff
@@ -27,29 +28,33 @@ interface VertexAttributeDescriptor {
  * Declares all the vertex attributes used for rendering a line using the [[SolidLineMaterial]].
  */
 const LINE_VERTEX_ATTRIBUTES: VertexAttributeDescriptor[] = [
-    { name: "texcoord", itemSize: 2, offset: 0 },
+    { name: "extrusionCoord", itemSize: 2, offset: 0 },
     { name: "position", itemSize: 3, offset: 2 },
     { name: "tangent", itemSize: 3, offset: 5 },
-    { name: "bitangent", itemSize: 4, offset: 8 }
+    { name: "bitangent", itemSize: 4, offset: 8 },
+    { name: "uv", itemSize: 2, offset: 12 },
```
ninok: we could squeeze uv and normal into one vec4 but maybe we can do this later b/c it will make the code harder to read.
What we should do right now is make the texture coordinates and normal optional and only add them if we have terrain.

atomicsulfate: done (making uv and normal optional).

PR #733, opened 2019-08-19
HARP-5144: Unit tests for picking

pit-rpg:


@here/harp-omv-datasource/test/MapViewPickingTest.ts
```diff
@@ -0,0 +1,326 @@
...
+        assert.notEqual(decodeTile, undefined);
+
+        assert.notEqual(decodeTile.textGeometries, undefined);
+        assert.notEqual(decodeTile.geometries, undefined);
+        assert.equal(decodeTile.geometries.length, 2);
+
+        const tileInfo = decodeTile.tileInfo as ExtendedTileInfo;
+        assert.notEqual(tileInfo, undefined);
+
+        describe("Test decodeTile, tileInfo data", () => {
```
okanishchev: There is no output for the test cases in this scope. It's not clear whether do they run or not.

@here/harp-omv-datasource/test/MapViewPickingTest.ts
```diff
@@ -0,0 +1,326 @@
...
+        const POINT = GEOJSON_DATA.features[3];
+        const pointCoordinates = ((POINT.geometry as any).coordinates as any) as number[];
+
+        const point = new GeoCoordinates(pointCoordinates[1], pointCoordinates[0]);
+        await displayLocation(mapView, point);
+
+        const screenPointLocation = mapView.getScreenPosition(point) as THREE.Vector2;
+        assert.notEqual(screenPointLocation, undefined);
+
+        mapView.worldRootObject.children.forEach(object => {
```
okanishchev: This is redundant. You can do `updateMatrixWorld(true)` only on top-level object, it's recursively updates all children, and the `true` is to `updateMatrix()` inside

okanishchev: You can check the source https://github.com/mrdoob/three.js/blob/dev/src/core/Object3D.js#L561

@here/harp-omv-datasource/package.json
```diff
@@ -29,6 +29,7 @@
     "license": "Apache-2.0",
     "dependencies": {
         "@here/harp-datasource-protocol": "^0.10.0",
+        "@here/harp-geojson-datasource": "^0.10.0",
```
zbigg: This should be in `devDependencies`...

@here/harp-omv-datasource/test/MapViewPickingTest.ts
```diff
@@ -0,0 +1,334 @@
...
+import * as sinon from "sinon";
+import * as THREE from "three";
+import { OmvTileDecoder } from "../index-worker";
+import { OmvDataSource } from "../lib/OmvDataSource";
+import { OmvTile } from "../lib/OmvTile";
+import { GEOJSON_DATA, THEME } from "./resources/geoJsonData";
+
+declare const global: any;
+
+// repeat a guard function until guard return true or repeat count <= 0
```
zbigg: It's very useful, please add it to `@here/harp-test-utils`.

This may be integrated with `lastWaiterError` defined there, so in case of `timeouts` you can display something useful like this:

"Failed waiting until `${fn.toString}` is true."

BTW, i would remove `repeat`
This is something like hardcoded timeout, and for this we have `this.timeout()` in mocha.

@here/harp-omv-datasource/test/resources/geoJsonData.ts
```diff
@@ -0,0 +1,135 @@
...
+    ]
+};
+
+export const THEME: Theme = {
+    defaultTextStyle: {
+        name: "defaultTextStyle",
+        color: "#6D7477",
+        fontCatalogName: "fira"
+    },
+    textStyles: [
```
zbigg: Not needed i guess ...

@here/harp-omv-datasource/test/MapViewPickingTest.ts
pit-rpg: I must leave the repeat counter because of node v8  event loop.
While some interval events exist in the node event loop, the node process won't stop. I need to clear the intervals to normally exit from the process.

@here/harp-omv-datasource/test/resources/geoJsonData.ts
pit-rpg: thanks, yes I cleaned up, but the empty array is required, maybe it's an issue?

@here/harp-omv-datasource/test/MapViewPickingTest.ts
zbigg: So this, is only hiding problem under carpet.
The real problem is that your code still works after mocha decides to `timeout` and abort test ...
Again, look at `willEventually` how it's done - there is `afterEach` handler added, that ensures that `willEventually` will abort if mocha done with current test.

BTW, Your use case can be implemented with
```
await willEventually(() => {
        assert.isTrue(mapView.visibleTileSet.allVisibleTilesLoaded);
});
```
so 1) you reuse code, 2) you have clear info when you hit timeout due to condition never ocurred 3) gain proper cleanup (already worked out in pain for `willEventually` and friends).

pit-rpg: Done.

PR #734, opened 2019-08-20
HARP-5144: added picking examples for polygons, lines, labels, POIs

pit-rpg:
HARP-5144: added picking examples for polygons, lines, labels, POIs
HARP-7129:  fix DataCloneError

### DIFF
--- a/@here/harp-datasource-protocol/lib/TileInfo.ts
+++ b/@here/harp-datasource-protocol/lib/TileInfo.ts
@@ -7,6 +7,7 @@ import { TileKey } from "@here/harp-geoutils";
 import { assert } from "@here/harp-utils";
 
 import { Env, MapEnv, Value } from "./Expr";
+import { makeDecodedTechnique } from "./StyleSetEvaluator";
 import { AttrEvaluationContext, evaluateTechniqueAttr } from "./TechniqueAttr";
 import {
     IndexedTechnique,
@@ -536,26 +537,12 @@ export class ExtendedTileInfoWriter {
             return infoTileTechniqueIndex;
         }
 
+        const decodedTechnique = makeDecodedTechnique(technique);
+
         infoTileTechniqueIndex = this.tileInfo.techniqueCatalog.length;
+        this.techniqueIndexMap.set(decodedTechnique._index, infoTileTechniqueIndex);
+        this.tileInfo.techniqueCatalog.push(decodedTechnique);
 
-        // add a new technique. Select the subset of features that should be stored (e.g., _index is
-        // not)
-        const storedTechnique = {} as any;
-
-        Object.getOwnPropertyNames(technique).forEach(property => {
-            if (!property.startsWith("_")) {
-                storedTechnique[property] = (technique as any)[property];
-            }
-        });
-
-        // Keep the index to identify the original technique later.
-        storedTechnique._index = technique._index;
-        storedTechnique._key = technique._key;
-        storedTechnique._styleSetIndex = technique._styleSetIndex;
-
-        this.techniqueIndexMap.set(technique._index, infoTileTechniqueIndex);
-
-        this.tileInfo.techniqueCatalog.push(storedTechnique);
         return infoTileTechniqueIndex;
     }
 
--- a/@here/harp-examples/src/datasource_object-picking.ts
+++ b/@here/harp-examples/src/datasource_object-picking.ts
@@ -1,0 +1,146 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { MapControls, MapControlsUI } from "@here/harp-map-controls";
+import { CopyrightElementHandler, CopyrightInfo, MapView, PickResult } from "@here/harp-mapview";
+import { APIFormat, OmvDataSource } from "@here/harp-omv-datasource";
+import { accessToken } from "../config";
+
+/**
+ * This example showcases how picking works.
+ *
+ * To enable line picking set `enableRoadPicking: true` in [[MapViewOptions]] and set
+ * `createTileInfo: true` in [[OmvWithRestClientParams]] or in [[OmvWithCustomDataProvider]].
+ * To enable polygon picking set `gatherFeatureIds: true` in
+ * [[OmvWithRestClientParams]] or in [[OmvWithCustomDataProvider]].
+ * To enable text element picking set `gatherFeatureIds: true` in
+ * [[OmvWithRestClientParams]] or in [[OmvWithCustomDataProvider]].
+ *
+ * Now, let's write an event that fires when the user clicks the map canvas:
+ * ```typescript
+ * [[include:datasource_object_picking_1.ts]]
+ * ```
+ *
+ * All the data handling is covered by the `handlePick` function. Here we find the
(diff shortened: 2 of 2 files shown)


@here/harp-examples/src/datasource_object-picking.ts
```diff
@@ -0,0 +1,141 @@
...
+ * For enable line picking set `enableRoadPicking: true` in [[MapViewOptions]] and set
+ * `createTileInfo: true` in [[OmvWithRestClientParams|OmvWithCustomDataProvider]].
+ * For enable polygon picking set `gatherFeatureIds: true` in
+ * [[OmvWithRestClientParams|OmvWithCustomDataProvider]].
+ * For enable text element picking set `gatherFeatureIds: true` in
+ * [[OmvWithRestClientParams|OmvWithCustomDataProvider]].
+ *
+ * Now, let's write an event that fires when the user clicks the map canvas:
+ * ```typescript
+ * [[include:datasource_geojson_picking1.ts]]
```
musculman: the name of the file is different, I would align it.

pit-rpg: done

Astrak:

@here/harp-examples/src/datasource_object-picking.ts
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        // Get userData from the first result;
+        currentUserData = usableIntersections[0].userData;
+
+        // Show helper box
+        element.style.visibility = "visible";
+
+        // Display userData inside of helper box
+        element.innerHTML = JSON.stringify(currentUserData, undefined, 2);
```
okanishchev: Non-html content should be stored as text
```suggestion
        element.textContent = JSON.stringify(currentUserData, undefined, 2);
```

pit-rpg: Done

@here/harp-examples/src/datasource_object-picking.ts
```diff
@@ -0,0 +1,144 @@
...
+
+import { MapControls, MapControlsUI } from "@here/harp-map-controls";
+import { CopyrightElementHandler, MapView, PickResult } from "@here/harp-mapview";
+import { APIFormat, OmvDataSource } from "@here/harp-omv-datasource";
+import { accessToken } from "../config";
+
+/**
+ * This example showcases how picking works.
+ *
+ * For enable line picking set `enableRoadPicking: true` in [[MapViewOptions]] and set
```
amsXYZ:
```suggestion
 * To enable line picking set `enableRoadPicking: true` in [[MapViewOptions]] and set
```

amsXYZ: And same for the 2 sentences below.

@here/harp-examples/src/datasource_object-picking.ts
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * All the data handling is covered by the `handlePick` function. Here we find the
+ * intersected objects, pick the first one in the array and display its
+ * data inside the helper box
+ * ```typescript
+ * [[include:datasource_object_picking_2.ts]]
+ * ```
+ */
+
+export namespace GeoJsonPickingExample {
```
amsXYZ: Maybe rename the namespace? This example doesn't have anything to do with GeoJson, and this is how it'll be displayed inside the documentation.

@here/harp-examples/src/datasource_object-picking.ts
```diff
@@ -0,0 +1,144 @@
...
+        });
+
+        // snippet:datasource_object_picking_1.ts
+        canvas.addEventListener("mousedown", event => {
+            handlePick(mapView, event.pageX, event.pageY);
+        });
+        // end:datasource_object_picking_1.ts
+
+        const omvDataSource = new OmvDataSource({
+            baseUrl: "https://xyz.api.here.com/tiles/osmbase/512/all",
```
ninok: Why not HERE VectorTiles (i.e. `https://xyz.api.here.com/tiles/herebase.02`)?

@here/harp-examples/src/datasource_object-picking.ts
pit-rpg: Done

@here/harp-examples/src/datasource_object-picking.ts
pit-rpg: Done

@here/harp-examples/src/datasource_object-picking.ts
pit-rpg: I only used a piece of code from other examples, but thanks for the hint using this data source found a bug HARP-7129.

@here/harp-examples/src/datasource_object-picking.ts
```diff
@@ -0,0 +1,145 @@
...
+            canvas,
+            theme: "resources/berlin_tilezen_base.json",
+            enableRoadPicking: true
+        });
+
+        CopyrightElementHandler.install("copyrightNotice")
+            .attach(mapView)
+            .setDefaults([
+                {
+                    id: "openstreetmap.org",
```
ninok: Now the copyright is not matching anymore.

@here/harp-datasource-protocol/lib/TileInfo.ts
```diff
@@ -543,6 +543,12 @@ export class ExtendedTileInfoWriter {
         const storedTechnique = {} as any;
 
         Object.getOwnPropertyNames(technique).forEach(property => {
+            if ((technique as any)[property] instanceof CallExpr) {
```
zbigg: This is wrong, 

1) For technique params dependent on feature, it freezez some random value that is not necessarily important for technique in general, see `AttrScope` enum
2) For technique params that are same for all features, they may be sent to main thread, but in that case you shouldn't evaluate them, but send them after serializing `Expr.toJSON` form but ...

It's not yet supported on main thread side, so it's better to just skip them ATM and think about fixing it in separate issue.

So, in general you should use `makeDecodedTechnique` from `StyleSetEvaluator`. to convert from technique used in decoder to technique variant that may be sent to main thread.

pit-rpg: Thanks, I fixed this. A block of code also was mowed to prevent unnecessarily iteration on object properties. Do we really need this code:
```
            if (property.startsWith("_")) {
                continue;
            }
 ...
        storedTechnique._index = technique._index;
        storedTechnique._key = technique._key;
        storedTechnique._styleSetIndex = technique._styleSetIndex;

```

zbigg: > Do we really need this code?

@pit-rpg No (speaking with rather  high confidence).

It's leftover from past, when we've kept more private `StyleSetEvaluator` stuff in technique. Now, it contains only what it needs and `_index`, `_key`, and `_styleSetIndex` are actually needed here and there in code.

So i think that plain shallow copy of all props _excluding_ instances of `Expr` is ok both in both places.

pit-rpg: Done

@here/harp-examples/src/datasource_object-picking.ts
pit-rpg: Done

@here/harp-examples/src/datasource_object-picking.ts
pit-rpg: Done

PR #735, opened 2019-08-20
HARP-6729: Disable MSAA for high pixelRatio devices.

amsXYZ:

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -248,7 +248,7 @@ export interface MapViewOptions {
      * Whether the native WebGL antialiasing should be enabled. It is better to disable it if the
      * MapView's MSAA is enabled.
      *
-     * @default `true`
+     * @default `true` for `pixelRatio` < `2.0`, `false` otherwise.
      */
     enableNativeWebglAntialias?: boolean;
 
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -826,7 +826,8 @@ export class MapView extends THREE.EventDispatcher {
         // Initialization of the renderer
         this.m_renderer = new THREE.WebGLRenderer({
             canvas: this.canvas,
-            antialias: this.m_options.enableNativeWebglAntialias !== false,
+            antialias:
+                this.m_options.enableNativeWebglAntialias !== false || this.pixelRatio <= 2.0,
```
Astrak: When `pixelRatio <= 2.0` is true then the antialias will be enabled, even if `this.m_options.enableNativeWebglAntialias` is false.

Astrak: Also it would be nice to document this new mechanism as users would not understand it before delving into the source code. Places for this could be in MapView's parameters, in the antialias parameters.

amsXYZ: True, thanks for spotting! Addressed in new patch-set.

PR #736, opened 2019-08-20
HARP-5953: Labels aren't smoothly rendered along rivers, roads and borders

okanishchev:

### DIFF
--- a/@here/harp-datasource-protocol/lib/TechniqueParams.ts
+++ b/@here/harp-datasource-protocol/lib/TechniqueParams.ts
@@ -1234,6 +1234,12 @@ export interface TextTechniqueParams extends BaseTechniqueParams {
      */
     mayOverlap?: boolean;
     /**
+     * If `true`, text path will be smoothen for a better look.
+     * Only for `line` geometry type.
+     * @default `false`
+     */
+    smoothing?: boolean;
+    /**
      * If `true`, element will reserve screen space, other markers of lower priority will not be
      * able to overlap.
      * @default `true`
--- a/@here/harp-mapview/lib/text/TextElement.ts
+++ b/@here/harp-mapview/lib/text/TextElement.ts
@@ -351,6 +351,12 @@ export class TextElement {
     userData?: any;
 
     /**
+     * Enable smoothing on `TextElement` with `path`.
+     * @default `false`
+     */
+    smoothing?: boolean;
+
+    /**
      * Is `true` if is a point label.
      */
     isPointLabel?: boolean;
--- a/@here/harp-text-canvas/lib/TextCanvas.ts
+++ b/@here/harp-text-canvas/lib/TextCanvas.ts
@@ -14,6 +14,7 @@ import { SdfTextMaterial } from "./rendering/TextMaterials";
 import { FontVariant, TextLayoutStyle, TextRenderStyle } from "./rendering/TextStyle";
 import { LineTypesetter } from "./typesetting/LineTypesetter";
 import { PathTypesetter, PathTypesettingParameters } from "./typesetting/PathTypesetter";
+import { SmoothPathTypesetter } from "./typesetting/SmoothPathTypesetter";
 import { TypesettingParameters } from "./typesetting/Typesetter";
 import { createSdfTextMaterial } from "./utils/MaterialUtils";
 
@@ -33,6 +34,7 @@ interface TextPlacementParameters {
     individualBounds?: THREE.Box2[];
     computeTextBuffer?: boolean;
     letterCaseArray?: boolean[];
+    smoothing?: boolean;
 }
 
 /**
@@ -59,6 +61,11 @@ export interface MeasurementParameters {
      * `SmallCaps`.
      */
     letterCaseArray?: boolean[];
+
+    /**
+     * Use smoothing on text path.
+     */
+    smoothing?: boolean;
 }
 
 /**
@@ -96,6 +103,11 @@ export interface AdditionParameters {
      * `SmallCaps`.
      */
     letterCaseArray?: boolean[];
+
+    /**
+     * Use smoothing on text path.
+     */
+    smoothing?: boolean;
 }
 
 /**
@@ -239,6 +251,7 @@ export class TextCanvas {
 
     private m_lineTypesetter: LineTypesetter;
     private m_pathTypesetter: PathTypesetter;
+    private m_smoothTypesetter: SmoothPathTypesetter;
 
     /**
      * Constructs a new `TextCanvas`.
@@ -290,6 +303,7 @@ export class TextCanvas {
 
         this.m_lineTypesetter = new LineTypesetter();
         this.m_pathTypesetter = new PathTypesetter();
+        this.m_smoothTypesetter = new SmoothPathTypesetter();
     }
 
     /**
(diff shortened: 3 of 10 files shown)


@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -1142,6 +1142,11 @@ export interface TextTechniqueParams extends BaseTechniqueParams {
      * @default `false`
      */
     mayOverlap?: boolean;
+    /**
+     * If `true`, icon is allowed to overlap other labels or icons of lower priority.
```
amsXYZ: Fix typedoc.

@here/harp-mapview/lib/text/TextElement.ts
```diff
@@ -350,6 +350,11 @@ export class TextElement {
      */
     userData?: any;
 
+    /**
+     * Optional user data. Will be retrieved during picking.
```
amsXYZ: Fix typedoc.

@here/harp-text-canvas/lib/TextCanvas.ts
```diff
@@ -59,6 +61,12 @@ export interface MeasurementParameters {
      * `SmallCaps`.
      */
     letterCaseArray?: boolean[];
+
+    /**
+     * Array containing info on whether the glyphs are upper or lower case. Needed to support
```
amsXYZ: Fix typedoc.

@here/harp-text-canvas/lib/TextCanvas.ts
```diff
@@ -96,6 +104,12 @@ export interface AdditionParameters {
...
```
amsXYZ: Fix typedoc.

okanishchev: Wow, missed that. Thanks

@here/harp-text-canvas/lib/typesetting/SmoothPathTypesetter.ts
```diff
@@ -0,0 +1,323 @@
...
+            this.m_tempScale;
+
+        const e = this.m_tempPoints.length - 1;
+
+        const v1 = new THREE.Vector2();
+        const v2 = new THREE.Vector2();
+        const normal = new THREE.Vector2();
+
+        let currentAngle = this.m_tempAngles[0];
+        for (let i = 0; i < this.m_tempGlyphs.length; i++) {
```
amsXYZ: Missing some documentation here. Seems like you're doing some low-pass filter with a 1D 5-sample neighborhood, but you should try to add some doc so it's easy to understand and mantain for all devs.

@here/harp-text-canvas/lib/typesetting/SmoothPathTypesetter.ts
```diff
@@ -0,0 +1,323 @@
...
+
+            const isMirrored = this.m_tempMirrored[i];
+
+            const verticalOffset =
+                glyphFontMetrics.lineHeight -
+                glyphFontMetrics.base -
+                glyphFontMetrics.distanceRange * 0.5;
+
+            // Compute the glyphs transformation matrix and apply to all corners of a glyph.
+            TypesettingUtils.computeGlyphTransform(
```
amsXYZ: I really don't like the fact that we have to transform the glyphs twice, specially when this technique only have visible effects around the vertices of the path and not the edges (since the angle delta stays the same inside an edge)...

Cannot we try to pre-compute some information regarding where the path vertices are found in the path's `[0...1]` range and then use this info to modify the glyph placement (in `placeGlyph`) in a similar way ONLY if we fall in the area of effect of any of the vertices?

Think of it as if we were implementing FXAA in the path domain: https://en.wikipedia.org/wiki/Fast_approximate_anti-aliasing#Processes

@here/harp-text-canvas/lib/typesetting/SmoothPathTypesetter.ts
```diff
@@ -0,0 +1,323 @@
...
+        this.reset();
+    }
+
+    arrangeGlyphs(params: PathTypesettingParameters): boolean {
+        if (super.arrangeGlyphs(params) === false) {
+            return false;
+        }
+
+        const result = this.commitLine();
+        this.reset();
```
amsXYZ: This adds a bit of GC pressure in a performance critical part of the draw loop. Would be interesting to compare how much the impact is in real scenes.

@here/harp-text-canvas/lib/TextCanvas.ts
```diff
@@ -14,6 +14,7 @@ import { SdfTextMaterial } from "./rendering/TextMaterials";
 import { FontVariant, TextLayoutStyle, TextRenderStyle } from "./rendering/TextStyle";
 import { LineTypesetter } from "./typesetting/LineTypesetter";
 import { PathTypesetter, PathTypesettingParameters } from "./typesetting/PathTypesetter";
+import { SmoothPathTypesetter } from "./typesetting/SmoothPathTypesetter";
```
amsXYZ: I think this PR would be a nice opportunity to explore how the API could support user-defined typesetters, so we could take this code (and maybe the path typesetter too) outside of the module and keep it as part of the application logic, so maintainability of this module doesn't suffer...

But I can live with doing this later.

@here/harp-text-canvas/lib/typesetting/SmoothPathTypesetter.ts
okanishchev: Do you think, emptying existing arrays would help? With splice or so.

@here/harp-text-canvas/lib/typesetting/SmoothPathTypesetter.ts
okanishchev: We are doing this once.

okanishchev:
It's good idea to reduce additional loops, but in this case I think it would be more expensive than two-phase algo.

In first pass we're only collecting the points and angles - no math here at all, and all cpu-expensive stuff is in the second pass. Also the first pass is doing glyph ordering with LRT, RTL and weak directions and skipping not printed glyphs.

So if we want to run in 1 pass, with current implementation we need to do ordering check 5 times to get right neighbours for each glyph (not sure if this is even possible), then compute angle and delta also 5 times.

@here/harp-text-canvas/lib/typesetting/SmoothPathTypesetter.ts
okanishchev: Added some notes.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
okanishchev: Done.

@here/harp-mapview/lib/text/TextElement.ts
okanishchev: Yep.

@here/harp-text-canvas/lib/TextCanvas.ts
okanishchev: Also done.

@here/harp-text-canvas/lib/TextCanvas.ts
okanishchev: Done.

PR #737, opened 2019-08-21
Add logs for Chrome and Firefox versions

ninok:


musculman:
Part of #729 so I am closing it

PR #739, opened 2019-08-22
Update GettingStartedGuide.md

ninok:


CraigBerlin:
> Either we need to update it from time to time, or we just put above a sentence saying something like `You need to run an installation command similar to the one shown below, although the exact version of three.js might vary`

Could we tell them to run:
`
$ npm view @here/harp.gl dependencies
`

PR #744, opened 2019-08-22
HARP-6082: Display landmarks along side extruded buildings

CraigBerlin:


@here/harp-omv-datasource/lib/OmvDataFilter.ts
```diff
@@ -847,6 +707,24 @@ export class OmvGenericFeatureModifier implements OmvFeatureModifier {
...
     }
 
+    protected static matchAttribute(
+        layerName: string,
+        env: MapEnv,
+        items: OmvFilterDescription[]
+    ): boolean {
+        for (const item of items) {
+            if (item.featureAttribute !== undefined) {
+                if (!OmvFilterString.matchString(layerName, item.layerName)) {
```
nzjony: minor, you could consider moving this into the if condition below, i.e. if matchstring && lookup... { return true; }

@here/harp-omv-datasource/test/OmvTest.ts
```diff
@@ -121,13 +121,11 @@ describe("OmvFeatureFilterDescriptionBuilder", function() {
...
-            "road",
-            OmvGeometryType.LINESTRING,
-            "street",
-            OmvFilterString.StringMatch.Match,
-            OmvFilterString.StringMatch.Match
-        );
+        filterBuilder.processLine({
+            layer: "road",
+            geomType: OmvGeometryType.LINESTRING,
+            featureClass: "street"
```
nzjony: Is it difficult to add a test to test the matchAttribute function?

CraigBerlin: Good idea, added thanks.

PR #745, opened 2019-08-23
HARP-6818: Add support for 'match' and 'case' operators

robertoraggi:


@here/harp-datasource-protocol/test/ExprEvaluatorTest.ts
```diff
@@ -270,4 +273,62 @@ describe("ExprEvaluator", function() {
...
+        });
+
+        it("serialize", function() {
+            const expr = ["match", ["get", "someText"], ["some text", "y"], 1, "z", 2, 3];
+            assert.equal(JSON.stringify(expr), JSON.stringify(Expr.fromJSON(expr)));
+            assert.equal(evaluate(expr), 1);
+        });
+    });
+
+    describe("Operator 'match'", function() {
```
zbigg: `case`

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -87,6 +89,38 @@ export abstract class Expr {
...
+            }
+
+            case "case": {
+                const branches: Array<[Expr, Expr]> = [];
+                for (let i = 1; i < node.length - 1; i += 2) {
+                    const condition = this.fromJSON(node[i]);
+                    const expr = this.fromJSON(node[i + 1]);
+                    branches.push([condition, expr]);
+                }
+                const caseFallback = this.fromJSON(node[node.length - 1]);
```
zbigg: Some error handling is needed.
In case of missing `fallback` element, expr from last pair will whereas `null/undefined` would be expected.

PR #747, opened 2019-08-23
Harp 6905: Use maximum view range for features fading calculus.

kkostecki:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -117,6 +117,8 @@ const logger = LoggerManager.instance.create("MapView");
 const DEFAULT_FONT_CATALOG = "./resources/fonts/Default_FontCatalog.json";
 const DEFAULT_CLEAR_COLOR = 0xefe9e1;
 const DEFAULT_FOV_CALCULATION: FovCalculation = { type: "dynamic", fov: 40 };
+const DEFAULT_CAM_NEAR_PLANE = 0.1;
+const DEFAULT_CAM_FAR_PLANE = 4000000;
 const MAX_FIELD_OF_VIEW = 140;
 const MIN_FIELD_OF_VIEW = 10;
 
@@ -639,6 +641,7 @@ export class MapView extends THREE.EventDispatcher {
 
     private m_focalLength: number;
     private m_lookAtDistance: number;
+    private m_maxVisibility: number;
     private m_pointOfView?: THREE.PerspectiveCamera;
 
     private m_pixelToWorld?: number;
@@ -861,12 +864,13 @@ export class MapView extends THREE.EventDispatcher {
         this.m_camera = new THREE.PerspectiveCamera(
             this.m_options.fovCalculation.fov,
             aspect,
-            0.1,
-            4000000
+            DEFAULT_CAM_NEAR_PLANE,
+            DEFAULT_CAM_FAR_PLANE
         );
         this.m_camera.up.set(0, 0, 1);
         this.m_lookAtDistance = 0;
         this.m_focalLength = 0;
+        this.m_maxVisibility = DEFAULT_CAM_FAR_PLANE;
         this.m_scene.add(this.m_camera); // ensure the camera is added to the scene.
         this.m_screenProjector = new ScreenProjector(this.m_camera);
 
@@ -1401,6 +1405,17 @@ export class MapView extends THREE.EventDispatcher {
      */
     get lookAtDistance(): number {
         return this.m_lookAtDistance;
+    }
+
+    /**
+     * The maximum visibility range that may be achieved with current camera settings.
+     * @note Visibility is directly related to camera [[ClipPlaneEvaluator]] used and determines
+     * the maximum possible distance of camera far clipping plane regardless of tilt, but may change
+     * whenever zoom level changes. Distance is meassured in world units which may be approximately
+     * euqal to meters, but this depends on the distortion related to projection type used.
+     */
+    get maxVisibility(): number {
+        return this.m_maxVisibility;
     }
 
     /**
@@ -2247,6 +2262,7 @@ export class MapView extends THREE.EventDispatcher {
         );
         this.m_camera.near = clipPlanes.near;
         this.m_camera.far = clipPlanes.far;
+        this.m_maxVisibility = clipPlanes.maximum;
 
         this.m_camera.updateProjectionMatrix();
         this.m_camera.updateMatrixWorld(false);
--- a/@here/harp-mapview/lib/Utils.ts
+++ b/@here/harp-mapview/lib/Utils.ts
@@ -23,7 +23,6 @@ const MINIMUM_ATTRIBUTE_SIZE_ESTIMATION = 56;
 
 // Caching those for performance reasons.
 const groundNormalPlanarProj = new THREE.Vector3(0, 0, 1);
-const groundNormalInvPlanarProj = groundNormalPlanarProj.clone().negate();
 const groundPlane = new THREE.Plane(groundNormalPlanarProj.clone());
 const groundSphere = new THREE.Sphere(undefined, EarthConstants.EQUATORIAL_RADIUS);
 const cameraZPosition = new THREE.Vector3(0, 0, 0);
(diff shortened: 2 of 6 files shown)


kkostecki:

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -385,6 +394,7 @@ export namespace FadingFeature {
         object: THREE.Object3D,
         fadeNear: number | undefined,
         fadeFar: number | undefined,
+        maxVisibility: number | undefined,
```
nzjony: it seems that usages of this are always defined, which means that the cameraToWorldDistance could just take the maxVisibility and doesn't need to take the camera... or do I miss something?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -367,17 +367,14 @@ export namespace MapViewUtils {
...
-        if (projection.type === geoUtils.ProjectionType.Planar) {
-            return groundNormalInvPlanarProj;
-        } else {
-            // Note using different vector then caller, otherwise vector data could be
-            // silently modified.
-            const normal: THREE.Vector3 = tmpVectors[1];
-            projection.surfaceNormal(position, normal);
-            return normal.negate();
-        }
+        // We can simply return [0,0,1] for planar projections because WebMercatorProjection
```
nzjony:
```suggestion
        // We can't simply return [0,0,1] for planar projections because WebMercatorProjection
```

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1390,6 +1392,16 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_lookAtDistance;
     }
 
+    /**
+     * The maximum visibility range (in meters) that may be achievied with current camera setttings.
```
nzjony: setttings -> settings

nzjony: I think the comment `(in meters)` is correct only at the equator, or for globe projection, for any planar projection, I think this is incorrect, because the values distort towards the equator. Maybe world units, or projected units is best?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -867,6 +868,7 @@ export class MapView extends THREE.EventDispatcher {
         this.m_camera.up.set(0, 0, 1);
         this.m_lookAtDistance = 0;
         this.m_focalLength = 0;
+        this.m_maxVisibility = 4000000;
```
nzjony: does this mean anything specifically? Would it be better to use MAX_SAFE_INTEGER?

@here/harp-materials/lib/MapMeshMaterials.ts
kkostecki: No, you did not miss anything, although I left a possibility to calculate fading in old fashion (way). You simply do not pass maxVisibility - then camera far plane applies.

@here/harp-mapview/lib/MapView.ts
kkostecki: True.

@here/harp-mapview/lib/MapView.ts
kkostecki: This is exactly the far plane distance setup initially for camera, before the cameras and clip planes evaluator update. See few lines above. I guess MAX_SAFE_INTEGER is not the best option for frustum settings and this need to be compliant - the same.

@here/harp-mapview/lib/Utils.ts
kkostecki: Ok.

@here/harp-materials/lib/MapMeshMaterials.ts
kkostecki: Should I remove this option?

nzjony: Yeah, if you don't need a particular code path, best remove, we can always add it back later.

@here/harp-mapview/lib/MapView.ts
nzjony: Now I see, ok, great you made a constant.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1390,6 +1394,17 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_lookAtDistance;
     }
 
+    /**
+     * The maximum visibility range that may be achievied with current camera settings.
+     * @note Visibility is directly related to camera [[ClipPlaneEvaluator]] used and determines
+     * the maximum possible distance of camera far clipping plane regardless of tilt, but may change
+     * whenever zoom level changes. Distance is meassured in world units which may be approximatelly
```
nzjony: appoximatelly -> approximately

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1390,6 +1394,17 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_lookAtDistance;
     }
 
+    /**
+     * The maximum visibility range that may be achievied with current camera settings.
```
nzjony: achievied -> achieved

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2234,6 +2249,7 @@ export class MapView extends THREE.EventDispatcher {
         );
         this.m_camera.near = clipPlanes.near;
         this.m_camera.far = clipPlanes.far;
+        this.m_maxVisibility = clipPlanes.maximum;
```
nzjony: what is the difference between maximum and far? Can maximum be further than far, or is far up to maximum?

kkostecki: Visibility range (maxVisibility) is the absolutely maximum distance that far plane may be put on. It is our internal limit for distant rendering. We need it for situations when camera may be tilt with look at vector close to parallel to the ground, in this conditions simple math would account for infinite distance.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -367,17 +366,14 @@ export namespace MapViewUtils {
...
-            const normal: THREE.Vector3 = tmpVectors[1];
-            projection.surfaceNormal(position, normal);
-            return normal.negate();
-        }
+        // We can't simply return [0,0,1] for planar projections because WebMercatorProjection
+        // has ground normal inverted - as opposite to other planar projections, so use of
+        // the projection interface is encouraged to acquire ground normal.
+        // Note using different (tmp) vector then caller, otherwise vector data could be
+        // silently modified.
+        const normal: THREE.Vector3 = tmpVectors[1];
```
ninok: I know it's not your code, but I would change the interface of this method to:
```
    function getGroundNormalInv(
        position: THREE.Vector3,	        position: THREE.Vector3,
        projection: geoUtils.Projection	        projection: geoUtils.Projection
        result?: THREE.Vector3
    ): THREE.Vector3
```
Otherwise the behavior of this method is quite confusing:
```
const normal1 = getGroundNormalInv(position1, projection);
const normal2 = getGroundNormalInv(position2, projection);
```
If result is undefined this method would allocate a new Vector3. It would be basically up to the user whether an allocation happens or not.

kkostecki: Will fix it too.

kkostecki:

@here/harp-mapview/lib/Utils.ts
```diff
@@ -348,36 +347,35 @@ export namespace MapViewUtils {
...
-            return groundNormalInvPlanarProj;
-        } else {
-            // Note using different vector then caller, otherwise vector data could be
-            // silently modified.
-            const normal: THREE.Vector3 = tmpVectors[1];
-            projection.surfaceNormal(position, normal);
-            return normal.negate();
-        }
+        // We can't simply return [0,0,1] for planar projections because WebMercatorProjection
+        // has ground normal inverted - as opposite to other planar projections, so use of
```
nzjony:
```suggestion
        // has ground normal inverted - as opposed to other planar projections, so use of
```

@here/harp-mapview/lib/Utils.ts
```diff
@@ -348,36 +347,35 @@ export namespace MapViewUtils {
...
-     *
-     * @note Function is not exposed cause it returns internal temporary vectors that
-     * can not be modified. Use with care and clone the result if required.
+     * Return normal to the ground surface (looking down) directly above camera position.
      *
      * @param position The position above the ground for which normal is calculated.
      * @param projection The projection used to transform geo coordinates to world space.
+     * @param result The pre-allocated vector for storing result value, if it is
+     * undefined function will allocate new vector and return it as result.
+     * @returns inverted ground normal coppied to [[result]] vector or newly allocated
```
nzjony:
```suggestion
     * @returns inverted ground normal copied to [[result]] vector or newly allocated
```

@here/harp-mapview/lib/Utils.ts
kkostecki: Ok.

@here/harp-mapview/lib/Utils.ts
kkostecki: Will fix.

kkostecki:

kkostecki:

PR #748, opened 2019-08-23
HARP-6249: Caps for lines endings

pit-rpg:
Added caps support for lines.
   The following types are supported now:
   "Square", "Round", "None", "TriangleOut", "TriangleIn"

### DIFF
--- a/@here/harp-datasource-protocol/lib/TechniqueParams.ts
+++ b/@here/harp-datasource-protocol/lib/TechniqueParams.ts
@@ -8,6 +8,12 @@ import { JsonExpr } from "./Expr";
 import { InterpolatedPropertyDefinition } from "./InterpolatedPropertyDefs";
 
 /**
+ * Available line caps types(`"None"`, `"Round"`, `"Square"`, `"TriangleOut"`, `"TriangleIn"`).
+ * Default is `"Round"`.
+ */
+export type LineCaps = "Square" | "Round" | "None" | "TriangleOut" | "TriangleIn";
+
+/**
  * The kind of geometry is used to
  *
  * a) Group objects together, allowing the group to be hidden or displayed.
@@ -916,6 +922,12 @@ export interface SolidLineTechniqueParams extends BaseTechniqueParams, Polygonal
      * The render order of the secondary line geometry object created using this technique.
      */
     secondaryRenderOrder?: number;
+
+    /**
+     * Describes line caps type (`"None"`, `"Round"`, `"Square"`, `"TriangleOut"`, `"TriangleIn"`).
+     * Default is `"Round"`.
+     */
+    caps?: LineCaps;
 }
 
 /**
@@ -975,6 +987,12 @@ export interface DashedLineTechniqueParams extends BaseTechniqueParams, Polygona
      * Clip the line outside the tile if `true`.
      */
     clipping?: boolean;
+
+    /**
+     * Describes line caps type (`"None"`, `"Round"`, `"Square"`, `"TriangleOut"`, `"TriangleIn"`).
+     * Default is `"Round"`.
+     */
+    caps?: LineCaps;
 }
 
 /**
--- a/@here/harp-datasource-protocol/package.json
+++ b/@here/harp-datasource-protocol/package.json
@@ -26,7 +26,6 @@
     "dependencies": {
         "@here/harp-geometry": "^0.10.0",
         "@here/harp-geoutils": "^0.10.0",
-        "@here/harp-lines": "^0.10.0",
         "@here/harp-utils": "^0.10.0"
     },
     "devDependencies": {
--- a/@here/harp-lines/lib/Lines.ts
+++ b/@here/harp-lines/lib/Lines.ts
@@ -36,21 +36,23 @@ interface VertexDescriptor {
 /** Optional normal and uv coordinates. */
 const NORMAL_UV_VERTEX_ATTRIBUTES: VertexDescriptor = {
     attributes: [
-        { name: "uv", itemSize: 2, offset: 12 },
-        { name: "normal", itemSize: 3, offset: 14 }
+        { name: "uv", itemSize: 2, offset: 13 },
+        { name: "normal", itemSize: 3, offset: 15 }
     ],
     stride: 5
 };
 
 /** Base line vertex attributes. */
 const LINE_VERTEX_ATTRIBUTES: VertexDescriptor = {
+    // The "extrusionCoord" its a vec3. Represents UV coordinates + line length for the third
+    // component.
     attributes: [
-        { name: "extrusionCoord", itemSize: 2, offset: 0 },
-        { name: "position", itemSize: 3, offset: 2 },
-        { name: "tangent", itemSize: 3, offset: 5 },
-        { name: "bitangent", itemSize: 4, offset: 8 }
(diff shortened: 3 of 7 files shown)


@here/harp-lines/lib/Lines.ts
```diff
@@ -151,6 +152,7 @@ export function createLineGeometry(
         sum = sum + len;
         segments[i + 1] = sum;
     }
+    const lineLength = segments[segments.length - 1];
```
amsXYZ: If this is constant, couldn't just that be a uniform passed down to the material?

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -111,14 +127,49 @@ void main() {
...
+    float lineEnds = max(vExtrusionCoord.x - vLength,- vExtrusionCoord.x);
+
     #if TILE_CLIP
     tileClip(vPosition.xy, tileSize);
     #endif
 
-    float dist = joinDist(vSegment, vExtrusionCoord) - vLinewidth;
+    float dist = 0.0;
+
+    #if defined(CAPS_NONE)
```
amsXYZ: Cool stuff! Would add it to the `LinesChunks.ts` file to clean up the main shader code, but pretty neato.

@here/harp-lines/lib/Lines.ts
pit-rpg: We can't use uniform because this method is used in LineGroup class that renders a bunch of lines at once, those lines will have different length.

@here/harp-materials/lib/SolidLineMaterial.ts
pit-rpg: Done

@here/harp-datasource-protocol/package.json
```diff
@@ -27,7 +27,8 @@
         "@here/harp-geometry": "^0.10.0",
         "@here/harp-geoutils": "^0.10.0",
         "@here/harp-lines": "^0.10.0",
-        "@here/harp-utils": "^0.10.0"
+        "@here/harp-utils": "^0.10.0",
+        "@here/harp-materials": "^0.10.0"
```
zbigg: Datasource protocol should't depend on "renderer" part of code, in particular `harp-materials`.
(example: we don't want shaders and stuff in decoder code  -minor issue, but an issue).

I know, this is for `LineCaps` definition.

I know our package architecture is mess, but don't make it worse...

All the code around "implements" features defined in `datasource-protocol` not the other way around, so please move `LineCaps` do `datasource-protocol` and reverse the dependency or ... even duplicate it as
* "material" `LineCaps` should be actually an enum 
* "theme" `LineCaps` is "string enum" as defined by user in theme file
* "string" to "enum" should be explicit mapping with sane fallback - what if user will pass `"FOO"` as caps ? at first look something will explode somewhere

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -934,6 +940,11 @@ export interface DashedLineTechniqueParams extends BaseTechniqueParams, Polygona
      * Clip the line outside the tile if `true`.
      */
     clipping?: boolean;
+
+    /**
+     * Describes line caps type. Default is Round.
```
zbigg: Use markdown: `"Round"`.

Add doc what kind of cap types are supported.

@here/harp-datasource-protocol/package.json
pit-rpg: Done.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
pit-rpg: Done.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -934,6 +946,12 @@ export interface DashedLineTechniqueParams extends BaseTechniqueParams, Polygona
      * Clip the line outside the tile if `true`.
      */
     clipping?: boolean;
+
```
ninok: Out of scope for your topic, but I think `DashedLineTechniqueParams` should extend `SolidLineTechniqueParams`

@here/harp-lines/lib/Lines.ts
```diff
@@ -48,9 +48,10 @@ const LINE_VERTEX_ATTRIBUTES: VertexDescriptor = {
         { name: "extrusionCoord", itemSize: 2, offset: 0 },
         { name: "position", itemSize: 3, offset: 2 },
         { name: "tangent", itemSize: 3, offset: 5 },
-        { name: "bitangent", itemSize: 4, offset: 8 }
+        { name: "bitangent", itemSize: 4, offset: 8 },
+        { name: "lineLength", itemSize: 1, offset: 12 }
```
ninok: Can we not squeeze it somewhere else, e.g. position

@here/harp-materials/package.json
```diff
@@ -20,6 +20,7 @@
     },
     "license": "Apache-2.0",
     "dependencies": {
+        "@here/harp-datasource-protocol": "^0.10.0",
```
ninok: Do we really need this dependeny? I would rather add the mapping from cap type to define to datasource protocoll.

pit-rpg: Could you please clarify what did you exactly mean? How it should be mapped?

@here/harp-lines/lib/Lines.ts
pit-rpg: Please clarify, you mean to put it after position attribute?

@here/harp-materials/package.json
ninok: As far as I understood you added this dependency so that you can auto-generate the mapping from cap type (defined in datasource-protocol) to DEFINE name. I would rather do that mapping by hand (i.e a switch statement) than adding this dependency

@here/harp-lines/lib/Lines.ts
ninok: no, add it to the position attribute (i.e. the .w component of position) or alternatively into the tangent attribute. Having only one float in a vertex attribute is quite wasteful (see e.g.: https://stackoverflow.com/questions/21676280/glsl-packing-4-float-attributes-into-vec4)

Please also add documentation to the vertex attributes. The documentation is quite poor right now.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
pit-rpg: It makes sense to me. Now there are other pool requests that touch this code.
https://github.com/heremaps/harp.gl/pull/754
It is possible to do this upon completion of this task.
Please specify what to do next.

@here/harp-materials/package.json
pit-rpg: Yes, I'm trying to create one source of truth, and I also want TypeScript to guarantee me in future that if there are changes in this code, they will be accurately processed in all the necessary places. I think an approach like this can help write more error resistant code. I can change it of course but if you agree I would like to live it as is.

@here/harp-lines/lib/Lines.ts
pit-rpg: Done.

pit-rpg:
Rebuild please

PR #754, opened 2019-08-27
HARP-6801: Implement dash-line color & outline

sasha240100:
Signed-off-by: Oleksandr Buzin <ext-oleksandr.buzin@here.com>

# CI Check

Happy contributing!


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -834,6 +834,17 @@ export class TileGeometryCreator {
...
 
+                            if ('outlineWidth' in technique && technique.outlineWidth !== undefined) {
+                              lineMaterial.outlineWidth =
+                                  getPropertyValue(
+                                    technique.outlineWidth,
+                                    mapView.zoomLevel,
+                                    mapView.pixelToWorld
+                                  ) *
+                                  unitFactor *
+                                  0.5;
```
zbigg: Why `0.5` here?
What is the difference between `Technique.outlIneWidth` and `SolidLineMaterial.outlineWidth` ?
They should be the same i guess.

(I guess this is our usual mess with semantics of `width` vs `extrusion ratio`, but at least in new code, we can have it done well.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -48,20 +49,20 @@ varying vec3 vColor;
 #include <extrude_line_vert_func>
 
 void main() {
-    vLinewidth = lineWidth;
+    vResultLineWidth = lineWidth + outlineWidth * 2.0;
```
zbigg: 1. This is actually `lineExtrusionFactor`. (Which is apparently half actual lineWidth)

2.
Why `* 2 .0` here ?

At first look they cancel out with code in `TileGeometryCreator`.

sasha240100: @zbigg `* 2.0` is b/c we have oulines from both sides (right, left) and outlineWidth is width of one of them

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -308,16 +401,41 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
         this.uniforms.diffuse.value = value;
     }
 
+    get dashColor(): THREE.Color {
+        return this.uniforms.dashColor.value as THREE.Color;
+    }
+    set dashColor(value: THREE.Color) {
+        this.uniforms.dashColor.value = value;
+        this.updateDashAlpha(!Boolean(value));
```
okanishchev: This is always false, as you don't accept undefined as argument

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -308,16 +401,41 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
         this.uniforms.diffuse.value = value;
     }
 
+    get dashColor(): THREE.Color {
+        return this.uniforms.dashColor.value as THREE.Color;
+    }
+    set dashColor(value: THREE.Color) {
+        this.uniforms.dashColor.value = value;
```
okanishchev: You can do
```suggestion
        this.uniforms.dashColor.value.copy(value);
```
here, and then just assign `dashColor = color` without weird `this.dashColor = this.dashColor;` stuff

@here/harp-materials/lib/SolidLineMaterial.ts
zbigg: @sasha240100

> outlineWidth is width of one of them

Yes, i understand, but ... why you divide it at start ?

So, you shouldn't multiply(here) and divide (there) at all.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -124,13 +131,40 @@ void main() {
     float segmentDist = mod(vExtrusionCoord.x, dashSize + gapSize) / dashSize;
     float dashDist = 0.5 - distance(segmentDist, halfSegment);
     float dashWidth = fwidth(dashDist);
-    alpha *= smoothstep(-dashWidth, dashWidth, dashDist);
+    float dashedBlendFactor = smoothstep(-dashWidth, dashWidth, dashDist);
+
+    #ifdef USE_DASH_ALPHA
+    if (outlineWidth <= 0.0) {
+      alpha *= dashedBlendFactor;
```
okanishchev: You should avoid `if`s in shader code. In this case you can use `step`.
```glsl
alpha *= step(outlineWidth, 0.) + step(-outlineWidth, 0.) * dashedBlendFactor;
```

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -834,6 +834,21 @@ export class TileGeometryCreator {
                                 unitFactor *
                                 0.5;
 
+                            if (
+                                "outlineWidth" in technique &&
+                                typeof technique.outlineWidth === "number" &&
```
okanishchev: If this property is static, you should move this `lineMaterial.outlineWidth` assignment out from update function.
Otherwise it's a wrong way to update interpolated property, because it's not a number, and to check that you should use `isInterpolatedProperty`.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -215,14 +265,31 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
...
+        if (params !== undefined && !Boolean(params.dashColor)) {
+            defines.USE_DASH_ALPHA = "";
+        }
+
+        if (
+            params !== undefined &&
+            typeof params.outlineWidth === "number" &&
+            params.outlineWidth > 0
+        ) {
+            defines.USE_OUTLINE = "";
```
okanishchev: You can use `updateOutline` here

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -215,14 +265,31 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
             defines.USE_DISPLACEMENTMAP = "";
         }
 
+        if (params !== undefined && !Boolean(params.dashColor)) {
+            defines.USE_DASH_ALPHA = "";
```
okanishchev: You can use `updateDashAlpha` here

@here/harp-materials/lib/SolidLineMaterial.ts
sasha240100: @okanishchev What's the point? Doesn't it make it complicated to read with no performance impact? I see no reason to do that.

@here/harp-materials/lib/SolidLineMaterial.ts
sasha240100: @okanishchev for what reason? One more function call w/o actual benefit

@here/harp-materials/lib/SolidLineMaterial.ts
sasha240100: @okanishchev Same as above

ninok:
@sasha240100  Please add image regressions tests for this. There are some tests for line rendering already, so should be easy to extent.

@here/harp-materials/lib/SolidLineMaterial.ts
sasha240100: @okanishchev I guess you mean `this.dashColor.set(params.dashColor)` instead of `this.dashColor = this.dashColor;`. as copy() doesn't remove the need of calling setter

@here/harp-materials/lib/SolidLineMaterial.ts
sasha240100: Ah, @zbigg got what you meant. Now I noticed that it is related to that `0.5` multiplication - fixed, will push a fix

@here/harp-materials/lib/SolidLineMaterial.ts
sasha240100: @okanishchev That makes it even worse as if you assign new color and then change it outside, it wouldn't work!!!. + It doesn't help solve the setter issue in any way.

Example:
```js
const myColor = new THREE.Color();
solidLineMaterial.dashColor = myColor;
myColor.set(0xff0000); // wouldn't work as dashColor was copied
```

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
sasha240100: @okanishchev Please explain how do you see the usage of `isInterpolatedProperty` here

@here/harp-materials/lib/SolidLineMaterial.ts
okanishchev: The branches in shader code can be very slow on some hardware, for example a bit of explanation can be found here: https://stackoverflow.com/questions/37827216/does-if-statements-slow-down-my-shader
And the constructions like `a * (1 - c) + b * c` is pretty usual in shader code, but for readability you can use additional variable:
```glsl
float enableDashedBlend = step(-outlineWidth, 0.);
alpha *= (1. - enableDashedBlend) + enableDashedBlend * dashedBlendFactor;
```

@here/harp-materials/lib/SolidLineMaterial.ts
okanishchev: Here you have special semantics on enabling defines in shader code:
 * you need to use special name `USE_DASH_ALPHA`
 * you need to use special value `""`

And none of it is checked by typescript nor tslint. What effect it will have if I'll write for example
`defines.USE_DASHED_ALPHA = "";` or even
`defines.USE_DASH_ALPHA = 0;`

So using pre-defined function is less error-prone here.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
okanishchev: First of all, the technique passed here can't be changed, and all the properties are final. So you can check `isInterpolatedProperty(outlineWidth)` outside of this update function, and if it's false - you only need to set it once, and avoid setting it each frame here. And vice versa - set the value every frame if the condition was true.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -311,16 +414,45 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
         this.uniforms.diffuse.value = value;
     }
 
+    get dashColor(): THREE.Color {
+        return this.uniforms.dashColor.value as THREE.Color;
+    }
+    set dashColor(value: THREE.Color) {
+        if (value instanceof THREE.Color) {
```
okanishchev: It's also wrong. The value is always instanceof THREE.Color

@here/harp-materials/lib/SolidLineMaterial.ts
okanishchev: Then you should call the `updateDashAlpha()` function explicitly when it's about to change, or add special property setter for this and also assign it explicitly.
Or you can even use `onBeforeRender` callback to update this values, but I'm not sure if it will not affect performance.

sasha240100: @okanishchev How is it related to the written above?

@here/harp-materials/lib/SolidLineMaterial.ts
sasha240100: @okanishchev I don't see any use case where it would save the developer as it is an internal code. We force a consumer to use `updateDashAlpha()`, but this case is internal optimization and is unrelated. 
Plus it would break the consistency as USE_FOG is written the same way

@here/harp-materials/lib/SolidLineMaterial.ts
sasha240100: @okanishchev Then I guess you should rise a separate topic as if-statement is used in pretty-much any fragmentSource in the project. (see `IconMaterial.ts`, `TextMaterials.ts`, and a lot more)

**UPD:** I will replace it with define (USE_OUTLINE)

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -127,13 +134,41 @@ void main() {
     float segmentDist = mod(vExtrusionCoord.x, dashSize + gapSize) / dashSize;
     float dashDist = 0.5 - distance(segmentDist, halfSegment);
     float dashWidth = fwidth(dashDist);
-    alpha *= smoothstep(-dashWidth, dashWidth, dashDist);
+    float dashedBlendFactor = smoothstep(-dashWidth, dashWidth, dashDist);
+
+    #ifndef USE_DASH_ALPHA
```
okanishchev: The name `USE_DASH_ALPHA` is poorly related to what it actually does. Maybe `USE_DASH_COLOR`?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
okanishchev: You still miss the interpolated outlineWidth case

@here/harp-materials/lib/SolidLineMaterial.ts
okanishchev: The internal code doesn't mean that you'll do better on understanding and maintaining the code. Here you can reuse the code you already have. But maybe in this particular case it's really doesn't matter.

@here/harp-materials/lib/SolidLineMaterial.ts
okanishchev: Still not perfect solution, and I think you should leave some comments why it's done this way - enabling dash color only if you've passed a style for that.

sasha240100:
@heremaps-bot Rebuild please

sasha240100:
@heremaps-bot Retest please

sasha240100:

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
sasha240100: @okanishchev fixed.

PR #755, opened 2019-08-27
HARP-6773: Extruded building overlay without sockets.

atomicsulfate:

First implementation of extruded building overlay without creating sockets. All building vertices are displaced using the elevation map texture.

Tested on fake terrain since ElevationProvider does not provide yet elevation maps for zoom levels where extruded buildings are visible.


@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -1086,6 +1089,7 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
                         );
                         if (texCoordType !== undefined) {
                             textureCoordinates.push(vertices[i + 2], vertices[i + 3]);
+                            normals.push(...tempVertNormal.toArray());
```
ninok: As discussed, normals should only be generated if elevation is enabled

atomicsulfate: done.

PR #757, opened 2019-08-29
HARP-6802: ES2017 TS target.

amsXYZ:

Happy contributing!

### DIFF
--- a/@here/harp.gl/tsconfig.json
+++ b/@here/harp.gl/tsconfig.json
@@ -1,7 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es6",
-        "lib": ["es2017", "dom"],
+        "target": "es2017",
         "module": "commonjs",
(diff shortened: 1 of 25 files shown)


@here/harp.gl/tsconfig.json
```diff
@@ -1,7 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es6",
-        "lib": ["es2017", "dom"],
+        "target": "es2017",
```
zbigg: This is the only one i would discuss.

Code generated by this particular `tsconfig.json` is used directly in browsers without transpilation and i'm unable to say from head if this change will not cause any troubles for `harp.js` users.

amsXYZ: Tested in major browsers across different devices, and seems like no major incompatibilities are present.

PR #759, opened 2019-08-29
Run github CI only for push on master

musculman:

### DIFF
--- a/.github/workflows/ci-macos.yaml
+++ b/.github/workflows/ci-macos.yaml
@@ -1,6 +1,12 @@
 name: CI (MacOS)
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
 
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,12 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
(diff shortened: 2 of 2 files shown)


.github/workflows/ci-macos.yaml
```diff
@@ -1,6 +1,12 @@
 name: CI (MacOS)
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
```
ninok: I was checking the documentation but it is not clear to me what this means. What branch is meant here, the target branch (than it should be master) or the source branch?

[Documentation](https://help.github.com/en/articles/events-that-trigger-workflows) says: 
_Triggered when a pull request is assigned, unassigned, labeled, unlabeled, opened, edited, closed, reopened, synchronize, ready_for_review, locked, unlocked or when a pull request review is requested or removed. See below to learn how this event works with forked repositories._
So maybe we should limit the CI to opened PRs only.

musculman: Hmmm you may be right, the documentation is not 100% clear. I want to make sure the pull_request CI is triggered for any PR, that is why I put a wildcard on the branch. In the previous section of the documentation, it mentions that for pull request one should just put `pull_request` so that is why I just added it like that.

ninok: Okay, I guess we just try it.

musculman: That is why it is non-blocking for now.... we need to learn to fine tune it first :) thanks!

PR #760, opened 2019-08-29
HARP-6123: ServiceType failed connection logs.

amsXYZ:
Thank you for contributing to harp.gl!

Before requesting a pull request, please remember to check the following documents:
* [contribution guidelines](https://github.com/heremaps/harp.gl/blob/master/CONTRIBUTING.md)
* [coding style](https://github.com/heremaps/harp.gl/blob/master/CODINGSTYLE.md)

If you are adding new functionality we would highly appreciate if you can describe what is the capability you are adding and even better if you can add some examples. Please also remember to add tests for it.

Our bots will check whether your PR can be directly integrated into the mainline. We have some internal integration tests running on the background, our bots will inform you of the next steps and someone from our team will take a look and help if needed!

And please do not forget to sign-off your commit! You can read more about DCO [here](https://julien.ponge.org/blog/developer-certificate-of-origin-versus-contributor-license-agreements/). But, in short, you just need to use `git commit -s` or append `--signoff` when you are committing to the repo.

Happy contributing!

PR #761, opened 2019-08-30
HARP-7009: Improve error handling when connecting to datasource

zbigg:
Followup of #760 

* `ConcurrentWorkerSet#connect` now
  * properly waits for all workers (and rejects if any of them failed)
  * throws meaningful message if trying to connect to unknown service
  * timeouts if created worker doesn't respond at all
* implementations of `DataSource`/`DataProvider` pass `connect` failures up
  to `MapView`
* `MapView#addDataSource` handles and logs `DS#connect` errors
* `OmvDataSource#connect` shows hint, if `omv-decoder` service is not
  registered in decoder
* `GeoJsonDataProvider` shows hint in console, if `omv-tiler` service
  is not registered in decoder

TODO:
* [x] - show hints only once

### DIFF
--- a/@here/harp-datasource-protocol/lib/WorkerServiceProtocol.ts
+++ b/@here/harp-datasource-protocol/lib/WorkerServiceProtocol.ts
@@ -70,6 +70,9 @@ export namespace WorkerServiceProtocol {
     /**
      * This message is sent by the main thread to [[WorkerServiceManager]] to dynamically create a
      * new service.
+     *
+     * May throw `UnknownServiceError` if service of given type is not registered in
+     * [[WorkerServiceManager]], see [[isUnknownServiceError]].
      */
     export interface CreateServiceRequest extends ServiceRequest {
         type: Requests.CreateService;
@@ -85,6 +88,13 @@ export namespace WorkerServiceProtocol {
          * The newly created service instance will be available under this id.
          */
         targetServiceId: string;
+    }
+
+    /**
+     * Test if `error` thrown by [[CreateServiceRequest]] was caused by unknown type of service.
+     */
+    export function isUnknownServiceError(error: Error): boolean {
+        return /unknown targetServiceType requested: /.test(error.message);
     }
 
     /**
--- a/@here/harp-geojson-datasource/lib/GeoJsonDataProvider.ts
+++ b/@here/harp-geojson-datasource/lib/GeoJsonDataProvider.ts
@@ -4,11 +4,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { GeoJson, ITiler } from "@here/harp-datasource-protocol";
+import { GeoJson, ITiler, WorkerServiceProtocol } from "@here/harp-datasource-protocol";
 import "@here/harp-fetch";
 import { TileKey } from "@here/harp-geoutils";
 import { ConcurrentTilerFacade } from "@here/harp-mapview";
 import { DataProvider } from "@here/harp-mapview-decoder";
+import { LoggerManager } from "@here/harp-utils";
+
+const logger = LoggerManager.instance.create("GeoJsonDataProvider");
 
 export interface GeoJsonDataProviderOptions {
     /**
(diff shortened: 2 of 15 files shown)


@here/harp-datasource-protocol/lib/WorkerServiceProtocol.ts
```diff
@@ -87,6 +90,13 @@ export namespace WorkerServiceProtocol {
         targetServiceId: string;
     }
 
+    /**
+     * Test if `error` thrown by [[CreateServiceRequest]] was caused by unknown type of service.
+     */
+    export function isUnknownServiceError(error: Error): boolean {
```
amsXYZ: Shouldn't a similar error be thrown as a ``warn`` when trying to destroy a service?

@here/harp-geojson-datasource/lib/GeoJsonDataProvider.ts
```diff
@@ -51,7 +56,22 @@ export class GeoJsonDataProvider implements DataProvider {
     }
 
     async connect(): Promise<void> {
-        await this.m_tiler.connect();
+        try {
+            await this.m_tiler.connect();
+        } catch (error) {
+            if (
+                WorkerServiceProtocol.isUnknownServiceError(error) &&
+                !missingTilerServiceInfoEmitted
```
amsXYZ: I get why might be nice to use to only output this message once at app init, but in more interactive apps where you might add/delete DataSources... Is this the intended behaviour? Might just log the error always.

@here/harp-omv-datasource/lib/OmvDataSource.ts
```diff
@@ -215,6 +225,24 @@ export class OmvDataSource extends TileDataSource<OmvTile> {
         };
     }
 
+    async connect() {
+        try {
+            await super.connect();
+        } catch (error) {
+            if (
+                WorkerServiceProtocol.isUnknownServiceError(error) &&
+                !missingOmvDecoderServiceInfoEmitted
```
amsXYZ: Same as with ``missingTilerServiceInfoEmitted``. Really needed?

@here/harp-geojson-datasource/lib/GeoJsonDataProvider.ts
zbigg: Note that error will be logged always, in `MapView.addDataSource` (before, it was silent).
This is indeed just for convenience, note that error is thrown anyways.

@here/harp-datasource-protocol/lib/WorkerServiceProtocol.ts
zbigg: Maybe yes, maybe not, this patch is about improving UX of our public API w.r.t connecting to decoders. 
So from one side is out of scope.
From other side, destroying unknown service is IMO, super internal error, which doesn't occur frequently whereas "not registered" `omv-tiler` or `omv-decoder` is quite probable, beginner level issue that's why it has dedicated error and logging.

@here/harp-geojson-datasource/lib/GeoJsonDataProvider.ts
amsXYZ: Hmm I guess you're right, as you cannot dinamically register services in the decoders at runtime, as they have to be included in the decoder before bundling.

@here/harp-datasource-protocol/lib/WorkerServiceProtocol.ts
amsXYZ: You're right, connecting is a way more common use-case and will affect more the UX, but think that warning on destruction is relevant. 

You're right it might be out of scope though,  so we have to make sure to tackle it in the future.

@here/harp-geojson-datasource/lib/GeoJsonDataProvider.ts
zbigg: Of course you can dynamically register new service types, but again this is super advanced (and not sure if makes sense) so i assume that one that does tricks like this understand consequences.

PR #764, opened 2019-09-02
Harp-6270: Gather tiles elevation, store their min/max and update clip planes.

kkostecki:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-examples/src/datasource_features_polygons.ts
+++ b/@here/harp-examples/src/datasource_features_polygons.ts
@@ -13,7 +13,7 @@ import {
 import { GeoCoordinates, sphereProjection } from "@here/harp-geoutils";
 import { MapControls } from "@here/harp-map-controls";
 import { CopyrightInfo, MapView } from "@here/harp-mapview";
-import { AltitudeBasedClipPlanesEvaluator } from "@here/harp-mapview/lib/ClipPlanesEvaluator";
+import { TopViewClipPlanesEvaluator } from "@here/harp-mapview/lib/ClipPlanesEvaluator";
 import { APIFormat, OmvDataSource } from "@here/harp-omv-datasource";
 import * as THREE from "three";
 import { accessToken } from "../config";
@@ -282,7 +282,7 @@ export namespace PolygonsFeaturesExample {
             canvas,
             theme: "resources/berlin_tilezen_night_reduced.json",
             projection: sphereProjection,
-            clipPlanesEvaluator: new AltitudeBasedClipPlanesEvaluator(300000)
+            clipPlanesEvaluator: new TopViewClipPlanesEvaluator(300000)
         });
         mapView.setCameraGeolocationAndZoom(new GeoCoordinates(20, 30), 3.9);
         mapView.renderLabels = false;
--- a/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
+++ b/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
@@ -26,11 +26,31 @@ export interface ViewRanges {
      */
     far: number;
     /**
-     * Maximum possible visibility range in camera space, regardless of camera orientation.
-     * @note Holds the maximum value that may be returned for [[far]] plane distance at
-     * certain camera position and orientation. You may use this value to calculate
-     * fog or other depth effects that are related to frustum (visibility), but should not
-     * change as dynamically as near/far plane distances on certain evaluators.
+     * Minimum distance that may be applied to near plane.
+     *
+     * Reflects minimum possible near plane distance regardless of camera orientation.
+     *
+     * @note Such constraint is always required because near plane can not be placed at zero
+     * distance from camera (regardless of rendering Api used). Moreover this value may be
+     * used as input for algorighms related to frustum planes, but rather based on visibility
+     * ranges such as it does not change during tilt. Frustum planes may change dynamically with
+     * camera orientation changes, while this value preserve constness for certain camera
+     * position and may be used for effects like near geometry fading.
+     */
+    minimum: number;
+    /**
+     * Maximum possible distance for far plane placement.
+     *
+     * This accounts for maximum visibility range in camera space, regardless of camera
+     * orientation. Far plane will never be placed beyond that distance, this value says
+     * about viewing distance limit, thus it is constrained for performance reasons and
+     * allows to compute effects applied at the end of viewing range such as fog or
+     * geometry fading.
+     *
+     * @note Holds the maximum distance that may be applied for [[far]] plane at
+     * certain camera position, it is const in between orientation changes. You may use this
+     * value to calculate fog or other depth effects that are related to frustum planes,
+     * but should not change as dynamically as current near/far planes distances.
      */
     maximum: number;
 }
@@ -126,88 +146,34 @@ export class InterpolatedClipPlanesEvaluator implements ClipPlanesEvaluator {
             assert(false, "Unsuported projection type");
         }
 
-        const viewRanges: ViewRanges = { near: nearPlane, far: farPlane, maximum: farPlane };
+        const viewRanges: ViewRanges = {
+            near: nearPlane,
+            far: farPlane,
+            minimum: this.nearMin,
+            maximum: farPlane
+        };
         return viewRanges;
     }
 }
 
 /**
- * Simple camera clip planes evaluator that computes near plane based on camera to ground distance.
+ * Abstract evaluator class that adds support for elevation constraints.
  *
- * This evaluator supports both planar and spherical projections, although it behaviour is
- * slightly different in each case. General algorithm sets near plane and far plane close
- * to ground level, but taking into account some predefined margins.
- * @note This simple evaluator does behave correctly when camera is always pointing downwards the
- * ground surface (top view) along surface normal, but does not preserve correct planes when using
- * modifying camera pitch (tilt) angle. Suitable only top view camera modes.
+ * Classes derived from this should implement algorithms that takes into account rendered
+ * features height (elevations), such as ground plane is no more flat (or spherical), but
+ * contains geometry that should be overlapped by frustum planes.
  */
-export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
-    /**
-     * Helper for reducing number of objects created at runtime.
-     */
-    protected m_tmpVectors: THREE.Vector3[] = [
-        new THREE.Vector3(),
-        new THREE.Vector3(),
-        new THREE.Vector3()
-    ];
-    /**
-     * Helper object for reducing performance impact.
-     */
-    protected m_tmpQuaternion: THREE.Quaternion = new THREE.Quaternion();
+export abstract class ElevationBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
     private m_maxElevation: number;
     private m_minElevation: number;
-    private m_minimumViewRange: ViewRanges;
 
-    /**
-     * Allows to setup near/far offsets (margins), rendered geometry elevation relative to sea
-     * level as also minimum near plane and maximum far plane distance constraints.
-     * It is strongly recommended to set some reasonable [[nearFarMargin]] (offset) between near
-     * and far planes to avoid flickering.
-     * @param maxElevation defines near plane offset from the ground in the surface normal
-     * direction, positive values allows to render elevated terrain features (mountains,
-     * buildings).
-     * @param minElevation defines far plane offset from the ground surface, negative values moves
-     * far plane below the ground level (use it to render depressions).
-     * @param nearMin minimum allowable near plane distance from camera, must be bigger then zero.
-     * @param nearFarMargin minimum distance between near and far plane, have to be significantly
-     * bigger then zero (especially if [[maxElevation]], [[minElevation]] are equal), otherwise you
-     * may notice flickering during rendering, or even render empty scene if frustum planes are
-     * almost equal.
-     * @param farMaxRatio maximum ratio between ground and far plane distance, allows to limit
-     * viewing distance at overall. Have to be bigger then 1.0.
-     * @note Keep in mind that this evaluator does not evaluate terrain (or building) elevation
-     * automatically, to keep such features rendered (between frustum planes) use [[minElevation]],
-     * [[maxElevation]] constraints. You may change this parameters at any time, but it requires
-     * repeating [[evaluatePlanes]] step, if your camera is moving you need to evaluate planes
-     * anyway.
-     * @note You may treat [[minElevation]] and [[maxElevation]] parameters as the maximum and
-     * minimum renderable elevation respectivelly along the surface normal, when camera is
-     * constantly looking downwards (top-down view). If you need [[ClipPlanesEvaluator]] for
-     * cameras that support tilt or yaw please use [[TiltBasedClipPlanesEvaluator]].
-     * @note [[nearFarMaxRatio]] does not limit far plane when spherical projection is in use,
-     * the algorithm used there estimates distance to point on tangent where line from camera
-     * touches the sphere horizont and there is no reason to clamp it.
-     */
-    constructor(
-        maxElevation: number = 200,
-        minElevation: number = -50,
-        readonly nearMin: number = 0.1,
-        readonly nearFarMargin: number = 1.0,
-        readonly farMaxRatio = 1.8
-    ) {
+    constructor(maxElevation: number, minElevation: number) {
         assert(maxElevation >= minElevation);
-        assert(nearMin > 0);
-        assert(nearFarMargin > epsilon);
-        assert(farMaxRatio > 1.0);
         this.m_minElevation = minElevation;
         this.m_maxElevation = maxElevation;
-        this.m_minimumViewRange = {
-            near: nearMin,
-            far: nearMin + nearFarMargin,
-            maximum: Math.max(nearMin * farMaxRatio, nearMin + nearFarMargin)
-        };
     }
 
+    abstract evaluateClipPlanes(camera: THREE.Camera, projection: Projection): ViewRanges;
     /**
      * Set maximum elevation above sea level to be rendered.
      * @param elevation the elevation (altitude) value in world units (meters).
@@ -259,6 +225,83 @@ export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
     get minElevation(): number {
         return this.m_minElevation;
     }
+}
+
+/**
+ * Top view, clip planes evaluator that computes view ranges based on ground distance and elevation.
+ *
+ * This evaluator supports both planar and spherical projections, although it behaviour is
+ * slightly different in each case. General algorithm sets near plane and far plane close
+ * to ground level, but taking into account maximum and minimum elevation of features on the ground.
+ *
+ * @note This evaluator supports only cameras which are always looking down the ground surface
+ * (top-down view) along surface normal and does not preserve correct clip planes when
+ * modifying camera pitch (tilt) angle. In simple words it is suitable only for top view camera
+ * settings.
+ */
+export class AltitudeBasedClipPlanesEvaluator extends ElevationBasedClipPlanesEvaluator {
+    /**
+     * Helper for reducing number of objects created at runtime.
+     */
+    protected m_tmpVectors: THREE.Vector3[] = [
+        new THREE.Vector3(),
+        new THREE.Vector3(),
+        new THREE.Vector3()
+    ];
+    /**
+     * Helper object for reducing performance impact.
+     */
+    protected m_tmpQuaternion: THREE.Quaternion = new THREE.Quaternion();
+    private m_minimumViewRange: ViewRanges;
+
+    /**
+     * Allows to setup near/far offsets (margins), rendered geometry elevation relative to sea
+     * level as also minimum near plane and maximum far plane distance constraints.
+     * It is strongly recommended to set some reasonable [[nearFarMargin]] (offset) between near
+     * and far planes to avoid flickering.
+     * @param maxElevation defines near plane offset from the ground in the surface normal
+     * direction, positive values allows to render elevated terrain features (mountains,
+     * buildings).
+     * @param minElevation defines far plane offset from the ground surface, negative values moves
+     * far plane below the ground level (use it to render depressions).
+     * @param nearMin minimum allowable near plane distance from camera, must be bigger then zero.
+     * @param nearFarMargin minimum distance between near and far plane, have to be significantly
+     * bigger then zero (especially if [[maxElevation]], [[minElevation]] are equal), otherwise you
+     * may notice flickering during rendering, or even render empty scene if frustum planes are
+     * almost equal.
+     * @param farMaxRatio maximum ratio between ground and far plane distance, allows to limit
+     * viewing distance at overall. Have to be bigger then 1.0.
+     * @note Keep in mind that this evaluator does not evaluate terrain (or building) elevation
+     * automatically, to keep such features rendered (between frustum planes) use [[minElevation]],
+     * [[maxElevation]] constraints. You may change this parameters at any time, but it requires
+     * repeating [[evaluatePlanes]] step, if your camera is moving you need to evaluate planes
+     * anyway.
+     * @note You may treat [[minElevation]] and [[maxElevation]] parameters as the maximum and
+     * minimum renderable elevation respectivelly along the surface normal, when camera is
+     * constantly looking downwards (top-down view). If you need [[ClipPlanesEvaluator]] for
+     * cameras that support tilt or yaw please use [[TiltBasedClipPlanesEvaluator]].
+     * @note [[nearFarMaxRatio]] does not limit far plane when spherical projection is in use,
+     * the algorithm used there estimates distance to point on tangent where line from camera
+     * touches the sphere horizont and there is no reason to clamp it.
+     */
+    constructor(
+        maxElevation: number = EarthConstants.MAX_ELEVATION,
+        minElevation: number = EarthConstants.MIN_ELEVATION,
+        readonly nearMin: number = 0.1,
+        readonly nearFarMargin: number = 10.0,
+        readonly farMaxRatio = 1.8
+    ) {
+        super(maxElevation, minElevation);
+        assert(nearMin > 0);
+        assert(nearFarMargin > epsilon);
+        assert(farMaxRatio > 1.0);
+        this.m_minimumViewRange = {
+            near: nearMin,
+            far: nearMin + nearFarMargin,
+            minimum: this.nearMin,
+            maximum: Math.max(nearMin * farMaxRatio, nearMin + nearFarMargin)
+        };
+    }
 
     /**
      * Get minimum view range that is possible to achieve with current evaluator settings.
@@ -305,11 +348,12 @@ export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
         // Apply the constraints.
         nearPlane = Math.max(nearPlane - this.nearFarMargin / 2, this.nearMin);
         farPlane = Math.min(farPlane, farMax);
-        farPlane = Math.max(farPlane, nearPlane + this.nearFarMargin);
+        farPlane = Math.max(farPlane + this.nearFarMargin / 2, nearPlane + this.nearFarMargin);
 
         const viewRanges: ViewRanges = {
             near: nearPlane,
             far: farPlane,
+            minimum: this.nearMin,
             maximum: Math.max(farMax, farPlane)
         };
         return viewRanges;
@@ -408,9 +452,14 @@ export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
         // In extreme cases the largest depression assumed may be further then tangent
         // based far plane distance, take it into account
         farPlane = Math.max(farPlane, cameraAltitude - this.minElevation);
-        farPlane = Math.max(farPlane, nearPlane + this.nearFarMargin);
+        farPlane = Math.max(farPlane + this.nearFarMargin / 2, nearPlane + this.nearFarMargin);
 
-        const viewRanges: ViewRanges = { near: nearPlane, far: farPlane, maximum: farPlane };
+        const viewRanges: ViewRanges = {
+            near: nearPlane,
+            far: farPlane,
+            minimum: this.nearMin,
+            maximum: farPlane
+        };
         return viewRanges;
     }
 }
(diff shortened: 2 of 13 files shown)


@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -25,6 +25,10 @@ export interface ViewRanges {
      * @note Should be always positive and bigger then [[near]] plane distance.
      */
     far: number;
+    /**
+     * Minimum viewing distance that may apply to near plane.
```
ninok: What does this mean?

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -124,11 +128,83 @@ export class InterpolatedClipPlanesEvaluator implements ClipPlanesEvaluator {
...
+    /**
+     * Get minimum elevation to be covered by camera frustum.
+     */
+    get minElevation(): number {
+        return this.m_minElevation;
+    }
+}
+
 /**
  * Simple camera clip planes evaluator that computes near plane based on camera to ground distance.
```
ninok: Please update the description. This evaluator is far from simple

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -139,7 +215,7 @@ export class InterpolatedClipPlanesEvaluator implements ClipPlanesEvaluator {
  * ground surface (top view) along surface normal, but does not preserve correct planes when using
  * modifying camera pitch (tilt) angle. Suitable only top view camera modes.
  */
-export class AltitudeBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
+export class AltitudeBasedClipPlanesEvaluator extends ElevationBasedClipPlanesEvaluator {
```
ninok: Why do we actually need these two? What is the difference?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -404,6 +404,12 @@ export class VisibleTileSet {
                 maxElevation = Math.max(maxElevation, tile.maxElevation);
             });
         });
+        // We process buildings elevation yet (via ElevationRangeSource or either way), so correct
+        // the highest elevation to account for highest possible building.
+        maxElevation =
+            maxElevation < EarthConstants.MAX_ELEVATION - EarthConstants.MAX_BUILDING_HEIGHT
+                ? maxElevation + EarthConstants.MAX_BUILDING_HEIGHT
+                : maxElevation;
```
ninok: Shouldn't it be:`: EarthConstants.MAX_ELEVATION` ?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -404,6 +404,12 @@ export class VisibleTileSet {
                 maxElevation = Math.max(maxElevation, tile.maxElevation);
             });
         });
+        // We process buildings elevation yet (via ElevationRangeSource or either way), so correct
```
ninok: I don't like the idea of just increasing the maxElevation by the height of the highest building b/c:
1. You always do this, even if no buildings are rendered (e.g. they are disabled in the style or we just don't render vector data but raster tiles)
2. You don't know how high the actual data is. In the style the building heights could be multiplied by two or in the case of geojson it could be an arbitrary height.

I think we have to improve the min-max elevation if there is no elevationProvider and if there is an elevationProvider the min/max height of the tile should be added.

@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -142,19 +170,24 @@ export class FrustumIntersection {
...
                         calculationFinal &&
                         range.calculationStatus === CalculationStatus.FinalPrecise;
                 }
 
                 let subTileArea = 0;
 
-                if (this.m_projection.type === ProjectionType.Spherical) {
+                const strictIntersections: boolean =
+                    this.m_projection.type === ProjectionType.Spherical;
+                // || useElevationRangeSource;
```
ninok: please remove

@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -142,19 +170,24 @@ export class FrustumIntersection {
...
+                if (strictIntersections) {
```
ninok: Not sure if strict is the proper name, b/c in planar projection the obb and the aabb are identical so they are both strict.

@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -142,19 +170,24 @@ export class FrustumIntersection {
 
                 const geoBox = getGeoBox(tilingScheme, childTileKey, offset);
 
+                // For tiles without elevation range source, default 0 will be propagated
```
ninok: So if there is no elevationRangeSource you might want to check if the tile is already is in the cache and get the elevation from there.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -363,13 +387,50 @@ export class VisibleTileSet {
...
+        }
+        this.m_elevationRange.calculationStatus = visibleTileKeysResult.allBoundingBoxesFinal
+            ? CalculationStatus.FinalPrecise
+            : CalculationStatus.PendingApproximate;
+
+        // If clip planes evaluator depends on the tiles elevation re-calculate
+        // frustum planes and update the camera near/far plane distances.
+        const clipPlanesEvaluator = this.options.clipPlanesEvaluator;
+        let viewRangesChanged: boolean = false;
+        if (clipPlanesEvaluator instanceof ElevationBasedClipPlanesEvaluator) {
```
ninok: This is quite ugly. Please add a setter for min and max elevation to the base ClipPlanesEvaluator interface. The ClipPlanesEvaluator will then decide whether it makes use of these values or not.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -124,11 +128,83 @@ export class InterpolatedClipPlanesEvaluator implements ClipPlanesEvaluator {
...
+            maximum: farPlane
+        };
         return viewRanges;
     }
 }
 
+/**
+ * Add support for elevation constraints.
+ */
+export abstract class ElevationBasedClipPlanesEvaluator implements ClipPlanesEvaluator {
```
ninok: I have the impression we are overengineering this stuff a bit. Why do we need so many evaluators. I think there should be one that solves everything and if the user needs to customize he/she could provide it's own.
Also be aware that this is all public API that we have to maintain. The only thing that should actually be in our API is the ClippingPlanesEvaluator base interface.

@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -66,12 +72,28 @@ export class FrustumIntersection {
...
+     */
+    get projection(): Projection {
+        return this.m_projection;
+    }
+
     /**
      * Updates the frustum to match the current camera setup.
      */
-    updateFrustum() {
+    updateFrustum(projectionMatrixOveride?: THREE.Matrix4) {
```
nzjony: I'm not sure this is a good idea, shouldn't the m_camera.projectionMatrix always be valid? Now you are creating a frustum which doesn't align with the camera perspective, this could create some problems?

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -588,27 +615,29 @@ export class TiltBasedClipPlanesEvaluator extends AltitudeBasedClipPlanesEvaluat
...
 
     private evaluateSphericalProj(camera: THREE.Camera, projection: Projection): ViewRanges {
         assert(projection.type === ProjectionType.Spherical);
-        const clipPlanes = { ...this.minimumViewRange };
+        const viewRanges = { ...this.minimumViewRange };
 
         // Near plance calculus is pretty straightforward and does not depent on camera tilt:
         const cameraAltitude = this.getCameraAltitude(camera, projection);
-        clipPlanes.near = cameraAltitude - this.maxElevation;
+        viewRanges.near = cameraAltitude - this.maxElevation;
```
amsXYZ: Just as a note, despite being a good solution, we need to know what are the consequences of not pushing the nearPlane as aggresively as possible in our current setup.

This pushes the ``nearPlane`` to minimum values at street levels (or any height smaller than the Burj Khalifa), which might be nice to avoid nearPlaneClipping, but absolutely destroys the [depth buffer precision](https://developer.nvidia.com/content/depth-precision-visualized):

```
The graph above shows the “standard”, vanilla depth mapping used in D3D and similar APIs. You can immediately see how the 1/z curve leads to bunching up values close to the near plane, and the values close to the far plane are quite spread out.

It's also easy to see why the near plane has such a profound effect on depth precision. Pulling in the near plane will make the d range skyrocket up toward the asymptote of the 1/z curve, leading to an even more lopsided distribution of values.
```

Against this problem, we have 2 possible solutions:
* Push the min nearPlane to some distance that provides enough depth precision and fade fragment in a certain distance from the nearPlane (simple).
* Support a proper **ReversedDepth Projection matrix** which distributes the precision more evenly across the whole frustum, removing the precision dependency on the nearPlane. This one I like, but we need to get a few PRs into three to properly make it work...

nzjony: @amsXYZ, what do you mean that pushing the near plane to minimum values at street level destroys the depth buffer precision? If the near plane is as close as possible to the scene, then we have the best precision, to the near plane, i.e. close to the streets, which is what we want. So if you have a max elevation of 10 and min of 0, then the near -> far plane range would be 10, I don't see how you can calculate a depth range better than that.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: I have added some more notes to documentation. Quickly this is required setting for any rendering Api, cause we can not set near plane to zero, so usually it will hold some "epsilon" number, but we leave the flexibility to the user. 
The entire concept of ViewRanges is quite known in rendering engines cause it allows several effects to be implemented, for example imagine that we want to render real geometry in some range but put some special effects far behind or before, this is one of possible scenarios... He we use those ranges for fading and fog calculations that need to be tilt independent, this means that we can not base this effects on pure near/far planes, because those will be adjusted whenever tilt changes in order to preserve depth buffer precision.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: Updated.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: I have provided the abstract class that allows for implementing other algorithms (classes) that still supports elevation, but maybe using different methods - simpler or more sophisticated. Later on we have two classes derived:
AltitudeBasedClipPlanesEvaluator - top view only (I would like it to rename it for that one, but not in the same patchset cause refactoring could make it even bigger),
TiltBasedClipPlanesEvaluator - full support for any camera angles.

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: Nope, we want to take into account all buildings possible that otherwise could be clipped out by near plane. Expression is created this way, that it always preserve EarthConstants.MAX_ELEVATION as absolute maximum, either way we already know that there is no buildings on Mt.Everest or any 8000er :).
Further on there is idea to sample tiles geometry and provide actual buildings data (maximum height on the tile), but this is not a part of that topic/PR/issue. Here we need to take terrain into account, yet without breaking with buildings rendering.

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: We do not yet have the ElevationRangeSource that samples buildings data, but yes it can be improved to use real data. Although it is not a part of this ticket/issue/PR and I would not like to spread it more.... I we do not account for building height they will be (building tops) clipped when camera closes to the ground, before my stuff there we some magic numbers used, so I assume this is still better then was. See also my previous comment.

@here/harp-mapview/lib/FrustumIntersection.ts
kkostecki: OK.

@here/harp-mapview/lib/FrustumIntersection.ts
kkostecki: Ok.

@here/harp-mapview/lib/FrustumIntersection.ts
kkostecki: I have corrected comment here, it means that getGeoBox always returns box with altitude min/max equal to zero.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: As you wish, but then lots of work goes to trash... Keep in mind that for globe projection AltitudeBased (top view) works just fine and for planar projections you may also use this one for all 2d like (top-view) maps - faster and easier to understand. The later ones are:
- Interpolated - kept only to have backward compatibility until everything is done (without it we should fix all with single commit - impossible + no body would review that)
- Fixed - that was there always, but hardcoded into MapView / demo app.
- TiltBased - this just adds tilt support to AltitudeBased.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -870,7 +937,25 @@ export class VisibleTileSet {
...
 
-        this.m_frustumIntersection.updateFrustum();
+        // If elevation is to be taken into account create extended frustum:
+        // (near ~0, far: maxVisibilityRange) that allows to consider tiles that
+        // are far below ground plane and high enough to intersect the frustum.
+        if (elevationRangeSource !== undefined) {
+            const fp = MapViewUtils.getCameraFrustumPlanes(this.m_frustumIntersection.camera);
+            fp.near = this.m_viewRange.minimum;
+            fp.far = this.m_viewRange.maximum;
+            this.m_projectionMatrixOverride.makePerspective(
```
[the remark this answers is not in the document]
kkostecki: Please read the documentation:
https://threejs.org/docs/#api/en/math/Matrix4.makePerspective
Stackoverflow:
https://stackoverflow.com/questions/18064852/threejs-makeperspective-method-has-been-removed-or-changed
or even better see the Matrix4.js sources.

As far as I see, this method is no longer supported, but still exposed in the .ts wrapper.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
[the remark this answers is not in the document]
kkostecki: Do I need to do it in this commit?

@here/harp-mapview/lib/FrustumIntersection.ts
kkostecki: Nope. It is required for our purposes, because we need to probe intersection on the extended camera frustum, just to get idea how frustum should look like precisely after taking into account tiles elevations. I have no other idea how we could solve this chicked-egg problem nicer, although tried about 4 approaches already.
I you have some concepts please share...

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: @amsXYZ I think I could try with first approach (near fading), the second one is bold (as you also know), but still I would opt to merge this stuff, cause it already fixes some issues. I am REALLY afraid of extending this PR to infinity, such I have with previous ticket, cause we already agreed to do things step by step, as long as they are not moving us back, and this one definitely moves us forward, just compare InterpolatedClipPlanesEvaluator (legacy code) with new approaches - easy to test on MapViewDemo app (play with MapControls->Frustum->Planes evaluator, enable Debug Camera, turn on terrain). You will quickly note the difference.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -651,6 +651,18 @@ export class TiltBasedClipPlanesEvaluator extends AltitudeBasedClipPlanesEvaluat
...
+            // Now we need to account for camera tilt and frustum volume, so the longest
+            // frustum edge does not instersects with sphere, it takes the worst case
+            // scenario regardless of camera tilt, so may be improved little bit with more
+            // sophisticated algorithm.
+            // Note, the fov is vertical, otherwise we would need to
+            // translate it using aspect ratio:
+            // let aspect = camera.aspect > 1 ? camera.aspect : 1 / camera.aspect;
+            const aspect = 1.0;
+            const halfFovAngle = MathUtils.degToRad((camera.fov / 2) * aspect);
+            viewRanges.near *= Math.cos(halfFovAngle);
```
ninok: This should account for the tilt but there is no tilt angle used. Can you please elaborate a bit on the math.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -651,6 +651,18 @@ export class TiltBasedClipPlanesEvaluator extends AltitudeBasedClipPlanesEvaluat
         // Near plance calculus is pretty straightforward and does not depent on camera tilt:
         const cameraAltitude = this.getCameraAltitude(camera, projection);
         viewRanges.near = cameraAltitude - this.maxElevation;
+        if (camera instanceof THREE.PerspectiveCamera) {
```
ninok: Tilting is not only restricted to PerspectiveCamera

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
ninok: What goes to trash? You just merge everything into one class.
From the user point of view it just makes no sense to differentiate for example between tilt based and non tilt based evaluator. If the user is tilting it should just work.

We can maybe move the fixed one out and make a dedicated example that shows how to customize the near and far plane, but I doubt that there is big interest in this if our internal implementation is serving all the needs. We also have to check if anybody uses this in map-view-demo. If yes than we might move it back there and keep coresdk clean.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
nzjony: > so usually it will hold some "epsilon" number, but we leave the flexibility to the user.

Why does an epsilon value need to be configurable

If you want to use the maximum for fog, I think best rename it to something like `fogStart` or equivalent.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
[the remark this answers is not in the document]
nzjony: It could be done as a pre-step, but if the evaluators are to be merged to handle all use cases, then it isn't necessary.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
nzjony: If I understand, you want a top view only far plane evaluator that supports evaluation. I don't think it makes sense to have 3D terrain if the user can't tilt.

amsXYZ: What's the difference between the altitudeBased and tiltBased whenever there's no tilt? Because results look really similar.

If they're indeed the same, and the altitudeBased can only be used in 2D mode, why not just using the tiltBased always?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -870,7 +937,25 @@ export class VisibleTileSet {
             }
         });
 
-        this.m_frustumIntersection.updateFrustum();
+        // If elevation is to be taken into account create extended frustum:
+        // (near ~0, far: maxVisibilityRange) that allows to consider tiles that
+        // are far below ground plane and high enough to intersect the frustum.
+        if (elevationRangeSource !== undefined) {
+            const fp = MapViewUtils.getCameraFrustumPlanes(this.m_frustumIntersection.camera);
+            fp.near = this.m_viewRange.minimum;
```
nzjony: This seems to be the only use of the `minimum`, why not just hard code `0.1` and we save a lot of unnecessary code.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -870,7 +937,25 @@ export class VisibleTileSet {
...
+            fp.far = this.m_viewRange.maximum;
+            this.m_projectionMatrixOverride.makePerspective(
+                fp.left,
+                fp.right,
+                fp.bottom,
+                fp.top,
+                fp.near,
+                fp.far
+            );
+            this.m_frustumIntersection.updateFrustum(this.m_projectionMatrixOverride);
```
nzjony: Would it be more efficient to compute the min / max height using the ElevationRangeSource. i.e. find the common ancestor of all TileKeys in the scene and get the min / max elevation from that single tile?

@here/harp-mapview/lib/FrustumIntersection.ts
nzjony: Ahh yes, I had this issue in Olympia, I understand what you mean now. I made a comment below, I think it needs to be mentioned that this is required to avoid the cyclic dependency between near / far plane and visible tile set.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
nzjony: @kkostecki, I agree, I think we should focus rather in getting this PR through than fixing these issues... @amsXYZ , maybe create some issues for this and they can be followed up?

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: > > so usually it will hold some "epsilon" number, but we leave the flexibility to the user.
> 
> Why does an epsilon value need to be configurable

Epsilon may be dependent from the depth buffer precision (and format, int32, float) and thus rendering Api, moreover when processing map features (geo-projection) we may still notice some inaccuracy and rounding errors, if you would like to see the effect change the margin to some very small value, suddenly you will note that some of the features may be missing, even though the far plane lays exactly on the ground level (best to test with TopView - AltitudeBased evaluator).  
> If you want to use the maximum for fog, I think best rename it to something like `fogStart` or equivalent.
Unfortunately we may not relate maximum visibility range to fog, because there are certainly more features that depends on that, take Labels fading or geometry fading as example. Maximum visibility is the distance that far plane may be placed on at extreme tilt angles, we will never see more then that, but for top-view far plane is just below the ground so it is not our visibility limit, but rather far plane limit for that tilt angle, if you rotate camera to horizon far plane moves further, but max visibility remains the same. On the old (legacy) evaluator max_visibility == far plane (ALWAYS), because we never thought about depth buffer precision and shrinking the camera frustum to minimum. Since now, we differentiate this two values, because we want to have frustum as small as possible at certain tilt angles, of course at critical angles (tilt) this values will equalize.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: Basically only computation cost keeps us from using only TiltBased evaluator. If you know that the camera does not allow for tilt (as it is in current globe implementation), then AltitudeBased wins a lot. In the next commit I will rename them both to:
- TopViewClipPlanesEvaluator,
- TiltViewClipPlanesEvaluator.
just to clarify the things. The names are quite old and comes from the formulas used, by the time both evaluators changed a lot so no need to keep the old naming.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: I mean that if we remove entirely AltitudeBasedClipPlanesEvaluator (TopView after refactor) then we resign from still simple and fast evaluator that may be successfully used in any 2D map, or top-view situation.
Keep in mind that is works not only for planar projections, i.e. all globe examples should use it now (see DATASOURCE: features / polygons example), cause we don't have a tilt there. Moreover it is much easier to debug/test new features that are view range dependent, thus I would strongly recommend to keep it for a while (at least).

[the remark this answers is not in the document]
kkostecki: OK. I suggest to do it in the next commit, there is bunch of them already in PR.

@here/harp-mapview/lib/FrustumIntersection.ts
kkostecki: Yes perfect catch, this actually for cyclic dependency stuff. I will explain it in solutioning page related to the ticket.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: I think we have already agreed, that Burj Khalifa as our near constraint (distance limit) will be removed soon, of course I vote for @amsXYZ comments and believe that near fading seems to be most reasonable solution. I wrote about it in solutioning page too:
https://confluence.in.here.com/pages/viewpage.action?pageId=810200202

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: Yes but for Perspective one we need to make little bit more calculus. Either way I have already found some nice solution to simplify the code in this section, but I would not like to extend this PR with another commit. I would be more then happy to work on this a little more but with some new PR (ticket), cause its already difficult to maintain this one (rebase/merge cycle).

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -25,8 +25,19 @@ export interface ViewRanges {
...
+    /**
+     * Minimum viewing distance that may be applied to near plane.
+     *
+     * @note Relfects minimum possible near plane distance regardles of camera and
+     * projection settings. Such constraint is always required because near plane
+     * can not be placed at zero distance from camera (regardless of rendering Api),
+     * moreover it may be used as input for algorighms based on visibility ranges
+     * such as it does not change during tilt.
+     */
+    minimum: number;
```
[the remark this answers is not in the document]
kkostecki: I hope that is was explained few comments above. The near plane limit (minimum value) is usually dependent from the world scale. If your world is in units of millions it's reasonably to assume some number close to 1, but if you shrink to world to 0-1 then it need to be really small, otherwise you loose you geometry in front of camera. Here the defaults suits our projections, but if used provide its own, it may be totally different - thus configurable.

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: I think explanation is few comments above, when we discuss the reason for configurable nearMin value.

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: This approach is presented in the second PR, but it seems that requires more computations and propagating results between calls, so I guess this one is better:
https://github.com/heremaps/harp.gl/pull/762

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
ninok: So we have it mainly for performance optimizations? Since we only do the near/far plane evaluation once per frame it is not that critical and we should not do premature-optimzations and we should for sure not put this burden on the user to select the proper `ClipPlanesEvaluator`. If we really need to optimize for performance we should do it under the hood (i.e. have some `if( tilt !== 0)`)

kkostecki:
@harp-bot retest please

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: New solution has been uploaded as last commit, but of course still Perspective and Orthographics camera projections are differentiated. Anyway the code for TiltBased... is much more simpler and I think clearer, the graphs presenting math used are on the solutioning page now... I will try to write also few explanatory sentences.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -25,8 +25,19 @@ export interface ViewRanges {
      * @note Should be always positive and bigger then [[near]] plane distance.
      */
     far: number;
+    /**
+     * Minimum viewing distance that may be applied to near plane.
+     *
+     * @note Relfects minimum possible near plane distance regardles of camera and
```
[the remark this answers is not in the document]
kkostecki: Done.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -870,7 +937,25 @@ export class VisibleTileSet {
             }
         });
 
-        this.m_frustumIntersection.updateFrustum();
+        // If elevation is to be taken into account create extended frustum:
+        // (near ~0, far: maxVisibilityRange) that allows to consider tiles that
+        // are far below ground plane and high enough to intersect the frustum.
+        if (elevationRangeSource !== undefined) {
```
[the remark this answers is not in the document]
kkostecki: It is never to far - because of ViewRange.maximum  - mysterious maximum value which has been asked few times in this review.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: All stuff in the solutioning page.

kkostecki:

kkostecki:

@here/harp-examples/src/datasource_features_polygons.ts
```diff
@@ -282,7 +282,7 @@ export namespace PolygonsFeaturesExample {
             canvas,
             theme: "resources/berlin_tilezen_night_reduced.json",
             projection: sphereProjection,
-            clipPlanesEvaluator: new AltitudeBasedClipPlanesEvaluator(300000)
+            clipPlanesEvaluator: new TopViewClipPlanesEvaluator(300000)
```
ninok: This should be `TiltViewClipPlanesEvaluator` since @Astrak will merge Globe gestures soon.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: Nino just to keep you informed, by default MapViewDemo app will still use InterpolatedEvaluator (legacy/old one). We have special GUI for switching evaluators, this is already merged into master. In order to test (and compare) new evaluators you need to use MapControls->Frustum->PlanesEvaluator combo. If everything is fine and integrates with globe controls with tilt, we will switch default setup for new evaluator.

kkostecki:

PR #765, opened 2019-09-03
HARP-6512: Label elevation overlay.

atomicsulfate:
- Overlaid text geometry is updated back in the tile.
- Overlay of labels in a file is spread through 60 frames.

### DIFF
--- a/@here/harp-mapview/lib/geometry/PhasedTileGeometryManager.ts
+++ b/@here/harp-mapview/lib/geometry/PhasedTileGeometryManager.ts
@@ -56,6 +56,12 @@ export class PhasedTileGeometryManager extends TileGeometryManagerBase {
             ? this.updateSomeTiles(tiles)
             : this.updateAllTilesTogether(tiles);
 
+        if (this.m_tileUpdateCallback) {
+            for (const tile of tiles) {
+                this.m_tileUpdateCallback(tile);
+            }
+        }
+
         // updateTileObjectVisibility() has always to be called.
         needUpdate = this.updateTileObjectVisibility(tiles) || needUpdate;
 
@@ -85,7 +91,6 @@ export class PhasedTileGeometryManager extends TileGeometryManagerBase {
         for (const tile of tiles) {
             const phasedGeometryLoader = tile.tileGeometryLoader as PhasedTileGeometryLoader;
 
-            let limitReached = false;
             if (phasedGeometryLoader !== undefined) {
                 if (
                     phasedGeometryLoader.update(
@@ -98,14 +103,8 @@ export class PhasedTileGeometryManager extends TileGeometryManagerBase {
                         this.m_maxUpdatedTilePerFrame > 0 &&
                         numTilesUpdated >= this.m_maxUpdatedTilePerFrame
                     ) {
-                        limitReached = true;
+                        break;
                     }
-                }
-                if (this.m_tileUpdateCallback) {
-                    this.m_tileUpdateCallback(tile);
-                }
-                if (limitReached) {
-                    break;
                 }
             }
         }
@@ -160,9 +159,6 @@ export class PhasedTileGeometryManager extends TileGeometryManagerBase {
                     )
                 ) {
                     needUpdate = true;
-                }
-                if (this.m_tileUpdateCallback) {
-                    this.m_tileUpdateCallback(tile);
                 }
             }
         }
--- a/@here/harp-mapview/lib/geometry/overlayOnElevation.ts
+++ b/@here/harp-mapview/lib/geometry/overlayOnElevation.ts
@@ -6,8 +6,17 @@
 
 import { GeometryKind } from "@here/harp-datasource-protocol";
 import * as THREE from "three";
+import { ElevationProvider } from "../ElevationProvider";
+import { TextElement } from "../text/TextElement";
 import { Tile, TileObject } from "../Tile";
 
+/**
+ * Overlays the specified object's geometry on the elevation represented by the given displacement
+ * map .
+ *
+ * @param object The object to be overlaid.
+ * @param displacementMap Texture representing the elevation data used to overlay the object.
+ */
 function overlayObject(object: TileObject, displacementMap: THREE.DataTexture): void {
     if (!("material" in object)) {
         return;
@@ -19,6 +28,58 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture): 
         (material as any).displacementMap = displacementMap;
     }
 }
+
+const worldCoords = new THREE.Vector3();
+
+/**
+ * Overlays the specified coordinates on top of elevation data if available.
+ *
+ * @param tileCoords World coordinates relative to the `tile` center.
+ * @param elevationProvider Used to get elevation data.
+ * @param tile The tile to which the coordinates are relative.
+ */
+function overlayPosition(
+    tileCoords: THREE.Vector3,
+    elevationProvider: ElevationProvider,
+    tile: Tile
+) {
+    worldCoords.copy(tileCoords).add(tile.center);
+    const geoCoords = tile.mapView.projection.unprojectPoint(worldCoords);
+    const height = elevationProvider.getHeight(geoCoords);
+
+    if (height !== undefined) {
+        geoCoords.altitude = height;
+        tile.mapView.projection.projectPoint(geoCoords, worldCoords);
+        tileCoords.copy(worldCoords).sub(tile.center);
+    }
+}
+
+/**
+ * Overlays a text element on top of elevation data if available.
+ *
+ * @param textElement The text element whose geometry will be overlaid.
+ * @param elevationProvider Used to get elevation data.
+ * @param tile The tile the text element belongs to.
+ */
+function overlayTextElement(
+    textElement: TextElement,
+    elevationProvider: ElevationProvider,
+    tile: Tile
+) {
+    // TODO: Move calculation of text element geoCoordinates to decoder.
+    if (textElement.path === undefined) {
+        overlayPosition(textElement.position, elevationProvider, tile);
+        return;
+    }
+
+    for (const position of textElement.path) {
+        overlayPosition(position, elevationProvider, tile);
+    }
+}
+
+// Overlay of text elements within a tile is spread through this many frames to minimize the time
+// spent per frame on elevation overlay.
+const TEXT_OVERLAY_SPREAD_FRAME_COUNT = 60;
 
 /**
  * Overlays the geometry in the given tile on top of elevation data if available.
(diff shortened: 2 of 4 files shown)


@here/harp-mapview/lib/geometry/PhasedTileGeometryManager.ts
```diff
@@ -56,6 +56,12 @@ export class PhasedTileGeometryManager extends TileGeometryManagerBase {
             ? this.updateSomeTiles(tiles)
             : this.updateAllTilesTogether(tiles);
 
+        if (this.m_tileUpdateCallback) {
+            for (const tile of tiles) {
+                this.m_tileUpdateCallback(tile);
```
nzjony: is this ok to call this on a tile which wasn't updated. @StefanDachwitz , can you comment?

nzjony: I mean to say, above, based on the isDynamicFrame flag, you could have just updated some tiles, and not all, but here you call the callback on all tiles....

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -20,6 +23,42 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture):
     }
 }
 
+function overlayTextPosition(
```
nzjony: Please add some documentation

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -20,6 +23,42 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture):
...
+    tileCoords: THREE.Vector3,
+    elevationProvider: ElevationProvider,
+    projection: Projection,
+    tile: Tile
+) {
+    // TODO: Calculate geoCoordinates in decoder.
```
nzjony: I think this comment should go where this function is used, this function by itself seems perfectly appropriate.

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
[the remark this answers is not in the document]
atomicsulfate: done

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -20,6 +23,42 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture):
...
 }
 
+function overlayPosition(
+    tileCoords: THREE.Vector3,
+    elevationProvider: ElevationProvider,
+    projection: Projection,
+    tile: Tile
+) {
+    // TODO: Calculate geoCoordinates in decoder.
+    const worldCoords = tileCoords.clone().add(tile.center);
```
ninok: We should reduce creation of temporary objects. Please create a follow up ticket, if you did not already, and add that.

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
atomicsulfate: done

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
atomicsulfate: done

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
atomicsulfate: i'll fix it here.

@here/harp-mapview/lib/text/TextElement.ts
```diff
@@ -491,7 +491,7 @@ export class TextElement {
      */
     constructor(
         readonly text: string,
-        readonly points: THREE.Vector3[] | THREE.Vector3,
+        public points: THREE.Vector3[] | THREE.Vector3,
```
ninok: Why? Your are only changing the array elements, not the array instance?!

@here/harp-utils/lib/GroupedPriorityList.ts
```diff
@@ -45,7 +45,7 @@ export type PriorityListGroupMap<T extends PriorityListElement> = Map<number, Pr
  */
 export class GroupedPriorityList<T extends PriorityListElement> {
     readonly groups: PriorityListGroupMap<T> = new Map();
-
+    private m_sortedGroups: Array<PriorityListGroup<T>> | undefined;
```
ninok: Why do we need this? Does it matter in which order we overlay the labels?

StefanDachwitz: This is a merging issue. I introduced this to fix the disappearing point labels (HARP-6843: Labels disappear after initial load)

@here/harp-mapview/lib/geometry/PhasedTileGeometryManager.ts
atomicsulfate: that's correct. i need to call the overlay function for all tiles anyway, even if they weren't updated, for two reasons:
1. The elevation data may have changed because the final terrain tile is now loaded.
2. The text overlay will continue after the tile is loaded, since it's spread to multiple frames.

So right now the callback is called for any tile that goes through the TileGeometryManager.updateTiles() method.

I agree it's confusing though. I'll try to think on how to improve this.

@here/harp-utils/lib/GroupedPriorityList.ts
atomicsulfate: not really, but I can't index the priority lists in the "groups" map, that's why I use the sorted groups array.

@here/harp-mapview/lib/text/TextElement.ts
atomicsulfate: done.

@here/harp-utils/lib/GroupedPriorityList.ts
```diff
@@ -164,6 +179,9 @@ export class GroupedPriorityList<T extends PriorityListElement> {
             const normalizedPriority = Math.floor(priority);
             group = new PriorityListGroup<T>(normalizedPriority);
             this.groups.set(normalizedPriority, group);
+            if (this.m_sortedGroups) {
```
StefanDachwitz: truthy

@here/harp-utils/lib/GroupedPriorityList.ts
```diff
@@ -97,6 +103,9 @@ export class GroupedPriorityList<T extends PriorityListElement> {
             const group = this.findGroup(otherGroup[1].priority);
             if (group === undefined) {
                 this.groups.set(Math.floor(otherGroup[1].priority), otherGroup[1].clone());
+                if (this.m_sortedGroups) {
```
StefanDachwitz: truthy

@here/harp-utils/lib/GroupedPriorityList.ts
```diff
@@ -85,6 +88,9 @@ export class GroupedPriorityList<T extends PriorityListElement> {
      */
     clear(): void {
         this.groups.clear();
+        if (this.m_sortedGroups) {
```
StefanDachwitz: truthy

@here/harp-utils/lib/GroupedPriorityList.ts
```diff
@@ -73,6 +73,9 @@ export class GroupedPriorityList<T extends PriorityListElement> {
                 if (group.elements.length === 0) {
                     const normalizedPriority = Math.floor(element.priority);
                     this.groups.delete(normalizedPriority);
+                    if (this.m_sortedGroups) {
```
StefanDachwitz: truthy

@here/harp-utils/lib/GroupedPriorityList.ts
```diff
@@ -108,15 +117,21 @@ export class GroupedPriorityList<T extends PriorityListElement> {
      * Return a sorted list of [[PriorityListGroup]]s.
      */
     get sortedGroups(): Array<PriorityListGroup<T>> {
-        const groups: Array<PriorityListGroup<T>> = [];
+        if (this.m_sortedGroups && this.m_sortedGroups.length > 0) {
```
StefanDachwitz: truthy

@here/harp-utils/lib/GroupedPriorityList.ts
```diff
@@ -108,15 +117,21 @@ export class GroupedPriorityList<T extends PriorityListElement> {
...
+            return this.m_sortedGroups;
+        }
+
+        if (!this.m_sortedGroups) {
```
StefanDachwitz: truthy

@here/harp-utils/lib/GroupedPriorityList.ts
StefanDachwitz: OK, my bad. No merging issue.

@here/harp-utils/lib/GroupedPriorityList.ts
```diff
@@ -108,15 +117,21 @@ export class GroupedPriorityList<T extends PriorityListElement> {
      * Return a sorted list of [[PriorityListGroup]]s.
      */
     get sortedGroups(): Array<PriorityListGroup<T>> {
```
StefanDachwitz: This gives you a sorted list, just like m_sortedGroups, no? What is the difference?

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -20,6 +29,58 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture):
     }
 }
 
+const worldCoords = new THREE.Vector3();
```
StefanDachwitz: better "tempWorldCoords"

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -20,6 +29,58 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture):
...
+}
+
+/**
+ * Overlays a text element on top of elevation data if available.
+ *
+ * @param textElement The text element whose geometry will be overlaid.
+ * @param elevationProvider Used to get elevation data.
+ * @param tile The tile the text element belongs to.
+ */
+function overlayTextElement(
```
StefanDachwitz: I guess it would be better if the labels fade in _after_ they have been put on the overlay, to reduce the issue of labels moving.

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -20,6 +29,58 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture):
     }
 }
 
+const worldCoords = new THREE.Vector3();
+
+/**
+ * Overlays the specified coordinates on top of elevation data if available.
+ *
+ * @param tileCoords World coordinates relative to the `tile` center.
```
StefanDachwitz: Please add something like: "tileCoords" will be modified by terrain height at that position.

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -20,6 +29,58 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture):
...
+    }
+}
+
+// Overlay of text elements within a tile is spread through this many frames to minimize the time
+// spent per frame on elevation overlay.
+const TEXT_OVERLAY_SPREAD_FRAME_COUNT = 60;
+
 /**
  * Overlays the geometry in the given tile on top of elevation data if available.
  *
```
StefanDachwitz: There is a property like "fadingEnabled" in MapView. If fading is disabled, I think that all labels should be placed at once.
Also, it has to be checked that the FRAME_COMPLETE event is sent _after_ all labels have been set on the "overlay".

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -6,8 +6,17 @@
 
 import { GeometryKind } from "@here/harp-datasource-protocol";
```
StefanDachwitz: I am a bit unhappy with the name "overlay" here. Does it really fit? Is this an actual overlay, or is the height of the label object just modified by the terrain? I would suggest to find a name that explains better what is being done here, otherwise we have to explain why this overlay is not the kind overlay the rest of the world is using, it just "overlays", which is terrible English.

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
atomicsulfate: there is a update requested everytime we update the label overlay, that should delay the FRAME_COMPLETE event.

PR #770, opened 2019-09-04
Remove bogus text technique attribute.

zbigg:
This attribute was introduced by accident in #308 and never supported.


zbigg:
> Doesn't seem to be used anywhere in the codebase... Leftover from original implementation?

@amsXYZ 
As in description, my research shown that this was really introduced by accident (i've even looked at original repo in gerrit).

PR #772, opened 2019-09-04
Move InterpolatedProperty handling outside of StyleSetEvaluator.

zbigg:
Interpolator related code moved away from `StyleSetEvaluator`.
Only code move, no functional change.
(Needed by #695)

### DIFF
--- a/@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
+++ b/@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
@@ -3,19 +3,30 @@
  * Licensed under Apache 2.0, see full license in LICENSE
  * SPDX-License-Identifier: Apache-2.0
  */
+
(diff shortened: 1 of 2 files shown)


@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
```diff
@@ -119,7 +130,7 @@ export function getPropertyValue<T>(
     return getInterpolatedLength(property, level, pixelToMeters);
 }
 
-function getInterpolatedLength(
+export function getInterpolatedLength(
```
amsXYZ: Why do you need to export ``getInterpolatedLength`` and ``getInterpolatedColor``? ``getPropertyValue`` is hiding these details from the user, which might not know how to properly interpolate the value of a certain property.

zbigg: Accident.
Removed exports in next patch.

zbigg:

PR #773, opened 2019-09-04
HARP-6886: Z-fighting improvements.

amsXYZ:
Signed-off-by: Andrés Valencia Téllez <ext-andres.valencia-tellez@here.com>

Improved present z-fighting by:
* Fixing extrusionFeature attribute copy (broken depth pre-pass for non animated buildings).
* Removing the polygonOffset applied to meshes and applying instead a VS depth offset for edge geometries.

Happy contributing!


amsXYZ:
@zbigg Polygon offset cannot be applied to all primitives, only to triangle primitives. 

What we actually wanted to do in our scene is to **PULL** the outlines geometry towards the screen, but since we cannot apply WebGL PolygonOffset to the GL_LINES, we ended up having to **PUSH** all the extruded geometries away from the scene. 

In an initial implementation, this was fine, but now that we always render the ``groundPlane`` depth to our scene, this translated in low buildings undesirably clipping through the ground and presenting z-fighting issue when depthBuffer precision was not sufficient...

Current implementation solves both issues and performs a similar operation to what polygonOffset previously did.

PR #775, opened 2019-09-04
Support for continous zoom level inteprpolation in RoadPicker.

zbigg:
Support for continous zoom level inteprpolation in RoadPicker.
    
Makes road picking accurate when tilting map.

Width properties of `solid-line`s can be dynamically interpolated
in scope of one zoom level, so `widths` of road data have to be
interpolated based on actual zoom level.
    
    
Improvement will be visible once #734, is merged.

### DIFF
--- a/@here/harp-mapview/lib/RoadPicker.ts
+++ b/@here/harp-mapview/lib/RoadPicker.ts
@@ -7,11 +7,10 @@
 import {
     ExtendedTileInfo,
     getPropertyValue,
+    isInterpolatedProperty,
     LineTechnique,
-    SolidLineTechnique,
-    Technique
+    SolidLineTechnique
 } from "@here/harp-datasource-protocol";
-import { SolidLineMaterial } from "@here/harp-materials";
 import { assert, LoggerManager, Math2D } from "@here/harp-utils";
 import * as THREE from "three";
 import { MapView } from "./MapView";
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/RoadPicker.ts
```diff
@@ -192,20 +194,4 @@ export class RoadPicker {
             roadPickResult.userData = { ...objInfos[index] };
         }
     }
-
-    private getLineWidthInWorldUnit(technique: Technique, level: number): number | undefined {
-        const solidLineTech = technique as SolidLineTechnique;
-
-        if (solidLineTech.metricUnit === "Pixel") {
```
zbigg: We've just lost support for `metricUnit == Pixel`, TBF.

@here/harp-examples/src/datasource_object-picking.ts
```diff
@@ -0,0 +1,144 @@
+/*
```
ninok: @pit-rpg  added the same example in https://github.com/heremaps/harp.gl/pull/734.
So is @pit-rpg PR obsolete?

zbigg: @ninok No, i've based this on #734 only to be able to showcase the improvement.
Will remove this commit from PR and wait for #734.

PR #778, opened 2019-09-06
HARP-7097: Simply cutting tilted path

okanishchev:
Cut path at the points with angle more than the given one, then take longest uncut path segment.

Has disadvantage in case of tightly placed points, when each consequent point at a small distance from previous rotates the path in a smaller than a reference amount.
Approach with this disadvantage fixed is introduced in #785.

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { LineMarkerTechnique, TextStyleDefinition, Theme } from "@here/harp-datasource-protocol";
-import { ProjectionType } from "@here/harp-geoutils";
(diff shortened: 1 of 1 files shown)


amsXYZ:
Looks good! But I'm just worried about a few things:
* Cutting the path increases GC pressure on typesetting, which is not great.
* When we cut the paths, some labels can stop being visible since they no longer fit in the cut path.
* When we cut the paths, the placement position for text changes, leading to jumps in text position between frames...

Maybe calculating an offset to the text placement position to have the text placed in the straightest area of the path could solve all these issues?

okanishchev:
Regarding GC, we do zero to one slice per path label, which (in webkit) doesn't create a new object, but just a limited reference which as the original object would be orphaned after the frame ends. No so big pressure. But we can save just path limits and pass them to typesetter to avoid this if needed.

And we have a lot of labels, much more than other maps.

And the jumping is a real problem. Not sure we can solve it by moving the offset - if there are huge bends between the label's current and offsetted positions, or if the distance between them is big - the transition will look weird.

PR #781, opened 2019-09-09
HARP-7107: Introduce the property 'layer'.

robertoraggi:
The newly introduced property 'layer' can be used to restrict the
matching of techniques to a specific layer. For example, the following
two style conditions are equivalent:

            {
                "when": [
                    "all",
                    ["==", ["get", "$layer"], "water"],
                    ["==", ["get", "$geometryType"], "polygon"]
                ],
                ...
            }

            {
                "layer": "water",
                "when": [
                    "all",
                    ["==", ["get", "$geometryType"], "polygon"]
                ],
                ...
            }

### DIFF
--- a/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
+++ b/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
@@ -236,4 +236,65 @@ describe("StyleSetEvaluator", function() {
             assert.deepEqual(techniquesTileA[0], techniquesTileC[1]);
         });
     });
+
+    it("Filter techniques by layer", function() {
+        const styleSet: StyleSet = [
+            {
+                when: [
+                    "all",
+                    ["==", ["get", "$layer"], "buildings"],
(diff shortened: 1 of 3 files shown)


@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
```diff
@@ -236,4 +236,65 @@ describe("StyleSetEvaluator", function() {
...
+        const ev = new StyleSetEvaluator(styleSet);
+
+        const techniques = ev.getMatchingTechniques(
+            new MapEnv({ $layer: "buildings", $geometryType: "polygon" })
+        );
+
+        assert.equal(techniques.length, 4);
+        assert.equal(techniques[0].name, "extruded-polygon");
+        assert.equal(techniques[1].name, "text");
+        assert.equal(techniques[2].name, "solid-line");
```
ninok: Can you add descriptions to the styles and add them b/c from these asserts it is not clear which of the sold-line styles did match

PR #782, opened 2019-09-10
HARP-6816: Add the interpolate builtin operator.

robertoraggi:


@here/harp-datasource-protocol/lib/operators/InterpolationOperators.ts
```diff
@@ -0,0 +1,77 @@
...
+            for (let i = 0; i < samples.length; i += 2) {
+                const step = samples[i];
+                if (!(step instanceof NumberLiteralExpr)) {
+                    throw new Error("expected a numeric literal");
+                }
+                zoomLevels.push(step.value);
+                values.push(context.evaluate(samples[i + 1]));
+            }
+
+            const def: InterpolatedPropertyDefinition<any> = {
```
zbigg: You should create `InterpolatedProperty` here, which is "prepared" version of `InterpolatedPropertyDefinition` which is a way to define interpolation in JSON.
Use `createInterpolatedProperty`.

PR #785, opened 2019-09-10
HARP-7097: Using rope to cutoff path bends

okanishchev:
The rope algorithm:
1. Taking the rope of length d, and factor k
2. Placing the rope on the path on each point until end
3. On each point check if the rope length compared to the distance between it's endpoints is bigger than k
4. If so, cut the path at this point
5. Take the longest uncut segment of a path

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -150,6 +150,17 @@ const MIN_AVERAGE_CHAR_WIDTH = 5;
  */
 const LABEL_DIFFERENT_THRESHOLD_SQUARED = 4;
 
+/**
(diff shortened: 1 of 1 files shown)


okanishchev:
It solves the tightly placed points problem of more simple approach of #778.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -2197,6 +2216,111 @@ export class TextElementsRenderer {
         return numRenderedTextElements;
     }
 
+    private findLongestStraightPath(
```
zbigg: I know, it's early draft but consider creating this as free-function.
There is no use for `this`
Looks like complicated enough to deserve unit test.

PR #793, opened 2019-09-12
Minor: Add grid of images

nzjony:
Alternative to other PR: https://github.com/heremaps/harp.gl/pull/792


musculman:
Our examples should show the corresponding copyright by using the copyrightInfo box. But for the GIF/PNG shown on the README I guess it is not so important as it might be a cropped version of the full example.
As for the change, I am not a big fan of the grid, I prefer a simple PNG or  GIF version, I am no designer though :)
I'd say we could ask some of our colleagues to chime in: @adraksle, @ixirt , @kosoyka, Moritz

nzjony:
@dbabbs , do you have any opinion on the topic? How can we improve our first impression?

dbabbs:
I like the grid and think we should definitely keep it. In the future, let's swap out the images for a few more visually exciting examples, but for now this is good.

I did see you modified the simple bundle example to expand the css into multiple lines. While this is the generally accepted styling convention, I would recommend collapsing them back into one line each. This way we can save space (7 lines) on an important document (`readme.md`) by eliminating some relatively trivial required css.

nzjony:
Thanks for noticing the other changes, vscode does that automatically, I will remove those changes.

PR #795, opened 2019-09-13
HARP-6728: Cleanup configure/setStyleSet flow in TileDataSource

zbigg:
`TileDataSource` and `OmvDataSource` now automatically updates
owned tiles when decoder configuration (style, languages, options)
is changed.

Additionally, configure decoder only after successfull `#connect`.

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvDataSource.ts
+++ b/@here/harp-omv-datasource/lib/OmvDataSource.ts
@@ -241,17 +241,7 @@ export class OmvDataSource extends TileDataSource<OmvTile> {
             }
             throw error;
         }
-    }
-
-    /**
-     * Set a theme for the data source instance.
-     * @param styleSet The [[Theme]] to be added.
-     */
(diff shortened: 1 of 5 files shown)


@here/harp-omv-datasource/lib/OmvDataSource.ts
```diff
@@ -252,6 +252,7 @@ export class OmvDataSource extends TileDataSource<OmvTile> {
             return;
         }
         this.configureDecoder(styleSet, languages, this.m_decoderOptions);
+        this.mapView.markTilesDirty(this);
```
ninok: Is there a reason why we overwrite `setStyleSet`? Looking at `OmvDataSource.configureDecoder` it looks exactely the same as `TileDataSource.setStyleSet` except for the `options` but since OptionsMap is not related to `OmvDataSource` we should just move it to `TileDataSource`

zbigg: Agree, will update fix.

PR #796, opened 2019-09-13
HARP-7103: Support for 'ref' in Expr.fromJSON

zbigg:
Support for 'compile'-time' (`Expr.fromJSON`) reference resolving.

MapView changes:

* `DataSource.setTheme` revived as main theming interface for
  datasources because decoder based datasources require
  `theme.definitions` too
* `Definitions` are communicated from `MapView` to decoder
* `StyleSetEvaluator` now resolves references using `Expr.fromJSON`
* `ThemeLoader` doesn't resolve references by default
* (minor) BackgroundDataSource reacts to theme change

Rationale
We expect that objects and expressions in `Theme.definitions` can be big and this should be sent as one to decoders and we should reuse their instances to save memory.

### DIFF
--- a/@here/harp-datasource-protocol/lib/Expr.ts
+++ b/@here/harp-datasource-protocol/lib/Expr.ts
@@ -38,6 +38,7 @@ export function isJsonExpr(v: any): v is JsonExpr {
 interface ReferenceResolverState {
     definitions: Definitions;
     lockedNames: Set<string>;
+    cache: Map<string, Expr>;
 }
 
 /**
@@ -56,12 +57,29 @@ export abstract class Expr {
         return expr;
     }
 
-    static fromJSON(node: unknown, definitions?: Definitions) {
+    /**
+     * Parse expression in JSON form.
+     *
+     * If `definitions` are defined, then references (`['ref', name]`) are resolved.
+     *
+     * Paass `definitionExprCache` to reuse `Expr` instances created from definitions across
+     * many `fromJSON` calls.
+     *
+     * @param node expression in JSON format to parse
+     * @param definitions optional set of definitions needed definition resolved by `ref` operator
+     * @param definitionExprCache optional cache of `Expr` instances derived from `definitions`
+     */
+    static fromJSON(
+        node: unknown,
+        definitions?: Definitions,
+        definitionExprCache?: Map<string, Expr>
+    ) {
         const referenceResolverState: ReferenceResolverState | undefined =
             definitions !== undefined
                 ? {
                       definitions,
-                      lockedNames: new Set()
+                      lockedNames: new Set(),
+                      cache: definitionExprCache || new Map<string, Expr>()
                   }
                 : undefined;
 
@@ -223,22 +241,30 @@ export abstract class Expr {
             throw new Error(`definition '${name}' not found`);
         }
 
+        const cachedEntry = referenceResolverState.cache.get(name);
+        if (cachedEntry !== undefined) {
+            return cachedEntry;
+        }
         let definitionEntry = referenceResolverState.definitions[name] as any;
         if (isSelectorDefinition(definitionEntry)) {
             definitionEntry = definitionEntry.value;
         }
+        let result: Expr;
+
         if (isValueDefinition(definitionEntry)) {
-            return Expr.fromJSON(definitionEntry.value);
+            result = Expr.fromJSON(definitionEntry.value);
         } else if (isJsonExpr(definitionEntry)) {
             referenceResolverState.lockedNames.add(name);
             try {
-                return Expr.fromJSONInt(definitionEntry, referenceResolverState);
+                result = Expr.fromJSONInt(definitionEntry, referenceResolverState);
             } finally {
                 referenceResolverState.lockedNames.delete(name);
             }
         } else {
             throw new Error(`unsupported definition ${name}`);
         }
+        referenceResolverState.cache.set(name, result);
+        return result;
     }
 
     /**
--- a/@here/harp-datasource-protocol/lib/Expr.ts
+++ b/@here/harp-datasource-protocol/lib/Expr.ts
@@ -7,6 +7,7 @@
 import { ExprEvaluator, ExprEvaluatorContext, OperatorDescriptor } from "./ExprEvaluator";
 import { ExprParser } from "./ExprParser";
 import { ExprPool } from "./ExprPool";
+import { Definitions, isSelectorDefinition, isValueDefinition } from "./Theme";
 
 const exprEvaluator = new ExprEvaluator();
 
@@ -31,6 +32,15 @@ export function isJsonExpr(v: any): v is JsonExpr {
 }
 
 /**
+ * Internal state needed by [[Expr.fromJSON]] to resolve `"ref"` expressions.
+ * @hidden
+ */
+interface ReferenceResolverState {
+    definitions: Definitions;
+    lockedNames: Set<string>;
+}
+
+/**
  * Abstract class defining a shape of a [[Theme]]'s expression
  */
 export abstract class Expr {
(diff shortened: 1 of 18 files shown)


@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -75,16 +118,56 @@ export abstract class Expr {
...
                 if (node[2] !== undefined) {
-                    return Expr.makeCallExpr(op, node);
+                    return Expr.makeCallExpr(op, node, referenceResolverState);
                 }
                 if (typeof node[1] !== "string") {
                     throw new Error(`expected the name of an attribute`);
                 }
                 return new VarExpr(node[1]);
 
+            case "ref":
```
robertoraggi: could you move this to its own function.

zbigg: done

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -449,6 +559,33 @@ export class CaseExpr extends Expr {
     }
 }
 
+/**
+ * Expressions that represent literal/immediate values.
+ *
+ * @hidden
+ */
+export type LiteralExpr =
```
robertoraggi: please introduced a base abstract class `LiteralExpr` that extends `Expr` instead.

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -449,6 +559,33 @@ export class CaseExpr extends Expr {
...
+    | ObjectLiteralExpr
+    | NullLiteralExpr;
+
+/**
+ * Check if `expr` represents literal/immediate values i.e conforms to [[LiteralExpr]].
+ *
+ * @hidden
+ */
+export function isLiteralExpr(expr: Expr): expr is LiteralExpr {
+    return (
```
robertoraggi: please replace the isLiteralExpr with just an instanceof LiteralExpr when if you add the abstract class LiteralExpr

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -119,7 +168,7 @@ export abstract class Expr {
                 if (!(node.length % 2)) {
                     throw new Error("fallback is missing in 'match' expression");
                 }
-                const value = this.fromJSON(node[1]);
+                const value = this.fromJSONInt(node[1], referenceResolverState);
```
robertoraggi: fromJSONInt sounds a bit strange as name. Please use `fromJSONInternal`, `fromJSONHelper` something else

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -312,6 +413,15 @@ export class NullLiteralExpr extends Expr {
     accept<Result, Context>(visitor: ExprVisitor<Result, Context>, context: Context): Result {
         return visitor.visitNullLiteralExpr(this, context);
     }
+
+    /**
+     * Fake `value` getter to be compatible with other literal expressions.
```
robertoraggi: why do you need this? also `fake value` sounds a bit strange for public API.

zbigg: Each `LiteralExpr` has `value` property. So, null needs to have one too.

@here/harp-datasource-protocol/lib/Expr.ts
zbigg: Not needed now, if we have one base class.

@here/harp-datasource-protocol/lib/Expr.ts
zbigg: replaced with `parseNode` as to after `parseCall`

@here/harp-datasource-protocol/lib/Expr.ts
zbigg: Is this public API on the other hand, if it's marked with `@hidden` in doc ?

(nevertheless, i've removed offending comment).

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -46,9 +57,41 @@ export abstract class Expr {
         return expr;
     }
 
-    static fromJSON(node: unknown): Expr {
+    /**
+     * Parse expression in JSON form.
+     *
+     * If `definitions` are defined, then references (`['ref', name]`) are resolved.
+     *
+     * Paass `definitionExprCache` to reuse `Expr` instances created from definitions across
```
amsXYZ: Small typo:
```suggestion
     * Pass `definitionExprCache` to reuse `Expr` instances created from definitions across
```

@here/harp-datasource-protocol/lib/Expr.ts
zbigg: done

@here/harp-mapview/lib/ThemeLoader.ts
```diff
@@ -40,7 +40,7 @@ export interface ThemeLoadOptions {
     /**
      * Whether to resolve `ref` expressions in `definition` and `styles` elements.
      *
-     * @default `true`
+     * @default `false`, as featured datasources resolve definitions in [[StyleSetEvaluator]].
```
amsXYZ: I think the used of ``featured`` here is a bit confusing... I'd just say:
```suggestion
     * @default `false`, as datasources resolve definitions in [[StyleSetEvaluator]].
```

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -112,17 +112,14 @@ export abstract class DataSource extends THREE.EventDispatcher {
...
      */
     set styleSetName(styleSetName: string | undefined) {
         this.m_styleSetName = styleSetName;
-        if (
-            this.m_mapView !== undefined &&
-            styleSetName !== undefined &&
-            this.m_mapView.theme.styles !== undefined
-        ) {
-            this.setStyleSet(this.m_mapView.theme.styles[styleSetName]);
+        const mapView = this.m_mapView;
```
amsXYZ: What the purpose of this line?

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -209,13 +206,26 @@ export abstract class DataSource extends THREE.EventDispatcher {
...
+     * If `DataSource` depends on a `styleSet` or `languages`, it must update its tiles' geometry.
+     *
+     * @deprecated, Use [[setTheme]].
      *
      * @param styleSet The new theme that [[MapView]] uses.
      * @param languages An optional list of languages for the `DataSource`.
      */
     // tslint:disable-next-line:no-unused-variable
-    setStyleSet(styleSet?: StyleSet, languages?: string[]): void {
+    setStyleSet(styleSet?: StyleSet, definitions?: Definitions, languages?: string[]): void {
```
amsXYZ: I believe this is still used in a lot of our public examples. Don't say that the examples update should be in the same change, but please create a ticket to tackle these in the future.

@here/harp-mapview/lib/ThemeLoader.ts
zbigg: applied

@here/harp-mapview/lib/DataSource.ts
zbigg: What line ? ;)
(removed)

@here/harp-datasource-protocol/lib/Expr.ts
zbigg: applied, thx

@here/harp-mapview/lib/DataSource.ts
zbigg: Following up in #830

PR #797, opened 2019-09-13
Support for test execution on windows.

zbigg:
* github actions windows deserves it's own job definition - reenable most checks for windows
* `node-fetch` on windows, needs proper `file://` URL
* increase `LicenseHeaderCheck` as `glob` is slower on windows
* `PerformanceTimer#now` test fixed

### DIFF
--- a/.github/workflows/ci-linux.yaml
+++ b/.github/workflows/ci-linux.yaml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         node_version: [10, 11, 12]
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         node_version: [10, 11, 12]
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
```
musculman: Maybe you could rename ci as ci-linux. Afterall, if we merge this PR we would have three files for three OS

zbigg:
Test comment from bot

The 'foo test failed', here is diff: diff

.github/workflows/ci.yml
zbigg: Done.

PR #800, opened 2019-09-16
Update copyright year

dbabbs:
Marketing pointed out broken copyright on harp.gl site


docs/index.html
```diff
@@ -129,7 +129,7 @@ <h1 class="footer-title">harp.gl</h2>
                 </div>
                 <div>
                     <span>
-                        &copy; 2018-<span id="year"></span> <a target="_blank"
+                        &copy; 2019 <a target="_blank"
```
musculman: Are you sure it should be 2019? In the code you need to put it from the moment you started, that is why we have 2017-2019.

Issue #699, opened 2019-08-08
GeoJSON Point and LineString features with same coordinates display slightly differently


tomasberkeley:
![Screen Shot 2019-08-07 at 4 30 28 PM](https://user-images.githubusercontent.com/53882996/62665780-4e535480-b935-11e9-880d-a92d771f3aa3.png)

In the picture both the red dots and the green lines share the same exact coordinates however the red dots do not appear exactly on top of the line and exhibit a somewhat "wobbly" behavior.

Astrak:
Hello Tomas, thank you for your report. 

Can you detail your setup? How are you visualizing your GeoJson? What datasource + dataprovider do you pass it through?

tomasberkeley:
Hi Adrien,

I am using GeoJsonDataProvider(name, url) and OmvDataSource to create the datasource. I then add it to a MapView using a simple style sheet where I assign color with 
"technique": "circles" and with "technique": "solid-line" when "$geometryType == 'point'" and "$geometryType == 'line'", respectively.

To create the GeoJSON file I have a sequence of coordinates which I insert as "type": "LineString" to the file (these would be the green lines). To visualize the exact end-points of each line segment I also insert the same exact coordinates that formed the LineString, but insert them each separately as "type": "Point" (these are the red dots). I would expect the red dots to line up precisely on top of the green lines without the "wobbly" effect since both features share the same exact coordinates.

Thank you for looking into it!

Tomas

ninok:
Hi Michael,

I guess the basic setup is similar to the one of our geojson example:

- https://www.harp.gl/docs/master/examples/#datasource_geojson_chloropleth.html
- https://github.com/heremaps/harp.gl/blob/master/%40here/harp-examples/src/datasource_geojson_chloropleth.ts

The only thing you have to change are the geojson file and the styleset as described by Tomas.

Best regards,
Nino

nzjony:
It should be tested after #801  is merged to see if this is a fix.

nzjony:
I managed to reproduce with the following geojson.

{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          13.383072316646576,
          52.53059023621278
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          13.383429050445555,
          52.53068487407586
        ]
      }
    },
[...]

PR #801, opened 2019-09-17
HARP-6058: Fix GeoJSON lat2tile computation.

amsXYZ:

Happy contributing!

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvTileInfoEmitter.ts
+++ b/@here/harp-omv-datasource/lib/OmvTileInfoEmitter.ts
@@ -54,7 +54,6 @@ export class OmvTileInfoEmitter implements IOmvEmitter {
         featureId: number | undefined
     ): void {
         const tileInfoWriter = this.m_tileInfoWriter;
-
         const tmpV = new THREE.Vector3();
 
         for (const technique of techniques) {
--- a/@here/harp-omv-datasource/lib/OmvUtils.ts
+++ b/@here/harp-omv-datasource/lib/OmvUtils.ts
@@ -7,6 +7,7 @@ import { EarthConstants, webMercatorProjection } from "@here/harp-geoutils";
 import * as THREE from "three";
 
 import { OmvDecoder } from "./OmvDecoder";
+import { VTJsonDataAdapterId } from "./VTJsonDataAdapter";
 
 /**
  * @hidden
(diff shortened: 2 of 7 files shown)


@here/harp-omv-datasource/lib/OmvUtils.ts
```diff
@@ -22,8 +22,9 @@ export function isArrayBufferLike(data: any): data is ArrayBufferLike {
 /**
  * @hidden
  */
-export function lat2tile(lat: number, zoom: number): number {
-    return Math.floor(
+export function lat2tile(lat: number, zoom: number, round: boolean = false): number {
+    const func = round ? Math.round : Math.floor;
```
robertoraggi: maybe instead of passing a boolean you can just change the signature of the function and pass Math.round and Math.floor directly.

export function lat2tile(lat: number, zoom: number, converter = Math.round)  {
 return converter(...)
}

@here/harp-omv-datasource/lib/OmvTileInfoEmitter.ts
```diff
@@ -53,8 +53,8 @@ export class OmvTileInfoEmitter implements IOmvEmitter {
         techniques: IndexedTechnique[],
         featureId: number | undefined
     ): void {
+        const isGeoJSON = context.env.lookup("adapter") === "vt-json";
```
robertoraggi: maybe we should use something different than `isGeoJSON` here. Deciding how numbers get converted to integers is more general than vt-geojson. Also, we should probably avoid calling `context.env.lookup("adapter") === "vt-json";` for each feature, a property for required rounding behavior in `OmvDecoder.DecodeInfo`, initialized at construction time, may be less expensive to evaluate.

@here/harp-omv-datasource/lib/OmvUtils.ts
amsXYZ: Done.

@here/harp-omv-datasource/lib/OmvTileInfoEmitter.ts
amsXYZ: For now, decided to pass the ``DataAdapter`` id through the ``DecodeInfo`` struct to avoid lookups, but the logic is still based on which adapter you use. We can maybe move to a more generic approach in the future, but for now we only need to do this to contrarrest geojson-vt lat2tile logic.

@here/harp-omv-datasource/test/OmvDecodedTileEmitterTest.ts
```diff
@@ -47,7 +47,7 @@ describe("OmvDecodedTileEmitter", function() {
         const projection = mercatorProjection;
         const tileSizeOnScreen = 100;
 
-        const decodeInfo = new OmvDecoder.DecodeInfo(projection, tileKey, tileSizeOnScreen);
+        const decodeInfo = new OmvDecoder.DecodeInfo("", projection, tileKey, tileSizeOnScreen);
```
robertoraggi: this empty string literal looks weird.

amsXYZ: Done.

amsXYZ:

nzjony:
Does this potentially fix #699 ?

amsXYZ:
@nzjony I believe this PR is not gonna fix that issue, but we can double check.

PR #808, opened 2019-09-18
HARP-7201: All examples from Labeling section are broken

okanishchev:
Create `TextElementsRenderer` only if there's actual text elements created for any of the tiles, or overlay text exists. The `MapView` knows if there's text elements on the tile from the `Tile::hasTextElements()` that is checked for every tile every frame before one is found. Then the renderer creation is delayed to avoid affecting frame time.

If `MapView`'s theme was changed, remove text renderer completely. That allows to create new renderer with font catalogs and text styles changed if any text will be rendererd. And disable text rendering at all if none.

Has a drawback of loading font catalog files (json and png) every time the theme changes, even if they are the same. Not really crucial as files are taken from browser cache, but still.
Also there's a small overhead if no text is going to be rendered - checking every frame every tile owning some text elements, but it's can be avoided by using `mapView.renderLabels = false`.

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -710,6 +710,7 @@ export class MapView extends THREE.EventDispatcher {
     private m_detectedFps: number = FALLBACK_FRAME_RATE;
 
     private m_textElementsRenderer?: TextElementsRenderer;
+    private m_textElementsRendererTimer?: any;
     private m_textRenderStyleCache = new TextRenderStyleCache();
     private m_textLayoutStyleCache = new TextLayoutStyleCache();
     private m_overlayTextElements?: TextElement[] = [];
@@ -1050,6 +1051,11 @@ export class MapView extends THREE.EventDispatcher {
             this.m_movementFinishedUpdateTimerId = undefined;
         }
 
+        if (this.m_textElementsRendererTimer !== undefined) {
+            clearTimeout(this.m_textElementsRendererTimer);
+            this.m_textElementsRendererTimer = undefined;
+        }
+
         if (this.m_animationFrameHandle !== undefined) {
             cancelAnimationFrame(this.m_animationFrameHandle);
             this.m_animationFrameHandle = undefined;
@@ -1099,7 +1105,10 @@ export class MapView extends THREE.EventDispatcher {
         this.m_visibleTiles.setNumberOfVisibleTiles(Math.floor(numVisibleTiles));
         this.updateImages();
         this.updateLighting();
-        this.updateTextRenderer();
+
+        if (this.m_textElementsRenderer) {
+            this.m_textElementsRenderer.placeAllTileLabels();
+        }
         this.updateSkyBackground();
         this.update();
     }
@@ -1230,7 +1239,7 @@ export class MapView extends THREE.EventDispatcher {
         this.m_textRenderStyleCache.clear();
         this.m_textLayoutStyleCache.clear();
 
-        this.updateTextRenderer();
+        this.resetTextRenderer();
 
         if (this.m_theme.styles === undefined) {
             this.m_theme.styles = {};
@@ -1803,7 +1812,8 @@ export class MapView extends THREE.EventDispatcher {
         if (this.m_overlayTextElements !== undefined) {
             this.m_overlayTextElements = this.m_overlayTextElements.concat(textElements);
         }
-        this.updateTextRenderer();
+        this.createTextRendererIfNeeded();
+        this.m_textElementsRenderer!.placeAllTileLabels();
         this.update();
     }
 
@@ -2618,9 +2628,20 @@ export class MapView extends THREE.EventDispatcher {
 
         const renderList = this.m_visibleTiles.dataSourceTileList;
 
+        const hasTextRenderer = this.m_textElementsRenderer !== undefined;
+
+        let hasTextToRender =
+            this.m_overlayTextElements !== undefined && this.m_overlayTextElements.length > 0;
+        // no need to check everything if we're not going to create text renderer.
+        let needsTileCheck = !hasTextToRender && !hasTextRenderer && this.renderLabels;
         renderList.forEach(({ zoomLevel, renderedTiles }) => {
             renderedTiles.forEach(tile => {
                 this.renderTileObjects(tile, zoomLevel);
+
+                if (needsTileCheck && tile.hasTextElements()) {
+                    needsTileCheck = false;
+                    hasTextToRender = true;
+                }
                 //We know that rendered tiles are visible (in the view frustum), so we update the
                 //frame number, note we don't do this for the visibleTiles because some may still be
                 //loading (and therefore aren't visible in the sense of being seen on the screen).
@@ -2629,6 +2650,10 @@ export class MapView extends THREE.EventDispatcher {
                 tile.frameNumLastVisible = this.m_frameNumber;
             });
         });
+
+        if (hasTextToRender && !hasTextRenderer && this.renderLabels) {
+            this.enqueueTextRendererCreation();
+        }
 
         this.m_mapAnchors.children.forEach((childObject: MapAnchor) => {
             if (childObject.geoPosition === undefined) {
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -3059,7 +3059,7 @@ export class MapView extends THREE.EventDispatcher {
      * Gradually initialize & update TextRenderer as assets arrive.
      */
     private updateTextRenderer() {
-        if (this.m_theme.fontCatalogs === undefined) {
+        if (this.m_theme.fontCatalogs === undefined && this.m_theme.textStyles === undefined) {
```
zbigg: This reverts "fix" from #791 by requiring `textStyles` to be defined in order to render any text.
`textStyles` are by no means required in theme.

So, without further rationale i see this as invalid patch.

okanishchev: So it means, a fontCatalogs must be defined in theme in order to render any labels?

zbigg: After offline discussion, it looks like bigger issue.
1) Theme change by `mapView.theme = ...` doesn't update existing instance of `TextElementsRenderer` and thus new values of `textStyles` and/or `fontCatalogs` are not obeyed

2) It's impossible to know when to initialize `TextElementsRenderer` before `Tile`s are loaded.

Thus, `TextElementsRenderer` should be initialized lazily only when first `TextElements`appear in MapView (either from `Tile`s or using APIs like `addOverlayText`).

Also, a followup change to update text config with new theme is needed.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1754,10 +1757,33 @@ export class MapView extends THREE.EventDispatcher {
...
-        this.updateTextRenderer();
+        this.createTextRenderer();
+        this.m_textElementsRenderer!.placeAllTileLabels();
         this.update();
     }
 
+    /**
+     * Gradually initialize TextRenderer if it wasn't done before.
+     */
+    createTextRenderer() {
```
zbigg: `createTextRendererIfNeeded`

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1754,10 +1757,33 @@ export class MapView extends THREE.EventDispatcher {
...
             this.m_overlayTextElements = this.m_overlayTextElements.concat(textElements);
         }
-        this.updateTextRenderer();
+        this.createTextRenderer();
+        this.m_textElementsRenderer!.placeAllTileLabels();
         this.update();
     }
 
+    /**
+     * Gradually initialize TextRenderer if it wasn't done before.
```
zbigg: Ensure `TextElementsRenderer` is initialized.

@here/harp-mapview/lib/MapView.ts
zbigg: `createTextRendererIfNeeded`

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -342,7 +342,17 @@ export class TileGeometryCreator {
...
+        if (
+            decodedTile.textPathGeometries === undefined &&
+            decodedTile.textGeometries === undefined
+        ) {
+            return;
+        }
+
         const mapView = tile.mapView;
+
+        mapView.createTextRenderer();
```
zbigg: Well, i think that's small violation of hollywood principle here.

TIle shouldn't know about these details of `MapView` implementation, i would rather add check in `MapView` that if there is some text to be rendererd to create `this.createTextRendererIfNeeded` ...

Not a major issue, but usuall smell coming from making some implementation detail public.

@here/harp-mapview/lib/MapView.ts
okanishchev: done

@here/harp-mapview/lib/MapView.ts
okanishchev: done

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
okanishchev: Done, but the code looks like telescopic stick now.

zbigg: > telescopic stick

My general idea, was to collect really current status of text in renderLoop. It's not much, esp. Something like this in `MapView#render`

```

        let hasTextElements =
            this.m_overlayTextElements !== undefined && this.m_overlayTextElements.length > 0;
        renderList.forEach(({ zoomLevel, renderedTiles }) => {
                // this exists already in code, we only need to add 
                this.renderTileObjects(tile, zoomLevel);
                // simple boolean sum over all tiles like this ...
                hasTextElements = hasTextElements || !tile.textElementGroups.empty();
                ...
            });
        });
        /// ...
        if (hasTextElements && !this.m_textElementsRenderer) {
            this.createTextElementsRendererIfNeeded();
        }
```

okanishchev: That should be much more simple than my approach. Thanks.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3078,7 +3092,27 @@ export class MapView extends THREE.EventDispatcher {
...
+    /**
+     * Delay creation of `TextElementsRenderer` to out-of-frame to avoid fps lag.
+     */
+    private enqueueTextRendererCreation(): void {
+        if (
+            this.m_textElementsRenderer === undefined &&
+            this.m_textElementsRendererTimer === undefined
+        ) {
+            this.m_textElementsRendererTimer = setTimeout(() => {
+                delete this.m_textElementsRendererTimer;
```
zbigg: Uch, och, ech ...

I'm not sure if this statement:
> It is generally recommended to avoid using delete whenever possible
(https://stackoverflow.com/questions/43594092/slow-delete-of-object-properties-in-js-in-v8/44008788)

still stands, (intuition and common knowledge how optimizers work, say yes), but it's probably better (and current practice) to set unused properties to `undefined`, so we're not asking V8 to possibly re-evaluate shape of our object

Another guy saying the same: https://ljn.io/posts/node-performance-tips/

Minor, but i thought it's good to share.

okanishchev: Sounds reasonable.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -680,6 +680,7 @@ export class MapView extends THREE.EventDispatcher {
     private m_detectedFps: number = FALLBACK_FRAME_RATE;
 
     private m_textElementsRenderer?: TextElementsRenderer;
+    private m_textElementsRendererTimer?: any;
```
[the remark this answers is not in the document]
okanishchev: Good catch, thanks.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2539,9 +2544,17 @@ export class MapView extends THREE.EventDispatcher {
 
         const renderList = this.m_visibleTiles.dataSourceTileList;
 
+        const hasTextRenderer = this.m_textElementsRenderer !== undefined;
+
+        // no need to check everything if we're not going to create it.
+        let hasTextToRender = hasTextRenderer || !this.renderLabels;
```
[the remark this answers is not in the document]
okanishchev: Not really. If we set .renderLabels = false, and then set theme, text renderer would be destroyed and won't created again. So this would work.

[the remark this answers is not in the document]
zbigg: Ok, i get the shortcut but it's heavily misleading because of variable name.

O, so maybe i misunderstand because name says "has text to render", but in reality it can be `true` in some weird circumstances like:
 - no text, but textRenderer is already created 
 - no text, but renderLabels = false
When name says  its true if
 - there is text on some tile

So, i see problem with naming (inaccurate at best) and convoluted semantics (still relying on other variables). 

Do we really want to name it "hasTextToRenderOrTileCheckNotNeeded" ... or just `flag` with big comment.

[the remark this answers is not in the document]
okanishchev: Yep, it's confusing a bit. Rewrited it in more clear way.

PR #815, opened 2019-09-20
HARP-7271: Tune the styling rules for the extruded buildings.

robertoraggi:
Show extruded buildings only when the zoom level is greater than 15.


@here/harp-examples/src/rendering_post-effects_all.ts
```diff
@@ -40,7 +40,7 @@ export namespace EffectsAllExample {
         const mapControls = new MapControls(mapView);
         mapControls.maxTiltAngle = 60;
         const NY = new GeoCoordinates(40.707, -74.01);
-        mapView.lookAt(NY, 4000, 50);
+        mapView.lookAt(NY, 3500, 50);
```
ninok: Maybe we should make these 3 constants and put it somewhere else

PR #817, opened 2019-09-20
HARP-6982: Fix bug with normal (might be temporary fix, but works)

sasha240100:
Fix of the following bug:
image

Signed-off-by: Oleksandr Buzin <ext-oleksandr.buzin@here.com>

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
+++ b/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
@@ -21,10 +21,9 @@ vExtrusionAxis = vec4(normalMatrix * extrusionAxis.xyz, extrusionAxis.w);
     extrusion_normal_fragment_begin: `
 #ifdef FLAT_SHADED
     vec3 normal = vec3(0.0);
-    if (vExtrusionAxis.w > 0.999999) {
+    if (vExtrusionAxis.w > 0.999999 && length(vExtrusionAxis.xyz) > 0.000001) {
         normal = normalize(vExtrusionAxis.xyz);
-    }
-    else  {
+    } else  {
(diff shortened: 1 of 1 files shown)


@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -21,10 +21,9 @@ vExtrusionAxis = vec4(normalMatrix * extrusionAxis.xyz, extrusionAxis.w);
     extrusion_normal_fragment_begin: `
 #ifdef FLAT_SHADED
     vec3 normal = vec3(0.0);
-    if (vExtrusionAxis.w > 0.999999) {
+    if (vExtrusionAxis.w > 0.999999 && length(vExtrusionAxis.xyz) > 0.000001) {
```
ninok: Please add a comment why. Nobody will understand that condition in a year.
How does this look like on globe? The initial purpose of this code was to use smooth shading for the roof of the extruded geometry to fix the artifacts in the geojson example but now you switch to flat shading again if there is no elevation.
Would be good to have dedicated image regression test for that. As far as I remember @zbigg was adding some for extruded geometry. Maybe you can sync with him to make sure this case is covered. Sorry for asking this to you. I should have asked @amsXYZ to add the tests together with his change. Maybe we can also just revert his change and let him redo it.

ninok: I created another PR that reverts @amsXYZ change:
https://github.com/heremaps/harp.gl/pull/818

amsXYZ:
Issue addressed in #828 . I'd recommend closing this one.

PR #820, opened 2019-09-23
Merge Linux, Windows and Mac workflows

musculman:
Try to merge the workflows into a single unified file by means of filtering the OS on the matrix variable.


.github/workflows/ci.yaml
```diff
@@ -35,11 +35,18 @@ jobs:
...
+        yarn
         firefox --version
         whereis firefox
         yarn test-browser --headless-firefox
         google-chrome --version
         whereis google-chrome
         yarn test-browser --headless-chrome
-        yarn run build
-        yarn run typedoc
+        if: matrix.os != macOS-latest
```
ninok: What does this if do? Will it not execute the whole run section?

musculman: I am testing it. The github support team told me this if should filter the whole section, but it is not working as expected. I have requested more feedback. I can't filter the rules by OS for some reason.

musculman: Update: the `if` syntax works as expected and it lets you filter by os, so I will try to merge the three CI files into one so it is easier to maintain and we reuse code.

PR #826, opened 2019-09-24
MINOR: Extends the Berlin base theme for Globe.

robertoraggi:
Instead of writing a new style from scratch prefer to extends the
berlin_tilezen_base.json style. The only changes needed for globe
are the skybox and the lighting set up.


DimitryRd:
Decoders tests please

PR #827, opened 2019-09-24
HARP-7291 Block labels with boxes, lines and world lines.

nzjony:
This example doesn't actually relate directly to HARP-7291, it is used
just to show what an API could look like to pre populate the screen
before labels are rendered.

Could however be used to test specific label features etc.


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2250,6 +2250,17 @@ export class MapView extends THREE.EventDispatcher {
         this.clearTileCache();
     }
 
+    /**
+     * Allows the user to block some areas of the screen.
+     * @note To clear, simply pass in an empty array.
+     * @param screenSpaceBoxes Boxes that are in pseudo screen space, i.e. the center of the screen
```
ninok: We should not expose this pseudo screen space. I don't know why we use it internally but the API should just take normal screen coordinates.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2250,6 +2250,17 @@ export class MapView extends THREE.EventDispatcher {
...
+     * is at 0,0, whilst the left top point is -width/2, -height/2 and the bottom right is width/2
+     * and height/2.
+     */
+    setScreenSpaceBoxes(screenSpaceBoxes: Math2D.Box[]) {
```
ninok: I suggest to rename it to sth like. `setLabelAvoidanceArea` to describe what it is actually doing.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -577,6 +577,12 @@ export class TextElementsRenderer {
         return memoryUsage;
     }
 
+    prepopulateScreen(boxes: Math2D.Box[]) {
+        for (const box of boxes) {
+            this.m_screenCollisions.allocate(box);
```
ninok: I wonder how we handle priorities for the screen collisions, but I guess that's not relevant for your ticket.

@here/harp-mapview/lib/MapView.ts
nzjony: Agreed, I will change. I thought to be performant, it would be good, but ease of use trumps in this regard.

@here/harp-mapview/lib/MapView.ts
nzjony: Done, I have a similar method for the lines, so I changed it a bit.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
nzjony: I will add a comment.

@here/harp-mapview/lib/ScreenCollisions.ts
```diff
@@ -15,6 +18,22 @@ const RBush = require("rbush");
 
 const logger = LoggerManager.instance.create("ScreenCollissions");
 
+export interface IBox {
+    minX: number;
+    minY: number;
+    maxX: number;
+    maxY: number;
+    type: string;
```
ninok: remove type from IBox and in addition have:
```
interface CollisionBox extends IBox {
    type: "box";
}
```
and also replace the type in `LineWithBounds` with `type: "line"`

This gives you some additional type safety.

@here/harp-mapview/lib/ScreenCollisions.ts
```diff
@@ -104,6 +232,35 @@ export class ScreenCollisions {
     isVisible(bounds: Math2D.Box): boolean {
         return this.screenBounds.intersects(bounds);
     }
+
+    /**
+     * Transfer from screen space to screen bounds space
```
ninok: What is bounds space? Please add some description.
Is there any reason why we need this "bounds" space? Can we not do everything in screen space?

@here/harp-mapview/lib/ScreenCollisions.ts
nzjony: Thanks, thats what I had in mind originally, just didn't come out quite right.

@here/harp-mapview/lib/ScreenCollisions.ts
nzjony: I think it is more optimal to do it in bounds space, it is just the modelViewProjection applied to some point (to get NDC) and then multiplied by the width / height, it avoids the extra addition and flip you need to do with the y. to get into screen coordinates.

I didn't change it because it would mean changing the entire text engine to move from this bounds space to screen space, which is outside of this ticket.

nzjony: I have added a comment to make it clearer.

@here/harp-mapview/lib/ScreenCollisions.ts
```diff
@@ -15,6 +17,26 @@ const RBush = require("rbush");
...
+    maxY: number;
+    type: string;
+}
+
+export interface CollisionBox extends IBox {
+    type: "box";
+}
+
+export interface LineWithBound extends IBox {
+    minX: number;
```
ninok: minX, maxX, ... are already present in IBox

@here/harp-mapview/lib/ScreenCollisions.ts
```diff
@@ -67,33 +89,142 @@ export class ScreenCollisions {
...
             minY: bounds.y,
             maxX: bounds.x + bounds.w,
-            maxY: bounds.y + bounds.h
+            maxY: bounds.y + bounds.h,
+            type: "box"
         };
 
-        // Re-use array to reduce allocations.
-        return this.rtree.collides(bbox);
+        const results = this.rtree.search(bbox);
```
ninok: Please measure perf impact of this. We do much more intersection checks now. Before we stopped after the first intersection and now we find all of them. Also a lot of temporary objects are created with this change.

@here/harp-mapview/lib/ScreenCollisions.ts
```diff
@@ -67,33 +89,142 @@ export class ScreenCollisions {
...
+        for (const result of results) {
+            switch (result.type) {
+                case "box":
+                    return true;
+                case "line": {
+                    const boundedLine = result as LineWithBound;
```
ninok: Move this code to a dedicated method

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -577,6 +577,68 @@ export class TextElementsRenderer {
...
+            const startLinePointProj = new THREE.Vector3();
+            const endLinePointProj = new THREE.Vector3();
+            for (const tile of renderListEntry.visibleTiles) {
+                for (const pathBlockingElement of tile.blockingElements) {
+                    if (pathBlockingElement.points.length < 2) {
+                        continue;
+                    }
+                    // Project to screen and store it in the ScreenCollisions.
+                    let startLinePoint = pathBlockingElement.points[0];
+                    for (let i = 1; i < pathBlockingElement.points.length; i++) {
```
ninok: Can we move the allocation of all this objects to the Tile creation and don't do it every frame?
You would only update the screen positions every frame.

@here/harp-mapview/lib/ScreenCollisions.ts
nzjony: Will do.

What do you mean that a lot of temporary objects are created with this change?

@here/harp-mapview/lib/text/TextElementsRenderer.ts
nzjony: I would do this in a follow up commit if performance is an issue. I don't want to have this stuck in github too long.

nzjony: If you are ok with that.

@here/harp-mapview/lib/ScreenCollisions.ts
nzjony: I compared collides with search over berlin, collide seems to be quite a bit slower, chrome shows 0.4% of the total time, but with the search it shows 0.0% of the total time.

@here/harp-mapview/lib/ScreenCollisions.ts
```diff
@@ -104,6 +168,107 @@ export class ScreenCollisions {
...
+    private intersectsLine(bbox: CollisionBox, boundedLine: LineWithBound): boolean {
+        const line = boundedLine.line;
+        // Transform line from screen space to screenBounds space.
+        const lineStartXTransformed = line.start.x + this.screenBounds.x;
+        const lineStartYTransformed = this.screenBounds.h / 2 - line.start.y;
+        const lineEndXTransformed = line.end.x + this.screenBounds.x;
+        const lineEndYTransformed = this.screenBounds.h / 2 - line.end.y;
+
+        // Test left side
+        let intersectionResult = checkIntersection(
```
ninok: Intersecting every box edge with the line segment would not catch the case when the box is completely containing the line segment.
I would go for another approach, i.e. check all the box corners with the infinite line equation(not the line segment) and if not all corners are on one side you have an intersection (this only works b/c we check if the bbox of the line interscets with the box before)

nzjony: This is a great idea, I will give it a try.

PR #835, opened 2019-09-25
Fix theme handling in TileDataSource.

zbigg:
TileDataSource shouldn't update styleset in `setTheme` if theme doesn't
define styleset with `styleSetName` of datasource.

This fixes backwards compatibility bug introduced in #796.

Among others fixes `datasource_features_polygons` demo.


okanishchev:
If you want to set the theme where there is no styles for current datasource and it should not render, it would still behave as no setTheme was called.

zbigg:
> If you want to set the theme where there is no styles for current datasource and it should not render, it would still behave as no setTheme was called.

But this particular patch reverts to how it worked since always.

PR #839, opened 2019-09-26
MINOR: Improve performance of Berlin theme

ninok:
* Show extruded buildings starting from zoomlevel 16
* show minor roads starting from zoomlevel 14
* use road_labels layer
* enable missing road labels or primary, secondary and tertiary roads

### DIFF
--- a/@here/harp-map-theme/resources/berlin_tilezen_base.json
+++ b/@here/harp-map-theme/resources/berlin_tilezen_base.json
@@ -9,6 +9,7 @@
             "type": "selector",
             "value": [
                 "all",
+                [">=", ["get", "$level"], 15],
                 ["==", ["get", "$layer"], "buildings"],
                 ["==", ["get", "$geometryType"], "polygon"]
             ]
@@ -328,7 +329,7 @@
                 "description": "primary-road-shield-2, level > 10",
                 "when": [
                     "all",
-                    ["==", ["get", "$layer"], "roads"],
+                    ["==", ["get", "$layer"], "road_labels"],
                     ["==", ["get", "kind"], "major_road"],
                     ["==", ["get", "kind_detail"], "primary"],
                     ["has", "ref"],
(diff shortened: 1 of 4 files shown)


@here/harp-map-theme/resources/berlin_tilezen_base.json
```diff
@@ -1932,6 +1999,51 @@
                 },
                 "renderOrder": 100
             }
+        ],
```
zbigg: Is this intended change ? (doesn't look like)

@here/harp-map-theme/resources/berlin_tilezen_base.json
```diff
@@ -464,7 +465,7 @@
                 "description": "primary-road-shield-5+, level > 10",
                 "when": [
                     "all",
-                    ["==", ["get", "$layer"], "roads"],
+                    ["==", ["get", "$layer"], "road_labels"],
```
musculman: Is there anyway to have a fallback in case road_labels does not exist? The layer `road_labels` is an optional one that could be removed in the future, so, to be safe it would be better to have a logic such as: use road_labels if available and if not, use roads.
Is that even possible?

ninok: As far as I know this is not possible b/c the when condition is only in scope of the current feature that is processed.
The only way I can imagine this to work is, if map compilation would add an additional attribute, e.g. `useForLabeling', to the roads and road_labels layer. That attribute we could check in the condition.

@here/harp-map-theme/resources/berlin_tilezen_base.json
ninok: True, I was actually stumbling upon all the slowness when enabling terrain (it is already slow w/o terrain, but w/ terrain it is even slower).
Not sure if you or @robertoraggi are looking into the styling for multiple data-sources right now, but it would be nice if we would not have to add the terrain block to every theme. What I could do for now, is to check if the theme contains a terrain block and if not, I add it on the fly (in MapViewDemo). What do you think?

ninok:

PR #840, opened 2019-09-26
HARP-7097: Drop labels with too curved paths

okanishchev:
In case of significant bends on label path, simply drop the label. 

Angle check is done by comparing distance between unit vectors rather than comparing angles itself. This would give an advantage of not calling costly Math.atan2. The reference angle is set via constant, but it can simply be given as an argument in case of need. Also the 20 degrees is taken empirically.

Also has more precise label fitting pre-computation using path length instead of bounding box.

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { LineMarkerTechnique, TextStyleDefinition, Theme } from "@here/harp-datasource-protocol";
-import { ProjectionType } from "@here/harp-geoutils";
+import { MathUtils as MathGeoUtils, ProjectionType } from "@here/harp-geoutils";
 import {
     AdditionParameters,
     DEFAULT_TEXT_CANVAS_LAYER,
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -2221,27 +2233,50 @@ export class TextElementsRenderer {
...
+
+            const pointA = screenPoint;
+            const pointB = screenPoints[i - 1];
+
+            const nextDeltaX = pointA.x - pointB.x;
+            const nextDeltaY = pointA.y - pointB.y;
+            const nextLength = Math.sqrt(nextDeltaX * nextDeltaX + nextDeltaY * nextDeltaY);
+            const nextLenInverse = 1 / nextLength;
+            const nextNormalX = nextDeltaX * nextLenInverse;
+            const nextNormalY = nextDeltaY * nextLenInverse;
```
StefanDachwitz: This could be nicely put into the utilities as a function "calculateLengthAndNormalFromPoints()" or something like that.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -2221,27 +2233,50 @@ export class TextElementsRenderer {
...
 
-                if (firstIndex >= 0) {
-                    indexOfFirstVisibleScreenPoint = firstIndex;
-                }
+            const prevDeltaX = pointB.x - pointC.x;
+            const prevDeltaY = pointB.y - pointC.y;
+            const prevLength = Math.sqrt(prevDeltaX * prevDeltaX + prevDeltaY * prevDeltaY);
+            const prevLenInverse = 1 / prevLength;
+            const prevNormalX = prevDeltaX * prevLenInverse;
+            const prevNormalY = prevDeltaY * prevLenInverse;
```
StefanDachwitz: ... and here "calculateLengthAndNormalFromPoints()" could be called again

PR #854, opened 2019-10-01
Fixing obviously wrong code

zbigg:
At first this innocent line causes, that we fill `imageTextures`
for _every_ point feature, even if technique is not poi.

This is bug for sure, to be checked what are consequences.


@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -283,7 +283,7 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
                     featureIds.push(featureId);
                     objInfos.push(env.entries);
                 }
-                if (isPoiTechnique) {
```
zbigg: Hint ...`isPoiTechnique` is function, so it evaluates to `true`

Typescript 3.6.3 errors on this, so we should update.

PR #856, opened 2019-10-01
HARP-5347: Compass button.

Astrak:
In progress.


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1642,6 +1644,11 @@ export class MapView extends THREE.EventDispatcher {
     get zoomLevel(): number {
         return this.m_zoomLevel;
     }
+    set zoomLevel(zoomLevel: number) {
```
robertoraggi: please add some documentation to  `get zoomLevel`/`set zoomlevel`. Thanks to this PR, changing the zoom level using the setter will probably be the preferred way to modify the zoom. So it would be good to document it.

PR #859, opened 2019-10-02
Minor update styles

robertoraggi:


@here/harp-map-theme/resources/berlin_tilezen_base.json
```diff
@@ -1100,7 +1085,8 @@
             ["ref", "waterPolygons"],
             {
                 "description": "water",
-                "when": ["==", ["get", "$layer"], "water"],
+                "layer": "water",
+                "when": ["has", "name"],
```
robertoraggi: our `when` conditions are mandatory so I need to put something here. Is the above condition correct, or should I use something else?

@here/harp-map-theme/resources/berlin_derived.json
```diff
@@ -15,11 +14,37 @@
         },
         "countryBorderLineWidth": {
             "type": "number",
-            "value": {
-                "interpolation": "Linear",
-                "zoomLevels": [2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14],
-                "values": [2000, 1400, 1000, 500, 220, 90, 50, 30, 20, 15, 10, 5, 2]
-            }
+            "value": [
```
robertoraggi: I wonder if we find an alternative way to `pretty print` our JSON theme files. The current `prettier` configuration that we use produces ugly results for interpolations.

@here/harp-map-theme/resources/berlin_tilezen_base.json
zbigg: If we have `layer` then `when` should be conditional, why not making it so ?

@here/harp-map-theme/resources/berlin_derived.json
zbigg: Not sure, prettier is  opinionaned  by design and intentionally rejects options to customiza layout more than they believe it's necessary. Brief look at https://prettier.io/docs/en/configuration.html didn't result in anything that may look like "that" option.

robertoraggi:
> As discussed on chat, i don't think we're ready for `[interpolator]` everywhere.
> 
> `JSON.stringify` in SSE issue have to be solved first. This should be easy for static interpolators, not easy though for those which have feature-dependent output values.

this should fixed by https://github.com/heremaps/harp.gl/pull/933

PR #864, opened 2019-10-04
HARP-7143: Label outlines are visible even if the outline size is set to 0

SergiiKulakovskyi:
Hide label outline if outline size is set to 0

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -1959,14 +1959,19 @@ export class TextElementsRenderer {
                         stats.fadeAnimationRunning || textIsFadingOut || textFading;
 
                     const opacity = textRenderState.opacity;
+                    const backgroundIsVisible =
+                        pointLabel.renderStyle!.backgroundOpacity > 0 &&
+                        textCanvas.textRenderStyle.fontSize.backgroundSize > 0;
 
                     temp.bufferAdditionParams.layer = pointLabel.renderOrder;
                     temp.bufferAdditionParams.position = tempPosition;
                     temp.bufferAdditionParams.scale = textScale;
                     temp.bufferAdditionParams.opacity =
                         opacity * distanceFadeFactor * pointLabel.renderStyle!.opacity;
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1272,7 +1272,10 @@ export class TileGeometryCreator {
                         ? getPropertyValue(technique.opacity, Math.floor(zoomLevel))
                         : defaultRenderParams.opacity,
                 backgroundOpacity:
-                    technique.backgroundOpacity !== undefined
+                    technique.backgroundSize !== undefined &&
```
ninok: I think this is the wrong place b/c `getRenderStyle` is called at decoding time but especially for interpolated properties the size and opacity could get zero at any time.

ninok: I also don't think we should modify the `backgroundOpacity`. Instead there should be a check when the label is about to be rendered (e.g. in `TextElementsRenderer`) whether `backgroundSize > 0  && backgroundOpacity > 0`

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1959,14 +1959,19 @@ export class TextElementsRenderer {
...
+                        textCanvas.textRenderStyle.fontSize.backgroundSize > 0;
 
                     temp.bufferAdditionParams.layer = pointLabel.renderOrder;
                     temp.bufferAdditionParams.position = tempPosition;
                     temp.bufferAdditionParams.scale = textScale;
                     temp.bufferAdditionParams.opacity =
                         opacity * distanceFadeFactor * pointLabel.renderStyle!.opacity;
-                    temp.bufferAdditionParams.backgroundOpacity =
-                        opacity * distanceFadeFactor * pointLabel.renderStyle!.backgroundOpacity;
+                    temp.bufferAdditionParams.backgroundOpacity = backgroundIsVisible
```
ninok: Will labels with opacity 0 or backgroundOpacity still be rendered?

PR #865, opened 2019-10-04
HARP-7419 Adding technique for world space line to block screen space labels.

nzjony:
Change-Id: Id2fb88f842fa6d6b4d8da49860760074e1043936 Signed-off-by:
   Jonathan Stichbury <2533428+nzjony@users.noreply.github.com>

### DIFF
--- a/@here/harp-datasource-protocol/lib/DecodedTile.ts
+++ b/@here/harp-datasource-protocol/lib/DecodedTile.ts
@@ -24,6 +24,7 @@ import { TileInfo } from "./TileInfo";
 export interface DecodedTile {
     techniques: Technique[];
     geometries: Geometry[];
+    pathGeometries?: PathGeometry[];
     textPathGeometries?: TextPathGeometry[];
     textGeometries?: TextGeometry[]; // ### deprecate
     poiGeometries?: PoiGeometry[];
@@ -49,10 +50,18 @@ export interface DecodedTile {
 }
 
 /**
+ * This object keeps the path of the geometry. The path is in local tile space.
+ */
+export interface PathGeometry {
+    path: number[];
+}
+
+/**
  * This object keeps textual data together with metadata to place it on the map.
  */
-export interface TextPathGeometry {
-    path: number[];
+export interface TextPathGeometry extends PathGeometry {
+    // The path length is specific to text path, as a means of prioritizing longer lines over
+    // shorter lines for text placement.
     pathLengthSqr: number;
     text: string;
     technique: number;
--- a/@here/harp-datasource-protocol/lib/DecodedTile.ts
+++ b/@here/harp-datasource-protocol/lib/DecodedTile.ts
@@ -11,6 +11,7 @@ import {
     OrientedBox3,
     Projection,
     sphereProjection,
+    Vector3Like,
     webMercatorProjection
 } from "@here/harp-geoutils";
 import { Technique } from "./Techniques";
@@ -54,7 +55,7 @@ export interface DecodedTile {
  * use case, so could be either world or local tile space.
  */
 export interface PathGeometry {
-    path: THREE.Vector3[];
+    path: Vector3Like[];
 }
 
 /**
--- a/@here/harp-datasource-protocol/lib/DecodedTile.ts
+++ b/@here/harp-datasource-protocol/lib/DecodedTile.ts
@@ -54,15 +54,14 @@ export interface DecodedTile {
  * use case, so could be either world or local tile space.
  */
 export interface PathGeometry {
-    path: number[];
+    path: THREE.Vector3[];
 }
 
 /**
  * This object keeps textual data together with metadata to place it on the map.
  */
-export interface TextPathGeometry extends PathGeometry {
-    // The path length is specific to text path, as a means of prioritizing longer lines over
-    // shorter lines for text placement.
+export interface TextPathGeometry {
+    path: number[];
     pathLengthSqr: number;
     text: string;
     technique: number;
--- a/@here/harp-datasource-protocol/lib/DecodedTile.ts
+++ b/@here/harp-datasource-protocol/lib/DecodedTile.ts
@@ -50,7 +50,8 @@ export interface DecodedTile {
 }
 
 /**
- * This object keeps the path of the geometry. The path is in local tile space.
+ * This object keeps the path of the geometry. Space of the path depends on the
+ * use case, so could be either world or local tile space.
  */
 export interface PathGeometry {
     path: number[];
--- a/@here/harp-datasource-protocol/lib/Techniques.ts
+++ b/@here/harp-datasource-protocol/lib/Techniques.ts
@@ -481,6 +481,10 @@ export interface TerrainTechnique extends MakeTechniqueAttrs<TerrainTechniquePar
     name: "terrain";
 }
 
+export interface LabelRejectionLineTechnique extends MakeTechniqueAttrs<BaseTechniqueParams> {
+    name: "label-rejection-line";
+}
+
 /**
  * Possible techniques that can be used to draw a geometry on the map.
  */
@@ -500,7 +504,8 @@ export type Technique =
     | StandardExtrudedLineTechnique
     | ExtrudedPolygonTechnique
     | ShaderTechnique
-    | TextTechnique;
+    | TextTechnique
+    | LabelRejectionLineTechnique;
 
 /**
  * Additional params used for optimized usage of `Techniques`.
@@ -658,6 +663,12 @@ export function isTextTechnique(technique: Technique): technique is TextTechniqu
  */
 export function isShaderTechnique(technique: Technique): technique is ShaderTechnique {
     return technique.name === "shader";
+}
+
+export function isLabelRejectionLineTechnique(
+    technique: Technique
+): technique is LabelRejectionLineTechnique {
+    return technique.name === "label-rejection-line";
 }
 
 /**
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -19,7 +19,7 @@ import {
     ProjectionType,
     TilingScheme
 } from "@here/harp-geoutils";
-import { assert, getOptionValue, LoggerManager, Math2D, PerformanceTimer } from "@here/harp-utils";
+import { assert, getOptionValue, LoggerManager, PerformanceTimer } from "@here/harp-utils";
 import * as THREE from "three";
 
 import { ViewRanges } from "@here/harp-datasource-protocol/lib/ViewRanges";
@@ -43,7 +43,7 @@ import { PickHandler, PickResult } from "./PickHandler";
 import { PoiManager } from "./poi/PoiManager";
 import { PoiTableManager } from "./poi/PoiTableManager";
 import { PolarTileDataSource } from "./PolarTileDataSource";
-import { LineWithBound, ScreenCollisions, ScreenCollisionsDebug } from "./ScreenCollisions";
+import { ScreenCollisions, ScreenCollisionsDebug } from "./ScreenCollisions";
 import { ScreenProjector } from "./ScreenProjector";
 import { SkyBackground } from "./SkyBackground";
 import { FrameStats, PerformanceStatistics } from "./Statistics";
@@ -770,8 +770,6 @@ export class MapView extends THREE.EventDispatcher {
     private m_lastTileIds: string = "";
     private m_languages: string[] | undefined;
     private m_copyrightInfo: CopyrightInfo[] = [];
-    private m_screenSpaceBoxes: Math2D.Box[] = [];
-    private m_screenSpaceLines: LineWithBound[] = [];
     private m_animatedExtrusionHandler: AnimatedExtrusionHandler;
 
     /**
@@ -2356,25 +2354,6 @@ export class MapView extends THREE.EventDispatcher {
     }
 
     /**
-     * Allows the user to block some areas of the screen.
-     * @note To clear, simply pass in an empty array.
-     * @param screenSpaceBoxes Boxes that are in screen space.
-     * @deprecated
-     */
-    setLabelAvoidanceWithBoxes(screenSpaceBoxes: Math2D.Box[]) {
-        this.m_screenSpaceBoxes = screenSpaceBoxes;
-    }
-
-    /**
-     * Allows the user to block some area of the screen.
-     * @note To clear, simply pass in an empty array.
-     * @deprecated
-     */
-    setLabelAvoidanceWithLines(screenSpaceLines: LineWithBound[]) {
-        this.m_screenSpaceLines = screenSpaceLines;
-    }
-
-    /**
      * Public access to [[MapViewFog]] allowing to toggle it by setting its `enabled` property.
      */
     get fog(): MapViewFog {
@@ -2903,8 +2882,6 @@ export class MapView extends THREE.EventDispatcher {
         // they are handled first. They will be rendered after the normal map objects and
         // TextElements
         this.m_textElementsRenderer.reset();
-        this.m_textElementsRenderer.prepopulateScreen(this.m_screenSpaceBoxes);
-        this.m_textElementsRenderer.prepopulateScreenWithLines(this.m_screenSpaceLines);
         this.m_textElementsRenderer.prepopulateScreenWithBlockingElements();
         this.m_textElementsRenderer.renderUserTextElements(time, this.m_frameNumber);
         this.m_textElementsRenderer.renderAllTileText(time, this.m_frameNumber);
--- a/@here/harp-mapview/lib/PathBlockingElement.ts
+++ b/@here/harp-mapview/lib/PathBlockingElement.ts
@@ -17,8 +17,15 @@ import * as THREE from "three";
  */
 
 export class PathBlockingElement {
-    // We allocate the memory here to avoid allocating each frame
+    /**
+     * Note, lines is only used as a performance improvement and contains no useful information.
+     */
     readonly lines: THREE.Line3[];
+
+    /**
+     * Constructs a path from a list of points.
+     * @param points Points in world coordinates.
+     */
     constructor(readonly points: THREE.Vector3[]) {
         this.lines = new Array<THREE.Line3>(points.length >= 2 ? points.length - 1 : 0);
         for (let i = 0; i < this.lines.length; i++) {
--- a/@here/harp-mapview/lib/PathBlockingElement.ts
+++ b/@here/harp-mapview/lib/PathBlockingElement.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { Vector3Like } from "@here/harp-geoutils";
 import * as THREE from "three";
 
 /**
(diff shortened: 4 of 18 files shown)


@here/harp-mapview/lib/ThemeLoader.ts
```diff
@@ -102,6 +102,8 @@ export class ThemeLoader {
             }
             theme = (await response.json()) as Theme;
             theme.url = themeUrl;
+            // Temporary fix until HARP-7424 is fixed.
+            theme = this.resolveUrls(theme);
```
nzjony: @zbigg , is this an acceptable short term solution?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -222,12 +226,30 @@ export class TileGeometryCreator {
...
+        const path: THREE.Vector3[] = [];
+        const point = new THREE.Vector3();
+        for (const flatpath of decodedTile.pathGeometries) {
+            for (let i = 0; i < flatpath.path.length; i += 3) {
+                point.set(flatpath.path[i], flatpath.path[i + 1], flatpath.path[i + 2]);
+                // Move to world coordinates, the accuracy isn't important, because we don't need to
+                // worry about z flickering (the path is not rendered). This also makes the code
+                // projecting to screen space in
+                // [[TextElementsRenderer.prepopulateScreenWithBlockingElements]] simpler because
+                // the shift before projection to NDC isn't required.
```
nzjony: @ninok , is this comment correct?

@here/harp-mapview/lib/ThemeLoader.ts
zbigg: At first look should be ok. But who knows all the consequences.

zbigg: @nzjony Fixed (properly) i belive in #866

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -481,6 +481,10 @@ export interface TerrainTechnique extends MakeTechniqueAttrs<TerrainTechniquePar
     name: "terrain";
 }
 
+export interface LabelRejectionLineTechnique extends MakeTechniqueAttrs<BaseTechniqueParams> {
+    name: "label-rejection-line";
```
ninok: Sorry for bringing this up so late (and not in the solutioning) but why do we need a separate technique?
This could also be just a property in the `SolidLineTechnique`. We could even reuse the buffer that is created by the `SolidLineTechnique`

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -222,12 +226,30 @@ export class TileGeometryCreator {
             }
             return filter(technique);
         };
+        this.createTextElements(tile, decodedTile, textFilter);
 
-        this.preparePois(tile, decodedTile);
+        this.createLabelRejectionElements(tile, decodedTile);
+    }
```
ninok: what happened to preparePois?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
ninok: The actual reason why we can use world coordinates for that is that we use double precision (b/c all the label stuff is done on the CPU).

@here/harp-datasource-protocol/lib/Techniques.ts
nzjony: Good point, never occurred to me, but makes complete sense, will look into it.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: line 215, this was also a refactoring, I moved textFilter to just above where it is needed.

I try to avoid doing this, but in this case, it made sense.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: but what does double precision give us? More accuracy to avoid z fighting, or something else?

@here/harp-datasource-protocol/lib/Techniques.ts
nzjony: I brought back the technique because it is needed to have a separate datasource with lower LOD.

nzjony:
The technique used on a separate data source with a -2 storage level offset provides very good performance, hence bringing back the technique.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -220,12 +224,30 @@ export class TileGeometryCreator {
...
+        for (const flatpath of decodedTile.pathGeometries) {
+            for (let i = 0; i < flatpath.path.length; i += 3) {
+                point.set(flatpath.path[i], flatpath.path[i + 1], flatpath.path[i + 2]);
+                // Move to world coordinates, the accuracy isn't important, because we don't need to
+                // worry about z flickering (the path is not rendered). This also makes the code
+                // projecting to screen space in
+                // [[TextElementsRenderer.prepopulateScreenWithBlockingElements]] simpler because
+                // the shift before projection to NDC isn't required.
+                point.add(tile.center);
+                path.push(point);
```
ninok: Is this working? I think you are missing a `point.clone()`.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -42,7 +42,8 @@ function decodedTileHasGeometry(decodedTile: DecodedTile) {
         decodedTile.geometries.length ||
         (decodedTile.poiGeometries !== undefined && decodedTile.poiGeometries.length) ||
         (decodedTile.textGeometries !== undefined && decodedTile.textGeometries.length) ||
-        (decodedTile.textPathGeometries !== undefined && decodedTile.textPathGeometries.length)
+        (decodedTile.textPathGeometries !== undefined && decodedTile.textPathGeometries.length) ||
```
ninok: Did you test this with a `Theme` that only has your new technique? Most likely it will not work b/c there is some weird handling of tiles that don't have geometries. I fixed that with:
https://github.com/heremaps/harp.gl/pull/902

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -220,12 +224,30 @@ export class TileGeometryCreator {
...
-        // TextElements do not get their geometry created by Tile, but are managed on a
-        // higher level.
-        this.createTextElements(tile, decodedTile, textFilter);
+    createLabelRejectionElements(tile: Tile, decodedTile: DecodedTile) {
+        if (decodedTile.pathGeometries === undefined) {
+            return;
+        }
+        const path: THREE.Vector3[] = [];
+        const point = new THREE.Vector3();
+        for (const flatpath of decodedTile.pathGeometries) {
```
ninok: `flatpath` is a bit weird name b/c `path` is also flat. Couldn't `decodedTile.pathGeometries` be in world space already so we would save this conversion in the main thread?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: I think it is fixed in a later commit, it is at least there when I check "Files Changed".

@here/harp-mapview/lib/Tile.ts
nzjony: Yes, see: https://gerrit.it.here.com/#/c/1232943/

I saw your fix, but didn't understand what the problem was unfortunately.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: Agreed, I will rename.

Agree it makes sense to do this in the decoder. I will fix.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -178,6 +180,7 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
     private readonly m_geometries: Geometry[] = [];
     private readonly m_textGeometries: TextGeometry[] = [];
     private readonly m_textPathGeometries: TextPathGeometry[] = [];
+    private readonly m_linePathGeometries: PathGeometry[] = [];
```
StefanDachwitz: m_linePathGeometries is called "pathGeometries", so I think it should be either "m_linePathGeometries" and "linePathGeometries", or "m_pathGeometries" and "pathGeometries" to be consistent.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -640,30 +619,29 @@ export class TextElementsRenderer {
...
                         continue;
                     }
-                    // Project to screen and store it in the ScreenCollisions.
                     let startLinePoint = pathBlockingElement.points[0];
                     for (let i = 1; i < pathBlockingElement.points.length; i++) {
                         const endLinePoint = pathBlockingElement.points[i];
-                        this.m_screenProjector.projectAndShift(startLinePoint, startLinePointProj);
-                        this.m_screenProjector.projectAndShift(endLinePoint, endLinePointProj);
+                        this.m_screenProjector.project3(startLinePoint, startLinePointProj);
+                        this.m_screenProjector.project3(endLinePoint, endLinePointProj);
```
StefanDachwitz: Can we not save the projected start point together with the unprojected "startLinePoint" to save one projection?

nzjony: yes, good point.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
nzjony: Done.

@here/harp-datasource-protocol/lib/Techniques.ts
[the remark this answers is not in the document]
nzjony: @ninok , do you have any preference?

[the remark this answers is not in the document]
nzjony: "label-separation-line" sounds a bit strange in my opinion, as i would question what it is separating. "Label-Reservation" together sounds a bit weird, it is a line that is reserving for labels, does that mean it reserves some space to make Labels visible?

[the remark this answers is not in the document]
StefanDachwitz: I was just a bit worried that the "rejection" could be understood as a line that keeps label from appearing, but I guess in the final implementation the labels will be moved to a different location, no? Alternatives (dict.cc is my friend...):

label-displacement-line
label-lock-out-line
label-relocation-line
label-push-back-line

... let the native speaker(s) decide!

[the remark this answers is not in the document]
nzjony: > I was just a bit worried that the "rejection" could be understood as a line that keeps label from appearing,

Good, because that is exactly what it does, it prevents labels from appearing.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -494,9 +494,16 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
                 for (const path of lines) {
+                    const worldPath: number[] = [];
+                    for (let i = 0; i < path.length; i += 3) {
+                        point.set(path[i], path[i + 1], path[i + 2]);
+                        point.add(this.m_decodeInfo.center);
+                        worldPath.push(point.x, point.y, point.z);
+                    }
                     this.m_linePathGeometries.push({
-                        path
+                        path: worldPath
```
ninok: Why do we flatten this array here and expand it again in the TileGeometryCreator? When the data is received the array will not contain proper THREE.Vector3 objects but Vector3Like but I guess that's enough?!

nzjony: This is in the decoder, so we need to flatten it to pass it back to the main thread? Or do I miss something.

ninok: There is no need to flatten it. If you pass an `THREE.Vector[]` from the worker to the main thread it would arrive in the main thread as `Vector3Like[]`(i.e. the methods are lost). Not sure if your code is relying on having full `THREE.Vector3` instances though.

@here/harp-datasource-protocol/lib/DecodedTile.ts
```diff
@@ -49,10 +50,18 @@ export interface DecodedTile {
 }
 
 /**
- * This object keeps textual data together with metadata to place it on the map.
+ * This object keeps the path of the geometry. The path is in local tile space.
```
zbigg: > The path is in local tile space

Yes, for normal text, but not for paths added in this patch - they are in world space.

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -481,6 +481,10 @@ export interface TerrainTechnique extends MakeTechniqueAttrs<TerrainTechniquePar
     name: "terrain";
 }
 
+export interface LabelRejectionLineTechnique extends MakeTechniqueAttrs<BaseTechniqueParams> {
```
zbigg: What is `width` of this line ?
As i understand, now we check collisions only against abstract infinitely-thin lines.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2355,25 +2353,6 @@ export class MapView extends THREE.EventDispatcher {
         this.clearTileCache();
     }
 
-    /**
-     * Allows the user to block some areas of the screen.
-     * @note To clear, simply pass in an empty array.
-     * @param screenSpaceBoxes Boxes that are in screen space.
-     * @deprecated
-     */
-    setLabelAvoidanceWithBoxes(screenSpaceBoxes: Math2D.Box[]) {
```
zbigg: What is new API for this functionality ?
For lines, we can create GeoJSON with new technique, what about boxes.

I see this as very useful API, not sure why removing it.

@here/harp-mapview/test/ScreenProjectorTest.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ScreenProjector } from "../lib/ScreenProjector";
+
+import { assert } from "chai";
+import * as THREE from "three";
+
+describe("screen projector test", () => {
```
zbigg: screen projector test -> ScreenProjector

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
nzjony: cool, I did not know, I will give it a try.

@here/harp-datasource-protocol/lib/DecodedTile.ts
nzjony: Good spotting!

@here/harp-datasource-protocol/lib/Techniques.ts
nzjony: Width is 0. I.e. we just check if the line intersects the box, without any consideration of its width.

@here/harp-mapview/lib/MapView.ts
nzjony: This was an API I added previously, unfortunately it had a bid snow ball affect on the inner code. It meant that, because the current ScreenCollisions class uses an rtree which is basically the NDC value scaled by height / width, that I had to convert from screen coordinates to this scaled NDC space (this is used, because it is the space generated by the project function in the ScreenProjector class).

By removing this API (which BTW was added as @deprecated), I can avoid this internal shift and was able to see an improvement in performance also.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
nzjony: it works well, thanks for the hint.

zbigg: > it works well, thanks for the hint.

Well, please consider that here:
```
export interface PathGeometry {
    path: THREE.Vector3[];
}
```
you're lying to Typescript compiler. What you get in main thread is only `Vector3Like` as mentioned by Nino.
To be on safe side, i would change typings in all `PathBlockingElement` related classed and interfaces to `Vector3Like` (it should work) and we're on safe side. 

>  Not sure if your code is relying on having full THREE.Vector3 instances though.

Not yet, but who knows how it will evolve. Unsafe silent conversion like we have in patch now will result in obscure runtime error, when it should be found by compiler.

nzjony: Your right, I will fix.

nzjony: @zbigg, done, please take a look at the latest commit.

@here/harp-mapview/lib/ScreenProjector.ts
```diff
@@ -69,7 +70,7 @@ export class ScreenProjector {
      * outside the near / far plane.
      */
     project3(
-        source: THREE.Vector3,
+        source: THREE.Vector3 | Vector3Like,
```
ninok: No need for THREE.Vector3 b/c it is a Vector3Like

@here/harp-mapview/lib/ScreenProjector.ts
```diff
@@ -88,7 +89,10 @@ export class ScreenProjector {
      * @param {THREE.Vector3} target The target vector.
      * @returns {THREE.Vector3} The projected vector (the parameter 'target').
      */
-    projectVector(source: THREE.Vector3 | THREE.Vector4, target: THREE.Vector3): THREE.Vector3 {
+    projectVector(
+        source: THREE.Vector3 | THREE.Vector4 | Vector3Like,
```
ninok: No need for Vecto3 and Vector4. They are both fulfilling the Vector3Like interface

@here/harp-mapview/lib/PathBlockingElement.ts
```diff
@@ -15,5 +18,19 @@
...
+    readonly lines: THREE.Line3[];
+
+    /**
+     * Constructs a path from a list of points.
+     * @param points Points in world coordinates.
+     */
+    constructor(readonly points: Vector3Like[]) {
+        this.lines = new Array<THREE.Line3>(points.length >= 2 ? points.length - 1 : 0);
+        for (let i = 0; i < this.lines.length; i++) {
+            this.lines[i] = new THREE.Line3(new THREE.Vector3(), new THREE.Vector3());
```
ninok: Would be better to initialize the Line segments with the proper points. I know the purpose here is to just allocate the objects but somebody might use that class for sth. else and it would be a surprise that all line segments are 0

nzjony: These lines are in screen space, so it can contain no useful information yet. I will add a comment to clarify this.

@here/harp-mapview/lib/ScreenProjector.ts
nzjony: See: 0716eb78a39efe80daee449146d5911b526e8dfb

@here/harp-mapview/lib/ScreenProjector.ts
nzjony: See: 0716eb78a39efe80daee449146d5911b526e8dfb

@here/harp-mapview/lib/PathBlockingElement.ts
nzjony: Check: 548a4ca2cb7575b090ecd2de5e1d0e86369ff728

@here/harp-mapview/lib/MapView.ts
zbigg: Ok, fair enough. 
But how is this PR is related to blocking lines which are only remotely related to blocking of whole areas.

PR #867, opened 2019-10-04
Test related tweaks.

zbigg:
* alias `yarn code-rules-tests` for code quality tests in `test/* folder`
* alias `pre-check` for them and `tslint` and `prettier` to align
  experience with CI `Pre check` step
* remove timeouts from code quality check, as they are sometimes very
  heavy and timeout is not a failure, run them with `--no-timeouts`
* add `--forbid-only` to `mocha` test runs to prevent accidental green,
  but actually not ran tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,33 +29,12 @@ jobs:
       with:
         version: ${{ matrix.node_version }}
     - name: Install dependencies
-      run: |
-        npm install -g chromedriver geckodriver
-      if: matrix.os != 'windows-latest'
-    - name: Install dependencies (Windows)
-      shell: bash
-      run: |
-        set -ex
-        # This is arbitrary version that happens to be installed in github CI env atm.
-        # As for now `chromedriver` supports only exact version, so this version have to be
-        # updated with Github CI windows infra.
-        #
-        # See
-        #  https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions#windows-server-2019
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -45,17 +44,13 @@ jobs:
         #  https://github.com/giggio/node-chromedriver - explanation
         #  https://sites.google.com/a/chromium.org/chromedriver/downloads - available versions
         export CHROMEDRIVER_VERSION=75.0.3770.140
+        yarn
```
musculman: This would run yarn and then export a variable and yarn again. Isn't it redundant or what did I miss?

zbigg: The idea was to have separate `yarn` steps for windows and for rest.

Previously, `yarn` was started in `pretest` step which was kind-a fishy, as it should be ran in "Install dependencies" step.

Moreover, "Install deps ..(Windows)" was de-facto no-op as onlyset env variable only for shell which immediately exits at end of step.

So, it looks like CHROMEDRIVER_VERSIOS is not needed at all, so i'll unify install dependencies ...

.github/workflows/ci.yaml
```diff
@@ -29,33 +29,12 @@ jobs:
       with:
         version: ${{ matrix.node_version }}
     - name: Install dependencies
-      run: |
```
zbigg: @musculman Looks like we don't need custom install dependencies step at all now.

musculman: Happy to remove code!

PR #869, opened 2019-10-07
HARP-7360: Implement the 'step' operator.

robertoraggi:


@here/harp-datasource-protocol/lib/operators/InterpolationOperators.ts
```diff
@@ -4,12 +4,24 @@
...
 import { createInterpolatedProperty } from "../InterpolatedProperty";
 import { InterpolatedPropertyDefinition } from "../InterpolatedPropertyDefs";
 
 const operators = {
+    zoom: {
+        call: (context: ExprEvaluatorContext, args: Expr[]): Value => {
+            if (context.scope === ExprScope.Condition) {
+                // A call to 'zoom' in a filter condition expression should
+                // return the storage level of the current tile.
+                return context.env.lookup("$level")!;
```
ninok: Why do we need to add zoom if we only support this case? It is also quite misleading b/c zoom(and level) contain the storage level offset of the data source

robertoraggi: This is the correct semantics. In 'filter-decoding time conditions' the operator 'zoom' must return the storage level. so user can write things like:
`when: ["step", ["zoom"], false, 14, true]`

@here/harp-omv-datasource/lib/OmvDecoder.ts
```diff
@@ -453,6 +460,10 @@ export class OmvDecoder implements IGeometryProcessor {
         return tileSizeOnScreen;
     }
 
+    private getZoomLevel(storageLevel: number) {
+        return Math.max(storageLevel - (this.m_storageLevelOffset || 0));
```
ninok: you are missing `, 0`

PR #872, opened 2019-10-07
HARP-7356: Added TextElementsRendererTest

ninok:


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -666,6 +666,227 @@ export class TextElementsRenderer {
         });
     }
 
+    renderTextElements(
```
ninok: This method was just moved up b/c it is now public and we have a tslint rule that public methods have to come first.

@here/harp-mapview/test/TextElementsRendererTest.ts
```diff
@@ -0,0 +1,279 @@
...
+        assert(renderedTextElements[0].textRenderState!.isFadingIn());
+
+        const lastOpacity = renderedTextElements[0].textRenderState!.opacity;
+
+        // Third frame. Label should be semi-transparent and still fading in.
+        renderedTextElements.splice(0);
+        numRenderedTextElements = textElementsRenderer.renderTextElements(
+            textElements,
+            time + 800,
+            frameNumber + 8,
```
nzjony: should this be +3?

@here/harp-mapview/test/TextElementsRendererTest.ts
```diff
@@ -0,0 +1,279 @@
...
+            secondChanceTextElements
+        );
+        expect(numRenderedTextElements).to.be.equal(1);
+        expect(renderedTextElements.length).to.be.equal(1);
+        expect(renderedTextElements[0].textRenderState!.opacity)
+            .to.be.greaterThan(lastOpacity)
+            .to.be.lessThan(1);
+        assert(renderedTextElements[0].textRenderState!.isFadingIn());
+
+        // Forth frame. Label should be visible and fading should be finished.
```
nzjony: Why does it take 4 frames to fade in? I tried to find where this is configured but couldn't, is it some default?

ninok: It is not the number of frames that matter, but the time. It is configurable but by default it is:
`export const DEFAULT_FADE_TIME = 800;` defined in TextElement.ts`
I will use this constant to make things more clear.

@here/harp-mapview/test/TextElementsRendererTest.ts
ninok: True. I actually don't know if the frameNumber is used for anything. I think the only requirement is, that it is monotonically increasing

ninok:

PR #876, opened 2019-10-08
Harp-7292: All data sources utilizes single cache for tiles.

kkostecki:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-lrucache/lib/LRUCache.ts
+++ b/@here/harp-lrucache/lib/LRUCache.ts
@@ -85,7 +85,8 @@ export class LRUCache<Key, Value> {
     /**
      * Iterates over all items from the most recently used item to the least recently used one.
      *
-     * **Note**: Results are undefined if the cache is modified during iteration.
+     * **Note**: Results are undefined if the entire cache is modified during iteration. You may
+     * although modify the current element in [[callbackfn]] function.
      *
      * @param callbackfn The callback to call for each item.
      * @param thisArg Optional this argument for the callback.
@@ -96,8 +97,9 @@ export class LRUCache<Key, Value> {
     ): void {
         let entry = this.m_newest;
         while (entry !== null) {
+            const older = entry.older;
             callbackfn.call(thisArg, entry.value, entry.key, this);
-            entry = entry.older;
+            entry = older;
         }
     }
 
@@ -166,6 +168,20 @@ export class LRUCache<Key, Value> {
     }
 
     /**
+     * Resets the cache capacity and function used to measure the element size.
+     *
+     * @param newCapacity The new capacity masured in units returned from [[sizeMeasure]] funtion.
+     * @param sizeMeasure Function that defines the size of element, if you want to measure
+     * number of elements only always return 1 from this function (default), you may also
+     * specify own function that measures entries by memory consumed, nubmer of sub-elements, etc.
+     */
+    setCapacityAndMeasure(newCapacity: number, sizeMeasure: (v: Value) => number = () => 1) {
+        this.m_capacity = newCapacity;
+        this.m_sizeFunction = sizeMeasure;
+        this.shrinkToCapacity();
+    }
+
+    /**
      * Updates the size of all elements in this cache. If their aggregated size is larger than the
      * capacity, items will be evicted until the cache shrinks to fit the capacity.
      */
@@ -204,7 +220,7 @@ export class LRUCache<Key, Value> {
             this.m_size = this.m_size - entry.size + valueSize;
             entry.value = value;
             entry.size = valueSize;
-            this.promote(entry);
+            this.promoteEntry(entry);
             this.evict();
         } else {
             if (valueSize > this.m_capacity) {
@@ -239,7 +255,7 @@ export class LRUCache<Key, Value> {
             return undefined;
         }
 
-        this.promote(entry);
+        this.promoteEntry(entry);
         return entry.value;
     }
 
@@ -279,6 +295,28 @@ export class LRUCache<Key, Value> {
     }
 
     /**
+     * Evict selected elements from the cache using [[selector]] function.
+     *
+     * @param selector The function for selecting elements for eviction.
+     * @param thisArg Optional _this_ object reference.
+     */
+    evictSelected(selector: (value: Value, key: Key) => boolean, thisArg?: any) {
+        const cb = this.evictionCallback;
+        let entry = this.m_newest;
+        while (entry !== null) {
+            const entryOlder = entry.older;
+            if (selector.call(thisArg, entry.value, entry.key)) {
+                if (cb !== undefined) {
+                    cb(entry.key, entry.value);
+                }
+                this.deleteEntry(entry);
+                this.m_map.delete(entry.key);
+            }
+            entry = entryOlder;
+        }
+    }
+
+    /**
      * Explicitly removes a key-value pair from the cache.
      *
      * **Note**: This is an explicit removal, thus, the eviction callback will not be called.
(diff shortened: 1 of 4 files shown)


@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -79,55 +79,216 @@ export interface VisibleTileSetOptions {
...
+        });
+    }
+
+    /**
+     * Get tile cached or __undefined__ if tile is not yet in cache.
+     *
+     * @param tileCode En unique tile code (its morton code).
+     * @param dataSource A [[DataSource]] the tile comes from.
+     */
+    get(tileCode: number, dataSource: DataSource): Tile | undefined {
```
ninok: I would prefer if this method would explicitely get all ingredients to compute the cache key, i.e.:
`get(mortonCode: number, offset:number, dataSource: DataSource): Tile | undefined`
Since the cache key is a string we could also remove the hack that adds the offset to the mortonCode

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -79,55 +79,216 @@ export interface VisibleTileSetOptions {
...
+ * only when cache reaches full saturation and tile is no longer visible.
+ * @note Currently cached entries (tiles) are identified by unique tile code (morton code) and
+ * data source name, thus it is required that each [[DataSource]] used should have unique
+ * name, but implementation could be improved to omit this limitation.
  */
 class DataSourceCache {
-    readonly tileCache: LRUCache<number, Tile>;
-    readonly disposedTiles: Tile[] = [];
+    static getKey(tileCode: number, dataSource: DataSource): string {
+        return `${dataSource.name}_${tileCode}`;
```
ninok: as mentioned below, I would prefer following cache key:
`${dataSource.name}_${mortonCode}_${tileOffset}`

@here/harp-lrucache/lib/LRUCache.ts
```diff
@@ -66,10 +66,15 @@ export class LRUCache<Key, Value> {
     private m_sizeFunction: (v: Value) => number;
 
     /**
-     * Creates a new instance of `LRUCache`. The optional sizeFunction can be used
-     * fine tune the size required to cache that item.
+     * Creates a new instance of `LRUCache`.
      *
-     * @param cacheCapacity The maximum number of entries to store in the cache.
+     * The optional [[sizeFunction]] can be used to fine tune the memory consumption of all cached
```
nzjony: this is part of the other pull request, so why is it here?

@here/harp-lrucache/lib/LRUCache.ts
```diff
@@ -80,7 +85,8 @@ export class LRUCache<Key, Value> {
     /**
      * Iterates over all items from the most recently used item to the least recently used one.
      *
-     * **Note**: Results are undefined if the cache is modified during iteration.
+     * **Note**: Results are undefined if the entire cache is modified during iteration. You may
+     * althought modify the current element in [[callbackfn]] function.
```
nzjony:
```suggestion
     * although modify the current element in [[callbackfn]] function.
```

@here/harp-lrucache/lib/LRUCache.ts
```diff
@@ -91,8 +97,9 @@ export class LRUCache<Key, Value> {
     ): void {
         let entry = this.m_newest;
         while (entry !== null) {
+            const older = entry.older;
```
nzjony: what does this do?

@here/harp-lrucache/lib/LRUCache.ts
nzjony: Nevermind, I see that you split the commits, ignore this.

nzjony:
Another comment, now that we have a single tile cache, is it possible that we have fighting for this cache in each frame, for example, we have two datasources, one which is the background data source i.e. very light memory usage, and another which is 3d city models, i.e. very heavy usage.

It is theoretically possible that the memory usage of the 3D city models in a single location could exceed the cache size limit, and then you would need to recreate the background tiles each frame because they would be pushed out each frame? At least with the previous approach there was no fighting between data sources for resources.

Maybe we could set an upper limit for each datasource, i.e. no datasource can have more than 80% of the total size?

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: Hmm.. sounds good!

@here/harp-lrucache/lib/LRUCache.ts
kkostecki: Fixed :), finally I have installed Code Spell Checker, so should limit the number of such "flowers"....

kkostecki:
> Another comment, now that we have a single tile cache, is it possible that we have fighting for this cache in each frame, for example, we have two datasources, one which is the background data source i.e. very light memory usage, and another which is 3d city models, i.e. very heavy usage.
> 
> It is theoretically possible that the memory usage of the 3D city models in a single location could exceed the cache size limit, and then you would need to recreate the background tiles each frame because they would be pushed out each frame? At least with the previous approach there was no fighting between data sources for resources.
> 
> Maybe we could set an upper limit for each datasource, i.e. no datasource can have more than 80% of the total size?

This is very good point, of course for now we need to live with such implementation, so cache size should be heuristically set (i.e. based on our experience with data we have), but...
I have some nice idea, that we could have adaptive LRU cache which adjusts its capacity according to biggest item stored (something similar to std::vector reallocation). Thus whenever something BIG is put into the cache the algorithm could spread capacity accordingly (let's say double the size of the biggest item in history). Of course we could then provide some specialized callback for this with "reallocation" (capacity change function) to be fully customizable.

@here/harp-lrucache/lib/LRUCache.ts
kkostecki: Key the oldest reference, before it will be changed in the callback itself... We do not have any guaranty that callback will not modify (say delete) currently processed item. Actually it was already done in VisibleTileSet.markDataSourceTilesDirty().

nzjony: Ok, sounds like a valid point. Please add a comment in the code to make it clear for others.
Also, the note above says that the results are undefined if the cache is modified during iteration, is that no longer correct? It seems that the cache can be modified and this will still work.

kkostecki:
> Please add tests

LRU cache test added, both for new and some old methods.

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: Because this may be hard change and it inevitably conflicts with renderList hold in VisibleTileSet.dataSourceTileList, I would strongly insist to split this to second PR, cause this change may cause some regression bugs, it will be easier to investigate and revert (if needed).

@here/harp-lrucache/lib/LRUCache.ts
kkostecki: Cache element may be modified, but please note that there is also Map<K, V> reference passed in, so entire cache can not be modified. Thus my update in method comments.

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: Ok done in separate patchset: 
https://github.com/heremaps/harp.gl/pull/884

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: Done in separate patchset: 
https://github.com/heremaps/harp.gl/pull/884

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -864,34 +1026,21 @@ export class VisibleTileSet {
...
-        }
-
-        return dataSourceCache;
-    }
-
     private markDataSourceTilesDirty(renderListEntry: DataSourceTileList) {
-        const dataSourceCache = this.m_dataSourceCache.get(renderListEntry.dataSource.name);
-        const retainedTiles: Set<number> = new Set();
+        const dataSourceCache = this.m_dataSourceCache;
+        const retainedTiles: Set<string> = new Set();
```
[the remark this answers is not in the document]
kkostecki: Added in follow up PR:
https://github.com/heremaps/harp.gl/pull/884

kkostecki:

PR #880, opened 2019-10-09
Add codecov to Travis

musculman:
* Add code coverage build to Travis
* Integrate with codecov.io
* Add badge on README


.travis.yml
```diff
@@ -18,14 +18,16 @@ branches:
...
 
 jobs:
   include:
     - name: "Test"
       script: |
         set -ex
         yarn pre-test
-        yarn test --forbid-only
+        yarn cov-test --forbid-only
+        yarn cov-report-html
```
zbigg: This is not needed, just wasted ... cpu cycles.

.travis.yml
```diff
@@ -64,3 +66,5 @@ jobs:
             all_branches: true
             condition: $TRAVIS_BRANCH =~ ^master|release$
 
+after_success:
+ - codecov
```
zbigg: Looks like this is not enough, in my PoC i did this:

```
nyc report --reporter=json && ls -al coverage/ && codecov -f coverage/*.json
```

So i think we must tell which files exactly should be uploaded.

.travis.yml
musculman: I took it from our github actions. Should it be removed there as well?

.travis.yml
musculman: That's interesting, but it should not be needed according to [the example](https://github.com/codecov/example-typescript) as the `nyc report` is already done with cov-report-html. But if we remove it, I'd rather not call `nyc` directly but through `yarn` to reduce maintenance overhead.
Only the `-f coverage/*.json` as it seems it is not able to find it otherwise

.travis.yml
zbigg: I just think that for this particular job, we need only report in `json` format so there is no reason to generate HTML.
OTOH, due to configuration, we already generate HTML report and `coverage/coverage-final.json` in `yarn  cov-test` step.

.travis.yml
zbigg: Agree, all the files are already generated, only `-f` is needed.

PR #884, opened 2019-10-09
Harp-7292: Single cache with morton code and offset used in tile identifier.

kkostecki:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/VisibleTileSet.ts
+++ b/@here/harp-mapview/lib/VisibleTileSet.ts
@@ -79,6 +79,8 @@ export interface VisibleTileSetOptions {
 
 const MB_FACTOR = 1.0 / (1024.0 * 1024.0);
 
+type TileCacheId = string;
+
 /**
  * Wrapper for LRU cache that encapsulates tiles caching for any [[DataSource]] used.
  *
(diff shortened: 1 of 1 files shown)


kkostecki:
This PR presents the approach suggested by @ninok , where tile identifiers contains morton code, tile offset and its data source name.
Please notice that this scheme is not perfectly compliant with other tiles lists used in VisibleTileSet, such as:
- renderedTiles,
- incompleteTiles (internal),
- checkedTiles (internal),
- visibleTileKeys (returned from getVisibleTileKeysForDataSources)

Although not so important from implementation point of view, it is worth to mention that we now follow different conventions. In order to totally get rid of TileKey concept, much bigger refactoring/changes are required, that are touching also other classes.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -91,11 +93,25 @@ const MB_FACTOR = 1.0 / (1024.0 * 1024.0);
...
+     */
+    static getKey(mortonCode: number, offset: number, dataSource: DataSource): TileCacheId {
+        return `${dataSource.name}_${mortonCode}_${offset}`;
+    }
+
+    /**
+     * Create unique tile identifier for caching based on tile object passed in.
+     * @param tile The tile for which key is generated.
+     */
+    static getTileKey(tile: Tile): TileCacheId {
```
ninok: The name is a bit missleading. Maybe `getKeyFromTile` would be more clear.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -931,16 +948,19 @@ export class VisibleTileSet {
 
                         if (!checkedTiles.has(parentCode) && !renderedTiles.get(parentCode)) {
                             checkedTiles.add(parentCode);
-                            const parentTile = tileCache.get(parentCode, dataSource);
+
+                            const {
+                                offset,
+                                mortonCode
+                            } = TileOffsetUtils.extractOffsetAndMortonKeyFromKey(parentCode);
```
ninok: So we still have this packing of offset into the morton code. Maybe we can remove this as well in a separate commit.

kkostecki: I have checked it, this would much much bigger change, cause we utilize this system almost everywhere, that is why I was not so prone to introduce this change. We may still leave the old system in cache if you wish...

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -91,11 +93,25 @@ const MB_FACTOR = 1.0 / (1024.0 * 1024.0);
...
-    static getKey(tileCode: number, dataSource: DataSource): string {
-        return `${dataSource.name}_${tileCode}`;
+    /**
+     * Creates unique tile key for caching based on morton code, tile offset and its data source.
+     * @param mortonCode The tile morton code.
+     * @param offset The tile offset.
+     * @param dataSource The [[DataSource]] from which tile was loaded.
+     */
+    static getKey(mortonCode: number, offset: number, dataSource: DataSource): TileCacheId {
+        return `${dataSource.name}_${mortonCode}_${offset}`;
```
nzjony: have you done a performance test on this? Is it quick to have strings as keys?

kkostecki: Here you have:
http://jsben.ch/ac9c9
Map with number keys is 35% faster then with strings, but do you have any idea how we could replace string while still preserving DataSource information to identify each tile within all data sources.

PR #895, opened 2019-10-11
HARP-7270: Change maxVisibleDataSourceTiles use from MapViewOptions.

kkostecki:
Since it is difficult to predict the number of tiles to be displayed cause
it depends on geo-projection used and camera pose, we decided to set this
constraint (limit of tiles) to extreme value and not relate it to cache
size.
The number of tiles should be anyway limited by proper frustum culling and
correct near/far frustum planes calculation (see ClipPlanesEvaluator).

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -609,7 +609,7 @@ export const MapViewDefaults = {
     projection: mercatorProjection,
     clipPlanesEvaluator: defaultClipPlanesEvaluator,
 
-    maxVisibleDataSourceTiles: 120,
+    maxVisibleDataSourceTiles: 500,
     extendedFrustumCulling: true,
 
     tileCacheSize: 200,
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -404,18 +404,18 @@ export class VisibleTileSet {
...
+     * Retrieves maximum number of visible tiles per data source.
      */
-    getNumberOfVisibleTiles() {
+    get maxVisibleDataSourceTiles() {
         return this.options.maxVisibleDataSourceTiles;
     }
 
     /**
-     * Sets maximum number of visible tiles.
+     * Sets maximum number of visible tiles oer data source.
```
ninok: typo `per data source`

kkostecki:
> I'm a bit afraid of what will happen if we commit this change standalone. Especially in the globe examples the performance will drop significantly b/c we render a shitload of tiles. So can you please check and maybe do the change for the globe examples as well (e.g. make storageLevelOffset depending on the latitude if `projection === sphereProjection`

@ninok I think it is not needed, or maybe we are considering some different scenarios. ClipPlanesEvaluator itself seems to be good enough to limit the tiles loaded (within the frustum). 
I have checked with different zoom levels (camera distance) on globe example and it seems that we never load more then ~80 tiles at once with TiltViewClipPlanesEvaluator, and no more then ~110 with old InterpolatedClipPlanesEvaluator (still used as default).
See my screenshots from globe example with memory info (benchmark code) attached the ticket.

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: Fixed.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -609,7 +609,7 @@ export const MapViewDefaults = {
     projection: mercatorProjection,
     clipPlanesEvaluator: defaultClipPlanesEvaluator,
 
-    maxVisibleDataSourceTiles: 120,
+    maxVisibleDataSourceTiles: 500,
```
StefanDachwitz: I cannot approve that setting - 120 is already WAY too high.

StefanDachwitz: This change is also not addressing the issue in the ticket.

kkostecki:
> is not need

Ok, I have checked the scenario you have mentioned, at certain zoom levels (above 7) with some tilt we may reach more then 300 tiles (about 500 at extreme). So what should be the desired approach here?
I think we can not just handle the situation in globe example, because sooner or later we will have full globe support in map view demo. Do you have any suggestions? The ticket itself specifies to remove that option, but what then? Any ideas?

PR #899, opened 2019-10-14
Gather actual max geometry height from tiles.

zbigg:
Fix problems with near clip plane, if geometry contains unnaturally tall extruded polygons.

Instead of hard-coding maximum height of real building, aggregate maximum height
of geometry in `OmvDecodedTileEmitter` and then use as input to `ClipPlanesEvaluator`.

Fixes "black circle" in https://www.harp.gl/docs/master/examples/#datasource_features_polygons.html

- [x] gather elevation from actual data
- [x] use proper clip planes evaluator in `datasource features polygons` example
- [x] check who shall call `updateClipPlanes` now it may be called twice per frame
- [x] cleanup & docs 

Depends on #987 (includes commit).

### DIFF
--- a/@here/harp-examples/src/datasource_features_polygons.ts
+++ b/@here/harp-examples/src/datasource_features_polygons.ts
@@ -13,7 +13,6 @@ import {
 import { sphereProjection } from "@here/harp-geoutils";
 import { MapControls, MapControlsUI } from "@here/harp-map-controls";
 import { MapView } from "@here/harp-mapview";
-import { TiltViewClipPlanesEvaluator } from "@here/harp-mapview/lib/ClipPlanesEvaluator";
 import { APIFormat, OmvDataSource } from "@here/harp-omv-datasource";
 import * as THREE from "three";
 import { accessToken, copyrightInfo } from "../config";
@@ -113,6 +112,8 @@ export namespace PolygonsFeaturesExample {
                 features.push(feature);
             }
             const featuresDataSource = new FeaturesDataSource();
+            featuresDataSource.maxGeometryHeight = 300000;
+
             const addPromise = map.addDataSource(featuresDataSource);
             addPromises.push(addPromise);
             addPromise.then(() => {
@@ -281,7 +282,6 @@ export namespace PolygonsFeaturesExample {
         const mapView = new MapView({
             canvas,
             projection: sphereProjection,
-            clipPlanesEvaluator: new TiltViewClipPlanesEvaluator(),
             theme: {
                 extends: "resources/berlin_tilezen_night_reduced.json",
                 definitions: {
--- a/@here/harp-examples/src/datasource_features_polygons.ts
+++ b/@here/harp-examples/src/datasource_features_polygons.ts
@@ -13,7 +13,7 @@ import {
 import { sphereProjection } from "@here/harp-geoutils";
 import { MapControls, MapControlsUI } from "@here/harp-map-controls";
 import { MapView } from "@here/harp-mapview";
-import { TopViewClipPlanesEvaluator } from "@here/harp-mapview/lib/ClipPlanesEvaluator";
+import { TiltViewClipPlanesEvaluator } from "@here/harp-mapview/lib/ClipPlanesEvaluator";
 import { APIFormat, OmvDataSource } from "@here/harp-omv-datasource";
 import * as THREE from "three";
 import { accessToken, copyrightInfo } from "../config";
@@ -281,7 +281,7 @@ export namespace PolygonsFeaturesExample {
         const mapView = new MapView({
             canvas,
             projection: sphereProjection,
-            clipPlanesEvaluator: new TopViewClipPlanesEvaluator(300000),
+            clipPlanesEvaluator: new TiltViewClipPlanesEvaluator(),
             theme: {
                 extends: "resources/berlin_tilezen_night_reduced.json",
                 definitions: {
--- a/@here/harp-mapview/lib/DataSource.ts
+++ b/@here/harp-mapview/lib/DataSource.ts
@@ -63,6 +63,11 @@ export abstract class DataSource extends THREE.EventDispatcher {
     private m_maxZoomLevel: number = 20;
 
     /**
+     * Current value of [[maxGeometryHeight]] property.
+     */
+    private m_maxGeometryHeight = 0;
+
+    /**
      * Storage level offset applied to this `DataSource`.
      */
     private m_storageLevelOffset: number = 0;
@@ -291,6 +296,20 @@ export abstract class DataSource extends THREE.EventDispatcher {
 
     set maxZoomLevel(level: number) {
         this.m_maxZoomLevel = level;
+    }
+
+    /**
+     * Maximum geometry height above ground level this `DataSource` can produce.
+     *
+     * Used in first stage of frustum culling before [[Tile.maxGeometryHeight]] data is available.
+     *
+     * @default 0.
+     */
+    get maxGeometryHeight() {
+        return this.m_maxGeometryHeight;
+    }
+    set maxGeometryHeight(value: number) {
+        this.m_maxGeometryHeight = value;
     }
 
     /**
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -32,7 +32,7 @@ import { ViewRanges } from "@here/harp-datasource-protocol/lib/ViewRanges";
 import { AnimatedExtrusionHandler } from "./AnimatedExtrusionHandler";
 import { BackgroundDataSource } from "./BackgroundDataSource";
 import { CameraMovementDetector } from "./CameraMovementDetector";
-import { ClipPlanesEvaluator, defaultClipPlanesEvaluator } from "./ClipPlanesEvaluator";
+import { ClipPlanesEvaluator, createDefaultClipPlanesEvaluator } from "./ClipPlanesEvaluator";
 import { IMapAntialiasSettings, IMapRenderingManager, MapRenderingManager } from "./composing";
 import { ConcurrentDecoderFacade } from "./ConcurrentDecoderFacade";
 import { CopyrightInfo } from "./copyrights/CopyrightInfo";
@@ -343,9 +343,9 @@ export interface MapViewOptions {
 
     /**
      * User-defined camera clipping planes distance evaluator.
-     * If not defined, [[DefaultClipPlanesEvaluator]] will be used by [[MapView]].
+     * If not defined, [[TiltViewClipPlanesEvaluator]] will be used by [[MapView]].
      *
-     * @default See [[MapViewDefaults.clipPlanesEvaluator]]
+     * @default [[TiltViewClipPlanesEvaluator]]
      */
     clipPlanesEvaluator?: ClipPlanesEvaluator;
 
@@ -621,7 +621,6 @@ export interface MapViewOptions {
  */
 export const MapViewDefaults = {
     projection: mercatorProjection,
-    clipPlanesEvaluator: defaultClipPlanesEvaluator,
 
     maxVisibleDataSourceTiles: 120,
     extendedFrustumCulling: true,
@@ -820,14 +819,16 @@ export class MapView extends THREE.EventDispatcher {
             ConcurrentDecoderFacade.defaultWorkerCount = this.m_options.decoderCount;
         }
 
-        this.m_visibleTileSetOptions = { ...MapViewDefaults };
+        this.m_visibleTileSetOptions = {
+            ...MapViewDefaults,
+            clipPlanesEvaluator:
+                options.clipPlanesEvaluator !== undefined
+                    ? options.clipPlanesEvaluator
+                    : createDefaultClipPlanesEvaluator()
+        };
 
         if (options.projection !== undefined) {
             this.m_visibleTileSetOptions.projection = options.projection;
-        }
-
-        if (options.clipPlanesEvaluator !== undefined) {
-            this.m_visibleTileSetOptions.clipPlanesEvaluator = options.clipPlanesEvaluator;
         }
 
         if (options.extendedFrustumCulling !== undefined) {
@@ -2939,7 +2940,7 @@ export class MapView extends THREE.EventDispatcher {
 
         if (canRenderTextElements && this.m_textElementsRenderer !== undefined) {
             // copy far value from scene camera, as the distance to the POIs matter now.
-            this.m_screenCamera.far = this.m_camera.far;
+            this.m_screenCamera.far = this.m_viewRanges.maximum;
             this.m_textElementsRenderer.renderText(this.m_screenCamera);
         }
     }
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2442,13 +2442,18 @@ export class MapView extends THREE.EventDispatcher {
         // constantHeight property of extruded polygon technique is set as default false,
         // otherwise the near plane margins will be bigger then required, but still correct.
         const projectionScale = this.projection.getScaleFactor(this.camera.position);
-        const maxHeight = EarthConstants.MAX_BUILDING_HEIGHT * projectionScale;
+        const maxGeometryHeightScaled =
+            projectionScale *
+            this.m_tileDataSources.reduce((r, ds) => Math.max(r, ds.maxGeometryHeight), 0);
+
         // Copy all properties from new view ranges to our readonly object.
         // This allows to keep all view ranges references valid and keeps up-to-date
         // information within them. Works the same as copping all properties one-by-one.
         Object.assign(
             this.m_viewRanges,
-            viewRanges === undefined ? this.m_visibleTiles.updateClipPlanes(maxHeight) : viewRanges
+            viewRanges === undefined
+                ? this.m_visibleTiles.updateClipPlanes(maxGeometryHeightScaled)
+                : viewRanges
         );
         this.m_camera.near = this.m_viewRanges.near;
         this.m_camera.far = this.m_viewRanges.far;
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -280,6 +280,11 @@ export class Tile implements CachedResource {
     readonly boundingBox = new OrientedBox3();
 
     /**
+     * Maximum height of geometry on this tile above ground level.
+     */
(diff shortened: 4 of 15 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2432,18 +2432,9 @@ export class MapView extends THREE.EventDispatcher {
...
-        // otherwise the near plane margins will be bigger then required, but still correct.
-        const projectionScale = this.projection.getScaleFactor(this.camera.position);
-        const maxHeight = EarthConstants.MAX_BUILDING_HEIGHT * projectionScale;
-        // Copy all properties from new view ranges to our readonly object.
-        // This allows to keep all view ranges references valid and keeps up-to-date
-        // information within them. Works the same as copping all properties one-by-one.
         Object.assign(
             this.m_viewRanges,
-            viewRanges === undefined ? this.m_visibleTiles.updateClipPlanes(maxHeight) : viewRanges
+            viewRanges === undefined ? this.m_visibleTiles.updateClipPlanes() : viewRanges
```
kkostecki: Consider what may happen when decoder will not calculate maxBuildingHeight - then we leave clip planes without accounting for any buildings. The base approach though would be to have tile.maxGeometryHeight set to EarthConstants.MAX_BUILDING_HEIGHT by default. See also my comments below.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2968,6 +2959,17 @@ export class MapView extends THREE.EventDispatcher {
...
+        // When calculating clip planes account for the highest building on the earth,
+        // multipling its height by projection scalling factor. This approach assumes
+        // constantHeight property of extruded polygon technique is set as default false,
+        // otherwise the near plane margins will be bigger then required, but still correct.
+        const projectionScale = this.projection.getScaleFactor(this.camera.position);
+        const maxHeight = EarthConstants.MAX_BUILDING_HEIGHT * projectionScale;
+        // Copy all properties from new view ranges to our readonly object.
+        // This allows to keep all view ranges references valid and keeps up-to-date
+        // information within them. Works the same as copping all properties one-by-one.
+        this.m_visibleTiles.updateClipPlanes(maxHeight);
```
kkostecki: I see no point of updating clip planes here. The evaluation using EarthConstants.MAX_BUILDING_HEIGHT as default buildings/features height is done for broad phase tiles elimination (frustum culling) and thus allows to see tiles on globe with high geometry behind earth horizon (surface), without maxHeight (even naive) defined is may clip those tiles out in the tile rejection phase, so those geometries (buildings) may be never visible or actually pop-out immediately from behind horizon.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -592,13 +592,16 @@ export class VisibleTileSet {
             maxElevation < EarthConstants.MAX_ELEVATION - maxBuildingHeight
                 ? maxElevation + maxBuildingHeight
                 : maxElevation;
+
         // Update elevation ranges stored.
+        let needUpdateClipPlanes = false;
```
kkostecki: Good point, with buildings height this is definitely needed.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -592,13 +592,16 @@ export class VisibleTileSet {
             maxElevation < EarthConstants.MAX_ELEVATION - maxBuildingHeight
                 ? maxElevation + maxBuildingHeight
                 : maxElevation;
```
kkostecki: If buildings height is now included (accounted) in tile elevation, there is no need to add it here.... My proposal would be to define tile.maxGeometryHeight = EarthConstants.MAX_ELEVATION as default. Then before adding it to tileMaxElevation account for projection scale. 
So in simple words if decoder will not provide accurate value, we will use Abu Khalifa, if accurate height is delivered from decoder then we will have better accuracy and more fine tuned near plane position.

@here/harp-examples/src/datasource_features_polygons.ts
```diff
@@ -281,7 +281,7 @@ export namespace PolygonsFeaturesExample {
         const mapView = new MapView({
             canvas,
             projection: sphereProjection,
-            clipPlanesEvaluator: new TopViewClipPlanesEvaluator(300000),
+            clipPlanesEvaluator: new TiltViewClipPlanesEvaluator(),
```
kkostecki: Good point, since globe controls have now tilt added, somebody should update this too.

ninok: Please remove this and make TiltViewClipPlanesEvaluator the default one

zbigg:
Yes, but it's much to too far w.r.t what we're trying to fix here.
We want to fix clipping planes in renderer, which are wrong. Finding tiles is another story and may be built on top of this.

@ninok The idea that we came up with @kkostecki is that, we should
 - introduce new "configurable" constant in `DataSource` - `maxGeometryHeight`
 - use this constant (aggregated from datasources) in first phase of frustum culling
 - use actual data from tiles in second phase, when rendering clip planes are calculated

So, 
> Actually the geometry height of a tile needs to be taken into account already when traversing the tile tree

For now, it's kind-of impossible because we're calculating frustum with no reference to datasource, that's why have

```
this.m_visibleTiles.updateClipPlanes(maxHeight)
```

In update camera in each map render.

I'm preparing improved version of patch and think it should get in, even if initial culling is not perfect.

kkostecki:
> I think this whole topic needs a bit more rethinking. Actually the geometry height of a tile needs to be taken into account already when traversing the tile tree (i.e. in `VisibleTileSet.getVisibleTileKeysForDataSources`). Instead of always using the flat bounding box computed out of the tile key we should check if a tile is in the cache and use it's proper bounding box and if it's not in cache we should fallback to the flat bounding box + some fallback height (maybe configurable).
> Unfortunately was there some smart caching/bucketing introduced to optimize the tile tree traversal. This needs to be removed/reworked to take into account that different data sources will have different bounding boxes, even if the tiling scheme is the same (we could e.g. always use the maximum bounding box)

I guess when @zbigg  will update PR with our new ideas, everything should be more clear.

@here/harp-examples/src/datasource_features_polygons.ts
kkostecki: I think it should be removed in the patch-set that changes the TiltViewClipPlanesEvaluator to default one. We need to firstly merge this one, in order not to break tests, etc....

kkostecki:
> * phase, when rendering c



> Yes, but it's much to too far w.r.t what we're trying to fix here.
> We want to fix clipping planes in renderer, which are wrong. Finding tiles is another story and may be built on top of this.
> 
> @ninok The idea that we came up with @kkostecki is that, we should
> 
> * introduce new "configurable" constant in `DataSource` - `maxGeometryHeight`
> * use this constant (aggregated from datasources) in first phase of frustum culling
> * use actual data from tiles in second phase, when rendering clip planes are calculated
> 
> So,
> 
> > Actually the geometry height of a tile needs to be taken into account already when traversing the tile tree
> 
> For now, it's kind-of impossible because we're calculating frustum with no reference to datasource, that's why have
> 
> ```
> this.m_visibleTiles.updateClipPlanes(maxHeight)
> ```
> 
> In update camera in each map render.
> 
> I'm preparing improved version of patch and think it should get in, even if initial culling is not perfect.

Please update PR as we agreed - DataSource with default max buildings height.

@here/harp-mapview/lib/MapView.ts
zbigg: Agreed, removed.

@here/harp-examples/src/datasource_features_polygons.ts
zbigg: TiltViewClipPlanesEvaluator is now default as part of this PR.
This should make #922 obsolete.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -597,39 +592,16 @@ export class VisibleTileSet {
             const tiles = renderListEntry.visibleTiles;
             tiles.forEach(tile => {
                 minElevation = Math.min(minElevation, tile.minElevation);
-                maxElevation = Math.max(maxElevation, tile.maxElevation);
+                maxElevation = Math.max(maxElevation, tile.maxElevation + tile.maxGeometryHeight);
```
kkostecki: This already contains a bug, see my comments above: maxElevation (getter) already sums up elevation and geometry height.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -597,39 +592,16 @@ export class VisibleTileSet {
...
         // If clip planes evaluator depends on the tiles elevation re-calculate
         // frustum planes and update the camera near/far plane distances.
         let viewRangesChanged: boolean = false;
-        if (elevationRangeSource !== undefined) {
-            const viewRanges = this.updateClipPlanes(maxElevation, minElevation);
-            viewRangesChanged = viewRanges === this.m_viewRange;
-            this.m_viewRange = viewRanges;
-        }
+        const viewRanges = this.updateClipPlanes(maxElevation, minElevation);
+        viewRangesChanged = viewRanges === this.m_viewRange;
```
kkostecki: We already discussed it - compare component wise (this was my bug before : ) ).

@here/harp-mapview/test/VisibleTileSetTest.ts
```diff
@@ -51,11 +52,10 @@ class Fixture {
...
             params.tileWrappingEnabled
         );
-        this.vts = new VisibleTileSet(
-            this.frustumIntersection,
-            this.tileGeometryManager,
-            MapViewDefaults
-        );
+        this.vts = new VisibleTileSet(this.frustumIntersection, this.tileGeometryManager, {
+            ...MapViewDefaults,
+            clipPlanesEvaluator: new TiltViewClipPlanesEvaluator()
```
kkostecki: Somehow it shows the need for default factory method  ... but as you wish....

@here/harp-examples/src/datasource_features_polygons.ts
```diff
@@ -281,7 +280,6 @@ export namespace PolygonsFeaturesExample {
         const mapView = new MapView({
             canvas,
             projection: sphereProjection,
-            clipPlanesEvaluator: new TopViewClipPlanesEvaluator(300000),
```
kkostecki: Here we will need to pass this visualization geometry height into (300000) DataSource as (default) maxGeometryHeight.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2419,18 +2420,15 @@ export class MapView extends THREE.EventDispatcher {
...
 
-        // When calculating clip planes account for the highest building on the earth,
-        // multipling its height by projection scalling factor. This approach assumes
-        // constantHeight property of extruded polygon technique is set as default false,
-        // otherwise the near plane margins will be bigger then required, but still correct.
-        const projectionScale = this.projection.getScaleFactor(this.camera.position);
-        const maxHeight = EarthConstants.MAX_BUILDING_HEIGHT * projectionScale;
-        // Copy all properties from new view ranges to our readonly object.
-        // This allows to keep all view ranges references valid and keeps up-to-date
-        // information within them. Works the same as copping all properties one-by-one.
```
kkostecki: I think we need to leave with this projection scale outside of updateClipPlanes call. We may not move it inside method, because after accurate (tile base) height evaluation we have values already scaled.

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -293,6 +298,20 @@ export abstract class DataSource extends THREE.EventDispatcher {
...
 
+    /**
+     * Maximum geometry height above ground level this `DataSource` can produce.
+     *
+     * Used in first stage of frustum culling before [[Tile.maxGeometryHeight]] data is available.
+     *
+     * @default [[EarthConstants.MAX_BUILDING_HEIGHT]].
+     */
+    get maxGeometryHeight() {
+        return this.m_maxGeometryHeight;
```
ninok: I think the default should be 0 and only for OmvDataSource it should be MAX_BUILDING_HEIGHT by default,
especially since the default for Tile.maxBuildingHeight is 0 as well.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2419,18 +2420,15 @@ export class MapView extends THREE.EventDispatcher {
...
+        const maxGeometryHeight = this.m_tileDataSources.reduce(
+            (r, ds) => Math.max(r, ds.maxGeometryHeight),
+            0
+        );
         Object.assign(
             this.m_viewRanges,
-            viewRanges === undefined ? this.m_visibleTiles.updateClipPlanes(maxHeight) : viewRanges
+            viewRanges === undefined
+                ? this.m_visibleTiles.updateClipPlanes(maxGeometryHeight)
```
ninok: I don't get this code. We update the clip planes only the first time?

@here/harp-mapview/lib/Tile.ts
```diff
@@ -664,7 +669,7 @@ export class Tile implements CachedResource {
      * @note Negative values indicates depressions.
      */
     get maxElevation(): number {
-        return this.m_maxElevation;
+        return this.m_maxElevation + this.maxGeometryHeight;
```
kkostecki: I think we may not leave it inconsistent, some new getter should be added, i.e. get maxHeight, otherwise simple unit test fails:
```
tile.maxElevation = 100;
assert_equal(tile.maxElevation, 100);
```
and more importantly it becomes misleading.

@here/harp-mapview/lib/VisibleTileSet.ts
ninok: Isn't that always true?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -597,39 +592,16 @@ export class VisibleTileSet {
...
 
         // If clip planes evaluator depends on the tiles elevation re-calculate
         // frustum planes and update the camera near/far plane distances.
         let viewRangesChanged: boolean = false;
-        if (elevationRangeSource !== undefined) {
-            const viewRanges = this.updateClipPlanes(maxElevation, minElevation);
-            viewRangesChanged = viewRanges === this.m_viewRange;
-            this.m_viewRange = viewRanges;
-        }
+        const viewRanges = this.updateClipPlanes(maxElevation, minElevation);
```
kkostecki: Would be nice to add at least some comments that elevation caching is possible, such as together with camera movement information we could possible omit update... But that's later story and I am not sure if worth entire effort.

@here/harp-mapview/lib/DataSource.ts
zbigg: ok

@here/harp-mapview/lib/MapView.ts
zbigg: I don't get old code too it's old code, but formatted.
By not fixing all the smells i just try to minimize patch and not modify unrelated code. I'll check if this can fixed easily.

@here/harp-mapview/lib/VisibleTileSet.ts
zbigg: Yes, again ...
By not fixing all the smells i just try to minimize patch and not modify unrelated code. I'll check if this can fixed easily.

@here/harp-mapview/test/VisibleTileSetTest.ts
zbigg: ok, done

@here/harp-examples/src/datasource_features_polygons.ts
zbigg: done

@here/harp-mapview/lib/Tile.ts
zbigg: fixed

@here/harp-mapview/lib/VisibleTileSet.ts
zbigg: As discussed above, this code never worked and i can't tell now, if fixing the issue will not being any regressions. Trying to fix it now, only increases chances of regressions, but let's see.

@here/harp-mapview/lib/VisibleTileSet.ts
zbigg: fixed, `Tile.maxElevation` contains ground elevation

@here/harp-mapview/lib/VisibleTileSet.ts
zbigg: Fixed

kkostecki:

@here/harp-mapview/test/MapViewTest.ts
```diff
@@ -59,8 +59,8 @@ describe("MapView", function() {
         addEventListenerSpy = sinon.stub();
         removeEventListenerSpy = sinon.stub();
         canvas = ({
-            clientWidth: 0,
-            clientHeight: 0,
+            clientWidth: 400,
```
kkostecki: Should be big enough to provide tested precision: probably 1920x1080 as in other tests.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -1173,3 +1155,9 @@ export class VisibleTileSet {
         return { tileKeys, allBoundingBoxesFinal };
     }
 }
+
+function viewRangesEqual(a: ViewRanges, b: ViewRanges) {
```
kkostecki: I would move this method to ViewRanges interface: equal(v: ViewRanges).

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -588,48 +583,35 @@ export class VisibleTileSet {
...
-        ) {
-            this.m_elevationRange.minElevation = minElevation;
-            this.m_elevationRange.maxElevation = maxElevation;
-        }
-        this.m_elevationRange.calculationStatus = visibleTileKeysResult.allBoundingBoxesFinal
-            ? CalculationStatus.FinalPrecise
-            : CalculationStatus.PendingApproximate;
 
+        if (minElevation === undefined) {
+            minElevation = EarthConstants.MIN_ELEVATION;
```
kkostecki: Probably:
minElevation = 0;
maxElevation = 0;
We do not wan't to extend planes (instead of required margins) if no elevation info is provided.

zbigg: done

zbigg:

zbigg:

PR #901, opened 2019-10-14
HARP-7161: Support for UrlResolvers in MapView.

zbigg:
Support for cutom URIs in Theme.

  * Adds `uriResolver` option to `MapView` that provides means to
      use resolve application specific URIs in theme.
    
  * `MapView` and `ThemeLoader` resolves all provided `URI`s using
      provided `UriResolver`
   * URL->URI in places when we support `URI`s in API
    
  * Added, basic, prefix-map based resolver - `PrefixMapUriResolver`.

Tasks

* [x] add `urlResolver` to MapView
* [x] add doc
* [x] whether to resolve als decoder/tiles urls, (urls passed to `OmvDataSource` cannot be resolved in custructor, as this.mapview is not yet known) -> not in scope
* [x] provide reference basic resolver
* [x] stabilize API
* [x] cleanup `URI` and `URL`s in API names
* [x] add tests


@here/harp-mapview/lib/MapView.ts
```diff
@@ -312,6 +319,48 @@ export interface MapViewOptions {
...
+     * 1) you should ensure that `@here/harp-fontcatalog/resources` is either packaged with your app
+     *   or available from some CDN;
+     *
+     * 2) `MapView` should be configured, where to find `@here/harp-fontcatalog` files, for example
+     * with [[createSimpleUrlResolver]]:
+     *
+     * ```
+     * uriResolver: createSimpleUrlResolver({
+     *    // typical, `harp-examples`-like deployment
+     *    "npm://@here/harp-fontcatalog/resources/": "/resources/fonts/",
```
ninok: That's a nice usecase but I think the main reason why we introduce this is to resolve the asset path like:
`font://FiraSans`, `icon://day/zoo`, ...

@here/harp-utils/lib/UrlResolver.ts
```diff
@@ -4,38 +4,45 @@
...
+ *       "npm://@here/harp-fontcatalog/":
+ *            "https://unpkg.com/@here/harp-fontcatalog/",
+ *       # but theme theme assets from same app base url
+ *       "npm://@here/harp-map-theme/resources": "/resources/"
+ *    })
  */
-export function setDefaultUrlResolver(resolver: UrlResolver | undefined) {
-    customDefaultUrlResolver = resolver;
+export function createSimpleUrlResolver(definition: SimpleUrlResolverDefinition) {
+    return (uri: string) => {
```
ninok: I would prefer to have a class for the UrlResolver

zbigg: done

@here/harp-mapview/lib/MapView.ts
zbigg: Documentation updated to be more abstract, see https://github.com/heremaps/harp.gl/pull/901/files

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1273,6 +1309,17 @@ export class MapView extends THREE.EventDispatcher {
...
 
+    /**
+     * [[UriResolver]] used to resolve application/deployment specific `URI`s into actual `URLs`
+     * that can be loaded with `fetch`.
+     */
+    get uriResolver(): UriResolver | undefined {
+        return this.m_uriResolver;
+    }
+    set uriResolver(uriResolver: UriResolver | undefined) {
+        this.m_uriResolver = uriResolver;
```
ninok: Not sure if it's a good idea to add this setter. To make this work properly the theme would need to be re-resolved.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -641,6 +670,8 @@ export const MapViewDefaults = {
  * linked to datasources.
  */
 export class MapView extends THREE.EventDispatcher {
+    static defaultUriResolver?: UriResolver;
```
ninok: Why is this one optional? Shouldn't there always be a default one?

zbigg: This was not necessary, removed completly.

PR #902, opened 2019-10-14
HARP-7356: Fixed wrong handling of tiles that only contain text eleme…

ninok:
…nts but no geometry.

If a tile contains no geometry but just text the Tile.hasGeometry flag was set to false (b/c Tile.forceGeometry was not called). Since hasGeometry is used in the VisibleTileSet computation to deduce if a tile is loaded(what is also totally wrong but will be adressed in a separate PR) no tile was rendered.
With this change forceGeometry will be called if the tile has no geometry but text.
Additionally the original decodedTile will be kept and it will not be replaced with a dummy empty decodedTile (this was also requested by OLP)

### DIFF
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -37,15 +37,6 @@ interface DisposableObject {
     material?: THREE.Material[] | THREE.Material;
 }
 
-function decodedTileHasGeometry(decodedTile: DecodedTile) {
-    return (
-        decodedTile.geometries.length ||
-        (decodedTile.poiGeometries !== undefined && decodedTile.poiGeometries.length) ||
-        (decodedTile.textGeometries !== undefined && decodedTile.textGeometries.length) ||
-        (decodedTile.textPathGeometries !== undefined && decodedTile.textPathGeometries.length)
-    );
-}
-
 /**
  * An interface for optional feature data that is saved in a `THREE.Object3D`'s `userData`
  * property.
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/Tile.ts
```diff
@@ -702,15 +685,34 @@ export class Tile implements CachedResource {
...
+        if (decodedTile.geometries.length === 0) {
             this.forceHasGeometry(true);
         }
+
+        if (decodedTile.boundingBox !== undefined) {
+            // If the decoder provides a more accurate bounding box than the one we computed from
+            // the flat geo box we take it instead.
+            this.boundingBox.copy(decodedTile.boundingBox);
+        }
+        this.invalidateResourceInfo();
```
CraigBerlin: Is this still needed after the call on line 689?

nzjony:
What was the root cause? Please add to the commit message, I can't see the bug from reading the code unfortunately

@here/harp-mapview/lib/Tile.ts
ninok: ah, true. To much copy paste in this change.

nzjony:
Also, how difficult would it be to add a test which shows the wrong behaviour?

ninok:
> What was the root cause? Please add to the commit message, I can't see the bug from reading the code unfortunately

sure, will do

ninok:
> Also, how difficult would it be to add a test which shows the wrong behaviour?

Again I will be the pioneer to add a test for this class :-( Anyways I will try

nzjony:
> > Also, how difficult would it be to add a test which shows the wrong behaviour?
> 
> Again I will be the pioneer to add a test for this class :-( Anyways I will try

Not quite, I introduce a test for Tile here: https://github.com/heremaps/harp.gl/pull/865/commits/88b9fbd43a9c1db687310f7ac872e6ebf8e7b322

ninok:
> > > Also, how difficult would it be to add a test which shows the wrong behaviour?
> > 
> > 
> > Again I will be the pioneer to add a test for this class :-( Anyways I will try
> 
> Not quite, I introduce a test for Tile here: [88b9fbd](https://github.com/heremaps/harp.gl/commit/88b9fbd43a9c1db687310f7ac872e6ebf8e7b322)
> 
> (technically it is a test for TileGeometryCreator, but I know it calls a method on Tile... so you are technically correct)

I see, but your change is not on master. Anyways it was quite easy to add this test. Thanks for pushing me to do it ;-)

PR #903, opened 2019-10-15
HARP-7392 Add urlParams to OmvRestClient

musculman:

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvRestClient.ts
+++ b/@here/harp-omv-datasource/lib/OmvRestClient.ts
@@ -216,6 +216,12 @@ export interface OmvRestClientParameters {
      * @deprecated Please use [[authenticationCode]].
      */
     getBearerToken?: () => Promise<string>;
+
+    /**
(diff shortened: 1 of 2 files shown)


@here/harp-omv-datasource/lib/OmvRestClient.ts
```diff
@@ -216,19 +216,29 @@ export interface OmvRestClientParameters {
      * @deprecated Please use [[authenticationCode]].
      */
     getBearerToken?: () => Promise<string>;
+
+    /**
+     * Array of query parameters to be appended at the end of the url.
+     * It is empty by default.
+     */
+    urlParams?: Array<[string, string]>;
```
ninok: why array of arrays? Would be more natural to write sth. like:
```
{
    "param": "value",
    ...
}
```

musculman: I just took the same value as in the private function, but agree, that's nicer, let me change it

PR #885, opened 2019-10-09
Minor: Cleaned-up line shader.

amsXYZ:
Did some changes to the line shader needed for future line animations tickets:
- [x] Make the PS work in **normalized line space** (goes from ``[0,0]`` to ``[lineLength, lineWidth]``, with out-of-bounds values reserved for outline lines).
- [X] Removed any conditional branching in the shader in favor of **value masking** (should improve performance).
- [X] Reduced the number of varying attributes passed down to the PS, which should reduce vector register pressure.
- [X] Renamed variables and commented out shader for better understanding of the algorithms in place.

PR #916, opened 2019-10-17
HARP-7204: Implemented animated lines.

amsXYZ:
* PR addressing animation of lines using ``SolidLineMaterial.drawRange``.
* Also addressed line joining for smooth animations when using ``GeoJsonDataProvider``.

Depends on #885


pit-rpg:
now there are a few bugs:
 - if enable the line caps they stay on the line ends.
 - continuous flickering against building edge lines.
 - artifacts on the edges less than 90 degrees. they are noticeable with slow animation speed

pit-rpg:
It also looks good to me.

PR #917, opened 2019-10-17
HARP-6636: Fix MapView.FrameComplete event

StefanDachwitz:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -745,6 +745,7 @@ export class MapView extends THREE.EventDispatcher {
     private m_previousFrameTimeStamp?: number;
     private m_firstFrameRendered = false;
     private m_firstFrameComplete = false;
+    private m_initialTextPlacementDone = false;
     private m_previousRequestAnimationTime?: number;
     private m_targetRequestAnimationTime?: number;
     private m_frameTimeIndex: number = 0;
@@ -1109,7 +1110,7 @@ export class MapView extends THREE.EventDispatcher {
         this.updateImages();
         this.updateLighting();
 
-        if (this.m_textElementsRenderer) {
+        if (this.m_textElementsRenderer !== undefined) {
             this.m_textElementsRenderer.placeAllTileLabels();
         }
         this.updateSkyBackground();
@@ -1506,13 +1507,13 @@ export class MapView extends THREE.EventDispatcher {
     /**
      * Get object describing frustum planes distances and min/max visibility range for actual
      * camera setup.
-     * Near and far plance distance are self explanatory while minimum and maximum visibility range
+     * Near and far plane distance are self explanatory while minimum and maximum visibility range
      * describes the extreme near/far planes distances that may be achieved with current camera
      * settings, meaning at current zoom level (ground distance) and any possible orientation.
      * @note Visibility is directly related to camera [[ClipPlaneEvaluator]] used and determines
      * the maximum possible distance of camera far clipping plane regardless of tilt, but may change
-     * whenever zoom level changes. Distance is meassured in world units which may be approximately
-     * euqal to meters, but this depends on the distortion related to projection type used.
+     * whenever zoom level changes. Distance is measured in world units which may be approximately
+     * equal to meters, but this depends on the distortion related to projection type used.
      */
     get viewRanges(): ViewRanges {
         return this.m_viewRanges;
@@ -1980,7 +1981,12 @@ export class MapView extends THREE.EventDispatcher {
      * Returns `true` if the current frame will immediately be followed by another frame.
      */
     get isDynamicFrame(): boolean {
-        return this.cameraIsMoving || this.animating || this.m_updatePending;
+        return (
+            this.cameraIsMoving ||
+            this.animating ||
+            this.m_updatePending ||
+            this.m_animatedExtrusionHandler.isAnimating
+        );
     }
 
     /**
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -997,6 +998,14 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_textElementsRenderer;
     }
 
+    /**
+     * @hidden
+     * Is `true` if the TextElementRenderer is soon to be created.
+     */
+    get textElementsRendererRequested(): boolean {
```
zbigg: Why exposing so obscure internal property esp, it's only used once.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2713,6 +2722,36 @@ export class MapView extends THREE.EventDispatcher {
...
+            if (!this.textElementsRendererRequested || textElementRendererFinished) {
+                if (textElementRendererFinished) {
+                    // Placement of text shall be done now.
+                    this.m_textElementsRenderer!.placeAllTileLabels();
+                }
+                this.m_initialTextPlacementDone = true;
+            }
+            // Either render the labels after placing them, or wait for the TextElementsRenderer to
+            // be created (delayed creation).
+            this.update();
```
zbigg: This doesn't look right, that we've just almost said "frame complete"  and in very same time request new update.
So, is it compete or not ?

By _almost_ i mean If we're here, then following new "if" condition is always true
 * `this.m_initialTextPlacementDone` true, as it's set here
 * `!this.m_firstFrameComplete` -> true
 * `(this.m_textElementsRenderer === undefined ||
                (this.m_textElementsRenderer.ready && !this.m_textElementsRenderer.loading)) ` -> true as checked above 
 * `!this.m_animatedExtrusionHandler.isAnimating` -> may be always true if dynamic extrusion is disabled -> true

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2713,6 +2722,36 @@ export class MapView extends THREE.EventDispatcher {
...
+            !this.m_initialTextPlacementDone &&
+            !this.m_themeIsLoading &&
+            this.m_visibleTiles.allVisibleTilesLoaded &&
+            this.m_connectedDataSources.size + this.m_failedDataSources.size ===
+                this.m_tileDataSources.length &&
+            this.m_poiTableManager.finishedLoading
+        ) {
+            const textElementRendererFinished =
+                this.m_textElementsRenderer !== undefined &&
+                this.m_textElementsRenderer.ready &&
```
zbigg: Maybe instead of juggling with partial states of TER just expose proper `TER.ready` which checks also `loading` state.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2713,6 +2722,36 @@ export class MapView extends THREE.EventDispatcher {
...
 
+        // Check if this is the time to place the labels for the first time. Pretty much everything
+        // should have been loaded, and no animation should be running.
+        if (
+            !this.m_initialTextPlacementDone &&
+            !this.m_firstFrameComplete &&
+            !this.isDynamicFrame &&
+            !this.m_initialTextPlacementDone &&
+            !this.m_themeIsLoading &&
+            this.m_visibleTiles.allVisibleTilesLoaded &&
```
zbigg: IMO, ImageCache check is still missing from this "litany"

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: TextElementsRenderer is public, there is no control over when/if it is created. Now it is accessible if it _will_ be available soon. Given the fact that we have/had so many issues with that first frame complete event, I think it is valuable.

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: I pondered that idea as well. But the ready state is final, while the loading state may change when new character sets are being loaded, no?

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: Not calling update may let us without an update if the TER is finally created, of if the TER is finally loaded. As far as I can see, these conditions make sense.

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: Yes, a litany it is. ImageCache may be a bit tricky, because not every imageItem in it is being loaded, since that is optional. But the images for the POIs should be handled. I would rather add that check and eventually fix it once we have POIs to render in our default themes.

@here/harp-mapview/lib/MapView.ts
ninok: Can we instead do it properly and always create the TextElementsRenderer?
I understand that we don't want to create it if we have no text b/c we would load the font for nothing, but I think we should load the font lazily instead of creating the whole TextElementsRenderer lazily

StefanDachwitz:

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: This will increase startup time. @Nino: Please, we need executive decision on that.

StefanDachwitz:

@here/harp-mapview/lib/MapView.ts
ninok: Why will this increase startup time? As I said, the text elements renderer should always be created but the font's inside the text elements renderer should be loaded lazily. Please sync with @atomicsulfate He is refactorin TextElementsRenderer right now.

PR #922, opened 2019-10-18
MINOR: Use TiltViewClipPlanesEvaluator by default

ninok:


kkostecki:
We need to firstly fix the unit test (failing) it looks (100% sure) that few tests have bugs there. I.e. canvas can not be set to 0, 0 when why try to raycast - it requires NDC coordinates which can not be calculated there.

zbigg:
Depends on #939

zbigg:

@here/harp-mapview/lib/MapView.ts
```diff
@@ -807,16 +806,18 @@ export class MapView extends THREE.EventDispatcher {
...
+        };
 
         if (options.projection !== undefined) {
             this.m_visibleTileSetOptions.projection = options.projection;
         }
 
-        if (options.clipPlanesEvaluator !== undefined) {
-            this.m_visibleTileSetOptions.clipPlanesEvaluator = options.clipPlanesEvaluator;
-        }
-
```
kkostecki: Why it is removed? This way we disallow users to specify it via options...

@here/harp-mapview/test/VisibleTileSetTest.ts
```diff
@@ -51,11 +52,10 @@ class Fixture {
...
             params.tileWrappingEnabled
         );
-        this.vts = new VisibleTileSet(
-            this.frustumIntersection,
-            this.tileGeometryManager,
-            MapViewDefaults
-        );
+        this.vts = new VisibleTileSet(this.frustumIntersection, this.tileGeometryManager, {
+            ...MapViewDefaults,
+            clipPlanesEvaluator: new TiltViewClipPlanesEvaluator()
```
kkostecki: Personally I would replace this with simple factory method (createDefaultClipPlanesEvaluator()), of course we all know the reason that default (instance) can not be used, but factory gives us flexibility of changing it once and everywhere if needed.

@here/harp-mapview/test/MapViewTest.ts
```diff
@@ -80,7 +80,13 @@ describe("MapView", function() {
...
     });
 
-    it("Correctly sets geolocation and zoom", function() {
+    //
+    // This test is broken, because `setCameraGeolocationAndZoom` doesn't behave as expected, i.e
+    // it offsets actual `geoCenter` a litte.
+    //
+    // TODO: check who is right? this test or `setCameraGeolocationAndZoom` implementation
+    //
+    it.skip("Correctly sets geolocation and zoom", function() {
```
kkostecki: I guess we should create a ticket for that, before we forget about this problem, I think there is some bug in setCameraGeoloactionAndZoom. - we need to investigate.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -892,9 +892,3 @@ export class FixedClipPlanesEvaluator implements ClipPlanesEvaluator {
         this.m_farPlane = farDist;
     }
 }
-
-/**
- * Default [[ClipPlanesEvaluator]] calculates near plane based on ground distance.
- */
-// tslint:disable-next-line: max-line-length
-export const defaultClipPlanesEvaluator: ClipPlanesEvaluator = new InterpolatedClipPlanesEvaluator();
```
kkostecki: I would exchange it for simple factory function:
createDefaultClipPlanesEvaluator().

@here/harp-examples/src/datasource_features_polygons.ts
```diff
@@ -281,7 +281,7 @@ export namespace PolygonsFeaturesExample {
         const mapView = new MapView({
             canvas,
             projection: sphereProjection,
-            clipPlanesEvaluator: new TopViewClipPlanesEvaluator(300000),
+            clipPlanesEvaluator: new TiltViewClipPlanesEvaluator(300000),
```
kkostecki: This may be removed if buildings (constructions) height will be embedded/evaluated in DataSource. Then datasource (in this example) should receive its defaultMaxHeight = 300000.

@here/harp-mapview/lib/MapView.ts
zbigg: It's still possible, look few lines above.

zbigg:
Delivered as part of #899

PR #924, opened 2019-10-18
HARP-7322: City center fixes

ninok:
* Fix wrong calculation of icon size
* Fix wrong handling of dynamic string properties that are not numeral
* Added interpolation to following properties:
  * xOffset
  *  yOffset
  * iconXOffset
  * iconYOffset
  * screenWidth
  * screenHeight
  * vAlignment
  * hAlignment

### DIFF
--- a/@here/harp-datasource-protocol/lib/TechniqueParams.ts
+++ b/@here/harp-datasource-protocol/lib/TechniqueParams.ts
@@ -548,19 +548,19 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
     /**
      * Horizontal offset (to the right) in screen pixels.
      */
-    xOffset?: number;
+    xOffset?: DynamicProperty<number>;
     /**
      * Vertical offset (up) in screen pixels.
      */
-    yOffset?: number;
+    yOffset?: DynamicProperty<number>;
     /**
      * Horizontal offset (to the right) in screen pixels.
      */
-    iconXOffset?: number;
+    iconXOffset?: DynamicProperty<number>;
     /**
      * Vertical offset (up) in screen pixels.
      */
-    iconYOffset?: number;
+    iconYOffset?: DynamicProperty<number>;
     /**
      * Scaling factor of icon.
      */
@@ -568,11 +568,11 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
     /**
      * Vertical height in pixels, controls vertical scaling. Overrides `iconScale`.
      */
-    screenHeight?: number;
+    screenHeight?: DynamicProperty<number>;
     /**
      * Horizontal height in pixels, controls horizontal scaling. Overrides `iconScale`.
      */
-    screenWidth?: number;
+    screenWidth?: DynamicProperty<number>;
     /**
      * Name of the POI table which should be used for this POI.
      */
@@ -682,11 +682,11 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
     /**
      * Text position regarding the baseline.
      */
-    hAlignment?: "Left" | "Center" | "Right";
+    hAlignment?: DynamicProperty<"Left" | "Center" | "Right">;
     /**
      * Text position inside a line.
      */
-    vAlignment?: "Above" | "Center" | "Below";
+    vAlignment?: DynamicProperty<"Above" | "Center" | "Below">;
 }
 
 export interface LineTechniqueParams extends BaseTechniqueParams {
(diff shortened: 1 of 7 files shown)


@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -682,11 +682,11 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
     /**
      * Text position regarding the baseline.
      */
-    hAlignment?: "Left" | "Center" | "Right";
+    hAlignment?: DynamicProperty<"Left" | "Center" | "Right">;
```
nzjony: How do you interpolate between strings?

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -491,7 +491,13 @@ export class PoiManager {
     ): TextElement {
         const priority = technique.priority !== undefined ? technique.priority : 0;
         const positions = Array.isArray(x) ? (x as THREE.Vector3[]) : new THREE.Vector3(x, y, z);
+
+        // The current zoomlevel of mapview. Since this method is called for all tiles in the
+        // VisibleTileSet we can be sure that the current zoomlevel matches the zoomlevel where
+        // the tile should be shown.
         const displayZoomLevel = this.mapView.zoomLevel;
+        const floorZoomLevel = Math.floor(displayZoomLevel);
+        assert(floorZoomLevel === tile.tileKey.level - tile.dataSource.storageLevelOffset);
```
nzjony: why do you have this assertion here? This would point to some issue in the VisibleTileSet?

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -348,13 +348,13 @@ export class PoiRenderer {
      *
      * @returns `True` if the space is not already allocated by another object (text label or POI)
      */
-    prepareRender(pointLabel: TextElement): boolean {
+    prepareRender(pointLabel: TextElement, zoomLevel: number): boolean {
```
nzjony: please add to documentation

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -590,7 +602,12 @@ export class PoiRenderer {
      * @param imageTexture Shared [[ImageTexture]], defines used area in atlas.
      * @param imageItem Shared [[ImageItem]], contains cached image for texture.
```
nzjony: please update params

@here/harp-mapview/lib/poi/PoiManager.ts
ninok: The assert is here to make it more clear and that the following code is relying on it.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
ninok: Only discrete interpolation is supported right now.

nzjony: Makes sense, DynamicProperty is a bit of a misleading name, it is a property that is based on some level correct?

@here/harp-mapview/lib/poi/PoiManager.ts
nzjony: When you mean the following code relies on it, you mean that the getPropertyValue method relies on the correct zoom level to give the correct value?

It just seems strange to have this check here, and not somewhere (for example in the VisibleTileSet), or in the render method on the MapView. Because the check only needs to be done one for each tile, and not for each POI in each tile, perhaps straight after the line:

```
```

Second question, this method is called when creating the poi (as far as I can tell), why does the MapView zoom level influence the creation? Just seems like a strange dependency to have.

ninok: You are right since we don't have a proper production build that removes asserts this could have some performance impact.
Regarding using the zoomlevel here: You are right the logic is broken. All properties should be evaluate every frame but with the current state of the label rendering everything would fall apart. Especially changing properties that drastically change the layout (like vAlignment and hAlignment) should not change between zoomlevels b/c it will flicker even more. Also the performance of our interpolants is really bad right now. @robertoraggi is refactoring them and once they are performant we can also refactor this code.

nzjony: Ok, at least it is known that this is broken and that it is in the pipeline to be fixed.

PR #925, opened 2019-10-18
MINOR: Enable tslint-override

ninok:


nzjony:
I tried locally and got:

ERROR: /Users/jonstich/dev/harp.gl/@here/harp-mapview/lib/MapViewPoints.ts:58:5 - Member with @override keyword does not override any base class member

PR #928, opened 2019-10-21
HARP-5644: Support for Political Views via FeatureModfier.

kkostecki:
When different political view is requested, special implementation of
FeatureModifier is initialized that converts border lines "kind"
attributes according to different point of view, i.e. re-writing generic
"kind" with a contents of "kind:xx", where "xx" denotes country code.
This results in assigning new kinds'v values:
- unrecognized_disputed,
- unrecognized_country,
to generic "kind". Of course this kinds require separate/unique styling
policy, because current themes does not define any styling for it.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -549,6 +549,18 @@ export interface MapViewOptions extends TextElementsRendererOptions, Partial<Loo
     languages?: string[];
 
     /**
+     * Sets the data sources to use specific country point of view (political view).
+     *
+     * This option may result in rendering different country borders then commonly accepted for
+     * some regions and it mainly regards to so called __disputed borders__. Although not all
+     * data sources or themes may support it.
+     *
+     * @note Country code should be coded in ISO 3166-1 alpha-2 standard, if this option
+     * is `undefined` the majority point of view will be used.
+     */
+    politicalView?: string;
+
+    /**
      * Set fixed pixel ratio for rendering. Useful when rendering on high resolution displays with
      * low performance GPUs that may be fill-rate limited.
      * @default `window.devicePixelRatio`
@@ -900,6 +912,7 @@ export class MapView extends THREE.EventDispatcher {
     private m_thisFrameTilesChanged: boolean | undefined;
     private m_lastTileIds: string = "";
     private m_languages: string[] | undefined;
+    private m_politicalView: string | undefined;
     private m_copyrightInfo: CopyrightInfo[] = [];
     private m_animatedExtrusionHandler: AnimatedExtrusionHandler;
 
@@ -991,6 +1004,7 @@ export class MapView extends THREE.EventDispatcher {
         this.m_options.enableStatistics = this.m_options.enableStatistics === true;
 
         this.m_languages = this.m_options.languages;
+        this.m_politicalView = this.m_options.politicalView;
 
         if (
             !isProduction &&
@@ -1460,6 +1474,36 @@ export class MapView extends THREE.EventDispatcher {
         this.m_tileDataSources.forEach((dataSource: DataSource) => {
             dataSource.setLanguages(this.m_languages);
         });
+        this.update();
+    }
+
+    /**
+     * Get currently presented political point of view - the country code.
+     *
+     * @note Country code is stored in ISO 3166-1 alpha-2 standard.
+     * @return Country code or undefined if default
+     * (majorly accepted) point of view is used.
+     */
+    get politicalView(): string | undefined {
+        return this.m_politicalView;
+    }
+
+    /**
+     * Set the political view (country code) to be used when rendering disputed features (borders).
+     *
+     * @note Country code should be encoded in ISO 3166-1 alpha-2 standard.
+     * @param pov The code of the country which point of view should be presented,
+     * `undefined` if default (commonly accepted) political view should be used.
+     */
+    set politicalView(pov: string | undefined) {
+        if (this.m_politicalView === pov) {
+            return;
+        }
+        this.m_politicalView = pov;
+        this.m_tileDataSources.forEach((dataSource: DataSource) => {
+            dataSource.setPoliticalView(pov);
+        });
+        this.clearTileCache();
         this.update();
     }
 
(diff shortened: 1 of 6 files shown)


kkostecki:
Please note that it is first approach for the Political View problem and only in draft state - just for discussion. 
The second approach will be based on feature environment object (MapEnv) that will map some attributes into different ones - will be soon provided for comparison.

@here/harp-omv-datasource/lib/OmvDecoder.ts
```diff
@@ -49,6 +49,7 @@ import {
     OmvDecoderOptions,
     OmvFeatureFilterDescription
 } from "./OmvDecoderDefs";
+import { OmvPointOfViewFeatureModifier } from "./OmvPointOfViewFeatureModifier";
```
ninok: I cannot see this file. Did you forget to commit it?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -767,6 +767,7 @@ export class MapView extends THREE.EventDispatcher {
     private m_thisFrameTilesChanged: boolean | undefined;
     private m_lastTileIds: string = "";
     private m_languages: string[] | undefined;
+    private m_politicalView: string | null;
```
zbigg: `null` -> `undefined`

@here/harp-omv-datasource/lib/OmvDecoder.ts
```diff
@@ -695,11 +697,32 @@ export class OmvTileDecoder extends ThemedTileDecoder {
...
+            );
             return new OmvTomTomFeatureModifier(filterDescription);
+        } else if (
+            respectedPov !== undefined &&
+            respectedPov !== null &&
+            respectedPov.length === 2
+        ) {
+            return new OmvPointOfViewFeatureModifier(filterDescription, [respectedPov]);
         } else {
+            assert(
```
zbigg: This validation belongs to constructor of `OmvPointOfViewFeatureModifier`

@here/harp-omv-datasource/lib/OmvDecoderDefs.ts
```diff
@@ -12,7 +12,11 @@ export enum FeatureModifierId {
     /**
      * Identifier to use the OmvTomTomFeatureModifier in the OmvDecoder.
      */
-    tomTom
+    tomTom,
```
zbigg: Not needed, we don't expose this modifier through ID, but by decoder options.

@here/harp-omv-datasource/lib/OmvDecoder.ts
kkostecki: Yes, my mistake... fixed now.

@here/harp-mapview/lib/MapView.ts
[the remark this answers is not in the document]
kkostecki: The problem with this is the functionality of ITileDecoder->configure() method, which does not override (re-write) options if they are undefined, thus setting default Political Point of View (which would be undefined) is not possible on decoder, because it will be simply ignored.
We may have two solutions on that:
- extend ITileDecoder->configure() which additional parameter for political view (the same as it's done with languages, but in my opinion breaks the interface, because political view is part of options - passed via OptionsMap).
- make a contract that empty string "" for point of view resets to default behavior.

@here/harp-omv-datasource/lib/OmvDecoder.ts
kkostecki: It depends on the solution chosen above, if we go with empty string for default point of view, I would not even create OmvPointOfViewFeatureModifier(), which gives overhead, but instead just use OmvGenericFeatureModifier.
Let's analyze it after choosing approach from above comments.

@here/harp-omv-datasource/lib/OmvPointOfViewFeatureModifier.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * C-tor.
+     *
+     * @param description
+     * @param respectedPov The codes (in ISO 3166-1 alpha-2 format) of countries which
+     * point of view should be taken firstly into account. The position on the lists
+     * relates to priority, so the first entries have precedence before the later.
+     */
+    constructor(description: OmvFeatureFilterDescription, respectedPov: string[]) {
```
ninok: Why is respectedPov an array?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1306,6 +1308,32 @@ export class MapView extends THREE.EventDispatcher {
...
+     * @param pov The code of country which point of view should be presented or
+     * null if default (commonly accepted) point of view should be set.
+     */
+    set politicalView(pov: string | null) {
+        console.log("MapView set politicalView: ", pov);
+        this.m_politicalView = pov;
+        this.m_tileDataSources.forEach((dataSource: DataSource) => {
+            dataSource.setPoliticalView(pov);
+        });
+        this.clearTileCache();
```
zbigg: Why?
It's DS responsibility to update tiles if they support POV at all.

@here/harp-omv-datasource/lib/OmvPointOfViewFeatureModifier.ts
kkostecki: Hm... firstly I was thinking about support for multiple POVs at once, but probably it is not needed... I will finally remove it... We just need to decide which implementation variant we will choose. I am preparing second PR which is based on MapEnv (feature environment) which will support localized (or political) views.

@here/harp-mapview/lib/MapView.ts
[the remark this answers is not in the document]
StefanDachwitz: we do not use "null" internally, we use "undefined" only. There must be a way to keep this principle.

@here/harp-omv-datasource/lib/OmvDataSource.ts
```diff
@@ -140,11 +140,21 @@ export interface OmvDataSourceParameters {
 
     /**
      * Identifier used to choose OmvFeatureModifier, if undefined [[OmvGenericFeatureModifier]] is
-     * used. This parameter gets applied to the decoder used in the [[OmvDataSource]] which might
+     * used only if [[politicalView]] is undefined, otherwise [[OmvPointOfViewFeatureModifier]] will
+     * be used.
+     * This parameter gets applied to the decoder used in the [[OmvDataSource]] which might
```
StefanDachwitz: The issue here is that implementing  POV as a feature modifier effectively removes the ability of a user to add his own feature modifier, or forces him to make it a subclass of the POV modifier. I do not think that this is a wise decision.

kkostecki:

PR #932, opened 2019-10-22
HARP-7609: Added opacity parameter to WebTileDataSource

ninok:


@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -410,10 +426,16 @@ export class WebTileDataSource extends DataSource {
...
+                    this.m_options.renderingOptions !== undefined
+                        ? this.m_options.renderingOptions.opacity
+                        : undefined;
                 const material = new THREE.MeshBasicMaterial({
                     map: texture,
                     depthTest: false,
-                    depthWrite: false
+                    depthWrite: false,
+                    opacity,
+                    transparent: opacity !== undefined ? true : false
```
kkostecki: What about opacity equal to 1.0, should it be transparent anyway? Maybe we could detect it too.

ninok: good point. I will add that check as well.

PR #936, opened 2019-10-23
HARP-7356: Fix textIsOptional property

ninok:
textIsOptional = false was not handled properly (i.e. the icon was rendered without text).
In addition the rendering was never finished (infinite animation loop)

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -1762,7 +1762,7 @@ export class TextElementsRenderer {
         let iconSpaceAvailable = true;
 
         // Check if icon should be rendered at this zoomLevel
-        const renderIcon =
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1980,6 +1980,7 @@ export class TextElementsRenderer {
 
             // If the text is not visible nor optional, we won't render the icon neither.
             else if (!renderIcon || !textIsOptional) {
+                renderIcon = false;
```
nzjony: why not keep renderIcon as const and add && !textIsOptional at the end of the condition?

nzjony: Resolving because this code will anyway be refactored later.

ninok:

PR #939, opened 2019-10-23
MINOR: Fix buildings outlines (edges) rendering.

kkostecki:
Depth offset was multiplied by homogeneous 'w' coordinate, which results
in constant offset in NDC space, although it seems to be correct,
because of depth buffer non-linear nature, this may promote some occluded
edges before their buildings geometry. This comes from fact that varying
offset (dependent on homogeneous w) applied in clip space provides huge
depth differences, thus resulting in z-fight (and win) of buildings
outlines rendered behind the actual geometry (in certeain frustum settings)

Second problem solved in this PR was promoting clipped geometry (outlines)
from outside the clip range -1 < z < 1 to inside of it, resulting in
rendering outlines of clipped buildings, this requires conditional
offseting - only if vertexes are not to be clipped in the next phase.

# CI Check

Happy contributing!


kkostecki:

kkostecki:

PR #940, opened 2019-10-23
OLPRPS-5209: Implemented watermark.

keaukraine:
OLPRPS-5209: Implemented watermark.


ninok:
@heremaps-bot ok to test

@here/harp-mapview/lib/Watermark.ts
```diff
@@ -0,0 +1,53 @@
...
+     * Default constructor.
+     *
+     * @param textureUrl URL to watermark image.
+     * @param m_size Watermark size.
+     * @param m_margin Distance from canvas corner to watermark logo.
+     */
+    constructor(textureUrl: string, private m_size = 32, private m_margin = 4) {
+        const geometry = new THREE.PlaneGeometry(this.m_size, this.m_size);
+        const map = new THREE.TextureLoader().load(textureUrl);
+        const material = new THREE.MeshBasicMaterial({
```
ninok: I think https://threejs.org/docs/index.html#api/en/objects/Sprite and 
https://threejs.org/docs/index.html#api/en/materials/SpriteMaterial would be mare appropriate for this usecase.

@here/harp-examples/config.ts
```diff
@@ -12,3 +12,7 @@ export const appCode = "9v2BkviRwi9Ot26kp2IysQ";
 
 /** @hidden */
 export const accessToken = "AGln99HORnqL1kfIQtsQl70";
+
```
ninok: Why not putting the svg file into the resources folder?

@here/harp-mapview/lib/Watermark.ts
```diff
@@ -0,0 +1,53 @@
...
+            transparent: true,
+            depthWrite: false,
+            depthTest: false,
+            map
+        });
+        this.m_watermarkMesh = new THREE.Mesh(geometry, material);
+        this.m_watermarkMesh.renderOrder = this.WATERMARK_RENDERORDER;
```
zbigg: I don't think this render order is needed at all, if there is only one object rendered in this batch.

@here/harp-mapview/lib/Watermark.ts
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Default constructor.
+     *
+     * @param textureUrl URL to watermark image.
+     * @param m_size Watermark size.
+     * @param m_margin Distance from canvas corner to watermark logo.
+     */
+    constructor(textureUrl: string, private m_size = 32, private m_margin = 4) {
+        const geometry = new THREE.PlaneGeometry(this.m_size, this.m_size);
+        const map = new THREE.TextureLoader().load(textureUrl);
```
zbigg: No handling of update after texture is actually loaded.
In case of remote images and static map this can cause that watermark will be never displayed.

OTOH, giving that watermark is kind-of business/legal measure it may be required that watermark is always displayed, so in this case we should wait with first render until watermark is really loaded.

Should it be even possible to  display map without watermark if it is configured ?

@here/harp-mapview/lib/Watermark.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as THREE from "three";
+
+/**
+ * Creates and handles rendering of watermark logo geometry for [[MapView]].
+ */
+export class Watermark {
```
zbigg: The way it's rendererd - just texture in corner (no opacity) and no real watermarking of whole area of image area doesn't fit common definition of watermark in digital imaging.

Consider more precise documentation what it really does and possible change name completly.

Maybe

@here/harp-mapview/lib/MapView.ts
```diff
@@ -602,6 +603,11 @@ export interface MapViewOptions {
      * @default false.
      */
     synchronousRendering?: boolean;
+
+    /**
+     * URL to watermark logo to be displayed in bottom left corner.
+     */
+    watermark?: string;
```
zbigg: Pass it through.
```
WatermarkOptions {
    imageUrl: ..
    width: ... // defaults to texture image width
    height: ... // defaults to texture image width
    margin ...
}
```
And make `Watermark` object accept more complex options.

Or just allow passing ready `Watermark` object.

@here/harp-examples/src/getting-started_orbiting-view.ts
```diff
@@ -60,7 +60,8 @@ export namespace CameraOrbitExample {
         const mapView = new MapView({
             canvas,
             projection: sphereProjection,
-            theme: "resources/berlin_tilezen_base_globe.json"
+            theme: "resources/berlin_tilezen_base_globe.json",
+            watermark: URL.createObjectURL(new Blob([watermarkLogo], { type: "image/svg+xml" }))
```
zbigg: This `URL.createObjectURL(new Blob(...))` really begs to be moved to common

zbigg:
From title it would suggest that this has something with I thought that it's real watermarking (https://en.wikipedia.org/wiki/Digital_watermarking)  which indeed have to be integrated into rendering of whole canvas.

After brief look, this is just image in corner, so hardly a watermark.

Can't we just add DOM overlay over canvas ?

@here/harp-examples/src/getting-started_orbiting-view.ts
keaukraine: Unfortunately it is impossible to use `Blob` in `config.ts` because this file is imported in `scripts/credentials.ts` which is run in Node.JS environment and has no `Blob`. Have to apply some hacks for this.

@here/harp-mapview/lib/Watermark.ts
nzjony: If you have repo version checked out, search for: terrain_raycast-and-getheight.html

I render a textured sprite there.

keaukraine:
After discussion with @ninok and @zbigg decided to use a simpler HTML+CSS overlay approach.

PR #942, opened 2019-10-23
WIP HARP-7302 Adding headless integration test for MapView

nzjony:


test/performance/MapViewIntegrationTest.ts
```diff
@@ -0,0 +1,131 @@
...
+    afterEach(() => {
+        sandbox.restore();
+        if (inNodeContext) {
+            delete global.window;
+            delete global.requestAnimationFrame;
+            delete global.cancelAnimationFrame;
+            delete global.navigator;
+            delete global.performance;
+        }
+    });
```
zbigg: add `mapView.dispose`, in `afterEach` so MapView instance releases all resources even if failed.

nzjony: the MapView instance is created in the it(...), so it should be released anyway once done is called? 

Or should i create it in the beforeEach ?

nzjony:
Please don't review, this is a POC, but not likely to be ever submitted. The dependencies are unlikely to be allowed (namely jsdom), at least according to my short assessment.

PR #944, opened 2019-10-24
MINOR: Extract methods for camera position and rotation in MapViewUtils

ninok:


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1877,28 +1877,23 @@ export class MapView extends THREE.EventDispatcher {
         const limitedTilt = Math.min(89, tiltDeg);
         // MapViewUtils#setRotation uses pitch, not tilt, which is different in sphere projection.
         // But in sphere, in the tangent space of the target of the camera, pitch = tilt. So, put
-        // the camera on the target, so the tilt can be passed to setRotation as a pitch.
-        this.geoCenter = target;
-        MapViewUtils.setRotation(this, -azimuthDeg, limitedTilt);
-        // Then deduce the actual geoCenter.
-        this.geoCenter = MapViewUtils.getCameraCoordinatesFromTargetCoordinates(
+        // the camera on the target, so the tilt can be passed to getRotation as a pitch.
```
Astrak: The 3 lines of comments are now wrong, they can be removed.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -239,11 +258,34 @@ export namespace MapViewUtils {
...
+
+    export function getCameraCoordinatesFromTargetCoordinates(
+        targetCoordinates: GeoCoordinates,
+        distance: number,
+        yawDeg: number,
+        pitchDeg: number,
+        mapView: MapView
+    ): GeoCoordinates {
+        return mapView.projection.unprojectPoint(
+            getCameraPositionFromTargetCoordinates(
```
Astrak: :v: 

Do we actually need this method now?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -381,34 +423,33 @@ export namespace MapViewUtils {
...
      */
-    export function setRotation(mapView: MapView, yawDeg: number, pitchDeg: number) {
-        const transform = {
-            xAxis: space.x,
-            yAxis: space.y,
-            zAxis: space.z,
-            position: cache.vector3[0]
-        };
-        mapView.projection.localTangentSpace(mapView.geoCenter, transform);
+    export function getCameraRotation(
```
Astrak: Isn't it a more general `getRotation` or `getRotationAtCoordinates` now? It doesn't deal with any object like the camera anymore. Also in this mindset, the argument `geoCenter` could be renamed `coordinates`.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -417,11 +458,32 @@ export namespace MapViewUtils {
...
+     * space. In particular, `MapView#geoCenter` needs to be set before calling `setRotation`.
+     *
+     * **Note:** `yaw == 0 && pitch == 0` will north up the map and you will look downwards onto the
+     * map.
+     *
+     * @param mapView Instance of MapView.
+     * @param yawDeg Yaw in degrees, counter-clockwise (as opposed to azimuth), starting north.
+     * @param pitchDeg Pitch in degrees.
+     */
+    export function setRotation(mapView: MapView, yawDeg: number, pitchDeg: number) {
```
Astrak: :v: 

Do we actually need this method now?

@here/harp-mapview/lib/Utils.ts
ninok: Yeah true, so I will rename `geoCenter` to `target` and the method `getCameraRotationAtTarget`.

@here/harp-mapview/lib/Utils.ts
ninok: I left it b/c it is public API. I will mark it as deprecated though

@here/harp-mapview/lib/MapView.ts
Astrak: The 3 lines can be removed. It was to justify was the geoCenter was set to the target to then call `setRotation`, but this is not done anymore.

ninok: Yes, and now they are there to justify why the target is actually a parameter to this method. Maybe I reword it a bit more.

PR #946, opened 2019-10-24
HARP-7639: Update ThreeJS to r109.

Astrak:
update to r109

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7723,11 +7723,6 @@ textextensions@^2.4.0:
   resolved "https://registry.yarnpkg.com/textextensions/-/textextensions-2.5.0.tgz#e21d3831dafa37513dd80666dff541414e314293"
   integrity sha512-1IkVr355eHcomgK7fgj1Xsokturx6L5S2JRT5WcRdA6v5shk9sxWuO/w/VbpQexwkXJMQIa/j1dBi3oo7+HhcA==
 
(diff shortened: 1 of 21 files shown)


yarn.lock
```diff
@@ -7723,11 +7723,6 @@ textextensions@^2.4.0:
   resolved "https://registry.yarnpkg.com/textextensions/-/textextensions-2.5.0.tgz#e21d3831dafa37513dd80666dff541414e314293"
   integrity sha512-1IkVr355eHcomgK7fgj1Xsokturx6L5S2JRT5WcRdA6v5shk9sxWuO/w/VbpQexwkXJMQIa/j1dBi3oo7+HhcA==
 
-three@^0.108.0:
```
ninok: Why is there only a removal? 108 should be replaced by 109

Astrak: I have to call `yarn` for this to be replaced properly, but then the new value is not using the same SHA algorithm, and it is referencing artifactory.in.here.com which is rejected by the test, so I didn't know how to bypass that another way. Since users have to install packages anyway, what is the use of maintaining `yarn.lock`-s files here and at the root of mapsdk?

ninok: This is b/c of your .nprc in mapsdk. Either you temporarily remove this file or you do your change in a standalone harp.gl checkout.

PR #948, opened 2019-10-25
HARP-7648: Improve performance of TextElementsRenderer.checkForSmallL…

atomicsulfate:
…abels.

- Add projectOnScreen method to screen projector which skips calculating the
screen coordinates if the projected point is outside of the screen.

- Reuse a temp vector to store the calculated screen points.
- Remove redundant check within the screen points loop that was checking
whether there was any visible screen point.

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -186,6 +186,7 @@ const tempBox2D = new Math2D.Box();
 
 const tempPosition = new THREE.Vector3();
 const tempScreenPosition = new THREE.Vector2();
+const tempScreenPoints: THREE.Vector2[] = [];
 const tempPoiScreenPosition = new THREE.Vector2();
 
 class TileTextElements {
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -2294,45 +2292,25 @@ export class TextElementsRenderer {
...
-        const screenPoints: THREE.Vector2[] = [];
-        let minX = Number.MAX_SAFE_INTEGER;
-        let maxX = Number.MIN_SAFE_INTEGER;
-        let minY = Number.MAX_SAFE_INTEGER;
-        let maxY = Number.MIN_SAFE_INTEGER;
+        screenPoints.length = 0;
+        let anyPointVisible = false;
+
         for (const pt of textElement.path!) {
             tempPosition.copy(pt).add(textElement.tileCenter!);
```
nzjony: I wonder if this can be avoided somehow, why can't the textElement path positions be pre-transformed to world, this is done on the cpu anyway, so we don't lose any precision.

atomicsulfate: agreed, there's many places in text rendering where we need to transform to world coordinates, so probably we're doing duplicated work. I'll add it to a the labeling performance ticket i created.

atomicsulfate:

PR #952, opened 2019-10-25
MINOR: Initial view is not Berlin.

Astrak:
If `lookAt`, `setCameraGeolocationAndZoom` or `geoCenter` are not specified in the client code, MapView currently sets the camera above Berlin. This PR makes the initial view a wide world view from afar that fits both planar and globe instead.


Astrak:

PR #956, opened 2019-10-25
DRAFT: Reimplement interpolations as dynamic expressions (needs to be splitted)

robertoraggi:


@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
```diff
@@ -754,7 +760,7 @@ export class StyleSetEvaluator {
             try {
                 const evaluatedValue = attrExpr.evaluate(
                     env,
-                    ExprScope.Value,
+                    ExprScope.Dynamic,
```
robertoraggi: The above call to `evaluate` should use the static scope `ExprScope.Value` instead of `ExprScope.Dynamic`.
Using static scopes is important because "statically evaluated expressions" preserve nested interpolations.

The follow two WIP changes are required by this PR:
 - support for partial evaluation of calls contains interpolations (or in general dynamic properties) in their sub expressions
 - change the renderer to consistently use `getPropertyValue`; alternatively change the way DecodedTile containing expressions are processed by the TileGeometryCreator.

PR #961, opened 2019-10-28
Change signatures of the style operator descriptors.

robertoraggi:
Prefer to pass the CallExpr instead of just the list of actual
arguments. This change will allow the operators to use properties
stored together with the CallExpr node. Also, it makes it possible
for evaluators to return the CallExpr node when partially evaluating
expressions.

### DIFF
--- a/@here/harp-datasource-protocol/lib/Expr.ts
+++ b/@here/harp-datasource-protocol/lib/Expr.ts
@@ -90,10 +90,10 @@ class ComputeExprDependencies implements ExprVisitor<void, ExprDependencies> {
     }
 
     visitCallExpr(expr: CallExpr, context: ExprDependencies): void {
-        if (expr.op === "zoom" && expr.children.length === 0) {
+        if (expr.op === "zoom" && expr.args.length === 0) {
             context.zoom = true;
         } else {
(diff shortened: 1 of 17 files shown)


@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -696,10 +696,14 @@ export class ContainsExpr extends Expr {
 export class CallExpr extends Expr {
     descriptor?: OperatorDescriptor;
 
-    constructor(readonly op: string, readonly children: Expr[]) {
+    constructor(readonly op: string, readonly args: Expr[]) {
         super();
     }
 
+    get children() {
```
zbigg: In some cases, we reference `expr.children` in others `expr.args`.
Is there difference?, can't we settle on one one property name ?

robertoraggi: `children` is the generic name that we use for "AST nodes" containing a list of child nodes. In this case I introduced `args` to make sure that the patch was smaller and easier to read (e.g. instead of `args` the code now uses `call.args`); at some time I added a gettter `children` to keep the code backward compatible.

PR #966, opened 2019-10-28
HARP-5644: Support for Political Views via features environment map.

kkostecki:
When different political view is requested, special implementation of
Feature Env (environment) is initialized that allows to access localized
attributes - attributes with posix in from of ":xx", where "xx" denotes
country code. The special (localized) environment exposes such local
attributes instead of general ones, for example:
- border lines with "kind" attributes are accessed according to different
point of view (by country), i.e. replacing generic "kind" with a contents
of "kind:xx" with a given point of view.
- features "name" atributes may be translated according to localized name
with "name:xx".

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-datasource-protocol/lib/Env.ts
+++ b/@here/harp-datasource-protocol/lib/Env.ts
@@ -20,6 +20,7 @@ export interface ValueMap {
  * @hidden
  */
 export class Env {
+    constructor(readonly entries: ValueMap) {}
     /**
      * Returns property in [[Env]] by name.
      *
@@ -28,6 +29,7 @@ export class Env {
     lookup(_name: string): Value | undefined {
         return undefined;
     }
+
     /**
      * Return an object containing all properties of this environment. (Here: empty object).
      */
@@ -41,8 +43,9 @@ export class Env {
  */
 export class MapEnv extends Env {
     constructor(readonly entries: ValueMap, private readonly parent?: Env) {
-        super();
+        super(entries);
     }
+
     /**
      * Returns property in [[Env]] by name.
      *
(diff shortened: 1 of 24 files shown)


kkostecki:
DO NOT REVIEW yet, it is just concept for solutioning to compare with:
https://github.com/heremaps/harp.gl/pull/928

@here/harp-omv-datasource/lib/OmvDataSource.ts
```diff
@@ -140,11 +140,21 @@ export interface OmvDataSourceParameters {
 
     /**
      * Identifier used to choose OmvFeatureModifier, if undefined [[OmvGenericFeatureModifier]] is
-     * used. This parameter gets applied to the decoder used in the [[OmvDataSource]] which might
+     * used only if [[politicalView]] is undefined, otherwise [[OmvPointOfViewFeatureModifier]] will
```
StefanDachwitz: OmvPointOfViewFeatureModifier is still used???

@here/harp-datasource-protocol/lib/Env.ts
```diff
@@ -71,3 +74,92 @@ export class MapEnv extends Env {
...
+     */
+    lookup(name: string): Value | undefined {
+        const country: string | undefined = this.m_locale.get(name);
+        if (country !== undefined) {
+            const localProp = super.lookup(`${name}:${country}`);
+            if (localProp !== undefined) {
+                return localProp;
+            }
+        }
+        return super.lookup(name);
```
StefanDachwitz: I am a bit worried about the performance of this function, since it is used heavily. Please measure performance and compare.

@here/harp-omv-datasource/lib/OmvDecoder.ts
```diff
@@ -167,14 +169,29 @@ export class OmvDecoder implements IGeometryProcessor {
...
+        const localEnvProcessor = new OmvLocalEnvironmentProcessor();
+        // TODO: Later on it is possible to localize features names using the Env interface.
+        //if (m_languages !== undefined && m_languages!.length > 0) {
+        //    localEnvProcessor.setLocale("name", m_languages[0]);
+        //}
+        // Setup localization (point of view) for kind attribute
+        if (politicalView !== undefined && politicalView.length > 0) {
+            localEnvProcessor.setLocale("kind", politicalView);
+        }
+        this.m_envProcessor = localEnvProcessor;
```
StefanDachwitz: Please verify that the performance is not degrading (measure, please). In case there is a significant change, please make sure that
1. the localeEnvProcessor is only used for lines (Ask Stephan Kosmella)
2. the localEnvProcessor is only used if a politicalView is set, such that the performance is not degrading in the default case. In the default case the MapEnv may be used.

PR #973, opened 2019-10-30
Avoid fetching Georgian Extended glyphs.

amsXYZ:
Currently, our font in ``harp-fontcatalog`` doesn't support the [Georgian Extended Unicode Block](https://en.wikipedia.org/wiki/Georgian_Extended), and missing glyphs are represented in some uppercased labels in Georgia.

After investigating the issue a bit, I've noticed that Georgian Extended is a quite [new and unsupported block](https://github.com/harfbuzz/harfbuzz/issues/1104), and some shapers will avoid to uppercase these glyphs. This PR does just that, but since it's a quite hardcoded solution, I'm not sure if we should merge it into master (ideally we should just a font to our ``harp-fontcatalog`` that supports Georgian Extended)...

### DIFF
--- a/@here/harp-text-canvas/lib/rendering/FontCatalog.ts
+++ b/@here/harp-text-canvas/lib/rendering/FontCatalog.ts
@@ -369,7 +369,9 @@ export class FontCatalog {
         );
         const glyphPromises: Array<Promise<GlyphData>> = [];
         for (const char of charset) {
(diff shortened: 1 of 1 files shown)


amsXYZ:
Decided that indeed this is not the best way to handle this, so closing this PR for now.

PR #988, opened 2019-10-31
MINOR: Fix the building animation going through ground.

Astrak:
Fixes the building animation going through the ground.

### DIFF
--- a/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
+++ b/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
@@ -12,7 +12,7 @@ uniform float extrusionRatio;
 varying vec4 vExtrusionAxis;
 `,
     extrusion_vertex: `
-transformed = transformed - extrusionAxis.xyz + extrusionAxis.xyz * extrusionRatio;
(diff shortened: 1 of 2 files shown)


@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -12,7 +12,10 @@ uniform float extrusionRatio;
 varying vec4 vExtrusionAxis;
 `,
     extrusion_vertex: `
-transformed = transformed - extrusionAxis.xyz + extrusionAxis.xyz * extrusionRatio;
+// animate only if the point is above the ground
+if(dot(position,vec3(extrusionAxis))>0.0){
```
ninok: I still don't get why we have to do this. Either this is a data issue then we should report it to map compilation but most likely it is a bug in our geometry generation code or we should do some clamping there instead of doing it in the shader

Astrak: You're right, I will rework it to fix the geometry instead, as this `extrusionAxis` is telling points on the ground to go below it.

PR #989, opened 2019-11-01
Fix free camera example.

Astrak:
The default was the new large view, too far away for the default far plane in this example. Switched to a close up of Geneva.

### DIFF
--- a/@here/harp-examples/src/getting-started_free-camera.ts
+++ b/@here/harp-examples/src/getting-started_free-camera.ts
@@ -69,6 +69,9 @@ export namespace FreeCameraAppDebuggingToolExample {
         constructor(readonly options: FreeCameraAppOptions) {
             this.mapView = new MapView(options);
             this.mapView.fog.enabled = false;
+            // Set the view over Geneva.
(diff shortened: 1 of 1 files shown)


@here/harp-examples/src/getting-started_free-camera.ts
```diff
@@ -280,7 +280,8 @@ Press 'V' to change the scene point of view<br>`;
         document.body.appendChild(message);
 
         const canvas = document.getElementById("mapCanvas") as HTMLCanvasElement;
-        const geoCenter = new GeoCoordinates(52.518611, 13.376111);
+        // Set the view over Geneva.
+        const geoCenter = new GeoCoordinates(46.207, 6.147, 2000);
```
ninok: How is this fixing the issue? How is it related to the location? Is it b/c of the altitude?
Maybe you can also switch this example to use the look-at API

nzjony: instead of changing the example, can the far plane be fixed? It otherwise makes this example hard to maintain . (because if someone changes this position in future, or uses it as a base for another example, they may not be aware of the restriction of the position).

PR #1003, opened 2019-11-05
HARP-7578: Evaluate interpolators only if zoom/pixelration has changed.

zbigg:
Temporary solution to increase performance of interpolators before fully featured dynamic expressions (#956) land in master.

### DIFF
--- a/@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
+++ b/@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
@@ -90,6 +90,37 @@ export function getPropertyValue<T>(
     level: number,
     pixelToMeters: number = 1.0
 ): any {
+    if (isInterpolatedProperty(property)) {
+        if (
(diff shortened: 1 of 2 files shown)


@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
```diff
@@ -89,6 +89,37 @@ export function getPropertyValue<T>(
...
+            level === property._lastZoomLevel &&
+            pixelToMeters === property._lastPixel2Meters
+        ) {
+            return property._lastValue;
+        }
+        const actualValue = getPropertyValueInt(property, level, pixelToMeters);
+        property._lastValue = actualValue;
+        property._lastZoomLevel = level;
+        property._lastPixel2Meters = pixelToMeters;
+        return actualValue;
```
robertoraggi: I don't think this is the right place for doing memoization.

PR #1007, opened 2019-11-06
Add support for evaluating expressions to 'getPropertyValue'.

robertoraggi:
This change adds an overload to getPropertyValue that accepts
enviroments. Also, it introduces a new evaluation scope
called ExprScope.Dynamic that can be used by the Expr operators
to provide different semantics depending on how the expression is used
(e.g. an expression used during decoding vs an expression
used at rending time can potentially have different values).

### DIFF
--- a/@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
+++ b/@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
@@ -7,8 +7,9 @@
 import { Color, CubicInterpolant, DiscreteInterpolant, LinearInterpolant } from "three";
 
 import { LoggerManager } from "@here/harp-utils";
+import { Env, MapEnv } from "./Env";
 import { ExponentialInterpolant } from "./ExponentialInterpolant";
-import { Expr, Value } from "./Expr";
+import { Expr, ExprScope, Value } from "./Expr";
 import {
     InterpolatedProperty,
     InterpolatedPropertyDefinition,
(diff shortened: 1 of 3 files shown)


@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
```diff
@@ -85,14 +111,45 @@ export function isInterpolatedProperty(p: any): p is InterpolatedProperty {
...
+    }
+
+    let level: number;
+
+    if (typeof envOrLevel === "number") {
+        level = envOrLevel;
+    } else {
+        level = envOrLevel.lookup("$zoom") as any;
+        pixelToMeters = envOrLevel.lookup("$pixelToMeters") as any;
+    }
```
robertoraggi: `$zoom` and and `$pixelToRatio` are guaranteed to be in the `Env` and have "number" value, so there's no need for additional checks here.

@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
```diff
@@ -85,14 +111,45 @@ export function isInterpolatedProperty(p: any): p is InterpolatedProperty {
...
+export function getPropertyValue(
+    property: Value | Expr | InterpolatedProperty | undefined,
+    envOrLevel: number | Env,
     pixelToMeters: number = 1.0
 ): any {
-    if (isInterpolatedPropertyDefinition<T>(property)) {
+    if (isInterpolatedPropertyDefinition(property)) {
         throw new Error("Cannot interpolate a InterpolatedPropertyDefinition.");
-    } else if (!isInterpolatedProperty(property)) {
+    }
```
robertoraggi: I don't see any reason for the above test. I think we should remove it.

@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
zbigg: > guaranteed
Maybe use typescript or at least assert for this guarantee ?

How about new `IntepolationEnv` (accepted by `getPropertyValue)` which has first class `zoom` and `pixelToMeters` members.

robertoraggi: what do you mean?

zbigg: nvm, please leav as it is as this is only temporary.

PR #1008, opened 2019-11-06
Add internal APIs to partially evaluate expressions.

robertoraggi:
A partial evaluation of an Expr is defined to return a new Expr
semantically equivalent to the original Expr but with a different
structure.

### DIFF
--- a/@here/harp-datasource-protocol/lib/Expr.ts
+++ b/@here/harp-datasource-protocol/lib/Expr.ts
@@ -502,6 +502,26 @@ export class VarExpr extends Expr {
 }
 
 export abstract class LiteralExpr extends Expr {
+    /**
+     * Create a [[LiteralExpr]] from the given value.
+     *
+     * @param value A constant value.
+     */
+    static fromValue(value: Value): Expr {
+        switch (typeof value) {
+            case "boolean":
+                return new BooleanLiteralExpr(value);
+            case "number":
+                return new NumberLiteralExpr(value);
+            case "string":
+                return new StringLiteralExpr(value);
+            case "object":
+                return value === null ? NullLiteralExpr.instance : new ObjectLiteralExpr(value);
+            default:
+                throw new Error(`failed to create a literal from '${value}'`);
+        } // switch
+    }
+
     abstract get value(): Value;
 }
 
--- a/@here/harp-datasource-protocol/lib/ExprEvaluator.ts
+++ b/@here/harp-datasource-protocol/lib/ExprEvaluator.ts
@@ -14,6 +14,7 @@ import {
     ExprScope,
     ExprVisitor,
     HasAttributeExpr,
+    LiteralExpr,
     MatchExpr,
     NullLiteralExpr,
     NumberLiteralExpr,
@@ -50,6 +51,8 @@ const operatorDescriptors = new Map<string, OperatorDescriptor>();
  * @hidden
  */
 export class ExprEvaluatorContext {
+    private m_partialEvaluation = false;
+
     constructor(
         readonly evaluator: ExprEvaluator,
         readonly env: Env,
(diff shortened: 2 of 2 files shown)


robertoraggi:
Please note that this change adds **only** the APIs needed by the ExprEvalutorContext/visitor to support partial evaluation. The actually partial evaluator will be introduced with follow up changes.

@here/harp-datasource-protocol/lib/ExprEvaluator.ts
```diff
@@ -50,19 +51,66 @@ const operatorDescriptors = new Map<string, OperatorDescriptor>();
  * @hidden
  */
 export class ExprEvaluatorContext {
+    private _partialEvaluation = false;
```
Astrak: It's minor but so far we've used `m_` everywhere for all private members.

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -502,6 +502,26 @@ export class VarExpr extends Expr {
...
+            case "number":
+                return new NumberLiteralExpr(value);
+            case "string":
+                return new StringLiteralExpr(value);
+            case "object":
+                return value === null ? NullLiteralExpr.instance : new ObjectLiteralExpr(value);
+            default:
+                throw new Error(`failed to create a literal from '${value}'`);
+        } // switch
+    }
```
Astrak: ?

@here/harp-datasource-protocol/lib/ExprEvaluator.ts
```diff
@@ -50,19 +51,66 @@ const operatorDescriptors = new Map<string, OperatorDescriptor>();
...
     }
+
+    /**
+     * Partially evaluate the given expression.
+     *
+     * @param expr The [[Expr]] to evaluate.
+     */
+    partiallyEvaluate(expr: Expr | undefined): Expr {
+        if (expr === undefined) {
+            throw new Error("Failed to evaluate expression");
```
Astrak: Why accept `undefined` if it should not be supported?

@here/harp-datasource-protocol/lib/ExprEvaluator.ts
```diff
@@ -50,19 +51,66 @@ const operatorDescriptors = new Map<string, OperatorDescriptor>();
...
+        }
+
+        const previousEvaluationMode = this._partialEvaluation;
+
+        this._partialEvaluation = true;
+
+        try {
+            const value = expr.accept(this.evaluator, this);
```
Astrak: As far as I understand, `this._partialEvaluation` is only `true` during that statement, and always come back to `false` otherwise. What is the use of it?

@here/harp-datasource-protocol/lib/ExprEvaluator.ts
robertoraggi: like for the above `evaluate`, possible `undefined` values may come for the recursive evaluation of expressions.

@here/harp-datasource-protocol/lib/ExprEvaluator.ts
robertoraggi: we use both, i will change it to m_, but in this module we usually prefer `_`

@here/harp-datasource-protocol/lib/Expr.ts
robertoraggi: when the code gets too nested and i'm using switch statements, i like to annotate where the 'switch' ends.

@here/harp-datasource-protocol/lib/ExprEvaluator.ts
robertoraggi: I renamed it, but in this module we usually prefer `_`.

PR #1023, opened 2019-11-11
HARP-7401: Method to retrive map.lookAt() options

pit-rpg:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -162,14 +162,11 @@ const FRAME_RATE_RING_SIZE = 12;
 const FALLBACK_FRAME_RATE = 30;
 
 /**
- * Zoom level to request terrain tiles for getting the height of the camera above terrain.
- */
-const TERRAIN_ZOOM_LEVEL = 4;
-
-/**
  * Style set used by [[PolarTileDataSource]] by default.
  */
 const DEFAULT_POLAR_STYLE_SET_NAME = "polar";
+
+const MAX_PITCH = 89;
 
 /**
  * The type of `RenderEvent`.
@@ -633,10 +630,10 @@ export class MapView extends THREE.EventDispatcher {
 
     private m_tileWrappingEnabled: boolean = true;
 
-    private m_zoomLevel: number = DEFAULT_MIN_ZOOM_LEVEL;
     private m_minZoomLevel: number = DEFAULT_MIN_ZOOM_LEVEL;
     private m_maxZoomLevel: number = DEFAULT_MAX_ZOOM_LEVEL;
     private m_minCameraHeight: number = DEFAULT_MIN_CAMERA_HEIGHT;
+    private m_lookAtDistance: number = 0.0;
 
     private readonly m_screenCamera = new THREE.OrthographicCamera(-1, 1, 1, -1);
 
@@ -651,7 +648,6 @@ export class MapView extends THREE.EventDispatcher {
     private readonly m_rteCamera = new THREE.PerspectiveCamera();
 
     private m_focalLength: number;
-    private m_lookAtDistance: number;
     private readonly m_viewRanges: ViewRanges = {
         near: DEFAULT_CAM_NEAR_PLANE,
         far: DEFAULT_CAM_FAR_PLANE,
@@ -831,7 +827,7 @@ export class MapView extends THREE.EventDispatcher {
             this.m_screenCollisions = new ScreenCollisionsDebug(this.m_collisionDebugCanvas);
         }
 
-        this.handleRequestAnimationFrame = this.renderFunc.bind(this);
+        this.handleRequestAnimationFrame = this.render.bind(this);
         this.handlePostponedAnimationFrame = this.postponedAnimationFrame.bind(this);
         this.m_pickHandler = new PickHandler(
             this,
@@ -893,7 +889,6 @@ export class MapView extends THREE.EventDispatcher {
             DEFAULT_CAM_FAR_PLANE
         );
         this.m_camera.up.set(0, 0, 1);
-        this.m_lookAtDistance = 0;
         this.m_focalLength = 0;
         this.m_scene.add(this.m_camera); // ensure the camera is added to the scene.
         this.m_screenProjector = new ScreenProjector(this.m_camera);
@@ -1426,7 +1421,7 @@ export class MapView extends THREE.EventDispatcher {
         const targetDistance = this.camera.position.distanceTo(target);
         const attitude = MapViewUtils.extractAttitude(this, this.camera);
         const pitchDeg = THREE.Math.radToDeg(attitude.pitch);
-        const azimuthDeg = -THREE.Math.radToDeg(attitude.yaw);
+        const headingDeg = -THREE.Math.radToDeg(attitude.yaw);
 
         this.m_visibleTileSetOptions.projection = projection;
         this.updatePolarDataSource();
@@ -1442,7 +1437,7 @@ export class MapView extends THREE.EventDispatcher {
             this.m_visibleTileSetOptions
         );
 
-        this.lookAt(targetCoordinates, targetDistance, pitchDeg, azimuthDeg);
+        this.lookAt(targetCoordinates, targetDistance, pitchDeg, headingDeg);
     }
 
     /**
@@ -1471,6 +1466,23 @@ export class MapView extends THREE.EventDispatcher {
      */
     get lookAtDistance(): number {
         return this.m_lookAtDistance;
+    }
+
+    /**
+     * Set distance from camera to the point of focus in world units.
+     */
+    set lookAtDistance(distance: number) {
+        const sphericalCoordinates = MapViewUtils.extractSphericalCoordinatesFromLocation(
+            this,
+            this.m_camera,
+            this.target
+        );
+        this.lookAt(
+            this.target,
+            distance,
+            THREE.Math.radToDeg(sphericalCoordinates.tilt),
+            THREE.Math.radToDeg(-sphericalCoordinates.heading)
+        );
     }
 
     /**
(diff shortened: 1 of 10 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2224,7 +2349,7 @@ export class MapView extends THREE.EventDispatcher {
      * in the [[MapViewOptions]]
      */
     renderSync() {
-        this.renderFunc(PerformanceTimer.now());
+        this.render(PerformanceTimer.now());
```
Astrak: @zbigg @ninok I never understood why we had `renderFunc` there and what it was. Does it serve any purpose?

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1598,7 +1598,8 @@ export class TextElementsRenderer {
         // rendered in screen space. This method is unaffected by near and far clipping planes
         // distances, but may be improved by taking FOV into equation or customizing the
         // focus point screen position based on horizont, actual ground, tilt ets.
-        let factor = this.m_mapView.lookAtDistance / distance;
```
Astrak: This should stay unchanged if `lookAtDistance` is always a number.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1558,6 +1577,85 @@ export class MapView extends THREE.EventDispatcher {
...
     }
 
+    /**
+     * Returns the camera focus point position in geo-coordinates. If the camera points to an
+     * invalid position will return `null`.
+     */
+    get geoTarget() {
+        const target = MapViewUtils.rayCastWorldCoordinates(this, 0, 0);
+        if (target === null) {
+            throw new Error("Incorrect view properties, the target is null.");
```
ninok: Why throwing an error here? This can easily happen if the user is modifying the three.js camera directly

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1558,6 +1577,85 @@ export class MapView extends THREE.EventDispatcher {
...
+        );
+        return THREE.Math.radToDeg(sphericalCoordinates.azimuth + Math.PI);
+    }
+
+    /**
+     * Set `azimuth` property of the [[MapView]] in degrees. Please note [[MapView.geoTarget]] must
+     * be not `null`
+     */
+    set azimuth(azimuthDeg: number) {
+        this.lookAt(this.geoTarget, this.lookAtDistance, this.tilt, azimuthDeg);
```
ninok: I think we should log an error if geoTarget is null

@here/harp-mapview/lib/MapView.ts
Astrak: I asked for it :) it's because we have that throughout the code already, and MapView doesn't work if the target is in the void (the method to retrieve the tilt and azimuth doesn't know the target, so lookAt doesn't work)

pit-rpg:
I personally prefer the approach where an error is thrown.
This keeps the code cleaner and also forces the user to use the API created for this, or forces him to do more careful the camera transformations.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2011,7 +2102,7 @@ export class MapView extends THREE.EventDispatcher {
             // lookAtDistance = (EQUATORIAL_CIRCUMFERENCE * focalLength) / (256 * zoomLevel^2);
             // lookAtDistance = abs(cameraPos.z) / cos(cameraPitch);
             // Here we may use precalculated distance (once pre frame):
-            const lookAtDistance = this.m_lookAtDistance;
+            const lookAtDistance = this.lookAtDistance;
```
ninok: Please adjust the comment. It is not precalculated anymore

@here/harp-mapview/lib/MapView.ts
ninok: looks fine to me to remove it.

@here/harp-mapview/lib/MapView.ts
ninok: that's not acceptable. We should fix the places where MapView relies on that (I only know that the TiltClipPlaneEvaluator does) and don't add more.
It's totally fine if the map controls have this restriction but MapView should not have it.

ninok:
> I personally prefer the approach where an error is thrown.
> This keeps the code cleaner and also forces the user to use the API created for this, or forces him to do more careful the camera transformations.

There is no way for the user to know if an error will be thrown in this method (we could add it to the documentation but that's easy to miss). Having a GeoCoordinates | undefined return value will make it clear to the user that there is not always a target point.

Also forcing the user to add a try catch is bad b/c it has a performance impact. My rule of thumb is, exceptions should only be used for sth. unexpected. Not having a target point is a totally valid case if the user modified the camera.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1511,7 +1529,97 @@ export class MapView extends THREE.EventDispatcher {
...
 
-        this.update();
+        this.updateCamera();
+    }
+
+    /**
+     * Returns the camera focus point position in geo-coordinates. If the camera points to an
+     * invalid position will return `null`.
+     */
+    get geoTarget() {
```
ninok: I would rather just name this `target`. Did you check the confluence page where we discussed the naming?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1511,7 +1529,97 @@ export class MapView extends THREE.EventDispatcher {
...
+     * Set `pitch` value of the current view in degrees.
+     */
+    set pitch(pitchDeg: number) {
+        this.setCameraGeolocationAndZoom(this.geoCenter, this.zoomLevel, this.yaw, pitchDeg);
+    }
+
+    /**
+     * Return `yaw` value of the current view in degrees.
+     */
+    get yaw() {
```
ninok: why not `azimuth`?

ninok: Ah, just saw it below. I would not add yaw and pitch getters/setters at all but we should stick to azimuth/tilt(or however we want to name them)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1609,24 +1717,24 @@ export class MapView extends THREE.EventDispatcher {
...
+        const zoomLevelDistance = cameraPosZ / Math.cos(Math.min(cameraPitch, Math.PI / 3));
+        return MapViewUtils.calculateZoomLevelFromDistance(zoomLevelDistance, this);
     }
+
     set zoomLevel(zoomLevel: number) {
-        this.m_zoomLevel = THREE.Math.clamp(zoomLevel, this.m_minZoomLevel, this.m_maxZoomLevel);
-        MapViewUtils.zoomOnTargetPosition(this, 0, 0, this.m_zoomLevel);
-        this.update();
+        THREE.Math.clamp(zoomLevel, this.m_minZoomLevel, this.m_maxZoomLevel);
+        this.lookAtDistance = MapViewUtils.calculateDistanceToGroundFromZoomLevel(this, zoomLevel);
```
ninok: that's not the inverse of zoomlevel, especially when the pitch is > 60°.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3294,4 +3332,49 @@ export class MapView extends THREE.EventDispatcher {
...
+            );
+        }
+
+        if (target === undefined) {
+            target = MapViewUtils.rayCastWorldCoordinates(this, 0, 0) || undefined;
+        }
+
+        if (target === undefined) {
+            target = new THREE.Vector3(0.0, 0.0, 1.0);
+            target.applyMatrix4(this.camera.matrixWorld).multiplyScalar(this.m_lookAtDistance);
```
ninok: Isn't that camera.getWorldDirection?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3294,4 +3332,49 @@ export class MapView extends THREE.EventDispatcher {
...
+            const rect = this.canvas.getBoundingClientRect();
+
+            target = this.elevationProvider.rayCast(
+                rect.left + rect.width / 2,
+                rect.top + rect.height / 2
+            );
+        }
+
+        if (target === undefined) {
+            target = MapViewUtils.rayCastWorldCoordinates(this, 0, 0) || undefined;
```
ninok: Maybe we should never do the raycasting but always use the formula below. Otherwise it could happen that distance between target and camera is bigger than m_lookAtDistance (especially if tilt > 60°)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1610,24 +1682,23 @@ export class MapView extends THREE.EventDispatcher {
...
      * Returns the storage level for the given camera setup.
      * Actual storage level of the rendered data also depends on [[DataSource.storageLevelOffset]].
      */
     get storageLevel(): number {
-        return THREE.Math.clamp(
-            Math.floor(this.m_zoomLevel),
-            this.m_minZoomLevel,
-            this.m_maxZoomLevel
-        );
+        return this.zoomLevel;
```
ninok: Why did you remove the floor? I think nobody is using this method but we should not break it's beheviour.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2200,6 +2261,23 @@ export class MapView extends THREE.EventDispatcher {
         this.drawFrame();
     }
 
+    /**
+     * Need to be called if the [[MapView.camera]] position, rotation, quaternion properties are
+     * changed manually.
+     */
+    updateCamera(lookAtDistance?: number) {
+        if (lookAtDistance === undefined) {
+            this.m_lookAtDistance = this.getTarget()
```
ninok: Hmm, this is a bit fishy. getTarget will return a vector with length m_lookAtDistance so m_lookAtDistance will effectively never change if the user is moving the camera and no target point is hit.
I think you need to compute the lookAtDistance from the distance to ground (like we do in the zoomlevel getter)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1610,24 +1682,23 @@ export class MapView extends THREE.EventDispatcher {
      * Returns the zoom level for the given camera setup.
      */
     get zoomLevel(): number {
-        return this.m_zoomLevel;
+        const cameraPitch = MapViewUtils.extractAttitude(this, this.m_camera).pitch;
+        const zoomLevelDistance =
+            this.m_lookAtDistance / Math.cos(Math.min(cameraPitch, Math.PI / 3));
```
ninok: I think that is wrong. m_lookAtDistance should already be distance to ground divided by cosine of the pitch

PR #1025, opened 2019-11-11
 HARP-6510 Running rendering tests locally in docker container. 

arekspiewak1991:
HARP-6510 This change allows to run rendering tests locally in docker container.

- add script `rendering-tests-reference-platform.sh` for running tests with specified reference platform (`Chrome Headless with software renderer / Linux`) 
- adds possibility to inject module as express middleware to existing express app instance in: `with-http-server.ts`
- uses image from https://github.com/SeleniumHQ/docker-selenium


arekspiewak1991:

arekspiewak1991:

arekspiewak1991:

@here/harp-test-utils/lib/rendering/ReferenceImageLocator.ts
```diff
@@ -22,18 +22,18 @@ export function getReferenceImageUrl(imageProps: TestImageProps) {
...
 export function defaultReferenceImageResolver(imageProps: TestImageProps) {
     if (imageProps.name && imageProps.module) {
-        const baseUrl = "../ibct-results/";
-
         const platform = getOverride("IBCT_REFERENCE_PLATFORM", imageProps.platform || "");
-        const platformPart = platform === "" ? "" : `${platform}/`;
-        const modulePart = imageProps.module ? imageProps.module + "-" : "";
         const extra = imageProps.extra || "";
-        return `${baseUrl}${platformPart}${modulePart}${imageProps.name}${extra}.png`;
+
```
zbigg: const queryString = Object.keys(imageProps).map(key => `key=${urlencode(imageProps[key])}`).join("&")

test/README.md
```diff
@@ -0,0 +1,65 @@
...
+### Prerequisites
+-   Linux
+-   Docker installed
+
+### Tests execution - build and run locally
+
+1. Run command:
+
+```shell
+    bash scripts/docker-tests-runner.sh
```
zbigg: It should be something like "yarn rendering-tests".
Scripts should be hidden.

test/README.md
```diff
@@ -0,0 +1,65 @@
+# Performance Tests
+
+# Rendering Tests
+
+Tests for rendering components, which compares image result and stored reference image.
```
zbigg: Reword.
Tests for components *that* render images. They render images to canvas and compare the result image with approved reference images.

test/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```shell
+    bash scripts/docker-tests-runner.sh
+    #OR
+    ./scripts/docker-tests-runner.sh
+```
+ 
+2. Local reference images
+
+Now, the problem is that your browser platform string doesn't look like the one that CI uses
```
zbigg: This text doesn't apply as we don't have CI.
You should describe local flow only, where first run just skips all tests because there are no reference images.

@here/harp-test-utils/lib/rendering/RenderingTestResultServer.ts
```diff
@@ -158,21 +165,56 @@ export function installMiddleware(app: express.Router, basePath: string) {
     });
 
     outputBasePath = basePath;
+    logger.log("installMiddleware output basepath:", outputBasePath);
+
     const jsonParser = bodyParser.json({ limit: 1024 * 1024 * 16 });
     app.get("/ibct-report", jsonParser, getIbctReport);
     app.post("/ibct-feedback", jsonParser, postIbctFeedback);
+    app.get("/reference-image", getReferenceImage);
```
zbigg: These changes should be in separate commits, but ok.
At least document all the featutures added in commit, because it's not only docker image.

scripts/docker-tests-runner.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
zbigg: Name is misleading.
It should be `rendering-tests-reference-platform`.

scripts/docker-tests-runner.sh
```diff
@@ -0,0 +1,11 @@
...
+
+yarn build-tests
+
+npx ts-node -- \
```
zbigg: Need to create another script that will run these tests against tests served by `webpack-dev-server`.

Also `webpack.tests.config` are missing `RenderingTestResultServer` middleware.

test/README.md
```diff
@@ -0,0 +1,86 @@
+# Rendering Tests
+
+Tests for components that render images. They render images to canvas and compare the result image
+with approved reference images.
+Tests are run in the browser headles mode, uses `mocha-webdriver-runner` to run `mocha` tests
```
musculman: Should be headless (not headles)

test/README.md
```diff
@@ -0,0 +1,86 @@
...
+in browser context.
+
+## What to do first? Where to start?
+
+Before start the work on implementing new features you should run rendering tests first to
+generate actual images and store them as reference images to have possibility of making comparision
```
musculman: of making comparison

test/README.md
```diff
@@ -0,0 +1,86 @@
...
+with right images after doing some changes in code.
+
+### Prerequisites
+-   Linux
```
musculman: Does it really work only on Linux? Is there a reason why the docker image does not work on windows/Mac?

package.json
```diff
@@ -48,6 +49,9 @@
         "start-tests": "webpack-dev-server -d --config webpack.tests.config.js",
         "test-browser": "ts-node -- ./scripts/with-http-server.ts -C dist/test -p 8079 -- mocha-webdriver-runner http://localhost:8079/index.html",
         "build-tests": "webpack -d --config webpack.tests.config.js",
+        "rendering-tests": "./scripts/rendering-tests-reference-platform.sh",
```
musculman: Shouldn't this be run-rendering-tests?

test/README.md
zbigg: There is no fundamental problem with running these on Windows/MacOS but it's not easy.

The man obstacle, is that we rely on `--network host` and it's [not available outside linux](https://docs.docker.com/network/host/).

Mac/Windows workflow would require lot of wiring, in order to open connection from  browser to test server running on localhost.

I think we need followup PR for this.

Some interesting links
 * https://stackoverflow.com/questions/40746453/how-to-connect-to-docker-host-from-container-on-windows-10-docker-for-windows
 * there is `host.docker.internal` -> https://docs.docker.com/docker-for-mac/networking & https://docs.docker.com/docker-for-windows/networking/

So seems, that with new dockers it's not so complex, but i would still propose another PR for this

musculman: Agreed with your proposal @zbigg , thanks for the explanation.

@here/harp-test-utils/lib/rendering/RenderingTestResultServer.ts
arekspiewak1991: please verity (changed commit message and added one line of documentation... not sure if should I write something more there)

test/README.md
arekspiewak1991: also - thanks Zbyszek answering that question...
I will try to add those changes with new PR.

scripts/docker-tests-runner.sh
arekspiewak1991: I think done.

test/README.md
arekspiewak1991: fixed

package.json
arekspiewak1991: ok, I've changed that

package.json
```diff
@@ -48,6 +49,9 @@
         "start-tests": "webpack-dev-server -d --config webpack.tests.config.js",
         "test-browser": "ts-node -- ./scripts/with-http-server.ts -C dist/test -p 8079 -- mocha-webdriver-runner http://localhost:8079/index.html",
         "build-tests": "webpack -d --config webpack.tests.config.js",
+        "run-rendering-tests": "./scripts/rendering-tests-reference-platform.sh",
+        "save-reference": "npx ts-node @here/harp-test-utils/lib/rendering/RenderingTestResultCli.ts save-reference",
```
musculman: I think the naming is not consistent as just by looking at the name you won't notice they are in fact used in combination.
What about?
* run-rendering-tests
* save-reference-rendering-tests
* approve-reference-rendering-tests

test/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Interactive mode
+
+1. Run command:
+
+```shell
+yarn start-tests
+```
+
+Starts `webpack-dev-server` which compiles your rendering tests on fly
```
musculman: on the fly

test/README.md
musculman: I have tried testing this PR in linux wsl and mac and failed, on both systems with no clear error message. As expected I imagine, but I have not yet verified it with plain linux.

arekspiewak1991: @zbigg Could you please verify that again and test on your side?

zbigg: @musculman #1106 will adapt scripts and instructions, so all runs on mac + Docker fo Mac.

arekspiewak1991:

arekspiewak1991:

PR #1028, opened 2019-11-12
HARP-7828: Update the npm dependencies

robertoraggi:


robertoraggi:
there's too much stuff going on here for one single commit:
 - typescript 3.7 is a bit stricter and it breaks the way we use sinon
 - new versions of prettier format the files a bit differently

I'm closing this PR and propose smaller changes.

PR #1030, opened 2019-11-12
HARP-7808 Clip polygons

nzjony:
Change-Id: I90ed4a795f01a02308697874d044bf5485b83d77

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -858,6 +858,11 @@ export class MapView extends THREE.EventDispatcher {
         });
         this.m_renderer.autoClear = false;
 
+        // This is a bit of a misnomer, this is required for the clipping planes
+        // to be enabled, and they are in world space, see:
+        // https://threejs.org/docs/#api/en/materials/Material.clippingPlanes
+        this.m_renderer.localClippingEnabled = true;
+
         // This is detailed at https://threejs.org/docs/#api/renderers/WebGLRenderer.info
         // When using several WebGLRenderer#render calls per frame, it is the only way to get
         // correct rendering data from ThreeJS.
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -718,30 +718,33 @@ export class TileGeometryCreator {
                             if (_material.clippingPlanes === null) {
                                 _material.clippingPlanes = this.clippingPlanes;
                                 // TODO: Add clipping for Spherical projection.
-                            } else if (mapView.projection.type === ProjectionType.Planar) {
+                            }
+                            if (mapView.projection.type === ProjectionType.Planar) {
+                                // This prevents 1 pixel errors in the IBCT
+                                const expandFactor = 1.001;
                                 const planes = _material.clippingPlanes;
                                 const rightConstant =
                                     tile.center.x -
                                     mapView.worldCenter.x +
-                                    tile.boundingBox.extents.x;
+                                    tile.boundingBox.extents.x * expandFactor;
                                 planes[0].constant = rightConstant;
 
                                 const leftConstant =
                                     tile.center.x -
                                     mapView.worldCenter.x -
-                                    tile.boundingBox.extents.x;
+                                    tile.boundingBox.extents.x * expandFactor;
                                 planes[1].constant = -leftConstant;
 
                                 const topConstant =
                                     tile.center.y -
                                     mapView.worldCenter.y +
-                                    tile.boundingBox.extents.y;
+                                    tile.boundingBox.extents.y * expandFactor;
                                 planes[2].constant = topConstant;
 
                                 const bottomConstant =
                                     tile.center.y -
                                     mapView.worldCenter.y -
-                                    tile.boundingBox.extents.y;
+                                    tile.boundingBox.extents.y * expandFactor;
                                 planes[3].constant = -bottomConstant;
                             }
                         }
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -48,7 +48,7 @@ import {
     SolidLineMaterial
 } from "@here/harp-materials";
 import { ContextualArabicConverter } from "@here/harp-text-canvas";
-import { LoggerManager } from "@here/harp-utils";
+import { chainCallbacks, LoggerManager } from "@here/harp-utils";
 import * as THREE from "three";
 
 import { AnimatedExtrusionTileHandler } from "../AnimatedExtrusionHandler";
@@ -105,6 +105,13 @@ export interface PolygonFadingParameters extends FadingParameters {
  */
 export class TileGeometryCreator {
     private static m_instance: TileGeometryCreator;
+    // These are used for clipping the geometry so that the overlapping geometry is removed.
+    private clippingPlanes = [
+        new THREE.Plane(new THREE.Vector3(-1, 0, 0), 0),
+        new THREE.Plane(new THREE.Vector3(1, 0, 0), 0),
+        new THREE.Plane(new THREE.Vector3(0, -1, 0), 0),
+        new THREE.Plane(new THREE.Vector3(0, 1, 0), 0)
+    ];
 
     /**
      * The `instance` of the `TileGeometryCreator`.
@@ -702,6 +709,43 @@ export class TileGeometryCreator {
 
                 if (srcGeometry.uuid !== undefined) {
                     object.userData.geometryId = srcGeometry.uuid;
+                }
+
+                if (isFillTechnique(technique)) {
+                    object.onBeforeRender = chainCallbacks(
+                        object.onBeforeRender,
+                        (_renderer, _scene, _camera, _geometry, _material) => {
+                            if (_material.clippingPlanes === null) {
+                                _material.clippingPlanes = this.clippingPlanes;
+                                // TODO: Add clipping for Spherical projection.
+                            } else if (mapView.projection.type === ProjectionType.Planar) {
+                                const planes = _material.clippingPlanes;
+                                const rightConstant =
+                                    tile.center.x -
+                                    mapView.worldCenter.x +
+                                    tile.boundingBox.extents.x;
+                                planes[0].constant = rightConstant;
+
+                                const leftConstant =
+                                    tile.center.x -
+                                    mapView.worldCenter.x -
+                                    tile.boundingBox.extents.x;
+                                planes[1].constant = -leftConstant;
+
+                                const topConstant =
+                                    tile.center.y -
+                                    mapView.worldCenter.y +
+                                    tile.boundingBox.extents.y;
+                                planes[2].constant = topConstant;
+
+                                const bottomConstant =
+                                    tile.center.y -
+                                    mapView.worldCenter.y -
+                                    tile.boundingBox.extents.y;
+                                planes[3].constant = -bottomConstant;
+                            }
+                        }
+                    );
                 }
 
                 if (


@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -901,7 +907,12 @@ export class VisibleTileSet {
                     tile.tileKey,
                     tile.offset
                 );
-                if (tile.hasGeometry || defaultSearchDirection === SearchDirection.NONE) {
+                tile.isFallback = false;
+                if (
+                    (tile.hasGeometry || defaultSearchDirection === SearchDirection.NONE) &&
+                    (!this.excludeTiles.has(tile.tileKey.mortonCode()) ||
```
nzjony: this change is just a debug change, it is to test the fallback, basically, if you put a breakpoint here, and add a morton code to the excludeTiles, then you can test the fallback manually.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -910,6 +911,7 @@ export class MapView extends THREE.EventDispatcher {
                     : this.m_options.powerPreference
         });
         this.m_renderer.autoClear = false;
+        this.m_renderer.localClippingEnabled = true;
```
nzjony: Required to enable clipping.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -126,6 +126,13 @@ export class TileGeometryCreator {
     private static m_colorMap: Map<string, THREE.Color> = new Map();
 
     private static m_instance: TileGeometryCreator;
+    // These are used for clipping the geometry so that the overlapping geometry is removed.
+    private clippingPlanes = [
```
ninok: These will not work in globe

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -231,6 +238,11 @@ export class TileGeometryCreator {
         this.createTextElements(tile, decodedTile, textFilter);
 
         this.createLabelRejectionElements(tile, decodedTile);
+
+        // The ground plane is required for when we change the zoom back and we fall back to the
+        // parent, in that case we reduce the renderOrder of the parent tile and this ground place
+        // ensures that parent doesn't come through.
+        TileGeometryCreator.instance.addGroundPlane(tile);
```
ninok: There should be a flag in the datasource options weather the plane is added or not. Especially for rendering streets on top of satellite you don't  want to have it.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1512,7 +1548,7 @@ export class TileGeometryCreator {
         const material = new MapMeshBasicMaterial({
             color: colorHex,
             visible: isVisible,
-            depthWrite: false
+            depthWrite: true
```
ninok: Why is that?

ninok:
Please split this PR in two, one for the clip planes and one for the fallback render order

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: see the comment below `// TODO: Add clipping for Spherical projection.`

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: Good point!

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: opps, maybe this is a left over, I will check

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: > Please split this PR in two, one for the clip planes and one for the fallback render order

The problem with splitting it is that if you have the render order reduced, the fallback has weird seams at the end, because the other tiles spill over. IMO this is a regression and shouldn't be split.

nzjony:
> Please split this PR in two, one for the clip planes and one for the fallback render order

The problem with splitting it is that if you have the render order reduced, the fallback has weird seams at the end, because the other tiles spill over. IMO this is a regression and shouldn't be split.

nzjony:
> Please also add an IBCT test for polygons with clip planes

Can this be in a separate PR? Otherwise it will take me longer to implement (I haven't written an IBCT test before) and I will need to communicate with Stephan that we have to delay the release.

nzjony:
> > Please also add an IBCT test for polygons with clip planes
> 
> Can this be in a separate PR? Otherwise it will take me longer to implement (I haven't written an IBCT test before) and I will need to communicate with Stephan that we have to delay the release.

And Anyway, the IBCT is to prevent regressions, sure it makes sense to implement, but I doubt there will be a regression between now and when we do the release early next week.

nzjony:
> > Please also add an IBCT test for polygons with clip planes
> 
> Can this be in a separate PR? Otherwise it will take me longer to implement (I haven't written an IBCT test before) and I will need to communicate with Stephan that we have to delay the release.

IBCT test is on the HARP-7808_clip_ibct branch (gerrit + github).

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -1038,6 +1038,11 @@ export interface FillTechniqueParams extends BaseTechniqueParams, PolygonalTechn
      * Width of the lines. Currently limited to the [0, 1] range.
      */
     lineWidth?: DynamicProperty<number>;
+
+    /**
+     * Whether to clip the polygon at the Tile boundary.
+     */
```
robertoraggi: Please mention what is the default value for clipping

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -724,6 +731,32 @@ export class TileGeometryCreator {
                     object.userData.geometryId = srcGeometry.uuid;
                 }
 
+                if (isFillTechnique(technique) && (technique.clipping || technique.transparent)) {
+                    object.onBeforeRender = (_renderer, _scene, _camera, _geometry, _material) => {
+                        if (_material.clippingPlanes === null) {
```
robertoraggi: i believe we already have some setting `object.onBeforeRender`, this code here may override the callback

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -724,6 +731,32 @@ export class TileGeometryCreator {
                     object.userData.geometryId = srcGeometry.uuid;
                 }
 
+                if (isFillTechnique(technique) && (technique.clipping || technique.transparent)) {
+                    object.onBeforeRender = (_renderer, _scene, _camera, _geometry, _material) => {
```
robertoraggi: this is not correct, it won't work when clipping/transparent is defined as an interpolation or a step function.
you should use `getPropertyValue(technique.clipping)` and you should re-evaluate getPropertyValue(technique.clipping/transparent) for each frame

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -724,6 +731,32 @@ export class TileGeometryCreator {
                     object.userData.geometryId = srcGeometry.uuid;
                 }
 
+                if (isFillTechnique(technique) && (technique.clipping || technique.transparent)) {
```
robertoraggi: i believe the test should be a bit different
`(clipping || (clipping === undefined && transparent === true)`

@here/harp-datasource-protocol/lib/Techniques.ts
```diff
@@ -331,7 +331,8 @@ const fillTechniqueDescriptor = mergeTechniqueDescriptor<FillTechnique>(
             color: AttrScope.TechniqueRendering,
             opacity: AttrScope.TechniqueRendering,
             transparent: AttrScope.TechniqueRendering,
-            lineWidth: AttrScope.TechniqueRendering
+            lineWidth: AttrScope.TechniqueRendering,
+            clipping: AttrScope.TechniqueRendering
```
zbigg: This should be `AttrScope.TechniqueGeometry` because _NOW_ there is no way to re-evaluate this property on each frame.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
nzjony: I guess the default is undefined?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: Good point @robertoraggi, and thanks for the chainCallbacks @zbigg .

@zbigg , what do you mean about this not be required?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: Great point, thanks.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
robertoraggi: what do you mean is `undefined`? you should write in the documentation something like
` Whether to clip the polygon at the Tile boundary. Default value is false` or true if it is true.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: isn't this the same? Because clipping === undefined is anyway false. Or do you mean I should avoid truthy?

@here/harp-datasource-protocol/lib/Techniques.ts
nzjony: I fixed it to use getPropertyValue... so this can stay the same if I understand correct?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
zbigg: @nzjony While reading code you've added i spotted that for planar projections, you always set same objects which are constant., so i assume in this case there is no need for `onBeforeRender`.
... at leas this is how i understand code you've written :).

@here/harp-datasource-protocol/lib/Techniques.ts
zbigg: Yep.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -731,30 +731,54 @@ export class TileGeometryCreator {
...
-                                tile.center.y - mapView.worldCenter.y + tile.boundingBox.extents.y;
-                            planes[2].constant = topConstant;
-
-                            const bottomConstant =
-                                tile.center.y - mapView.worldCenter.y - tile.boundingBox.extents.y;
-                            planes[3].constant = -bottomConstant;
+                if (isFillTechnique(technique)) {
+                    object.onBeforeRender = chainCallbacks(
+                        object.onBeforeRender,
+                        (_renderer, _scene, _camera, _geometry, _material) => {
```
robertoraggi: yes, now it is correct and also very expensive we do a function call for each filled polygon :)

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
zbigg: Ok, but now we have `onBeforeRender` in all `fill` object even if they have completely static and disabled clipping and/or transparency.

So this whole `onBeforeRender` installation should be guarded by variation of `if (isInterpolated(technique.clipping) || isInterpolated(technique.transparent)` ... 

Oh no, this should be even more complex
 - even non-interpolated transparency and clipping & sphere should have onBeforeRender
 - interpolated transparency or clipping should have onBeforeRender
 - else (static transparency and clipping and planar projection) statically set material attributes

That's at least how i see it from review perspective :)

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
zbigg: @nzjony NVM, sorry i've mistakenly took === as !==.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -724,6 +731,56 @@ export class TileGeometryCreator {
...
+                            const clip =
+                                clipping || (clipping === undefined && transparent === true);
+                            if (!clip) {
+                                return;
+                            }
+                            if (_material.clippingPlanes === null) {
+                                _material.clippingPlanes = this.clippingPlanes;
+                                // TODO: Add clipping for Spherical projection.
+                            } else if (mapView.projection.type === ProjectionType.Planar) {
+                                const planes = _material.clippingPlanes;
```
ninok: I wonder if we could do this only once per material in the tile and not for every object in the tile. Therefore we would need sth. like a Tile.onBeforeRender or Tile.updateTechniques callback. I remember this was planned anyways by @robertoraggi , not sure what's the state though.

robertoraggi: there's nothing in our code preventing our users to share a material with objects coming from different tiles, so unless we explicitly prevent sharing materials we need to have a per-object call for setting up the clipping planes.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -731,30 +731,54 @@ export class TileGeometryCreator {
...
+                                mapView.zoomLevel
+                            );
+                            const transparent = getPropertyValue(
+                                technique.transparent,
+                                mapView.zoomLevel
+                            );
+                            const clip =
+                                clipping || (clipping === undefined && transparent === true);
+                            if (!clip) {
+                                return;
```
robertoraggi: is it correct to simply `return` in this case? I mean, what happen when `clipping/transparent` is true and them it becomes false?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
robertoraggi: no, it's not the same. clipping can be `true`, `false` or `undefined`. if users explicitly sets `clipping` to `false` it doesn't matter if transparent is `true` or `false`, clipping should be disabled, on the other hand if the `clipping` is left `undefined` then we should clip if `transparent` is true.

nzjony: My logic failed me, thank you for explaining.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: @zbigg , I will change this logic, I discussed with @ninok and we agreed to instead apply clipping to descendants of children that are currently falling back.....

Wait, that won't work, even if I clip the descendants, the children off other tiles could overlap.

nzjony: I discussed with @ninok he suggested to just enable clipping if any tile is falling back, I think this is a good compromise for now, and later we can add some special neighbour traversal code if performance is an issue. Regarding the other guards, I will take a look @zbigg , thanks.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
ninok: What do you mean? The materials are created by us and internal to the tile objects. We don't expose them. Of course a user can dig trough the objects of the tile and get access to the materials but that is such an advanced usecase where the user is most likely aware that the material needs to be cloned.
Maybe we can discuss this tomorrow face 2 face but this was my understanding how we will also do the interpolations in the future (i.e. Tile.updateTechniques will evaluate all interpolations and update the materials accordingly)

robertoraggi: All it's required by our current DataSource/Tile APIs is that a tile `Tile` provides a list of threejs objects for each `tile.objects`. For example, the DebugTIleDataSource defines only one material `debugMaterial` https://github.com/heremaps/harp.gl/blob/master/%40here/harp-debug-datasource/lib/DebugTileDataSource.ts#L19 and it uses it for each tile https://github.com/heremaps/harp.gl/blob/master/%40here/harp-debug-datasource/lib/DebugTileDataSource.ts#L71

many examples and apps using harp.gl do the same.

ninok: @robertoraggi But that's a different usecase. These datasources don't use a Technique. This change is adding the clipping planes only to the materials that are created for the fill technique-

nzjony:
GPU tests please

@here/harp-datasource-protocol/lib/TechniqueParams.ts
nzjony: I removed this from the style, so setting to resolved.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: As discussed in standup, enabling clipping for fallback tiles would require a recompilation of the shader, so we just enable clipping for all tiles.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: I don't fully understand this conversation, but it sounds like it is referring to some refactoring we should do in the material, and their handling? And is orthogonal, though somewhat related to this patch?

nzjony:

nzjony:

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -704,6 +711,45 @@ export class TileGeometryCreator {
...
+                if (isFillTechnique(technique)) {
+                    object.onBeforeRender = chainCallbacks(
+                        object.onBeforeRender,
+                        (_renderer, _scene, _camera, _geometry, _material) => {
+                            if (_material.clippingPlanes === null) {
+                                _material.clippingPlanes = this.clippingPlanes;
+                                // TODO: Add clipping for Spherical projection.
+                            }
+                            if (mapView.projection.type === ProjectionType.Planar) {
+                                const expandFactor = 1.001;
```
robertoraggi: Why do you need to expand the bounding box?

nzjony: see: 1757b05049ccfe6ed659827f753da5b9901b43a4

nzjony:
Submitting after @MPursche  showed that worst case this introduces 0.6ms GPU time on a 4K display, so the performance impact is therefore deemed adequat.

PR #1036, opened 2019-11-13
HARP-7808 Add ibct test to show overlapping

nzjony:


nzjony:
Not needed after discussion with @zbigg

PR #1038, opened 2019-11-14
HARP-7808 Fallback has render order.

nzjony:
Note, this introduces a regression, which will be fixed by introducing
clipping planes (follow up PR)

Change-Id: I0d96b10242b133f8d8bdbf6a187e04d444b28886

### DIFF
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -37,8 +37,9 @@ export type TileObject = THREE.Object3D & {
      * This stores the THREE.Object3D renderOrder property, we need to back it up because we need to
      * reduce it if the tile is used as fall back. When it is used normally, the renderOrder needs
      * to be reset.
+     * @hidden
      */
-    backupRenderOrder?: number;
+    _backupRenderOrder?: number;
 };
 
 interface DisposableObject {
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -32,6 +32,13 @@ export type TileObject = THREE.Object3D & {
      * Distance of this object from the [[Tile]]'s center.
      */
     displacement?: THREE.Vector3;
+
+    /**
+     * This stores the THREE.Object3D renderOrder property, we need to back it up because we need to
+     * reduce it if the tile is used as fall back. When it is used normally, the renderOrder needs
+     * to be reset.
+     */
+    backupRenderOrder?: number;
 };
 
 interface DisposableObject {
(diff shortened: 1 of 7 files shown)


nzjony:
Screenshot 2019-11-14 at 10 04 48

This is the regression that will be introduced.

@here/harp-omv-datasource/lib/OmvDataSource.ts
```diff
@@ -212,6 +212,7 @@ export class OmvDataSource extends TileDataSource<OmvTile> {
         });
 
         this.cacheable = true;
+        this.addGroundPlane = true;
```
ninok: I think this should be configurable (i.e. in the OmvDataSourceOptions). Especially for hybrid usecase.

nzjony: Which hybrid use cases do you mean? Are you referring to some DataSource which extends OmvDataSource and includes some raster tiles?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -231,6 +231,14 @@ export class TileGeometryCreator {
         this.createTextElements(tile, decodedTile, textFilter);
 
         this.createLabelRejectionElements(tile, decodedTile);
+
+        // HARP-7899, disable ground plane for globe
+        if (tile.dataSource.addGroundPlane && tile.projection.type === ProjectionType.Planar) {
+            // The ground plane is required for when we change the zoom back and we fall back to the
+            // parent, in that case we reduce the renderOrder of the parent tile and this ground place
+            // ensures that parent doesn't come through.
+            TileGeometryCreator.instance.addGroundPlane(tile, -10000);
```
CraigBerlin: Can we avoid magic number?  Since it is related to fallback maybe use instead:
``
-MapView.FALLBACK_RENDER_ORDER_OFFSET / 2 
``

nzjony: Done, and added comment.

@here/harp-omv-datasource/lib/OmvDataSource.ts
nzjony: Done

@here/harp-mapview/lib/Tile.ts
```diff
@@ -32,6 +32,13 @@ export type TileObject = THREE.Object3D & {
      * Distance of this object from the [[Tile]]'s center.
      */
     displacement?: THREE.Vector3;
+
+    /**
+     * This stores the THREE.Object3D renderOrder property, we need to back it up because we need to
+     * reduce it if the tile is used as fall back. When it is used normally, the renderOrder needs
+     * to be reset.
+     */
+    backupRenderOrder?: number;
```
robertoraggi: If this is an internal property that users are not supposed to use and modify you should mark it with @hidden in comment and rename it to _backupRenderOrder

nzjony: resolved in latest commit: b9c2c3920d9b0d435e8ade3cb8b1ff7b722bfb88

PR #1041, opened 2019-11-14
MINOR: Fix three.js buffer conversion when index is null

ninok:


@here/harp-datasource-protocol/lib/ThreeBufferUtils.ts
```diff
@@ -77,7 +77,10 @@ export namespace ThreeBufferUtils {
             vertexAttribute.name = name;
             vertexAttributes.push(vertexAttribute);
         }
-        const index = fromThreeBufferAttribute(bufferGeometry.index);
+        const index =
+            bufferGeometry.index !== null
+                ? fromThreeBufferAttribute(bufferGeometry.index)
```
nzjony: I see this is exported, would it make sense to do this check internally, incase there are others using this function?

ninok: I tried to do this and also added some tests and I realized that even more special handling for the index buffer is needed.

PR #1045, opened 2019-11-18
HARP-7517: Partial support for HEX encoded RGBA colors.

kkostecki:
This patch-set provides support for defining theme colors in extended HEX
formats that contains alpha channel, such as
- #RGBA (#F008 - semi opaque red),
- #RRGGBBAA (#FF000088 - as above).
Although such colors may be defined in themes, there is no internal engine
support for alpha channel specified, which is silently ignored.

Note:
This also fixes RGBA colors parsing that were not passing via
getPropertyValue() call - static (not interpolated) color values.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
+++ b/@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
@@ -31,7 +31,7 @@ export interface StringEncodedNumeralFormat {
     // TODO: Add target/output array as parameter to minimize arrays creation.
     decoder: (encodedValue: string) => number[];
 }
-export const StringEncodedMeters: StringEncodedNumeralFormat = {
+const StringEncodedMeters: StringEncodedNumeralFormat = {
     type: StringEncodedNumeralType.Meters,
     size: 1,
     regExp: /((?=\.\d|\d)(?:\d+)?(?:\.?\d*))m/,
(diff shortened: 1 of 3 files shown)


kkostecki:
Please note that some tests fails because we are no longer passing the color to THREE.js directly, instead we are doing HSL approximation (all the time), which provides inaccuracies.
The idea is to remove HSL conversion entirely, and rebase this change on top of that - see HARP-7898.

@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
```diff
@@ -48,16 +48,38 @@ export const StringEncodedPixels: StringEncodedNumeralFormat = {
...
+            // #RRGGBB or #RRGGBBAA
+            tmpColor.setRGB(
+                parseInt(hex.charAt(0) + hex.charAt(1), 16) / 255,
+                parseInt(hex.charAt(2) + hex.charAt(3), 16) / 255,
+                parseInt(hex.charAt(4) + hex.charAt(5), 16) / 255
+            );
+        } else {
+            throw new Error(`unsupported hex color '${encodedValue}'`);
+        }
+        tmpColor.getHSL(tmpHSL);
```
robertoraggi: this test looks a bit too expensive, we do two regex tests. Please consider that we have 100s of interpolations that get evaluated for each frame and for each one the interpolations that is using literals as value we run the above tests.

kkostecki:

@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
kkostecki: This is comming with a new patchset - single evaluation will be performed. I will update comment here when it will be available.

kkostecki: Here is desired optimization, please review this PR to solve it (dependency).

pit-rpg:
looks good to me, but it would be nice for someone else to take a look also

PR #1048, opened 2019-11-19
HARP-7373:  Refactor TextElementsRenderer dependencies.

atomicsulfate:
Pass all required MapView state parameters as part of a ViewState
interface to try to decouple the text renderer from MapView.

# CI Check

Happy contributing!


@here/harp-mapview/lib/MapView.ts
```diff
@@ -3007,27 +3008,6 @@ export class MapView extends THREE.EventDispatcher {
         }
     }
 
-    /**
-     * Check if the `textElementsChanged` flag in any tile has been set to `true`. If any flag was
-     * `true`, this function returns `true`, and resets the flag in all tiles to `false`.
-     */
-    private checkIfTextElementsChanged() {
```
atomicsulfate: Moved to TextElementsRenderer

@here/harp-mapview/lib/text/ViewState.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export interface ViewState {
+    worldCenter: THREE.Vector3; // View's center world coordinates.
+    cameraIsMoving: boolean; // Whether view's camera is currently moving.
+    maxVisibilityDist: number; // Maximum far plane distance.
+    zoomLevel: number; // View's zoom level.
+    frameNumber: number; // Current frame number.
+    lookAtDistance: number; // Distance to the lookAt pooint.
+    isDynamic: boolean; // Whether a new frame for the view is already requested.
+    hiddenGeometryKinds: GeometryKindSet | undefined; // Kinds of geometries that are disabled.
```
ninok: minor: `hiddenGeometryKinds?`

@here/harp-mapview/lib/text/MapViewState.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import { GeometryKindSet } from "@here/harp-datasource-protocol";
+import { MapView } from "../MapView";
+import { ViewState } from "./ViewState";
+
+/**
+ * View state obtained from a MapView instance.
+ */
+export class MapViewState implements ViewState {
```
ninok: Wouldn't it be better to have a copy of the values instead of fetching them? I guess it's fine for JavaScript b/c we don't have multi threading and mapview cannot change while we are rendering text.

@here/harp-mapview/lib/text/ViewState.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * State parameters of a view that are required by the text renderer.
+ */
+export interface ViewState {
+    worldCenter: THREE.Vector3; // View's center world coordinates.
+    cameraIsMoving: boolean; // Whether view's camera is currently moving.
+    maxVisibilityDist: number; // Maximum far plane distance.
+    zoomLevel: number; // View's zoom level.
+    frameNumber: number; // Current frame number.
+    lookAtDistance: number; // Distance to the lookAt pooint.
```
nzjony:
```suggestion
    lookAtDistance: number; // Distance to the lookAt point.
```

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -928,12 +952,11 @@ export class TextElementsRenderer {
 
         this.m_cacheInvalidated = false;
 
-        const renderList = this.m_mapView.visibleTileSet.dataSourceTileList;
-
-        this.checkIfOverloaded();
+        this.checkIfOverloaded(dataSourceTileList);
 
+        const worldMaxX = projection.worldExtent(0, 0).max.x;
```
nzjony: I understand the need to move this out for performance reasons, however moving it has made it harder to track what this is used for, because you have decoupled the offset from this value. Please add a comment to say this is used in conjunction with the tile offset.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -307,19 +326,20 @@ export class TextElementsRenderer {
...
+     * @param time Current frame time.
      * @param tilesChanged Indicates whether there's been any change in the tiles rendered since the
      * last call to this method (last frame).
-     * @param time Current frame time.
-     * @param frameNumber Current frame number.
      */
     placeText(
-        tileTextElementsChanged: boolean,
-        tilesChanged: boolean,
+        dataSourceTileList: DataSourceTileList[],
```
nzjony: Did you consider to add this to the ViewState? If yes, is there a reason it is here?

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -307,19 +326,20 @@ export class TextElementsRenderer {
...
+        projection: Projection,
```
nzjony: ditto, could this be part of the ViewState?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2833,10 +2834,10 @@ export class MapView extends THREE.EventDispatcher {
         }
 
         this.m_textElementsRenderer.placeText(
-            this.checkIfTextElementsChanged(),
-            this.checkIfTilesChanged(),
+            this.m_visibleTiles.dataSourceTileList,
+            this.projection,
             time,
-            this.m_frameNumber
+            this.checkIfTilesChanged()
```
nzjony: Why not delay evaluating this until you need it, I see that the placeText can early exit before needing to compute this value.

@here/harp-mapview/lib/text/ViewState.ts
atomicsulfate: See TileGeometryManager.updateTileObjectVisibility(). Objects whose kind is in the hiddenGeometryKinds set are set as invisible. Same for their labels.

@here/harp-mapview/lib/text/ViewState.ts
atomicsulfate: done

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: done

@here/harp-mapview/lib/text/MapViewState.ts
atomicsulfate: So far I'm creating this object once and passing it to the text renderer in the constructor. So it's meant to be a dynamic object, it gives the state of the view at the moment of the call.

The only reason I see to copy anything there would be as perf improvement if some of the getters are called too often or take too long. So far I haven't see any issue there when profiling.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: I moved to ViewState all properties that are used throughout many functions in TextElementsRenderer. Since the m_viewState is a member of TextElementsRenderer, any function can use those properties without having to pass them as parameter.

For the tile list, only a couple of functions need to use it, so I think it's better to pass it explicitly only to the functions that need it.

atomicsulfate: It seems also natural that when you call placeText, you pass as parameter the tiles containing the text.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: Similar reason. I didn't want to give all methods in TextElementsRenderer access to the projection, it's only needed for one thing (to compute the world offset for the tiles).

@here/harp-mapview/lib/MapView.ts
atomicsulfate: done

@here/harp-mapview/lib/text/ViewState.ts
ninok: Sorry for the confusion but the `?` was not a question but an optional property. It is equivalent to `|undefined` but shorter

@here/harp-mapview/lib/text/TextElementsRenderer.ts
nzjony: ok, fair enough

@here/harp-mapview/lib/text/TextElementsRenderer.ts
nzjony: good point

@here/harp-mapview/lib/text/ViewState.ts
atomicsulfate: done

PR #1047, opened 2019-11-18
HARP-7898: Color interpolation switches to RGB space instead of HSL.

kkostecki:
Default conversion to HSL color space causes inaccuracy's even when
using static colors in themes. Current unit tests (and methods) works only
because of passing color definition (string encoded) directly into to
THREE.js library (without HSL conversion).
In order to support for RGBA color channels and provide common mechanism
for colors decoding, color values may not be converted between spaces
unless needed or requested by special interpolation methods (to be
implemented).

Signed-off-by: Krystian Kostecki <ext-krystian.kostecki@here.com>

Thank you for contributing to harp.gl!

Before requesting a pull request, please remember to check the following documents:
* [contribution guidelines](https://github.com/heremaps/harp.gl/blob/master/CONTRIBUTING.md)
* [coding style](https://github.com/heremaps/harp.gl/blob/master/CODINGSTYLE.md)

If you are adding new functionality we would highly appreciate if you can describe what is the capability you are adding and even better if you can add some examples. Please also remember to add tests for it.

# CI Check

Our bots will check whether your PR can be directly integrated into the mainline. We have some internal integration tests running on the background, our bots will inform you of the next steps and someone from our team will take a look and help if needed!

And please do not forget to sign-off your commit! You can read more about DCO [here](https://julien.ponge.org/blog/developer-certificate-of-origin-versus-contributor-license-agreements/). But, in short, you just need to use `git commit -s` or append `--signoff` when you are committing to the repo.

Happy contributing!

PR #1049, opened 2019-11-19
HARP-7898: Decode color in RGB and use HSL only for interpolation.

kkostecki:
Default conversion to HSL color space causes inaccuracy's even when
using static colors in themes. Current unit tests (and methods) works only
because of passing color definition (string encoded) directly into to
THREE.js library (without HSL conversion).
In order to support for RGBA color channels and provide common mechanism
for colors decoding, color values may not be converted between spaces
unless needed or requested by interpolation methods (HCL methods to be
implemented).

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
+++ b/@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
@@ -6,7 +6,6 @@
 import { Color } from "three";
 
 const tmpColor = new Color();
-const tmpHSL = { h: 0, s: 0, l: 0 };
 
 /**
  * Enumeration of supported string encoded numerals.
@@ -29,6 +28,7 @@ export interface StringEncodedNumeralFormat {
     size: number;
     regExp: RegExp;
     mask?: number;
+    // TODO: Add target/output array as parameter to minimize arrays creation.
     decoder: (encodedValue: string) => number[];
 }
 export const StringEncodedMeters: StringEncodedNumeralFormat = {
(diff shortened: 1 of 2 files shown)


@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
```diff
@@ -64,14 +64,12 @@ export const StringEncodedRGB: StringEncodedNumeralFormat = {
...
         const channels = StringEncodedRGB.regExp.exec(encodedValue)!;
-        tmpColor
-            .setRGB(
-                parseInt(channels[1], 10) / 255,
-                parseInt(channels[2], 10) / 255,
-                parseInt(channels[3], 10) / 255
-            )
-            .getHSL(tmpHSL);
-        return [tmpHSL.h, tmpHSL.s, tmpHSL.l];
+        tmpColor.setRGB(
```
robertoraggi: are all these regex tests executed twice? can we remove that ? these are quite expensive tests and we will run several 100s of them for each render call.

kkostecki: Could we agree to do this in separate PR/ticket? I have already another PR that is dependent on that and making that optimization here will conflict related stuff a lot.

kkostecki: BTW, there is another slight optimization that I would like to do, mentioned in the comment above... Pass the output array to decoder as parameter instead of creating temp objects all the time.

robertoraggi: sure.

kkostecki:

kkostecki:

amsXYZ:
Despite I like this approach (since it keeps visual coherency with the previous implementation), the precision issue for RGB colors will still be present for interpolated RGB values, while it will not be with the solution provided in #1047 . 

Because of this, I'd stick to #1047 and later implement a way of specifying the interpolation space as part of the interpolation arguments.

Thanks for providing alternate solution though! ;)

kkostecki:
Chosen different solution fully based on RGB colors for interpolation and static techniques parameters.
See #1047

PR #1050, opened 2019-11-20
Fix resolving of decoder script in harp.js

zbigg:
Ensure that `getBundleScriptUrl` is called during main script load and
that it loaded `harp(.min).js`.

Fixes regression introduced in #901.

Related-to: HARP-7933


zbigg:

PR #1058, opened 2019-11-21
HARP-7699: Fix animateExtrusion property handling

SergiiKulakovskyi:


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -902,8 +902,10 @@ export class TileGeometryCreator {
                     isExtrudedPolygonTechnique(technique) &&
                     animatedExtrusionHandler !== undefined
                 ) {
-                    let animateExtrusionValue = technique.animateExtrusion;
-
+                    let animateExtrusionValue = getPropertyValue(
+                        technique.animateExtrusion,
+                        mapView.zoomLevel
```
pit-rpg: `mapView.zoomLevel` is a getter, maybe would be better to cache this value outside of the for loops

PR #1063, opened 2019-11-21
HARP-7000: Yeoman generator is using old versions

CraigBerlin:


@here/generator-harp.gl/generators/app/index.js
```diff
@@ -27,16 +27,40 @@ module.exports = class extends Generator {
...
                 packagename: this.answers.packagename,
                 access_token: this.answers.access_token,
                 generator_version: version
             }
         );
     }
+
+    install() {
+        // npmInstall conflicts with running install-peerdeps, use spawnCommandSync instead
+        this.spawnCommandSync("npx", ["npm", "install", "--no-save", "install-peerdeps"]);
```
ninok: why do we need to call `npx npm install ...`. Is the npx necessary?

CraigBerlin: Nope, good point thanks

PR #1067, opened 2019-11-21
HARP-7904: Restrict string encoded literals

kkostecki:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
+++ b/@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
@@ -20,6 +20,7 @@ import {
     StringEncodedColorFormats,
     StringEncodedMetricFormats,
     StringEncodedNumeralFormat,
+    StringEncodedNumeralFormatMaxSize,
     StringEncodedNumeralFormats,
     StringEncodedNumeralType
 } from "./StringEncodedNumeral";
@@ -34,6 +35,7 @@ const interpolants = [
 ];
 
 const tmpColor = new THREE.Color();
+const tmpBuffer = new Array<number>(StringEncodedNumeralFormatMaxSize);
 
 /**
  * Checks if a property is interpolated.
@@ -333,9 +335,11 @@ function procesStringEnocodedNumeralInterpolatedProperty(
             : StringEncodedColorFormats;
 
     for (let valueIdx = 0; valueIdx < prop.values.length; ++valueIdx) {
+        let matched = false;
         for (const valueFormat of allowedValueFormats) {
             const value = prop.values[valueIdx];
-            if (!valueFormat.regExp.test(value)) {
+            matched = valueFormat.decoder(value, tmpBuffer);
+            if (!matched) {
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


kkostecki:
Soon I will add (update) unit tests that proves the concept.

kkostecki:
> Soon I will add (update) unit tests that proves the concept.

Unit test for StringEncodedLiterals are done and PR updated.

@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
```diff
@@ -153,6 +176,13 @@ export const StringEncodedNumeralFormats: StringEncodedNumeralFormat[] = [
     ...StringEncodedColorFormats
 ];
 
+export const StringEncodedNumeralFormatMaxSize = Math.max(
```
robertoraggi: Do you think you could find a way to remove these exports?

@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
```diff
@@ -24,105 +24,118 @@ export enum StringEncodedNumeralType {
...
-            tmpColor.setRGB(
-                parseInt(hex.charAt(0) + hex.charAt(1), 16) / 255,
-                parseInt(hex.charAt(2) + hex.charAt(3), 16) / 255,
-                parseInt(hex.charAt(4) + hex.charAt(5), 16) / 255
-            );
+            target[0] = parseInt(hex.charAt(0) + hex.charAt(1), 16) / 255;
+            target[1] = parseInt(hex.charAt(2) + hex.charAt(3), 16) / 255;
+            target[2] = parseInt(hex.charAt(4) + hex.charAt(5), 16) / 255;
         } else {
+            // Impossible for given reg-exp
```
robertoraggi: Why not returning false instead of throwing an exception?

@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
kkostecki: Ok I reduced them to single export, Color and Metric format sizes are not used outside this file scope.
Anyway I could remove them all, but then we may pass any array and expand it dynamically when decoder receives something smaller - it is of course possible in JS:
`
decoder: (encodedValue: string, target: number[]) => {
    if (target.length < this.size) {
       target.length = this.size'
    }
    ...
}
`
Or even simpler, just write the values for certain indexes and JS will grow array automatically, but I am not sure how does this mechanism may affect performance on all platforms.
`
decoder: (encodedValue: string, target: number[]) => {
   ...
    target[0] = r;
    target[1] = g;
    target[2] = b;
    ...
`
Summarizing I am not sure if this is worth effort of hiding one constant. Please decide which approach you would prefer to see.

@here/harp-datasource-protocol/lib/StringEncodedNumeral.ts
kkostecki: I changed this to assert, cause it should never happen given that we have proper RegExp, we may also remove it entirely and then rely merely on unit tests.

@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
```diff
@@ -344,12 +348,16 @@ function procesStringEnocodedNumeralInterpolatedProperty(
...
+            for (let i = 0; i < valueFormat.size; ++i) {
+                propValues[valueIdx * valueFormat.size + i] = tmpBuffer[i];
             }
             break;
         }
+        if (!matched) {
+            throw Error(
+                `Not all interpolation values match the same format: ${JSON.stringify(prop)}`
+            );
+        }
```
kkostecki: @robertoraggi I think this was missing entirely (even in the old flow/logic), because you could create interpolated property like this:
`
 "color": {
                        "interpolation": "Linear",
                        "zoomLevels": [0, 11, 12, 15, 16, 20],
                        "values": ["#f00", "2m", "#f00", "#f00", "#f00", "#f00"]
                    }
`

>

PR #1069, opened 2019-11-22
HARP-7105 Updated builds script to create minified code

MPursche:
Use webpack optimizations for production build.

# CI Check

Happy contributing!


package.json
```diff
@@ -43,8 +43,8 @@
         "cov-report-html": "nyc report --reporter=html",
         "start": "ts-node ./scripts/credentials.ts -- . && webpack-dev-server -d --config @here/harp-examples/webpack.config.js",
         "build": "npm run build-bundle && npm run build-examples",
-        "build-examples": "ts-node ./scripts/credentials.ts -- dist/examples && webpack -d --config @here/harp-examples/webpack.config.js",
-        "build-bundle": "cd @here/harp.gl && webpack",
+        "build-examples": "ts-node ./scripts/credentials.ts -- dist/examples && webpack -p --config @here/harp-examples/webpack.config.js",
+        "build-bundle": "cd @here/harp.gl && webpack && NODE_ENV=production webpack",
```
pit-rpg: `&& webpack && NODE_ENV=production webpack` I guess here webpack run twice once in dev mode and second in production.

MPursche: Yes, this creates minified bundles for production(*.min.js) and unminified bundles(*.js).

zbigg: @MPursche Look at `package.json` in `@here/harp.gl` - we already build two bundles in `prepare` step used by `npm pack`.
So, this  change `build-bundle` is not necessary.

@here/harp-examples/package.json
```diff
@@ -15,7 +15,7 @@
     "scripts": {
         "test": "exit 0",
         "prebuild": "node scripts/postinstall.js",
-        "build": "webpack",
+        "build": "webpack -p",
```
zbigg: This makes production build the default. Not sure we want this.
I think, that main improvement we want from this change is that bundles deployed in `harp.gl` are minified.
So, i guess we need adaptation of apropriate deploy scripts (in travis-ci.yml) to set `NODE_ENV=production` and use this variable in `webpack.config`s.

MPursche: I changed the package scripts back to building without the production flag. The production environment is now set in the Travis deploy script. I also fixed the Webpack config of the examples to copy the right `harp.gl` bundle depending on the configured environment.

@here/harp-examples/webpack.config.js
```diff
@@ -195,9 +198,23 @@ const assets = [
...
+    }
+].filter(asset => {
+    if (asset === undefined || asset === null) {
+        return false;
+    } else if (typeof asset === "string") {
+        return true;
+    } else if (typeof asset === "object") {
+        return asset.from;
+    }
+}); // ignore stuff that is not found
```
pit-rpg: maybe it would be better:
```
.filter(asset => asset !== undefined && asset !== null && asset.from !== undefined);
```

MPursche: I don't think this will work. The array has a mix of strings and { from: string, to: string } objects. Plain strings should also pass the test. So the from property is only required if it is not a string.

@here/harp-examples/webpack.config.js
```diff
@@ -195,9 +198,23 @@ const assets = [
...
+        from: resolveOptional(`@here/harp.gl/dist/harp-decoders${harpBundleSuffix}.js`),
+        to: "harp-decoders.js"
+    }
+].filter(asset => {
+    if (asset === undefined || asset === null) {
+        return false;
+    } else if (typeof asset === "string") {
+        return true;
+    } else if (typeof asset === "object") {
+        return asset.from;
```
kkostecki: Just a blind shot.... do not know the topic well, but maybe we should check asset.from and asset.to, both for undefined, but also if it is empty - if ever possible.

PR #1073, opened 2019-11-25
HARP-7808 Fixing issue with tile offset + some artifacts

nzjony:
- Increased the expand factor to 1% when zooming out, the factor was not enough to prevent some aliasing at the tile borders
- fixed the calculation of the right / left offsets to take into consideration the tile's offsets.

### DIFF
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -729,14 +729,14 @@ export class TileGeometryCreator {
                     object.onBeforeRender = chainCallbacks(
                         object.onBeforeRender,
                         (_renderer, _scene, _camera, _geometry, _material) => {
-                            const worldOffsetX =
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -729,24 +729,29 @@ export class TileGeometryCreator {
                     object.onBeforeRender = chainCallbacks(
                         object.onBeforeRender,
                         (_renderer, _scene, _camera, _geometry, _material) => {
+                            const worldOffsetX =
```
CraigBerlin: Can be moved inside Planar projection if statement

nzjony: Good idea, done.

nzjony: 2e71b584c8bcd1314bf4c9ade4b83fdf23505ff5

nzjony:

PR #1075, opened 2019-11-25
MINOR: Added second path for add object example to show cube behind b…

ninok:
…uildings

### DIFF
--- a/@here/harp-examples/src/threejs_add-object.ts
+++ b/@here/harp-examples/src/threejs_add-object.ts
@@ -40,14 +40,33 @@ export namespace ThreejsAddSimpleObject {
     // snippet:harp_gl_threejs_add_simple_object_0.ts
     const scale = 100;
     const geometry = new THREE.BoxGeometry(1 * scale, 1 * scale, 1 * scale);
-    const material = new THREE.MeshStandardMaterial({
-        color: 0x00ff00fe
+    const prePassMaterial = new THREE.MeshStandardMaterial({
+        color: "#ff00fe",
+        opacity: 0.3,
+        depthTest: false,
+        transparent: true
     });
-    function createPinkCube(): MapAnchor<THREE.Mesh> {
+    const material = new THREE.MeshStandardMaterial({
+        color: "#ff00fe",
+        opacity: 0.9,
+        transparent: true
+    });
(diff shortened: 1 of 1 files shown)


@here/harp-examples/src/threejs_add-object.ts
```diff
@@ -40,14 +40,34 @@ export namespace ThreejsAddSimpleObject {
...
+    const prePassMaterial = new THREE.MeshStandardMaterial({
+        color: 0x00ff00fe,
+        opacity: 0.5,
+        depthTest: false,
+        transparent: true
+    });
     const material = new THREE.MeshStandardMaterial({
-        color: 0x00ff00fe
+        color: 0x00ff00fe,
+        opacity: 1.9,
```
nzjony: 1.9 opacity?

@here/harp-examples/src/threejs_add-object.ts
```diff
@@ -40,14 +40,34 @@ export namespace ThreejsAddSimpleObject {
...
+        const cube = new THREE.Object3D();
+
+        const prePassMesh = new THREE.Mesh(geometry, prePassMaterial);
+        prePassMesh.renderOrder = Number.MAX_SAFE_INTEGER - 1;
+        cube.add(prePassMesh);
+
+        const mesh = prePassMesh.clone();
         mesh.renderOrder = Number.MAX_SAFE_INTEGER;
-        return mesh;
+        mesh.material = material;
```
nzjony: does clone not already do this?

@here/harp-examples/src/threejs_add-object.ts
ninok: ups, should be 0.9

@here/harp-examples/src/threejs_add-object.ts
ninok: no, geometry and material are reused when cloning a mesh.

nzjony: yes of course, you are changing the material....😅

nzjony: in saying that, you could avoid 2 lines and just use new THREE.MESH(geometry, material), would save a line

ninok: ah, true. Will do

PR #1076, opened 2019-11-25
Adapted before-render callback for ThreeJS 0.110.0

vega-jose:
Adapted the definition of the `onBeforeRender` callback for the
fill technique, in order to make it compliant with the latest definition
for the `Object3D` class in ThreeJS 0.110.0

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -728,7 +728,7 @@ export class TileGeometryCreator {
                 if (isFillTechnique(technique)) {
                     object.onBeforeRender = chainCallbacks(
                         object.onBeforeRender,
-                        (_renderer, _scene, _camera, _geometry, _material) => {
+                        (_renderer, _scene, _camera, _geometry, _material, _group) => {
                             if (_material.clippingPlanes === null) {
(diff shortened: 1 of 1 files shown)


nzjony:

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -728,7 +728,7 @@ export class TileGeometryCreator {
                 if (isFillTechnique(technique)) {
                     object.onBeforeRender = chainCallbacks(
                         object.onBeforeRender,
-                        (_renderer, _scene, _camera, _geometry, _material) => {
+                        (_renderer, _scene, _camera, _geometry, _material, _group) => {
```
zbigg: Why is that even needed?
We don't use this group parameter so, it can be safely removed
Typescript accepts callback that has less arguments than required.

vega-jose: At least it does fail for other types of "tsconfigs". We're building harp.gl from source for MapLab and that change throws a `TS2345: Argument of type [...] is not assignable to parameter of type [...]`, I assume, because of something either stricter on our side or permissive on harp.gl's. Seems like a small enough but helpful change to support more TS configurations

ninok:
Already fixed by other PR

PR #1080, opened 2019-11-26
HARP-7808 Fix globe polygons not rendering

nzjony:
Don't apply the clippingPlanes if the projection is not planar.

### DIFF
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -727,7 +727,10 @@ export class TileGeometryCreator {
                     object.userData.geometryId = srcGeometry.uuid;
                 }
 
-                if (isFillTechnique(technique)) {
+                if (
+                    isFillTechnique(technique) &&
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -729,15 +729,16 @@ export class TileGeometryCreator {
                     object.onBeforeRender = chainCallbacks(
```
ninok: Why registering this callback at all if globe projection is active?

nzjony: Good question, maybe if the projection changes at runtime. But if this is something we don't need to support, then yes, all good.

In saying that though, are the objects recreated if the projection changes? I guess so, so yes, this would make sense.

ninok: yes, if the projection changes we clear the tile cache, so no need for runtime change

nzjony:

nzjony:

PR #1082, opened 2019-11-27
HARP-7985 Correct offset for fallback tiles

nzjony:
Fallback tiles now store the offset from the requested tile, for example
Grandparent tiles are -2, parent are -1 and children's children are +1

This ensures that grandparent and parent tiles don't overlap and cause
artifacts as shown in HARP-7985

There are cases, theoretically where the child's child's child overlaps
the child's child, this is also handled (because +2 has a higher render
order than +1).


nzjony:

PR #1085, opened 2019-11-27
HARP-7410 Changed algorithm to compute near and far plane for globe projection.

MPursche:
* Use maximum view distance to compute far plane for spherical projection.
* Increased number of visible tiles to avoid new visual artifacts.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
+++ b/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
@@ -294,9 +294,9 @@ export class TopViewClipPlanesEvaluator extends ElevationBasedClipPlanesEvaluato
 
     evaluateClipPlanes(mapView: MapView): ViewRanges {
         if (mapView.projection.type === ProjectionType.Spherical) {
-            return this.evaluateDistanceSphericalProj(mapView.camera, mapView.projection);
+            return this.evaluateDistanceSphericalProj(mapView);
         } else if (mapView.projection.type === ProjectionType.Planar) {
-            return this.evaluateDistancePlanarProj(mapView.camera, mapView.projection);
+            return this.evaluateDistancePlanarProj(mapView);
         }
         assert(false, "Unsupported projection type");
         return { ...this.minimumViewRange };
@@ -319,7 +319,8 @@ export class TopViewClipPlanesEvaluator extends ElevationBasedClipPlanesEvaluato
         return projection.groundDistance(camera.position);
     }
 
-    private evaluateDistancePlanarProj(camera: THREE.Camera, projection: Projection): ViewRanges {
+    protected evaluateDistancePlanarProj(mapView: MapView): ViewRanges {
+        const { camera, projection } = mapView;
         assert(projection.type !== ProjectionType.Spherical);
 
         let nearPlane: number = this.nearMin;
@@ -351,10 +352,8 @@ export class TopViewClipPlanesEvaluator extends ElevationBasedClipPlanesEvaluato
         return viewRanges;
     }
 
-    private evaluateDistanceSphericalProj(
-        camera: THREE.Camera,
-        projection: Projection
-    ): ViewRanges {
+    protected evaluateDistanceSphericalProj(mapView: MapView): ViewRanges {
+        const { camera, projection } = mapView;
         assert(projection.type === ProjectionType.Spherical);
 
         let nearPlane: number = this.nearMin;
@@ -424,7 +423,35 @@ export class TopViewClipPlanesEvaluator extends ElevationBasedClipPlanesEvaluato
         return viewRanges;
     }
 
-    private getTangentBasedFarPlane(
+    /**
+     * Calculate distance from a point to the tangent point of a sphere.
+     * Returns zero if point is below surface or only very slightly above surface of sphere.
+     * @param d Distance from point to center of sphere
+     * @param r Radius of sphere
+     */
+    protected getTangentDistance(d: number, r: number): number {
+        // There may be situations when maximum elevation still remains below sea level
+        // (elevation < 0) or it is negligible (elevation ~ epsilon)
+        if (d - r < epsilon) {
+            return 0;
+        }
+
+        // The distance to tangent point may be described as:
+        // t = sqrt(d^2 - r^2)
+        return Math.sqrt(d * d - r * r);
+    }
+
+    /**
+     * Calculate far plane depending on furthest visible distance from
+     * camera position. Furthest visible distance is assumed to be distance from camera to horizon
+     * plus distance from elevated geometry to horizon(so that high objects behind horizon
+     * remain visible).
+     * @param camera The camera of the mapview
+     * @param d Distance from camera to origin
+     * @param r Radius of earth
+     * @param alpha Angle between camera eye vector and tangent
+     */
+    protected getTangentBasedFarPlane(
         camera: THREE.PerspectiveCamera,
         d: number,
         r: number,
@@ -447,9 +474,8 @@ export class TopViewClipPlanesEvaluator extends ElevationBasedClipPlanesEvaluato
         //             d |  /
         //               |/
         //               C
-        // The distance to tangent point may be described as:
-        // t = sqrt(d^2 - r^2)
-        const t = Math.sqrt(d * d - r * r);
+        const t = this.getTangentDistance(d, r);
+
         // Because we would like to see elevated geometry that may be visible beyond
         // the tangent point on ground surface, we need to extend viewing distance along
         // the tangent line by te (see graph above). Knowing that OTE forms right angle,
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -717,6 +712,12 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
         return viewRanges;
     }
 
+    private getDistanceToHorizon(height: number): number {
```
ninok: Please add some typedoc:
* what height?
* be more precise about distance to horizon (distance from where to where)?

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -736,105 +747,79 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
...
-        // const farPlane = new THREE.Plane().setFromCoplanarPoints(p1, p2, p3);
-        // viewRanges.far = farPlane.distanceToPoint(camera.position);
-
-        // This of course may be simplified by moving calculus entirely to camera space,
-        // because far plane is indeed defined in that space anyway:
-        const farPlane = this.m_tmpPlane.setFromCoplanarPoints(tdrx, tdry, tdrx2);
-        viewRanges.far = farPlane.constant;
-        // Take into account largest depression assumed, that may be further then
-        // tangent based far plane distance.
+        const sinAlpha2 = r / (r + cameraAltitude);
```
kkostecki: Could check how much '(r + cameraAltitude)' differs from 'cameraToOrigin.length()'  (before normalization), because it should be exactly equal... Maybe that is the reason of problems...

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: I would go further, because it is actually calculates the distance from point P to the sphere tangent point T that lays on the line between P and T.
Maybe getTangentDistance()... I don't know, most importantly this need to assert when negative value is passed as height! Point below sea level - depression.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -736,105 +747,79 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
...
+            this.getDistanceToHorizon(this.maxElevation);
+
+        // Calculate angle between look at and surface normal
+        const cameraToOrigin = this.m_tmpVectors[0]
+            .set(0, 0, 0)
+            .sub(camera.position)
+            .normalize();
+        const lookAt = this.m_tmpVectors[1];
+        camera.getWorldDirection(lookAt).normalize();
+        const cosAlpha1 = cameraToOrigin.dot(lookAt);
```
kkostecki: This whole stuff could be enclosed in getCameraPitch() method, assuming that we will not need cameraToOirgin later on....
Keep in mind that we already introduced some change to TopViewEvaluator that optimize far distance whenever this angle is smaller then FOV/2, I hope to see this code also here at some point - would be nice, fingers crossed :).

@here/harp-mapview/lib/MapView.ts
```diff
@@ -582,7 +582,7 @@ export interface MapViewOptions extends TextElementsRendererOptions {
 export const MapViewDefaults = {
     projection: mercatorProjection,
 
-    maxVisibleDataSourceTiles: 120,
+    maxVisibleDataSourceTiles: 300,
```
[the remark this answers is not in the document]
MPursche: If the view is tilted far enough more than 120 tiles become visible. Problem is that it is semi random which tiles get discarded for spherical projection. For planar projection the area of a tile on screen is used for ordering, but for spherical projection the area is not actually calculated but set to 1 if within view frustum. So sometimes tiles near to the camera get discarded instead of only discarding tiles that are further away.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -736,105 +747,79 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
...
+        //     near      |      /
+        //             h |    / t = distance camera to horizon
+        //               |  /
+        //     look at ↑ |/  ↗ sightline to horizon
+        //               C
+
+        // Use distance to geometric horizon as base for max view distance
+        const maxViewDistance =
+            this.getDistanceToHorizon(cameraAltitude) +
+            this.getDistanceToHorizon(this.maxElevation);
```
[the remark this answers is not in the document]
MPursche: I'm currently rebasing my changes on top of the current master. I could reuse more code from `TopViewEvaluator`. I'm also changing some of my variable names to keep inline with existing names.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -736,105 +747,79 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
...
+
+        // Calculate angle between look at and surface normal
+        const cameraToOrigin = this.m_tmpVectors[0]
+            .set(0, 0, 0)
+            .sub(camera.position)
+            .normalize();
+        const lookAt = this.m_tmpVectors[1];
+        camera.getWorldDirection(lookAt).normalize();
+        const cosAlpha1 = cameraToOrigin.dot(lookAt);
+        const alpha1 = Math.acos(THREE.Math.clamp(cosAlpha1, -1.0, 1.0));
```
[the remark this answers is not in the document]
MPursche: Yes, because of FP inaccuracy length of the normalized vector was very slightly > 1.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
MPursche: Yes this can be simplified to cameraToOrigin.length().

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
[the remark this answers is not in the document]
kkostecki: Great! Good luck.

@here/harp-mapview/lib/MapView.ts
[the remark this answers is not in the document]
kkostecki: Ok, good to know, I would suggest to try with some semi-accurate approach, instead of calculating the area of tiles after view projection we may simply use their distance to camera inverted (instead of area in screen space), maybe that would help us - at least gives better results with small overhead.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -462,38 +488,21 @@ export class TopViewClipPlanesEvaluator extends ElevationBasedClipPlanesEvaluato
...
-        const cameraFwdVec = this.m_tmpVectors[2].negate();
-        // Apply quaternion to forward vector, creating tangent vector (in tmpVectors[1]).
-        const tangentVec = this.m_tmpVectors[1]
-            .copy(cameraFwdVec)
-            .applyQuaternion(this.m_tmpQuaternion);
-        // Give it a proper length
-        tangentVec.multiplyScalar(t + te);
-        // Calculate tangent projection onto camera eye vector, giving far plane distance.
-        return tangentVec.dot(cameraFwdVec);
+        // Calculate tangent projection onto camera eye vector, giving far plane distance
```
MPursche: I know this seems like a huge change but the code is actually equivalent.
We know that `tangentVec.dot(cameraFwdVec) = Math.cos(alpha) * tangentVec.length() * cameraFwdVec.length()`.
Because `cameraFwdVec.length() = 1` and ` tangentVec.length() == t + te` the formula can be simplified to ` Math.cos(alpha) * (t + te)`.

@here/harp-mapview/lib/FrustumIntersection.ts
```diff
@@ -230,39 +228,109 @@ export class FrustumIntersection {
...
-    private computeSubTileArea(tileBounds: THREE.Box3) {
-        if (
-            (!this.m_extendedFrustumCulling ||
-                this.m_mapTileCuller.frustumIntersectsTileBox(tileBounds)) &&
-            this.m_frustum.intersectsBox(tileBounds)
-        ) {
-            const contour = [
-                new THREE.Vector3(tileBounds.min.x, tileBounds.min.y, 0).applyMatrix4(
-                    this.m_viewProjectionMatrix
+    private computeSubTileArea(tileBounds: THREE.Box3 | OrientedBox3) {
```
MPursche: This change was necessary so that when the max visible tile limit is hit, tiles that are less visible(are smaller on screen) are discarded first. But there are still problems at the pole region.

kkostecki: Of course I understand the need for it, we will probably improve it anyway when implementing mixed LOD.
Anyway both methods are far from being accurate because only small portion of projected polygon may be actually visible. Imagine situation then only the corner vertex of entire tile passes the screen clipping (the rest is beyond frustum), then this tile is hardly visible (most probably just below camera) but contributes to loading a lot (gets highest priority because it is close and big after projection).
Thus we need to sacrifice here anyway, before finding some better solution. The question is if it will not affect our performance too much - need to be tested.
One minor change I would suggest is to split this method into two, they are used separately anyway and gets clearer interface (for ABB, OBB).

PR #1090, opened 2019-11-28
HARP-7729: MapViewUtils zoom and distance calculations give wrong value

SergiiKulakovskyi:

### DIFF
--- a/@here/harp-mapview/lib/Utils.ts
+++ b/@here/harp-mapview/lib/Utils.ts
@@ -62,6 +62,7 @@ const cache = {
 };
 
 export namespace MapViewUtils {
+    export const MAX_TILT_DEG = 89;
     /**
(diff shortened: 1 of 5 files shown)


@here/harp-mapview/lib/Utils.ts
```diff
@@ -686,6 +686,21 @@ export namespace MapViewUtils {
...
     }
 
+    /**
+     * Calculates the zoom level, which corresponds to the current distance from
+     * camera to the ground.
+     *
+     * @param mapView Instance of MapView.
+     * @param groundDistance Distance from the ground to the camera
+     */
+    export function calculateZoomLevelFromDistanceToGround(
```
ninok: Please add some tests that shows that going from zoomlevel to distance and back to zoomlevel actually works

PR #1097, opened 2019-11-29
Travis CI: Run on Node.JS 12

zbigg:
Run tests on  Node.JS 10 & as 12 is current LTS release.
Ensure that build for deplyment is ran only once on Node 12.

Related-to: HARP-7844

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 dist: bionic
 node_js:
   - "10"
+  - "12"
 cache:
   yarn: true
 
@@ -33,6 +34,7 @@ jobs:
         ./scripts/test-npm-packages.sh
         git status # just in case test-npm-packages leaves some garbage
     - name: "Build & Deploy"
+      node_js: "12"
       script: |
         set -ex
         yarn run build


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 dist: bionic
 node_js:
-  - "10"
+  - "12"
```
ninok: Not sure if we need another patch release, so I don't know if it is a good idea to do this now.
What travis.yml file will be picked up anyways for the release branch? The one from master or the one from the release branch?

zbigg: Travis always uses job definition from commit being tested. This isn't going to break current release .. not sure about future ones though :)

OTOH, this can wait no rush.

musculman: While I like having node 12, I dislike not having node 10... In fact I'd rather have node 10, 11 and 12 like in github ci actions...

zbigg: Hmm, agree, changing to 10 & 12.

zbigg: @musculman Node 11 is not supported anymore, not sure if we should waste our time on running CI on this. https://nodejs.org/en/about/releases/

PR #1100, opened 2019-11-29
HARP-7850: Add support for filtering style rules based on the zoom level

robertoraggi:
This change adds the property 'minZoomLevel' and 'maxZoomLevel'
to the style rules. These properties can be used to filter the rule
based on the zoom level.

### DIFF
--- a/@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
+++ b/@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
@@ -320,6 +320,7 @@ export class StyleSetEvaluator {
     private readonly m_emptyEnv = new Env();
     private m_layer: string | undefined;
     private m_geometryType: string | undefined;
+    private m_zoomLevel: number | undefined;
 
     constructor(styleSet: StyleSet, definitions?: Definitions) {
         this.m_definitions = definitions;
@@ -355,6 +356,8 @@ export class StyleSetEvaluator {
 
         const searchedStyleSet = this.getOptimizedStyleSet(optimizedSubSetKey);
 
+        const previousZoomLevel = this.changeZoomLevel(env.lookup("$zoom") as number | undefined);
+
         // set the requested $layer as the current layer.
         const previousLayer = this.changeLayer(
             typeof currentLayer === "string" ? currentLayer : undefined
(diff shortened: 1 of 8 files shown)


@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
```diff
@@ -355,6 +358,8 @@ export class StyleSetEvaluator {
 
         const searchedStyleSet = this.getOptimizedStyleSet(optimizedSubSetKey);
 
+        const previousZoomLevel = this.changeZoomLevel(env.lookup("$zoom") as number | undefined);
```
zbigg: This executes `env.lookup` for each feature, and most calls to `getMatchingTechniques` from OMVDecoder already know actual zoomLevel.
I would add `zoomLevel` to argumentlist of `getMatchingTechniques` ...

@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
```diff
@@ -372,6 +377,7 @@ export class StyleSetEvaluator {
 
         this.changeLayer(previousLayer); // restore the layer
         this.changeGeometryType(previousGeometryType); // restore the geometryType
+        this.changeZoomLevel(previousZoomLevel);
```
zbigg: This is actually question to #1096 ...
Is it still necessary to restore previous zoom/geomType/layer when we don't have nested styles ?

robertoraggi: Yes, now that the nested stylesets are gone the StyleSetEvaluator can be greatly simplified.
We could remove a lot of code and like the context handling (m_layer, m_geometrytype and m_zoomlevel and the change<Context> methods). I was trying to keep the change small to make it easier to review but of course, if we want we could also refactor that code in the context of this PR.

@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
robertoraggi: `$zoom` is something that we compute based the tile storage level and the storage level offsets, currently we compute $zoom in `createFeatureEnv()` and pass the Env to the the visit methods. So we need to refactor that code in order to remove the lookup for $zoom.

zbigg: Ok, will be in followup PR, which will set zoom at start of decoding of tile.

PR #1113, opened 2019-12-03
Fix: Raycaster is a `PickingRaycaster` object

vega-jose:
Adding point geometries from XYZ led to an error being thrown
when trying to map-pick. This was due to the fact that the raycast
method in the `MapViewPoints` class expected a raycaster object of
type `PickingRaycaster` in order to access the mapView object.
Providing it instead with a `THREE.Raycaster` object would result
in an error.

This commit addresses that by changing the type of raycaster the
`m_raycaster` object in the `MapView` class.

Signed-off-by: ESTIB <jose.vega@here.com>

# CI Check

Happy contributing!


zbigg:
@heremaps-bot ok to test

PR #1114, opened 2019-12-03
HARP-7881: Add support for javascript projects with harp.gl generator

CraigBerlin:


@here/generator-harp.gl/generators/app/templates/javascript/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+// react on resize events from the browser.
+window.addEventListener("resize", () => {
+    mapView.resize(window.innerWidth, window.innerHeight);
+});
+
+// center the camera to New York
+mapView.lookAt(new GeoCoordinates(40.70398928, -74.01319808), 1500, 40, 0);
+
+// make sure the map is rendered
```
zbigg: Wow, IMO `lookAt` should call `update`. Of couse, this is not scope of this PR.

PR #1116, opened 2019-12-04
HARP-7373: Add methods to wait for TextElementsRenderer initializatio…

atomicsulfate:
…n and loading.

Also, move MapView update call to the end of TextElementsRenderer initialization.
Update must be called after the renderer has been completely initialized.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -59,7 +59,7 @@ import {
 import { TextElementState } from "./TextElementState";
 import { TextElementStateCache } from "./TextElementStateCache";
 import { TextElementType } from "./TextElementType";
-import { TextStyleCache } from "./TextStyleCache";
+import { TextElementStyle, TextStyleCache } from "./TextStyleCache";
 import { UpdateStats } from "./UpdateStats";
 import { ViewState } from "./ViewState";
 
@@ -169,12 +169,6 @@ class TextElementLists {
     }
 }
 
-enum InitState {
-    Uninitialized,
-    Initializing,
-    Initialized
-}
-
 function checkIfTextElementsChanged(dataSourceTileList: DataSourceTileList[]) {
     let textElementsChanged = false;
 
@@ -212,7 +206,10 @@ function isPlacementTimeExceeded(startTime: number | undefined): boolean {
  * Internal class to manage all text rendering.
  */
 export class TextElementsRenderer {
-    private m_initState: InitState = InitState.Uninitialized;
+    private m_initialized: boolean = false;
+    private m_initPromise: Promise<void> | undefined;
+    private m_glyphLoadingCount: number = 0;
+    private m_loadPromise: Promise<any> | undefined;
     private m_initializedTextElementCount = 0;
     private readonly m_options: TextElementsRendererOptions;
 
@@ -485,11 +482,19 @@ export class TextElementsRenderer {
      * `true` if any resource used by any `FontCatalog` is still loading.
      */
     get loading(): boolean {
-        let isLoading = this.m_fontCatalogLoader.loading;
-        for (const textRenderer of this.m_textRenderers) {
-            isLoading = isLoading || textRenderer.textCanvas.fontCatalog.isLoading;
+        return this.m_fontCatalogLoader.loading || this.m_glyphLoadingCount > 0;
+    }
+
+    /**
+     * Waits till all pending resources from any `FontCatalog` are loaded.
+     */
+    async waitLoaded(): Promise<void> {
+        this.waitInitialized();
+
+        if (this.m_loadPromise !== undefined) {
+            await this.m_loadPromise;
         }
-        return isLoading;
+        return;
     }
 
     /**
@@ -520,11 +525,27 @@ export class TextElementsRenderer {
     }
 
     get initialized(): boolean {
-        return this.m_initState === InitState.Initialized;
+        return this.m_initialized;
     }
 
     get initializing(): boolean {
-        return this.m_initState === InitState.Initializing;
+        return this.m_initPromise !== undefined;
+    }
+
+    /**
+     * Waits until initialization is done.
+     * @returns Promise resolved to true if initialization was done, false otherwise.
+     */
+    async waitInitialized(): Promise<boolean> {
+        if (this.initialized) {
+            return true;
+        }
+
+        if (this.m_initPromise === undefined) {
+            return false;
+        }
+        await this.m_initPromise;
+        return true;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -713,61 +735,7 @@ export class TextElementsRenderer {
                 }
             }
 
-            // Trigger the glyph load if needed.
-            if (textElement.loadingState === undefined) {
-                textElement.loadingState = LoadingState.Requested;
-
```
atomicsulfate: Code moved to new method initializeGlyphs()

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -533,15 +554,16 @@ export class TextElementsRenderer {
...
-            this.invalidateCache(); // Force cache update after initialization.
+        if (!this.initialized && !this.initializing && textElementsAvailable) {
             this.initializeDefaultAssets();
-            this.initializeTextCanvases().then(() => {
-                this.m_initState = InitState.Initialized;
+            this.m_initPromise = this.initializeTextCanvases().then(() => {
+                this.m_initialized = true;
+                this.m_initPromise = undefined;
+                this.invalidateCache(); // Force cache update after initialization.
+                this.m_viewUpdateCallback();
```
atomicsulfate: Trigger map view update once initialization is completed.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -533,15 +554,16 @@ export class TextElementsRenderer {
...
             });
         }
-        return this.m_initState === InitState.Initialized;
+        return this.initialized;
```
kkostecki: Maybe it could be simpler just to return Promise instead of bool, and get rid of waitInitialized() method, of course it changes the usage too. Depends from the usage scope. Where do you want to use waitLoaded() (and thus waitInitialized)? I can not find this...

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -520,11 +525,27 @@ export class TextElementsRenderer {
...
+    /**
+     * Waits until initialization is done.
+     * @returns Promise resolved to true if initialization was done, false otherwise.
+     */
+    async waitInitialized(): Promise<boolean> {
+        if (this.initialized) {
+            return true;
+        }
+
+        if (this.m_initPromise === undefined) {
```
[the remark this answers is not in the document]
atomicsulfate: TextElementsRenderer has 3 states:

uninitialized, (we only initialize lazily once the first text is found).
initializing,
initialized

!this.initializing is true for both (uninitialized and initialized).

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: Initialization is handled completely within the class and done lazily only when there's some text to render. That's why initialize() is a private method.
MapView just blindly calls placeText() and renderText() on each frame, that way we keep client code from having to know too much about the state of the text renderer. Initialize is called from placeText() when needed, and there we don't want to wait for initialization.

The wait methods are there mainly to simplify testing. Synchronizing the test with the class initialization makes the test more understandable and less error prone.

atomicsulfate:
> For me looks good, only minor comments left, but would be better if somebody else would look on it too.

I trust your judgement. There's actually nobody with experience with this code because I've rewritten most of it.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
kkostecki: > Initialization is handled completely within the class and done lazily only when there's some text to render. That's why initialize() is a private method.
> MapView just blindly calls placeText() and renderText() on each frame, that way we keep client code from having to know too much about the state of the text renderer. Initialize is called from placeText() when needed, and there we don't want to wait for initialization.

That part was clear for me, sorry maybe I have asked to generic question.

>
> The wait methods are there mainly to simplify testing. Synchronizing the test with the class initialization makes the test more understandable and less error prone.

This is something I am missing - do we have `waitLoaded()` method used somewhere now (tests)? Could not find it. Maybe it would be nice to write some test code using it.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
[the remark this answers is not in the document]
kkostecki: So maybe:
`
get initializing(): boolean {
        return this.m_initPromise !== undefined && !this.m_initialized;
}
`
To better distinguish between states.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -520,11 +525,27 @@ export class TextElementsRenderer {
     }
 
     get initialized(): boolean {
-        return this.m_initState === InitState.Initialized;
+        return this.m_initialized;
     }
 
     get initializing(): boolean {
-        return this.m_initState === InitState.Initializing;
+        return this.m_initPromise !== undefined;
```
kkostecki:
```suggestion
        return this.m_initPromise !== undefined && !this.initialized;
```

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: I have a commit on top of this one adding unit testing for TextElementsRenderer, which uses this wait methods:

https://github.com/heremaps/harp.gl/commit/a47fb656967006963e8a6966720d02cd77343eb1

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -485,11 +482,19 @@ export class TextElementsRenderer {
...
+     * Waits till all pending resources from any `FontCatalog` are loaded.
+     */
+    async waitLoaded(): Promise<void> {
+        this.waitInitialized();
+
+        if (this.m_loadPromise !== undefined) {
+            await this.m_loadPromise;
         }
-        return isLoading;
+        return;
```
kkostecki: Here I would also return Promise<bool>, thus, the code would look like this:
`
this.waitInitialized();

if (this.m_loadPromise === undefined) {
    return false
}	 
await this.m_loadPromise;
return true;
`
this will be more consistent with another `waitInitilized()' method.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: Adding that condition there is unnecessary, initialized is always false when "this.m_initPromise !== undefined".

@here/harp-mapview/lib/text/TextElementsRenderer.ts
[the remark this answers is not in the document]
atomicsulfate: I'll make your suggested change in this method.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
kkostecki: Ok got ya.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
kkostecki: Ok, I see now.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: done

PR #1119, opened 2019-12-04
HARP-7517:  Transparency support via RGBA colors defined in techniques (themes).

kkostecki:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -6,7 +6,10 @@
 
 import {
     BufferAttribute,
+    ColorUtils,
+    Expr,
     getPropertyValue,
+    InterpolatedProperty,
     isExtrudedLineTechnique,
     isExtrudedPolygonTechnique,
     isInterpolatedProperty,
@@ -15,13 +18,18 @@ import {
     isTerrainTechnique,
     isTextureBuffer,
     parseStringEncodedColor,
+    ShaderTechnique,
     Technique,
+    techniqueDescriptors,
     TEXTURE_PROPERTY_KEYS,
-    TextureProperties
+    TextureProperties,
+    TRANSPARENCY_PROPERTY_KEYS,
+    Value
 } from "@here/harp-datasource-protocol";
-import { Expr } from "@here/harp-datasource-protocol/lib/Expr";
 import {
     CirclePointsMaterial,
+    disableBlending,
+    enableBlending,
     HighPrecisionLineMaterial,
     MapMeshBasicMaterial,
     MapMeshStandardMaterial,
@@ -111,6 +119,13 @@ export function createMaterial(
 
     if (isExtrudedPolygonTechnique(technique)) {
         material.flatShading = true;
+        // We do not support mixing vertex colors (static) and material colors (may be dynamic)
+        // mixture. Vertex colors are stored in VBO and are not modifiable - some solution for
+        // this problem is proposed in the HARP-8289 and PR #1164.
+        // TODO: Remove when problem with substitute (vertex & material) colors will be solved.
+        if (technique.vertexColors === true) {
+            delete technique.color;
+        }
     }
 
     material.depthTest = isExtrudedPolygonTechnique(technique) && technique.depthTest !== false;
@@ -212,18 +227,10 @@ export function createMaterial(
     }
 
     if (isShaderTechnique(technique)) {
-        // special case for ShaderTechnique.
-        // The shader technique takes the argument from its `params' member.
-        const params = technique.params as { [key: string]: any };
-        Object.getOwnPropertyNames(params).forEach(property => {
-            const prop = property as keyof typeof params;
-            if (prop === "name") {
-                // skip reserved property names
-                return;
-            }
-            applyTechniquePropertyToMaterial(material, prop, params[prop]);
-        });
+        // Special case for ShaderTechnique.
+        applyShaderTechniqueToMaterial(technique, material);
     } else {
+        // Generic technique.
         applyTechniqueToMaterial(technique, material, options.level, options.skipExtraProps);
     }
 
@@ -409,6 +416,71 @@ export function getMaterialConstructor(technique: Technique): MaterialConstructo
 }
 
 /**
+ * Allows to easy parse/encode technique's base color property value as number coded color.
+ *
+ * Function takes care about property parsing, interpolation and encoding if neccessary. If
+ * you wish to get default value without interpolation simply ignore @param zoom when calling.
+ *
+ * @see ColorUtils
+ * @param technique the technique where we search for base (transparency) color value
+ * @param zoomLevel zoom level used for value interpolation.
+ * @returns [[number]] encoded color value (in custom #TTRRGGBB) format or [[undefined]] if
+ * base color property is not defined in the technique passed.
+ */
+export function evaluateBaseColorProperty(
+    technique: Technique,
+    zoomLevel?: number
+): number | undefined {
+    const baseColorProp = getBaseColorProp(technique);
+    if (baseColorProp !== undefined) {
+        return evaluateColorProperty(baseColorProp, zoomLevel);
+    }
+    return undefined;
+}
+
+/**
+ * Apply [[ShaderTechnique]] parameters to material.
+ *
+ * @param technique the [[ShaderTechnique]] which requires special handling
+ * @param material material to which technique will be applied
+ */
+function applyShaderTechniqueToMaterial(technique: ShaderTechnique, material: THREE.Material) {
+    // The shader technique takes the argument from its `params' member.
+    const params = technique.params as { [key: string]: any };
+    // Remove base color and transparency properties from the processed set.
+    const baseColorPropName = getBaseColorPropName(technique);
+    const hasBaseColor = baseColorPropName && baseColorPropName in technique.params;
+    const props = Object.getOwnPropertyNames(params).filter(propertyName => {
+        // Omit base color and related transparency attributes if its defined in technique
+        if (
+            baseColorPropName === propertyName ||
+            (hasBaseColor && TRANSPARENCY_PROPERTY_KEYS.indexOf(propertyName) !== -1)
+        ) {
+            return false;
+        }
+        const prop = propertyName as keyof typeof params;
+        if (prop === "name") {
+            // skip reserved property names
+            return false;
+        }
+        return true;
+    });
+
+    // Apply all technique properties omitting base color and transparency attributes.
+    props.forEach(propertyName => {
+        // TODO: Check if properties values should not be interpolated, possible bug in old code!
+        // This behavior is kept in the new version too, level is set to undefined.
+        applyTechniquePropertyToMaterial(material, propertyName, params[propertyName]);
+    });
+
+    if (hasBaseColor) {
+        const propColor = baseColorPropName as keyof THREE.Material;
+        // Finally apply base color and related properties to material (opacity, transparent)
+        applyBaseColorToMaterial(material, material[propColor], technique, params[propColor]);
+    }
+}
+
+/**
  * Apply generic technique parameters to material.
  *
  * Skips non-material [[Technique]] props:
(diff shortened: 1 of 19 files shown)


@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -501,56 +522,84 @@ function applyTechniquePropertyToMaterial(
...
+/**
+ * Apply technique color to material taking special care with transparent (RGBA) colors.
+ *
+ * @note Transparent colors should be processed as the very last technique attributes,
+ * since their effect on [[THREE.Material]] properties like __opacity__ and __transparent__
+ * could be overridden by other technique params.
+ *
+ * @param technique an technique the applied color comes from
+ * @param material the material to which color is applied
+ * @param prop technique property (color) name
```
zbigg: s/technique/material/ property name?

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -501,56 +522,84 @@ function applyTechniquePropertyToMaterial(
...
+ * @param value color value
+ * @param level optional, tile zoom level for zoom-level dependent properties are evaluated.
+ */
+function applyTechniqueColorToMaterial(
```
zbigg: This is almost same as `applyColorToMaterial` (DecodedTileHelpers) added in #1104, https://github.com/heremaps/harp.gl/pull/1104/files#diff-4414054ad69d5d73485b5a4b589904e7R1573

Shouldn't we merge these two ?

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -457,26 +458,49 @@ export function applyTechniqueToMaterial(
...
         const m = material as any;
         if (typeof m[prop] === "undefined") {
-            return;
+            return false;
         }
+        return true;
+    });
+
+    // Remove transparent color from the firstly processed properties set.
+    let transparentColor = getTechniqueTransparentColor(technique);
```
zbigg: This is actually `transparentColorPropName`.

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -457,26 +458,49 @@ export function applyTechniqueToMaterial(
...
+    if (transparentColor !== undefined) {
+        const idx = commonProps.indexOf(transparentColor);
+        transparentColor = idx >= 0 ? commonProps.splice(idx, 1)[0] : undefined;
```
zbigg: >commonProps.splice(idx, 1)[0]

Huh? Why even searching for this, if we can basically ... check it later as.

```
if (transparentColorPropName !== undefined && transparentColorPropName in technique) {
}
```

@here/harp-mapview/lib/DecodedTileHelpers.ts
zbigg: And this should be named `applyMainTechniqueColorToMaterial` because this particular color should have some "afjective" and main is good :) (IMO)

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -457,26 +458,49 @@ export function applyTechniqueToMaterial(
     level?: number,
     skipExtraProps?: string[]
 ) {
-    Object.getOwnPropertyNames(technique).forEach(propertyName => {
+    const commonProps = Object.getOwnPropertyNames(technique).filter(propertyName => {
         if (
             propertyName.startsWith("_") ||
```
zbigg: `transparentColor` (if exists) and `opacity` and ``transparent` should be skipped here IMO

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -457,26 +458,49 @@ export function applyTechniqueToMaterial(
     level?: number,
     skipExtraProps?: string[]
 ) {
-    Object.getOwnPropertyNames(technique).forEach(propertyName => {
+    const commonProps = Object.getOwnPropertyNames(technique).filter(propertyName => {
```
zbigg: > `commonProps` -> `genericProps`

These props  `generic` as they're excercising generic/normal flow in contrast to filtered ones which are ignored or ones that require specific flow.

@here/harp-mapview/lib/DecodedTileHelpers.ts
kkostecki: Fixed.

@here/harp-mapview/lib/DecodedTileHelpers.ts
kkostecki: I guess you meant TileGeometryCreator.applyColorToMaterial(..), that's the fact that both methods are almost the same, the main three differences are:
- In TileGeometryCreator zoom level (`level` param) is always defined - don't ask me why it's not the same in DecodedTileHelpers - I am not the owner of original code/interfaces.
- In TileGeometryCreator all technique properties are always interpolated, which is not true for DecodedTileHelpers (see above for comments :) )
- In DecodedTileHelpers we could potentially use pre-calculated Material.opacity value (without call to getPropertyValue) assuming that transparent color transfer to material goes as the last,
- DecodedTileHelpers triggers notifying change (again legacy code) on material:
`m[prop] = m[prop];`
which is not done in TileGeometryCreator.
Summarizing, if we could use one general method for both, that would need to be DecodedTileHelpers.applyColorToMaterial(...) with all the consequences - lack of simplification and with all `level` checks and so...
I am also voting for this, but need some hints if it would be still correct and desired...

@here/harp-mapview/lib/DecodedTileHelpers.ts
kkostecki: Fixed.

@here/harp-mapview/lib/DecodedTileHelpers.ts
kkostecki: This can be done, but with some changes to the logic, imagine that technique has transparent color attribute (in descriptor), this color is defined, but does not contain the transparency octet (alpha value), then with current implementation material.opacity & material.transparent will never be defined (transferred from technique to material).
Of course we may do this, with really small overhead: 2 number multiplications regardless of the color defined.

@here/harp-mapview/lib/DecodedTileHelpers.ts
kkostecki: Yes but we need to remove (by index or by name) this property from the for each loop (of course I wouldn't like to check this condition:
`if propertyName !== transparentColorPropName)` in each loop call, moreover things complicate even more if we want to ignore transparency and opacity values as you suggested in next comments (2 more ifs) . So I guess find and splice is the best we can do with the lack of Array.removeIf utility function.

@here/harp-mapview/lib/DecodedTileHelpers.ts
kkostecki: Ok, never mind, I have changed logic/flow a lot, so transparent/opacity material properties are now removed from the first stage and applied with _main/base_ color if any, otherwise nothing is change and all props are applied as usually.

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -421,67 +480,249 @@ export function getMaterialConstructor(technique: Technique): MaterialConstructo
...
+    const { r, g, b, a } = ColorUtils.getRgbaFromHex(value);
+    // Override material opacity and transparency by mixing technique defined opacity
+    // with main color transparency
+    const tech = technique as any;
+    let opacity = a;
+    if (tech.opacity !== undefined) {
+        opacity *= evaluateProperty(tech.opacity, level);
+    }
+    opacity = THREE.Math.clamp(opacity, 0, 1);
+    let transparent = opacity !== 1.0;
```
ninok: As discussed this has the side effect of changing the render order. (despite of the fact that there are still artifacts that @robertoraggi is looking into). I would rather not touch the transparent flag but just make sure that blending is enabled.

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -422,65 +491,239 @@ export function getMaterialConstructor(technique: Technique): MaterialConstructo
...
+    // with main color transparency
+    const tech = technique as any;
+    let opacity = a;
+    if (tech.opacity !== undefined) {
+        opacity *= evaluateProperty(tech.opacity, level);
+    }
+
+    opacity = THREE.Math.clamp(opacity, 0, 1);
+    material.opacity = opacity;
+
```
kkostecki: If think we should also add transparent coping (support) from technique:
```
    // TODO: Remove setting material.transparent if we do now want to support it anymore
    // from the technique params.
    if (tech.transparent !== undefined) {
        material.transparent = evaluateProperty(tech.transparent, level);
    }
```

@here/harp-mapview/lib/DecodedTileHelpers.ts
zbigg: Done, we're using `enable/disableBlending` that is compatible with #1144.

@here/harp-mapview/lib/DecodedTileHelpers.ts
zbigg: Done

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -422,65 +491,239 @@ export function getMaterialConstructor(technique: Technique): MaterialConstructo
...
+) {
+    const colorValue = evaluateColorProperty(techniqueColor, level);
+
+    const { r, g, b, a } = ColorUtils.getRgbaFromHex(colorValue);
+    // Override material opacity and blending by mixing technique defined opacity
+    // with main color transparency
+    const tech = technique as any;
+    let opacity = a;
+    if (tech.opacity !== undefined) {
+        opacity *= evaluateProperty(tech.opacity, level);
```
ninok: Is all this supposed to work if opacity or the base color are dynamic properties?

ninok: As far as I can see it is (i.e. applyBaseColorToMAterial is called in the onBeforeRender callbacks).
Please rename level to zoomLevel to make it more clear.

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -422,65 +491,239 @@ export function getMaterialConstructor(technique: Technique): MaterialConstructo
...
+    genericProps.forEach(propertyName => {
+        const value = technique[propertyName as keyof Technique];
+        if (value !== undefined) {
+            applyTechniquePropertyToMaterial(material, propertyName, value, level);
         }
-        applyTechniquePropertyToMaterial(material, prop, value);
     });
+
+    // Finally apply base (possibly transparent) color itself, modifying material.opacity and
+    // material.transparent attributes too.
```
ninok: This comment is not accurate anymore. The transparent attribute stays untouched

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -422,65 +491,239 @@ export function getMaterialConstructor(technique: Technique): MaterialConstructo
...
+    const tech = technique as any;
+    let opacity = a;
+    if (tech.opacity !== undefined) {
+        opacity *= evaluateProperty(tech.opacity, level);
+    }
+
+    opacity = THREE.Math.clamp(opacity, 0, 1);
+    material.opacity = opacity;
+
+    const opaque = opacity === 1.0;
```
ninok: >= 1.0

kkostecki: Greater then 1.0 should never happen... see clamp above.

@here/harp-mapview/lib/DecodedTileHelpers.ts
kkostecki: Fixed.

@here/harp-mapview/lib/DecodedTileHelpers.ts
kkostecki: Fixed.

PR #1122, opened 2019-12-04
Harp 8066: Fix bug with fallback + IBCT tests.

nzjony:
This fixes a bug in the fallback logic where if a parent was already added to the set of renderedTiles, then subsequent siblings would search for their children, giving strange results. Now if there is a parent tile found, it is added and all siblings don't search for their children.
Added tests also, see gerrit.

### DIFF
--- a/@here/harp-mapview/lib/VisibleTileSet.ts
+++ b/@here/harp-mapview/lib/VisibleTileSet.ts
@@ -863,13 +863,10 @@ export class VisibleTileSet {
      * flickering by rendering already loaded tiles from upper/higher zoom levels.
      */
     private fillMissingTilesFromCache() {
-        const tileCache = this.m_dataSourceCache;
         this.dataSourceTileList.forEach(renderListEntry => {
             const dataSource = renderListEntry.dataSource;
-            const tilingScheme = dataSource.getTilingScheme();
             const displayZoomLevel = renderListEntry.zoomLevel;
             const renderedTiles = renderListEntry.renderedTiles;
-            const checkedTiles: Set<number> = new Set<number>();
 
             // Direction in quad tree to search: up -> shallower levels, down -> deeper levels.
             enum SearchDirection {
@@ -894,7 +891,7 @@ export class VisibleTileSet {
                     ? SearchDirection.UP
                     : SearchDirection.NONE;
 
-            let incompleteTiles: Map<number, SearchDirection> = new Map();
+            const incompleteTiles: Map<number, SearchDirection> = new Map();
 
             renderListEntry.visibleTiles.forEach(tile => {
                 const tileCode = TileOffsetUtils.getKeyForTileKeyAndOffset(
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -916,89 +913,138 @@ export class VisibleTileSet {
...
-                        searchDirection === SearchDirection.UP
+                        this.findUp(
+                            tileKeyCode,
+                            displayZoomLevel,
+                            renderedTiles,
+                            checkedTiles,
+                            dataSource
+                        )
                     ) {
-                        const parentCode = TileOffsetUtils.getParentKeyFromKey(tileKeyCode);
```
nzjony: this moved into findUp

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -916,89 +913,138 @@ export class VisibleTileSet {
...
-                            if (nextLevelDiff < this.options.quadTreeSearchDistanceUp) {
-                                nextLevelCandidates.set(parentCode, SearchDirection.UP);
-                            }
-                        }
+                        // Continue to next entry so we don't search down.
+                        continue;
                     }
+                }
 
-                    if (
```
nzjony: this is moved into findDown

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -916,89 +913,138 @@ export class VisibleTileSet {
...
+        tileKeyCode: number,
+        displayZoomLevel: number,
+        renderedTiles: Map<number, Tile>,
+        checkedTiles: Map<number, boolean>,
+        dataSource: DataSource
+    ): boolean {
+        const parentCode = TileOffsetUtils.getParentKeyFromKey(tileKeyCode);
+        // Check if another sibling has already added the parent.
+        if (renderedTiles.get(parentCode) !== undefined) {
+            return true;
```
nzjony: This is what fixes the bug mentioned, all siblings return true, which then continues, so the children of the siblings aren't searched.

nzjony:

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -916,89 +913,138 @@ export class VisibleTileSet {
...
-            // iterate over incomplete (not loaded tiles)
-            // and find their parents or children that are in cache that can be rendered temporarily
-            // until tile is loaded
-            while (incompleteTiles.size !== 0) {
-                const nextLevelCandidates: Map<number, SearchDirection> = new Map();
-
-                incompleteTiles.forEach((searchDirection, tileKeyCode) => {
+            // Minor optimization for the fallback search, only check parent tiles once, otherwise
+            // the recursive algorithm checks all parent tiles multiple times, the key is the code
+            // of the tile that is checked and the value is whether
```
ninok: whether what?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -916,89 +913,138 @@ export class VisibleTileSet {
...
+        renderedTiles: Map<number, Tile>,
+        checkedTiles: Map<number, boolean>,
+        dataSource: DataSource
+    ): boolean {
+        const parentCode = TileOffsetUtils.getParentKeyFromKey(tileKeyCode);
+        // Check if another sibling has already added the parent.
+        if (renderedTiles.get(parentCode) !== undefined) {
+            return true;
+        } else if (checkedTiles.has(parentCode)) {
+            return checkedTiles.get(parentCode)!;
```
ninok: Maybe only use get and check for undefined to avoid double lookup.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -916,89 +913,140 @@ export class VisibleTileSet {
...
+    private findUp(
+        tileKeyCode: number,
+        displayZoomLevel: number,
+        renderedTiles: Map<number, Tile>,
+        checkedTiles: Map<number, boolean>,
+        dataSource: DataSource
+    ): boolean {
+        const parentCode = TileOffsetUtils.getParentKeyFromKey(tileKeyCode);
+        // Check if another sibling has already added the parent.
+        if (renderedTiles.get(parentCode) !== undefined) {
```
ninok: I don't fully understand the difference between `checkedTiles` and `renderedTiles`. When would a tile be in `checkedTiles` and not in `renderedTiles`?

PR #1125, opened 2019-12-05
HARP-8111 Refactored tile loader to fix bug and remove unecessary tile reloads.

MPursche:
Signed-off-by: Marcel Pursche marcel.pursche@here.com

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview-decoder/lib/TileLoader.ts
+++ b/@here/harp-mapview-decoder/lib/TileLoader.ts
@@ -104,6 +104,9 @@ export class TileLoader {
     loadAndDecode(): Promise<TileLoaderState> {
         switch (this.state) {
             case TileLoaderState.Loading:
+            case TileLoaderState.Loaded:
+            case TileLoaderState.Decoding:
+                // tile is already loading
                 this.countRequests++;
                 return this.donePromise!;
 
@@ -111,18 +114,9 @@ export class TileLoader {
             case TileLoaderState.Failed:
             case TileLoaderState.Initialized:
             case TileLoaderState.Canceled:
+                // restart loading
                 this.countRequests++;
-                this.ensureLoadingStarted();
-                return this.donePromise!;
-
-            case TileLoaderState.Loaded:
-                this.countRequests++;
-                this.startDecodeTile();
-                return this.donePromise!;
-
-            case TileLoaderState.Decoding:
-                this.cancelDecoding();
-                this.ensureLoadingStarted();
+                this.startLoading();
                 return this.donePromise!;
         }
     }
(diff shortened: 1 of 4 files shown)


@here/harp-mapview-decoder/lib/TileLoader.ts
```diff
@@ -104,25 +104,19 @@ export class TileLoader {
     loadAndDecode(): Promise<TileLoaderState> {
         switch (this.state) {
             case TileLoaderState.Loading:
+            case TileLoaderState.Loaded:
+            case TileLoaderState.Decoding:
+                // tile is already loading
```
nzjony: it could also be Decoding

@here/harp-mapview-decoder/lib/TileLoader.ts
```diff
@@ -104,25 +104,19 @@ export class TileLoader {
     loadAndDecode(): Promise<TileLoaderState> {
         switch (this.state) {
             case TileLoaderState.Loading:
+            case TileLoaderState.Loaded:
```
nzjony: I looked into the code, it seems that the Loaded state is set for a very short time period. Maybe it would be worth at least adding a comment that this is potentially no longer necessary.

@here/harp-mapview-decoder/lib/TileLoader.ts
```diff
@@ -104,25 +104,19 @@ export class TileLoader {
...
-                this.ensureLoadingStarted();
-                return this.donePromise!;
-
-            case TileLoaderState.Loaded:
-                this.countRequests++;
-                this.startDecodeTile();
-                return this.donePromise!;
-
-            case TileLoaderState.Decoding:
-                this.cancelDecoding();
```
nzjony: I'm not sure this is exactly correct, if you are decoding a Tile, and then call loadAndDecode, it probably still makes sense to cancel the request. For example, if you are changing styles, and you don't cancel the current request, you will get tiles with different styles.

MPursche:
The main issue is that in `TileLoader.ensureLoadingStarted()` the `TileLoaderState.Failed` isn't handled. So if loading of a tile fails once(i.e. no network connection), it will never be loaded, because the `TileLoader` can't recover from the failed state.
I refactored the code and removed  `TileLoader.ensureLoadingStarted()` because it is only called from `TileLoader.loadAndDecode()` and it doesn't make sense to check the state of the `TileLoader` in both methods. It makes more sense to only check the state in `loadAndDecode()` and handle the different branches there.

@here/harp-mapview-decoder/lib/TileLoader.ts
MPursche: When the style is changed the tiles of a datasource are marked as dirty and evicted from the cache. This also cancels running requests. There still seems to be a bug where tiles are displayed with different styles but it can also be triggered with the changes in this pull request.

nzjony:

@here/harp-mapview-decoder/lib/TileLoader.ts
nzjony: You're right, ok cool.

PR #1128, opened 2019-12-09
HARP-7872: Add APIs to substitude variables in an Expr.

robertoraggi:


@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -16,6 +17,8 @@ export * from "./Env";
 
 const exprEvaluator = new ExprEvaluator();
 
+const instantiateExpr = new InstantiateExpr();
```
ninok: `ExprInstantiator` would match the naming scheme of `ExprEvaluator` better.

PR #1132, opened 2019-12-09
HARP-7872: Add 'Expr.isDynamic()'

robertoraggi:
This change adds the method 'isDynamic()' to the class Expr. This
method can be used to classify expressions based on their expected
execution context. For example, ["zoom"] is expected to always be
executed using a dynamic context.

### DIFF
--- a/@here/harp-datasource-protocol/lib/Expr.ts
+++ b/@here/harp-datasource-protocol/lib/Expr.ts
@@ -230,6 +230,8 @@ export abstract class Expr {
         return parseNode(node, referenceResolverState);
     }
 
+    private m_isDynamic?: boolean;
+
     /**
      * Evaluate an expression returning a [[Value]] object.
      *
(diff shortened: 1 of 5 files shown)


@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -521,6 +575,14 @@ export class MatchExpr extends Expr {
     accept<Result, Context>(visitor: ExprVisitor<Result, Context>, context: Context): Result {
         return visitor.visitMatchExpr(this, context);
     }
+
+    protected exprIsDynamic() {
+        return (
+            this.value.isDynamic() ||
+            this.branches.some(([_, branch]) => branch.isDynamic()) ||
```
ninok: if the value is not dynamic the expression would be only dynamic if the according match would be dynamic

robertoraggi: good point, I will restrict the behavior of `isDynamic` for `Case` and `Match` in a follow up PR.

PR #1134, opened 2019-12-09
HARP-7551: Adding harp.gl webpack configuration helper

CraigBerlin:


@here/harp-webpack-config/index.js
```diff
@@ -0,0 +1,98 @@
...
+    const userConfig = config ? config : {};
+    const commonConfig = {
+        output: {
+            filename: "[name].bundle.js",
+            path: path.join(__dirname, "dist")
+        },
+        devtool: "source-map",
+        externals: [
+            {
+                three: "THREE"
```
CraigBerlin: Actually this break pure javascript projects created with the generator.
We may need to add a parameter for project type.

@here/harp-mapview/package.json
```diff
@@ -47,17 +47,25 @@
...
+        "copy-webpack-plugin": "^5.0.5",
         "cross-env": "^6.0.3",
+        "html-webpack-plugin": "^3.2.0",
         "mocha": "^6.1.4",
         "sinon": "^7.1.1",
         "source-map-support": "^0.5.2",
         "three": "^0.110.0",
-        "typescript": "^3.7.2"
+        "typescript": "^3.7.2",
+        "webpack": "^4.41.2",
```
zbigg: Putting all the deps here, doesn't help user (one that npm-requires `@here/harp-mapview`) as it will not have `devDependencies` installed.

Adding this do `dependencies` is also no-go as we don't want user to be forced to use webpack.

I think, this script should either declare all these deps as `peerDependecies` or we should create separate package like mentioned somewhere `@here/harp-webpack-utils` or something.

@here/harp-mapview/scripts/HarpWebpackConfig.ts
```diff
@@ -0,0 +1,119 @@
...
+                stats: {
+                    all: false,
+                    timings: true,
+                    exclude: "/resources/",
+                    errors: true,
+                    entrypoints: true,
+                    warnings: true
+                }
+            },
+            userConfig
```
zbigg: Does it work if `userConfig` is already "array of bundles" ?

@here/generator-harp.gl/generators/app/index.js
```diff
@@ -55,8 +55,6 @@ module.exports = class extends Generator {
         this.spawnCommandSync("npm", [
             "install",
             "--save-dev",
-            "copy-webpack-plugin",
-            "html-webpack-plugin",
```
zbigg: I don't thing this can be removed now. In current state of patch, `copy-webpack-plugin` and second one will be unresolved because @here/harp-mapview specifies them as `devDependencies`.

docs/GettingStartedGuide.md
```diff
@@ -147,6 +126,21 @@ OmvTileDecoderService.start();
...
 
+### Webpack configuration
+
+Add the following lines to your `webpack.config.js`:
+
+```javascript
+const { addHarpWebpackConfig } = require("@here/harp-mapview/scripts/HarpWebpackConfig");
+const myConfig = {};
+const harpConfig = addHarpWebpackConfig(myConfig, "./index.js", "./harp-gl-decoders.js", "./index.html");
+module.exports = [ ...harpConfig ];
```
zbigg: This could be shortened to
`module.exports = addHarpWebpackConfig( ...)`

@here/harp-mapview/scripts/HarpWebpackConfig.ts
```diff
@@ -0,0 +1,119 @@
...
+            extensions: [".webpack.js", ".web.js", ".js"]
+        },
+        performance: {
+            hints: false
+        },
+        mode
+    };
+    const typescriptConfig: Configuration = {
+        externals: [
+            {
```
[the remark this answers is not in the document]
CraigBerlin: Ok will do. I removed it for javascript as the sample app worked without it.

@here/harp-mapview/package.json
CraigBerlin: ok, actually I had it in a separate package at first. Was trying to simplify things but you are right I overlooked that dev dependencies are not installed transitively. Will move back to separate package.

@here/harp-mapview/scripts/HarpWebpackConfig.ts
CraigBerlin: Nope. I had a check for that before I switched to typescript, I will restore it so we throw an exception.
The user will need to handle merging themselves in that case.

@here/generator-harp.gl/generators/app/index.js
CraigBerlin: Correct, will add a dependency on the harp-webpack-utils module instead..

@here/harp-webpack-utils/scripts/HarpWebpackConfig.ts
```diff
@@ -0,0 +1,118 @@
...
+
+function createPlugins(htmlTemplate?: string): Plugin[] {
+    const plugins = [
+        new CopyWebpackPlugin([
+            {
+                from: "node_modules/@here/harp-map-theme/resources",
+                to: "resources",
+                toType: "dir"
+            },
+            require.resolve("three/build/three.min.js")
```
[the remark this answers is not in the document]
CraigBerlin: True, not needed here. Moved to generator where it is required for example project.

@here/harp-webpack-utils/package.json
```diff
@@ -0,0 +1,48 @@
+{
+    "name": "@here/harp-webpack-utils",
+    "version": "0.11.1",
```
robertoraggi: 0.1.0?

CraigBerlin: To limit the git tags per release we try to keep the public packages in sync.

PR #1027, opened 2019-11-12
HARP-7399 Added camera properties to MapView constructor

MPursche:
Signed-off-by: Marcel Pursche <marcel.pursche@here.com>

Thank you for contributing to harp.gl!

Before requesting a pull request, please remember to check the following documents:
* [contribution guidelines](https://github.com/heremaps/harp.gl/blob/master/CONTRIBUTING.md)
* [coding style](https://github.com/heremaps/harp.gl/blob/master/CODINGSTYLE.md)

If you are adding new functionality we would highly appreciate if you can describe what is the capability you are adding and even better if you can add some examples. Please also remember to add tests for it.

# CI Check

Our bots will check whether your PR can be directly integrated into the mainline. We have some internal integration tests running on the background, our bots will inform you of the next steps and someone from our team will take a look and help if needed!

And please do not forget to sign-off your commit! You can read more about DCO [here](https://julien.ponge.org/blog/developer-certificate-of-origin-versus-contributor-license-agreements/). But, in short, you just need to use `git commit -s` or append `--signoff` when you are committing to the repo.

Happy contributing!

PR #1139, opened 2019-12-10
HARP-7907: Limit number of raycasts used for Camera tilt calculus.

kkostecki:
In some cases focus point (and tilt angle) may not be calculated because
Camera is looking to the void. In this scenarios the last known focus point
is used and the highest possible tilt value is exposed (90 degrees).
This allows to keep camera pivot consistent (when changing the projection
type, or manipulating camera manually) and computations based on tilt
stable (using maximum tilt), this may be: fog, fading effect, etc.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-map-controls/lib/MapControls.ts
+++ b/@here/harp-map-controls/lib/MapControls.ts
@@ -355,34 +355,34 @@ export class MapControls extends THREE.EventDispatcher {
             return;
         }
         this.camera.getWorldDirection(this.m_currentViewDirection);
-        const maxDistance = MapViewUtils.calculateDistanceToGroundFromZoomLevel(
+        let maxDistance = MapViewUtils.calculateDistanceToGroundFromZoomLevel(
             this.mapView,
             this.mapView.minZoomLevel
         );
-        const minDistance = MapViewUtils.calculateDistanceToGroundFromZoomLevel(
+        let minDistance = MapViewUtils.calculateDistanceToGroundFromZoomLevel(
             this.mapView,
             this.mapView.maxZoomLevel
         );
         this.m_currentViewDirection.multiplyScalar(amount);
         if (this.mapView.projection.type === geoUtils.ProjectionType.Planar) {
-            this.camera.position.z += this.m_currentViewDirection.z;
-            this.camera.position.z = Math.max(
+            const distance = THREE.Math.clamp(
+                this.camera.position.z + this.m_currentViewDirection.z,
                 minDistance,
-                Math.min(maxDistance, this.camera.position.z)
+                maxDistance
             );
+            this.camera.position.z = distance;
         } else if (this.mapView.projection.type === geoUtils.ProjectionType.Spherical) {
             const zOnVertical =
                 Math.cos(this.camera.position.angleTo(this.m_currentViewDirection)) *
                 this.m_currentViewDirection.length();
-            this.camera.position.setLength(
-                Math.max(
-                    minDistance + geoUtils.EarthConstants.EQUATORIAL_RADIUS,
-                    Math.min(
-                        maxDistance + geoUtils.EarthConstants.EQUATORIAL_RADIUS,
-                        this.camera.position.length() + zOnVertical
-                    )
-                )
+            minDistance += geoUtils.EarthConstants.EQUATORIAL_RADIUS;
+            maxDistance += geoUtils.EarthConstants.EQUATORIAL_RADIUS;
+            const distance = THREE.Math.clamp(
+                this.camera.position.length() + zOnVertical,
+                minDistance,
+                maxDistance
             );
+            this.camera.position.setLength(distance);
         }
 
         // In sphere, we may have to also orbit the camera around the position located at the
@@ -392,17 +392,18 @@ export class MapControls extends THREE.EventDispatcher {
             this.mapView.projection.type === geoUtils.ProjectionType.Spherical &&
             this.m_maxTiltAngle !== undefined
         ) {
-            const centerScreenTarget = MapViewUtils.rayCastWorldCoordinates(this.mapView, 0, 0);
-            if (centerScreenTarget !== null) {
-                const tilt = MapViewUtils.extractSphericalCoordinatesFromLocation(
-                    this.mapView,
-                    this.camera,
-                    this.mapView.projection.unprojectPoint(centerScreenTarget)
-                ).tilt;
-                const deltaTilt = tilt - this.m_maxTiltAngle;
-                if (deltaTilt > 0) {
-                    MapViewUtils.orbitFocusPoint(this.mapView, 0, deltaTilt, this.m_maxTiltAngle);
-                }
+            // Use pre-calculated camera target, otherwise we could get it via:
+            // centerScreenTarget = MapViewUtils.getTargetPositionFromCamera(camera, projection)
+            // and convert to geo-coordinates via:
+            // this.mapView.projection.unprojectPoint(centerScreenTarget)
+            const tilt = MapViewUtils.extractSphericalCoordinatesFromLocation(
+                this.mapView,
+                this.camera,
+                this.mapView.target
+            ).tilt;
+            const deltaTilt = tilt - this.m_maxTiltAngle;
+            if (deltaTilt > 0) {
+                MapViewUtils.orbitFocusPoint(this.mapView, 0, deltaTilt, this.m_maxTiltAngle);
             }
         }
 
@@ -418,16 +419,15 @@ export class MapControls extends THREE.EventDispatcher {
      * of changing the yaw (which would be the camera rotating on itself).
      */
     pointToNorth() {
-        const target = MapViewUtils.rayCastWorldCoordinates(this.mapView, 0, 0);
-        if (target === null) {
-            throw new Error("MapView does not support a view pointing in the void.");
-        }
+        // Use pre-calculated target coordinates, otherwise we could call utility method to evaluate
+        // geo-coordinates here:
+        // targetGeoCoords = MapViewUtils.getTargetCoordinatesFromCamera(camera, projection)
         this.m_startAzimuth =
             Math.PI +
             MapViewUtils.extractSphericalCoordinatesFromLocation(
                 this.mapView,
                 this.camera,
-                this.mapView.projection.unprojectPoint(target)
+                this.mapView.target
             ).azimuth;
         // Wrap between -PI and PI.
         this.m_startAzimuth = Math.atan2(
@@ -579,15 +579,7 @@ export class MapControls extends THREE.EventDispatcher {
     }
 
     private get currentTilt(): number {
-        const target = MapViewUtils.rayCastWorldCoordinates(this.mapView, 0, 0);
-        if (target === null) {
-            throw new Error("MapView does not support a view pointing in the void.");
-        }
-        return MapViewUtils.extractSphericalCoordinatesFromLocation(
-            this.mapView,
-            this.camera,
-            this.mapView.projection.unprojectPoint(target)
-        ).tilt;
+        return MapViewUtils.extractCameraTilt(this.mapView.camera, this.mapView.projection);
     }
 
     private get targetedTilt(): number {
@@ -693,7 +685,7 @@ export class MapControls extends THREE.EventDispatcher {
         const initialTilt = this.currentTilt;
         const deltaAngle = this.m_currentTilt - initialTilt;
         const oldCameraDistance = this.mapView.camera.position.z / Math.cos(initialTilt);
-        const newHeight = Math.cos(this.currentTilt) * oldCameraDistance;
+        const newHeight = Math.cos(initialTilt) * oldCameraDistance;
 
         MapViewUtils.orbitFocusPoint(
             this.mapView,
--- a/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
+++ b/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
@@ -704,15 +704,7 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
         // opposite to elevating ground level.
         const halfPiLimit = Math.PI / 2 - epsilon;
         const cameraAltitude = this.getCameraAltitude(camera, projection);
-        const target = MapViewUtils.rayCastWorldCoordinates(mapView, 0, 0);
-        if (target === null) {
-            throw new Error("MapView does not support a view pointing in the void.");
-        }
-        const cameraTilt = MapViewUtils.extractSphericalCoordinatesFromLocation(
-            mapView,
-            camera,
-            projection.unprojectPoint(target)
-        ).tilt;
+        const cameraTilt = this.getCameraTilt(mapView);
         // Angle between z and c2
         let topAngleRad: number;
         // Angle between z and c1
@@ -875,7 +867,7 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
 
         // Apply the constraints.
         const farMin = cameraAltitude - this.minElevation;
-        const farMax = mapView.lookAtDistance * this.farMaxRatio;
+        const farMax = mapView.targetDistance * this.farMaxRatio;
         viewRanges.near = Math.max(viewRanges.near, this.nearMin);
         viewRanges.far = MathUtils.clamp(viewRanges.far, farMin, farMax);
 
@@ -952,6 +944,10 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
         const cameraPitch = Math.acos(MathUtils.clamp(cosAlpha1, -1.0, 1.0));
 
         return cameraPitch;
+    }
+
+    private getCameraTilt(mapView: MapView): number {
+        return MapViewUtils.extractCameraTilt(mapView.camera, mapView.projection);
     }
 }
 
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -140,6 +140,7 @@ const DEFAULT_CAM_NEAR_PLANE = 0.1;
 const DEFAULT_CAM_FAR_PLANE = 4000000;
 const MAX_FIELD_OF_VIEW = 140;
 const MIN_FIELD_OF_VIEW = 10;
+export const MAX_TILT_ANGLE = 89;
 // All objects in fallback tiles are reduced by this amount.
 export const FALLBACK_RENDER_ORDER_OFFSET = 20000;
 
@@ -704,7 +705,10 @@ export class MapView extends THREE.EventDispatcher {
     private readonly m_rteCamera = new THREE.PerspectiveCamera();
 
     private m_focalLength: number;
-    private m_lookAtDistance: number;
+    private m_targetDistance: number;
+    private m_targetGeoPos = MapViewDefaults.target.clone();
+    // Focus point world coords may be calculated after setting projection, use dummy value here.
+    private m_targetWorldPos = new THREE.Vector3();
     private readonly m_viewRanges: ViewRanges = {
         near: DEFAULT_CAM_NEAR_PLANE,
         far: DEFAULT_CAM_FAR_PLANE,
@@ -944,13 +948,14 @@ export class MapView extends THREE.EventDispatcher {
             DEFAULT_CAM_FAR_PLANE
         );
         this.m_camera.up.set(0, 0, 1);
-        this.m_lookAtDistance = 0;
+        this.projection.projectPoint(this.m_targetGeoPos, this.m_targetWorldPos);
         this.m_focalLength = 0;
         this.m_scene.add(this.m_camera); // ensure the camera is added to the scene.
         this.m_screenProjector = new ScreenProjector(this.m_camera);
 
         // setup camera with initial position
         this.setupCamera(options);
+        this.m_targetDistance = this.m_camera.position.distanceTo(this.m_targetWorldPos);
 
         this.m_movementDetector = new CameraMovementDetector(
             this.m_options.movementThrottleTimeout,
@@ -1482,9 +1487,12 @@ export class MapView extends THREE.EventDispatcher {
     set projection(projection: Projection) {
         // The geo center must be reset when changing the projection, because the
         // camera's position is based on the projected geo center.
-        const target = MapViewUtils.rayCastWorldCoordinates(this, 0, 0);
+        let target = MapViewUtils.getWorldTargetFromCamera(this.camera, this.projection);
         if (target === null) {
-            throw new Error("MapView does not support a view pointing in the void.");
+            logger.warn(
+                "MapView does not support a view pointing in the void, using last focus point."
+            );
+            target = this.worldTarget;
         }
         const targetCoordinates = this.projection.unprojectPoint(target);
         const targetDistance = this.camera.position.distanceTo(target);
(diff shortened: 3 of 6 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1471,11 +1476,29 @@ export class MapView extends THREE.EventDispatcher {
...
+    /**
+     * Get world coordinates of camera focus point.
+     *
+     * @note The focus point coordinates are updated with each camera update so you don't need
+     * to re-calculate it, although if the camera started looking to the void, the last focus
+     * point is stored.
+     *
+     * @returns world coordinates of last camera focus point.
+     */
+    get lookAtPoint(): THREE.Vector3 {
```
ninok: This should be called target. Se also: https://github.com/heremaps/harp.gl/pull/1027

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2438,13 +2464,14 @@ export class MapView extends THREE.EventDispatcher {
 
         const cameraPitch = MapViewUtils.extractAttitude(this, this.m_camera).pitch;
         const cameraPosZ = this.getCameraHeightAboveTerrain(TERRAIN_ZOOM_LEVEL);
+        const zoomLevelDistance = cameraPosZ / Math.cos(Math.min(cameraPitch, Math.PI / 3));
```
ninok: I think this is all messed up. `zoomLevelDistance` should be the distance to the target point, but currently this is not true for globe and also if the tilt is > `60°`.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2438,13 +2464,14 @@ export class MapView extends THREE.EventDispatcher {
...
+        this.m_zoomLevel = MapViewUtils.calculateZoomLevelFromDistance(zoomLevelDistance, this);
+        this.m_fog.update(this, this.m_viewRanges.maximum);
 
         const target = MapViewUtils.rayCastWorldCoordinates(this, 0, 0);
```
ninok: Instead of raycasting we should compute the target as `camera.position + camera.view * zoomLevelDistance' 

That would at least ensure consistence (i.e. distance between target and camera is always equal to `zoomLevelDistance`

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1471,11 +1476,29 @@ export class MapView extends THREE.EventDispatcher {
 
     /**
      * Get distance from camera to the point of focus in world units.
+     *
+     * @note If camera does not point to any ground anymore the last focus point distance is
+     * then returned.
+     *
+     * @returns Last known focus point distance.
      */
     get lookAtDistance(): number {
```
ninok: Why do we even expose this? I think we should only expose zoom

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2438,13 +2464,14 @@ export class MapView extends THREE.EventDispatcher {
...
+        this.m_zoomLevel = MapViewUtils.calculateZoomLevelFromDistance(zoomLevelDistance, this);
+        this.m_fog.update(this, this.m_viewRanges.maximum);
 
         const target = MapViewUtils.rayCastWorldCoordinates(this, 0, 0);
         if (target !== null) {
-            this.m_lookAtDistance = target.sub(this.camera.position).length();
-            const zoomLevelDistance = cameraPosZ / Math.cos(Math.min(cameraPitch, Math.PI / 3));
-            this.m_zoomLevel = MapViewUtils.calculateZoomLevelFromDistance(zoomLevelDistance, this);
-            this.m_fog.update(this, this.m_viewRanges.maximum);
+            this.m_lookAtPoint.copy(target);
```
ninok: we should have a better fallback instead of taking the last frames target. This will lead to non-deterministic behavior.

@here/harp-mapview/lib/MapView.ts
kkostecki: Ok.

@here/harp-mapview/lib/MapView.ts
kkostecki: This allows to keep the same pivot point for camera rotations. Imagine that we have focused on some point on globe (let's say north pole), when we are rotating camera around this point (pivot point) we could do it until ray is reaching the pole ground (earth surface) after cam angle (tilt) exceeds the 90 degrees ray hits the outer space, so there is not reference pivot point to rotate around. The proposed behavior would keep that last pivot and keep rotation around it.... Of course maybe there is something better then this... Here I am just trying to keep old controls implementation without re-writing entire code.
Any suggestions appreciated.

@here/harp-mapview/lib/MapView.ts
kkostecki: This value is there for a long time, it allows for distance scaling computations, used in TextElementsRenderer, allows to mimic perspective for 2d (screen space) rendered objects.

@here/harp-mapview/lib/MapView.ts
kkostecki: > Instead of raycasting we should compute the target as `camera.position + camera.view * zoomLevelDistance'
> 
> That would at least ensure consistence (i.e. distance between target and camera is always equal to `zoomLevelDistance`

I am afraid that this is not true for globe projection, this simple trigonometry can not be applied here,
cos(pitch) = altitude / zoomLevelDistance,
zoomLevelDistance = altitude / cos(pitch)
This comes from right (angle) triangle formed by the camera position, camera position projected on the ground and target point on the ground.
Unfortunately when we consider globe geometry, we do not have flat ground anymore, so by using this simplification our target point (rotation pivot) will be floating above the ground.
I am not sure if we want this, aren't you?

@here/harp-mapview/lib/MapView.ts
ninok: Okay, but then we can pass it into the TextElementsRenderer. No need to have a public API for it.
I know it's not your code but we should remove it before people start using it and we never get rid of it.

@here/harp-mapview/lib/MapView.ts
ninok: Yeah, I was also thinking about that and wondering if we can have a similar formula for globe that would not rely on raycasting but just the camera position and camera pitch.
In that case we could just clamp the pitch and always compute the distance and with that distance we could compute "some" target point that might not match the real target point.

That would happen whenever the raycasting would also fail but that's fine b/c you can only get into this situation when you modify the camera directly. We just have to make sure that we can still compute everything we need internally (e.g. zoomlevel, near/far plane, ...)

@here/harp-mapview/lib/MapView.ts
ninok: I tried this locally:
```suggestion
        let target = null;
        if (this.projection.type === ProjectionType.Spherical) {
            target = MapViewUtils.rayCastWorldCoordinates(this, 0, 0);
        }

        if (target !== null) {
            this.m_lookAtDistance = target.sub(this.camera.position).length();
        } else {
            const cameraPosZ = this.getCameraHeightAboveTerrain(TERRAIN_ZOOM_LEVEL);
            const cameraPitch = MapViewUtils.extractAttitude(this, this.m_camera).pitch;
            const limitedPitch = Math.min(cameraPitch, 1.553343034); //Limit pitch to 89°
            this.m_lookAtDistance = cameraPosZ / Math.cos(limitedPitch);
        }

        this.m_zoomLevel = MapViewUtils.calculateZoomLevelFromDistance(this.m_lookAtDistance, this);
        this.m_fog.update(this, this.m_viewRanges.maximum);
```
So you would always rely on raycasting for globe but never for flat. Not sure how it will behave when exiting or entering the raycastable surface.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -33,6 +33,8 @@ const groundNormalPlanarProj = new THREE.Vector3(0, 0, 1);
 const groundPlane = new THREE.Plane(groundNormalPlanarProj.clone());
 const groundSphere = new THREE.Sphere(undefined, EarthConstants.EQUATORIAL_RADIUS);
 const rayCaster = new THREE.Raycaster();
+const maxTiltAngleAllowed = Math.PI / 2;
```
ninok: There is already a hardcoded value of 89 in MapView.lookAt. Please move this constant to a central place and reuse it.

@here/harp-mapview/lib/MapView.ts
kkostecki: Ohhh yes, that is all true, see also all my comments below (I started responding bottom-up).
- for globe we could improve the calculations the have exact distance to point on the ground (this it actually `lookAtDistance`.
- as for 60 deg constraint, I guess that it was provided to limit effect of vanishing (target) point, when zoom level dist goes to infinity when camera pitches close to 90 degrees.  
So having said all that here and below. What would you suggest?
I have a feeling that there are two possible approaches:
- we need to "cache" some of the params of the equation (last target/pivot point, or its last distance), when camera rotates and looking to the void,
- we could calculate "proper" target point on the ground (as it is now, but without raycasting) until the camera "ray" closes to horizon, then for globe it is limited by sphere tangent (furthest point visible) and for planar projections we should constraint it with some angle (89,9), because at 90 degrees our ray goes to infinity.
Do you have any other ideas?

@here/harp-mapview/lib/MapView.ts
kkostecki: That may be quite problematic, because we need to pass it every frame, do you want me to prepare special setter for it (in TextElementsRenderer), it takes all up-to-date MapView params via MapViewState, and this state is not frame persistent, but rather takes everything on demand from MapView, so it is kind of breaking entire interface/data-flow for single param. Please check it out, if it is worth doing. Maybe we could just @hide it in MapView interface?

@here/harp-mapview/lib/Utils.ts
kkostecki: Ok, saw it, although it is slight different :).

@here/harp-mapview/lib/MapView.ts
kkostecki: So I have similar idea, with that difference that when rendering globe, our `limitedPitch` would be the the pitch to sphere tangent, so in other words, look at distance would be the distance to the furthest point on horizon (tangent point).
Although I would provide those changes progressively, splitting improvements into 2 commits, first with just raycasting simplification - the current one and the second that introduces new calculus. This way it will we easier to revert in case of regression.

MPursche:
I think there are still some caveats. I can break the `MapControls`:
1. Set `mapControls.maxTiltAngle = 90` in globe example.
2. Turn the camera tilt to 90 degrees(right mouse button)
3. Use the zoom wheel to zoom in and out.

**Expected:**
Camera moves forward or backward to zoom in on target position
**Actual:**
Camera moves up and down relative to earth surface.

Another issue is that changing the camera tilt with the middle mouse button is not working reliably for me. Most of the times the camera points just straight to the ground (tilt=0) no matter what the initial tilt was and the azimuth is changed instead.

kkostecki:

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1507,9 +1513,27 @@ export class MapView extends THREE.EventDispatcher {
...
+     *
+     * @note The focus point coordinates are updated with each camera update so you don't need
+     * to re-calculate it, although if the camera started looking to the void, the last focus
+     * point is stored.
+     *
+     * @returns world coordinates of last camera focus point.
      */
-    get lookAtDistance(): number {
-        return this.m_lookAtDistance;
+    get targetPoint(): THREE.Vector3 {
```
ninok: Maybe call it targetWorld to make more clear that it is in world space

ninok: Please also add a getter for the target in geo space as agreed here:
https://confluence.in.here.com/display/HARP/Consistent+naming+of+camera+properties

ninok: There is also big overlap with this PR: https://github.com/heremaps/harp.gl/pull/1023

ninok: I also created a draft PR that might help to continue the discussion: https://github.com/heremaps/harp.gl/pull/1178

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -846,7 +838,7 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
         const r = EarthConstants.EQUATORIAL_RADIUS;
         const d = cameraToOrigin.length();
         let farPlane: number;
-        let farMax = mapView.lookAtDistance * this.farMaxRatio;
+        let farMax = mapView.targetDistance * this.farMaxRatio;
```
ninok: I agree that `targetDistance` is a better name but I would stick to lookAtDistance to not break the API

@here/harp-mapview/lib/Utils.ts
```diff
@@ -174,12 +169,9 @@ export namespace MapViewUtils {
         mapView: MapView,
         deltaAzimuthDeg: number,
         deltaTiltDeg: number,
-        maxTiltAngleRad = Math.PI / 2
+        maxTiltAngleRad = maxTiltAngleAllowed
```
ninok: Shouldn't this be < 90 degrees. If we allow 90 degrees we will get div. by 0 errors

@here/harp-mapview/lib/Utils.ts
```diff
@@ -201,6 +193,59 @@ export namespace MapViewUtils {
...
+        return null;
+    }
+
+    /**
+     * Calculate target (focus) point world coordinates for given camera position and orientation.
+     * @param camera The camera looking on target point.
+     * @param projection The geo-projection used.
+     * @param elevation Optional elevation above (or below) sea level in world units.
+     */
+    export function getTargetPositionFromCamera(
```
ninok: I would rather name the two methods:
- getGeoTargetFromCamera
- getWorldTargetFromCamera

@here/harp-mapview/lib/Utils.ts
```diff
@@ -201,6 +193,59 @@ export namespace MapViewUtils {
...
+            groundSphere.radius += elevation;
+        }
+        const targetWorldPos = new THREE.Vector3();
+        const result =
+            projection.type === ProjectionType.Planar
+                ? rayCaster.ray.intersectPlane(groundPlane, targetWorldPos)
+                : rayCaster.ray.intersectSphere(groundSphere, targetWorldPos);
+        if (elevation !== undefined) {
+            groundPlane.constant = 0;
+            groundSphere.radius -= elevation;
```
ninok: Maybe a bit paranoid, but due to floating point precision resetting it to the original value would be better.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -486,6 +534,44 @@ export namespace MapViewUtils {
         );
     }
 
+    /**
+     * Extracts current camera tilt angle in radians.
+     *
+     * @param mapView The [[MapView]] in use.
+     */
+    export function extractCameraTilt(mapView: MapView): number {
+        const { camera, projection } = mapView;
```
ninok: Why passing whole mapview?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -486,6 +534,44 @@ export namespace MapViewUtils {
...
+        // at every point on the ground (ignoring terrain fluctuations), so we may
+        // simply use inverted ground normal for tilt calculation. This simplifies
+        // the more generic calculus used for spherical projections.
+        if (projection.type === ProjectionType.Planar) {
+            const lookAt: THREE.Vector3 = camera.getWorldDirection(cache.vector3[0]).normalize();
+            const normal: THREE.Vector3 = projection
+                .surfaceNormal(camera.position, cache.vector3[1])
+                .negate();
+            const cosTheta = lookAt.dot(normal);
+            return Math.acos(THREE.Math.clamp(cosTheta, -1, 1));
```
ninok: Why this clamping? The cosine is by definition between -1 and 1.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -603,6 +687,46 @@ export namespace MapViewUtils {
...
+     * Note: this method can be used to get the direction that an object points to, when `location`
+     * is the target of that object, by adding PI to it. Otherwise it only returns the tilt angle
+     * (in radians) of `object` in the tangent space of `location`.
+     *
+     * @param mapView The [[MapView]] instance to consider.
+     * @param object The object to get the coordinates from.
+     * @param location The reference point.
+     */
+    export function extractTiltAngleFromLocation(
+        mapView: MapView,
```
ninok: don't pass whole MapView but only what you need

@here/harp-mapview/lib/text/MapViewState.ts
```diff
@@ -33,7 +33,7 @@ export class MapViewState implements ViewState {
         return this.m_mapView.frameNumber;
     }
     get lookAtDistance(): number {
-        return this.m_mapView.lookAtDistance;
+        return this.m_mapView.targetDistance;
```
ninok: As mentioned above: Don't rename 'lookAtDistance` for now.
I was also thinking to reverse the logic of the MapViewState, i.e.
MapViewState should be a (private) property of MapView and could contain more stuff that is needed internally only. The MapView getters would then just forward the MapViewState properties.

This would allow us to store more properties and pass them around w/o polluting the public MapView API. lookAtDistance is a good example of a property that should have never been exposed b/c it is for internal use only.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -201,6 +193,59 @@ export namespace MapViewUtils {
...
+        const cameraPos = cache.vector3[0].copy(camera.position);
+        const cameraLookAt = camera.getWorldDirection(cache.vector3[1]);
+        rayCaster.set(cameraPos, cameraLookAt);
+        if (elevation !== undefined) {
+            groundPlane.constant = -elevation;
+            groundSphere.radius += elevation;
+        }
+        const targetWorldPos = new THREE.Vector3();
+        const result =
+            projection.type === ProjectionType.Planar
```
ninok: Maybe check my PR: https://github.com/heremaps/harp.gl/pull/1178/files#diff-ab66d6ee9a32de1a17a5622106fd3b08R2432-R2522

I compute and cache the targetWorld in MapView.updateCameras. I think this you should also do since the target point is used in many places

@here/harp-mapview/lib/Utils.ts
MPursche: Probably because of floating point precision issues. I ran into the same problem where the length of the normalized vector was very slightly > 1. Which caused the dot product to be out of  [-1,1] range.

@here/harp-mapview/lib/Utils.ts
kkostecki: Fixed.

@here/harp-mapview/lib/Utils.ts
kkostecki: @MPursche  -> true!

@here/harp-mapview/lib/MapView.ts
kkostecki: Ok, I have provided the already used convention for getters:
- geoCenter / worldCenter
- geoTarget / worldTarget.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: I have checked its only used internally and I have marked it as (@internal) too. I think we need to rename it ASAP before someone else will propagate this stuff.

@here/harp-mapview/lib/Utils.ts
kkostecki: It is smaller then 90 degrees, actually set in MapView (const MAX_TILT_ANGLE) as 89 deg.

@here/harp-mapview/lib/Utils.ts
kkostecki: Done.

@here/harp-mapview/lib/Utils.ts
kkostecki: I don't like it too... but look at whole Utils.ts, almost everything there takes MapView, so I kindly adapted to _standard_ :). I changed mine stuff, but please don't ask me to fix all those methods :).

@here/harp-mapview/lib/Utils.ts
kkostecki: As said before, I have changed/fixed only mine methods to use the minimum required input set, but rest of the methods left untouched.

@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -355,34 +355,34 @@ export class MapControls extends THREE.EventDispatcher {
...
         if (this.mapView.projection.type === geoUtils.ProjectionType.Planar) {
-            this.camera.position.z += this.m_currentViewDirection.z;
-            this.camera.position.z = Math.max(
+            const distance = THREE.Math.clamp(
+                this.camera.position.z + this.m_currentViewDirection.z,
                 minDistance,
-                Math.min(maxDistance, this.camera.position.z)
+                maxDistance
             );
+            this.camera.position.z = distance;
```
ninok: I don't get what this method is doing. The name says it is moving the position along the view direction  but I don't see this happening(i.e. it only takes the z of the view direction).

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1505,11 +1516,40 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_focalLength;
     }
 
-    /**
+    /** @internal
      * Get distance from camera to the point of focus in world units.
+     *
+     * @note If camera does not point to any ground anymore the last focus point distance is
```
ninok: I still don't like this b/c it introduces in-deterministic behavior but we will address it in the follow up ticket

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1505,11 +1516,40 @@ export class MapView extends THREE.EventDispatcher {
...
+    /** @internal
+     * Get geo coordinates of camera focus point.
+     *
+     * @see targetWorldPos
+     *
+     * @returns geo coordinates of the camera focus point.
      */
-    get lookAtDistance(): number {
-        return this.m_lookAtDistance;
+    get geoTarget(): GeoCoordinates {
```
ninok: According to the naming discussion just call it `target`

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2956,7 +3002,7 @@ export class MapView extends THREE.EventDispatcher {
             this.m_camera.lookAt(this.scene.position);
         }
 
-        this.m_lookAtDistance = defaultGeoCenter.altitude!;
+        this.m_targetDistance = defaultGeoCenter.altitude!;
```
ninok: I think the usage of `MapViewDefaults.geoCenter` is a bit inconsistent. Here it is used as camera position but above it is used to initialize the camera target:
```
    private m_targetGeoPos = new GeoCoordinates(
        MapViewDefaults.geoCenter.latitude,
        MapViewDefaults.geoCenter.longitude,
        0
    );
```

@here/harp-mapview/lib/MapView.ts
kkostecki: Ok. thanx!

@here/harp-mapview/lib/MapView.ts
kkostecki: Ok, will move this discussion to follow up ticket, where we will implement the target point and zoom level distance proper relation as also free camera mode.
See HARP-8371.

kkostecki:
retest this please

kkostecki:

kkostecki:

PR #1140, opened 2019-12-11
HARP-7373: Point label persistence across zoom levels.

atomicsulfate:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementStateCache.ts
+++ b/@here/harp-mapview/lib/text/TextElementStateCache.ts
@@ -11,11 +11,32 @@ import { TextElementState } from "./TextElementState";
 import { TextElementType } from "./TextElementType";
 
 /**
- * Label distance threshold squared in meters. Point labels with the same name that are closer in
- * world space than this value are treated as the same label. Valid within a single datasource.
- * Used to identify duplicate labels in overlapping tiles.
+ * Label distance tolerance squared in meters. Point labels with the same name that are closer in
+ * world space than this value are treated as the same label. Used to identify duplicate labels in
+ * overlapping tiles and label replacements at different storage levels.
  */
-const MAX_LABEL_DUPLICATE_DIST_SQ = 100;
+function getDedupSqDistTolerance(zoomLevel: number) {
+    // Defining here a minimum tolerance of 10m at zoom level 13 or higher.
+    const minSqTol = 100;
+    const minSqTolLevel = 13;
+    const maxLevelDelta = 4;
+    const levelDelta = Math.min(
+        maxLevelDelta,
+        minSqTolLevel - Math.min(minSqTolLevel, Math.floor(zoomLevel))
+    );
+    // Distance tolerance computed applying a factor over an arbitrary minimum tolerance for a
+    // chosen zoom level. The factor is an exponential function on zoom level delta wrt minimum
+    // tolerance zoom level.
+    // error = sqrt(sqError) = sqrt(minSqError* 2^(4d)) = minError*2^(2d)
+
+    //tslint:disable-next-line: no-bitwise
+    return minSqTol << (levelDelta << 2);
+}
+
+const tmpCachedDuplicate: { entries: TextElementState[]; index: number } = {
+    entries: [],
+    index: -1
+};
 
 /**
  * Caches the state of text element groups currently rendered as well as the text element states
@@ -78,26 +99,46 @@ export class TextElementStateCache {
     /**
      * Updates state of all cached groups, discarding those that are not needed anymore.
      * @param time The current time.
-     * @param clearVisited `True` to clear visited flag in group states.
      * @param disableFading `True` if fading is currently disabled, `false` otherwise.
+     * @param findReplacements `True` to replace each visible unvisited text element with a
+     * visited duplicate.
+     * @param zoomLevel Current zoom level.
      * @returns `True` if any textElementGroup was evicted from cache, false otherwise.
      */
-    update(time: number, clearVisited: boolean, disableFading: boolean) {
+    update(time: number, disableFading: boolean, findReplacements: boolean, zoomLevel: number) {
+        const replaceCallback = findReplacements
+            ? this.replaceElement.bind(this, zoomLevel)
+            : undefined;
+
         let anyEviction = false;
         for (const [key, groupState] of this.m_referenceMap.entries()) {
-            const visible = groupState.updateFading(time, disableFading);
+            const visible = groupState.updateFading(
+                time,
+                disableFading,
+                groupState.visited ? undefined : replaceCallback
+            );
             const keep: boolean = visible || groupState.visited;
 
             if (!keep) {
                 this.m_referenceMap.delete(key);
                 this.m_sortedGroupStates = undefined;
                 anyEviction = true;
-            } else if (clearVisited) {
-                groupState.visited = false;
             }
         }
-        this.m_textMap.clear();
         return anyEviction;
+    }
+
+    /**
+     * Clears visited state for all text element groups in cache.
+     */
+    clearVisited() {
+        for (const groupState of this.m_referenceMap.values()) {
+            groupState.visited = false;
+        }
+    }
+
+    clearTextCache() {
+        this.m_textMap.clear();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


@here/harp-mapview/lib/text/TextElementStateCache.ts
```diff
@@ -82,24 +103,42 @@ export class TextElementStateCache {
      * @param disableFading `True` if fading is currently disabled, `false` otherwise.
      * @returns `True` if any textElementGroup was evicted from cache, false otherwise.
      */
-    update(time: number, clearVisited: boolean, disableFading: boolean) {
+    update(time: number, disableFading: boolean, findReplacements: boolean, zoomLevel: number) {
```
nzjony: please add the extra params to the documentation

@here/harp-mapview/lib/text/TextElementStateCache.ts
```diff
@@ -194,4 +233,43 @@ export class TextElementStateCache {
...
+        }
+
+        const maxSqDistError = getDedupSqDistTolerance(zoomLevel);
+
+        // Other labels found with the same text. Check if they're near enough to be considered
+        // duplicates.
+        const entryCount = cachedEntries.length;
+        const elementPosition = element.points as THREE.Vector3;
+        let duplicateIndex: number = -1;
+        for (let i = 0; i < entryCount; ++i) {
```
nzjony: how do you prevent this checking against itself? as far as i can read, the TextElementState is part of this cache, so it is possible that the element is checked to see if it is a duplicate with itself?

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -871,7 +872,7 @@ export class TextElementsRenderer {
                 ++this.m_initializedTextElementCount;
             }
         }
-        return textElement.loadingState === LoadingState.Initialized;
+        return textElement.glyphs !== undefined;
```
nzjony: why has this changed?

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1687,7 +1692,12 @@ export class TextElementsRenderer {
             const textNeedsDraw = !textRejected || textRenderState!.isFading();
 
             if (textNeedsDraw) {
-                if (pointLabel.textReservesSpace) {
+                // Don't allocate space for rejected text. When zooming, this allows placement of a
+                // lower priority text element that was displaced by a higher priority one (not
+                // present in the new zoom level) before an even lower priority one takes the space.
+                // Otherwise the lowest priority text will fade in and back out.
+                // TODO: Add a unit test for this scenario.
+                if (pointLabel.textReservesSpace && !textRejected) {
```
nzjony: is this related?

@here/harp-mapview/lib/text/TextElementStateCache.ts
```diff
@@ -194,4 +233,43 @@ export class TextElementStateCache {
         this.m_referenceMap.set(textElementGroup, textElementGroupState);
         this.m_sortedGroupStates = undefined;
     }
+
+    private findDuplicate(
+        elementState: TextElementState,
+        zoomLevel: number
+    ): { entry: TextElementState[]; index: number } | undefined {
```
nzjony: should entry be "entries"

@here/harp-mapview/lib/text/TextElementStateCache.ts
```diff
@@ -115,61 +154,61 @@ export class TextElementStateCache {
...
-
+            cacheResult.entry[cacheResult.index] = elementState;
             cachedDuplicate.reset();
             return true;
         }
 
         return false;
     }
 
+    replaceElement(zoomLevel: number, elementState: TextElementState): void {
```
nzjony: please add some documentation to this function.

@here/harp-mapview/lib/text/TextElementStateCache.ts
nzjony: would it make sense to also return the duplicate? I notice there are a few cases where it writes: cacheResult.entry[cacheResult.index]; wouldn't it be a bit cleaner / simpler to just return this item?

@here/harp-mapview/lib/text/TextElementStateCache.ts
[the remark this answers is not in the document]
atomicsulfate: done.

@here/harp-mapview/lib/text/TextElementStateCache.ts
```diff
@@ -82,24 +103,42 @@ export class TextElementStateCache {
      * @param disableFading `True` if fading is currently disabled, `false` otherwise.
```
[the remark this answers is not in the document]
atomicsulfate: done

@here/harp-mapview/lib/text/TextElementStateCache.ts
atomicsulfate: The duplicate cache is cleared after every frame and refilled again in a loop that iterates once through  every visited label. See the call to `deduplicateElement` in `prepareTextElementGroup`. So every visited label is added at most once per frame.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: This is not related to your other comment above. This is a minor improvement that avoids flickering in particular scenario:
Imagine we are look at "EUROPE" label, then we zoom in to a level where continent labels are disabled and capitals become visible. "London" starts fading in to the west side (it doesn't collide with "EUROPE") as "EUROPE" keeps fading out. As soon as "EUROPE" becomes invisible, the label "DEUTSCHLAND" that was in collision both with "EUROPE" and "London" starts fading in, since it has less priority than "EUROPE" but more than "London", and causes the rejection of "London" that starts fading out.
If we let "DEUTSCHLAND" fade in as soon as "EUROPE" starts fading out, then "London" will never fade in in the first place.

@here/harp-mapview/lib/text/TextElementStateCache.ts
atomicsulfate: done

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: When a label is replaced, the new label must inherit temporally the glyphs of the old label while its own glyphs are loaded asynchronously. Otherwise the replacement won't be rendered in the meantime. Nevertheless, `initializeGlyphs` will load the final glyphs for a label replacement and replace the inherited ones as soon as they're loaded.
So `initializeGlyphs` must return true as soon as a label has some glyphs assigned (either the final ones or the temporally inherited).

atomicsulfate: I'll document this in the code.

@here/harp-mapview/lib/text/TextElementStateCache.ts
atomicsulfate: returning only  cacheResult.entry[cacheResult.index] won't be enough cause sometimes the caller needs to overwrite that particular entry, i.e, it needs to do:
cacheResult.entry[cacheResult.index] = newLabel;

We could return the three things: the entries array, the index and the content of the entry addressed by the index, then the object we use for the result would look like this:
`const tmpCachedDuplicate: { entries: TextElementState[]; index: number; entry:TextElementState|undefined } = {
    entries: [],
    index: -1,
    entry:undefined
};`
But then it's confusing because there's more than one way to get the same stuff, you can get `entry` also by doing `entries[index]`. Also, someone might feel tempted to do `entry = newElementState` to replace the content of an entry, which wouldn't work. And we can't even set `entry` to `readonly` cause the object is reused on every call.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: There some comment in `TextElement.replace()`, but I'll add more in `initializeGlyphs`

@here/harp-mapview/lib/text/TextElementStateCache.ts
atomicsulfate: done

@here/harp-mapview/lib/text/TextElementsRenderer.ts
nzjony: Ok, please move to a separate PR, if there is any regression, it would be easier to revert this single line than have to revert everything.

@here/harp-mapview/lib/text/TextElementStateCache.ts
nzjony: thanks for taking a look, I understand.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: I've moved it to a separate commit, so that it could reverted if necessary as you say.
But I left it in the same PR, i won't squash it when merging. The change is related to zoom level persistence since the scenario I described is reproducible as soon as zoom level persistence is introduced.

nzjony: good idea, yes it can be a separate commit, cool!

nzjony: wait, separate commit is cool, but please merge the other two commits.

atomicsulfate: will do

PR #1144, opened 2019-12-12
HARP-8188: Use custom blending for solid lines.

robertoraggi:
This change removes the special code that overrides the `transparent`
property of the material before rendering solid lines.


ninok:
Obsolete with #1150

PR #1162, opened 2019-12-18
Hash JS bundles in harp-examples.

zbigg:
Add content-based hashes JS bundles created from examples for
better cooperation with caching in CDNs.

### DIFF
--- a/@here/harp-examples/codebrowser.html
+++ b/@here/harp-examples/codebrowser.html
@@ -9,7 +9,6 @@
 
 <body>
     <pre id="code"></pre>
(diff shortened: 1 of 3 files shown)


zbigg:
> Why do we do it on this level? I would rather publish to a new folder every time (like we do for the releases already)
> That would also allow us to check older versions of master and find regressions easier.

Because actual examples are served from stable URL

   https://www.harp.gl/docs/master/examples/#getting-started_free-camera.html

I don't think, we should have any hashes here.
If we want to compare several versions, then we have means for this (selector in lower section of harp.gl site) - nothing changes in that matter.

PR #1167, opened 2019-12-20
HARP-7551: Updates examples to use our webpack utils

CraigBerlin:
- Convert helper script to javascript to greatly simplify CI
    - Allow users to specify own entry point names
    - Do not hardcode location of map themes
    - Do not override user typescript module rules


@here/harp-webpack-utils/scripts/HarpWebpackConfig.js
```diff
@@ -96,11 +110,14 @@ export function addHarpWebpackConfig(config?: Configuration, harpConfig?: HarpWe
     return bundles;
```
zbigg: Return bundle[0] if there is only one bundle, so you don't have to resort to
```
addHarpConfig(...)[0];
```
later.

@here/harp-webpack-utils/scripts/HarpWebpackConfig.js
```diff
@@ -4,29 +4,45 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+/*
```
zbigg: We shouldn't abandon TS.

It's easy to include `.*ts` in our webpack.configs, just like [here](https://github.com/heremaps/harp.gl/blob/master/webpack.tests.config.js#L112)

@here/harp-webpack-utils/scripts/HarpWebpackConfig.js
```diff
@@ -4,29 +4,45 @@
...
         throw new Error("config passed to addHarpWebpackConfig must be a Configuration object");
     }
     const userConfig = config !== undefined ? config : {};
     const mode = process.env.NODE_ENV === "production" ? "production" : "development";
+    const devtool = process.env.PREPARE_ONLY === "true" ? undefined : "source-map";
     const mainEntry = harpConfig === undefined ? undefined : harpConfig.mainEntry;
-    const baseConfig: Configuration = {
+    const baseConfig /*: Configuration*/ = {
         output: {
             filename: "[name].bundle.js"
```
zbigg: General comment: we shouldn't overwrite this particular webpack default (no reason for that).

@here/harp-webpack-utils/scripts/HarpWebpackConfig.js
```diff
@@ -35,18 +51,20 @@ export function addHarpWebpackConfig(config?: Configuration, harpConfig?: HarpWe
         },
         mode
     };
-    const typescriptConfig: Configuration = {
+    const typescriptConfig /*: Configuration*/ = {
         resolve: {
             extensions: [".web.ts", ".ts", ".tsx"]
```
zbigg: > `tsx`

Again, harp.gl doesn't require `tsx` anywhere.
(Some of our apps, use it indded, but it's out of scope of this tool).

@here/harp-webpack-utils/scripts/HarpWebpackConfig.js
```diff
@@ -67,26 +85,22 @@ export function addHarpWebpackConfig(config?: Configuration, harpConfig?: HarpWe
...
-                }
-            },
-            bundles[0]
-        );
+        const entry = typeof mainEntry === "string" ? { mapview: mainEntry } : mainEntry;
+        bundles[0] = WebpackMerge.smart({ entry }, bundles[0]);
     }
     if (harpConfig !== undefined && harpConfig.decoderEntry !== undefined) {
-        const decoderConfig = harpConfig.decoderEntry.endsWith(".ts")
+        const decoderConfig = isTypescript(harpConfig.decoderEntry)
```
zbigg: General, `typescript` should be also part of `HarpWebpackConfig` interface, so we not only rely on automagic detection.

@here/harp-examples/webpack.config.js
```diff
@@ -123,31 +103,32 @@ function filterExamples(pattern) {
...
+const exampleBrowserConfig = addHarpWebpackConfig(commonConfig, {
+    mainEntry: {
         "example-browser": "./example-browser.ts"
     }
-});
+})[0];
 
-const codeBrowserConfig = merge(commonConfig, {
-    entry: {
+const codeBrowserConfig = addHarpWebpackConfig(commonConfig, {
```
zbigg: Now that's total misuse.
This entry has nothing to do with `harp` and it created two configs only for second to be abandoned.

I think it shows, that we should have "two" entries, or option to configure if decoder should be created.

It adds CopyWebpackPlugin for`resources` which is completly unnecessary!

Possible examples:

```
const codeBrowserConfig = addHarpWebpackConfig(commonConfig, {
    noDecoder: true,
    noResources: ???,
    noWhatever: ...
})```

Or just expose "harpSourceCommon" config (with typescript, resolve options etc) without "application" level extras like decoder and default resources.

File: @here/harp-webpack-utils/scripts/HarpWebpackConfig.js [hunk 7]
```diff
@@ -96,11 +110,14 @@ export function addHarpWebpackConfig(config?: Configuration, harpConfig?: HarpWe
...
+function createPlugins(htmlTemplate /*?: string*/) /*: Plugin[]*/ {
     const plugins = [
         new CopyWebpackPlugin([
             {
-                from: "node_modules/@here/harp-map-theme/resources",
+                from: path.join(
+                    path.dirname(require.resolve("@here/harp-map-theme/package.json")),
+                    "resources"
+                ),
                 to: "resources",
```
zbigg: General.
Although, we commonly use `resources` in our demo/example apps, here we're in namespace of user application and i think we should create all names in such way, so we avoid potential clashes, so this should be something like
`harp-resources` or just `harp-map-theme`.

@robertoraggi It would be good, if this "default" naming scheme would be discussed with above concern, so defaults for `font://`, `poi://` are aligned with tools like this one.

@here/harp-webpack-utils/scripts/HarpWebpackConfig.js
```diff
@@ -35,18 +51,20 @@ export function addHarpWebpackConfig(config?: Configuration, harpConfig?: HarpWe
...
-    const mainConfig =
-        mainEntry !== undefined && mainEntry.endsWith(".ts")
-            ? WebpackMerge.smart(baseConfig, typescriptConfig)
-            : baseConfig;
+    const mainConfig = isTypescript(mainEntry)
+        ? WebpackMerge.smart(baseConfig, typescriptConfig)
+        : baseConfig;
     const bundles = [
         WebpackMerge.smart(
             {
```
zbigg: General comment.
`stats` is also very specific to application. We're overwriting too much of `webpack`s defaults.

I think only "exclude: /resources/" should remain.

@here/harp-webpack-utils/scripts/HarpWebpackConfig.js
CraigBerlin: Right, I only abandoned TS due to simplify CI / build issues. We now need to have the js for webpack-utils generated first before starting builds. Will look into using prepublish or similar.

CraigBerlin:
> Many comments to usability of `addHarpWebpackConfig`.
> This still needs more work and polishing, especially that giving that this should be Developer Experience "enchancer" - this patch IMO doesn't show readability or simplicity gain vs original boilerplate.

Thanks for review @zbigg,
Looking at this patch again after a couple of weeks I agree. "addHarpWebpackConfig" was originally designed so the harp.gl generator would produce a very clean and simple webpack config for generated example projects.
This patch just tries to force that implementation on our more complex webpack configurations, we should instead redesign "addHarpWebpackConfig" to be more useful for anyone using harp.gl.

CraigBerlin:
Closing this for now, to proceed will need to redesign "addHarpWebpackConfig".

PR #1168, opened 2019-12-20
Added configurable environment to OlpCopyrightProvider.

keaukraine:
Added configurable environment to OlpCopyrightProvider.

Signed-off-by: Oleksandr Popov <ext-oleksandr.popov@here.com>

### DIFF
--- a/@here/harp-olp-utils/lib/OlpCopyrightProvider.ts
+++ b/@here/harp-olp-utils/lib/OlpCopyrightProvider.ts
@@ -7,6 +7,7 @@
 import { AreaCopyrightInfo, CopyrightCoverageProvider } from "@here/harp-mapview";
 import {
     DataRequest,
+    EnvironmentName,
     HRN,
     OlpClientSettings,
     VersionedLayerClient
@@ -44,6 +45,8 @@ export interface OlpCopyrightProviderParams {
      * Version of the catalog
      */
     version: number;
+    /** The name of environment. If omitted, environment from catalog HRN will be used. */
+    environment?: EnvironmentName;
 }
 
 const DEFAULT_LAYER = "copyright";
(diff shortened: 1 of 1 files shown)


@here/harp-olp-utils/lib/OlpCopyrightProvider.ts
```diff
@@ -6,6 +6,7 @@
 
 import { AreaCopyrightInfo, CopyrightCoverageProvider } from "@here/harp-mapview";
 import { CatalogClient, DataStoreContext, HRN } from "@here/olp-sdk-dataservice-read";
+import { EnvironmentName } from "@here/olp-sdk-dataservice-read/dist/@here/olp-sdk-dataservice-read/lib/getEnvLookupUrl";
```
ninok: Is this a temporary hack and this will be in the public API of olp-sdk-dataservice-read at some point?

keaukraine: New version of `olp-sdk-dataservice-read` has a better structure so this can be imported from it's `index.ts`. However, new version of `olp-sdk-dataservice-read` has significantly changed API so the whole `OlpCopyrightProvider` and `OlpDataProvider` should be rewritten to adapt to it. This is a quick fix which uses version specified in `package.json` and doesn't require major refactoring.

zbigg: We've found solution offline:
```
import { ..., DataStoreContextParams, ... } from "@here/olp-sdk-dataservice-read";

export interface OlpCopyrightProviderParams {
    ...
   environment?: DataStoreContextParams["environment"];
}
```

that is both typesafe and uses published interfaces only and most importantly doesn't depend on arbitrary physical layout of 3rdparty package.

zbigg:
Rebuild please

zbigg:
@heremaps-bot ok to test

keaukraine:
@CraigBerlin I've updated code to latest API,

PR #1169, opened 2019-12-23
Styling tests for RGBA support

zbigg:
Map styling tests (GeoJson + OmvDataSource + MapView) of following
techniques with focus on transparency.

 - text
 - solid-line
 - fill
 - standard
 - extruded-polygon


@here/harp-materials/lib/Utils.ts
```diff
@@ -38,12 +45,26 @@ ${tabChar}#include <${insertedShaderName}>`
 
  * @param material `Material` that should use blending
  */
-export function enforceBlending(material: THREE.Material | THREE.ShaderMaterialParameters) {
+export function enforceBlending(
+    material: (THREE.Material | THREE.ShaderMaterialParameters) & ForcedBlending
+) {
     if (material.transparent) {
```
kkostecki: I would also add here:
```
if (material.transparent && material.forcedBlending !== true)
```

PR #1179, opened 2020-01-02
Fix z-fighting of animated buildings with 0 height.

zbigg:
Don't animate appearance of extruded buildings with height near 0 (or very small)
to prevent flickering caused by float precision issues when panning & overlapping tiles

Fixes:
* flickering of whole building area of appearing building with height ~0
* most of flickering of said buildings on tile borders

At expense of _not animating_ exposure or height of those buildings at all.

Fallback tiles still cause flickering from time to time.

Related-to: HARP-8279

### DIFF
--- a/@here/harp-materials/lib/MapMeshMaterialsDefs.ts
+++ b/@here/harp-materials/lib/MapMeshMaterialsDefs.ts
@@ -1,0 +1,23 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export namespace ExtrusionFeatureDefs {
+    /**
+     * Minimum ratio value for extrusion effect
+     */
+    export const DEFAULT_RATIO_MIN: number = 0.0;
+    /**
+     * Maximum ratio value for extrusion effect
+     */
+    export const DEFAULT_RATIO_MAX: number = 1;
+
+    /**
+     * Buildings height used whenever no height-data is present or height is very small.
+     *
+     * Used to avoid z-fighting between ground plane and building.
+     */
+    export const MIN_BUILDING_HEIGHT = 0.01;
+}
--- a/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
+++ b/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
@@ -4,15 +4,27 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { ExtrusionFeatureDefs } from "../MapMeshMaterialsDefs";
+
+const MIN_BUILDING_HEIGHT_SQUARED =
+    ExtrusionFeatureDefs.MIN_BUILDING_HEIGHT * ExtrusionFeatureDefs.MIN_BUILDING_HEIGHT;
+
 export default {
     extrusion_pars_vertex: `
 // Extrusion axis (xyz: vector, w: factor).
 attribute vec4 extrusionAxis;
 uniform float extrusionRatio;
 varying vec4 vExtrusionAxis;
+varying float vExtrusionRatio;
+
 `,
     extrusion_vertex: `
-transformed = transformed + extrusionAxis.xyz * (extrusionRatio - 1.0);
+// Cancel extrusionRatio (meaning, force to 1) if extrusionAxisLen < MIN_BUILDING_HEIGHT.
+const float MIN_BUILDING_HEIGHT_SQUARED = ${MIN_BUILDING_HEIGHT_SQUARED};
+float extrusionAxisLenSquared = dot(extrusionAxis.xyz, extrusionAxis.xyz);
+vExtrusionRatio = max(1.0 - step(extrusionAxisLenSquared, MIN_BUILDING_HEIGHT_SQUARED), extrusionRatio);
+
+transformed = transformed + extrusionAxis.xyz * (vExtrusionRatio - 1.0);
 vExtrusionAxis = vec4(normalMatrix * extrusionAxis.xyz, extrusionAxis.w);
 `,
     // Modified version of THREE <normal_fragment_begin> shader chunk which, for flat shaded
@@ -50,10 +62,10 @@ vExtrusionAxis = vec4(normalMatrix * extrusionAxis.xyz, extrusionAxis.w);
 vec3 geometryNormal = normal;
 `,
     extrusion_pars_fragment: `
-uniform float extrusionRatio;
+varying float vExtrusionRatio;
 varying vec4 vExtrusionAxis;
 `,
     extrusion_fragment: `
-gl_FragColor.a *= smoothstep( 0.0, 0.25, extrusionRatio );
+gl_FragColor.a *= smoothstep( 0.0, 0.25, vExtrusionRatio );
 `
 };
--- a/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
+++ b/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
@@ -75,6 +75,7 @@ import {
 } from "@here/harp-datasource-protocol/lib/TechniqueAttr";
 // tslint:disable-next-line:max-line-length
 import { SphericalGeometrySubdivisionModifier } from "@here/harp-geometry/lib/SphericalGeometrySubdivisionModifier";
+import { ExtrusionFeatureDefs } from "@here/harp-materials/lib/MapMeshMaterialsDefs";
 
 const logger = LoggerManager.instance.create("OmvDecodedTileEmitter");
 
@@ -99,12 +100,6 @@ const tmpPointC = new THREE.Vector3();
 const tmpPointD = new THREE.Vector3();
 const tmpPointE = new THREE.Vector3();
 const tmpLine = new THREE.Line3();
-
-/**
- * Height buildings take whenever no height-data is present. Used to avoid z-fighting and other
- * graphics artifacts.
- */
-const MIN_BUILDING_HEIGHT = 1e-18;
 
 /**
  * Minimum number of pixels per character. Used during estimation if there is enough screen space
@@ -1042,9 +1037,7 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
 
         // Prevent that extruded buildings are completely flat (can introduce errors in normal
         // computation and extrusion).
-        if (height === floorHeight) {
-            height += MIN_BUILDING_HEIGHT;
-        }
+        height = Math.max(floorHeight + ExtrusionFeatureDefs.MIN_BUILDING_HEIGHT, height);
 
         const styleSetConstantHeight = getOptionValue(
             extrudedPolygonTechnique.constantHeight,
--- a/@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
+++ b/@here/harp-mapview/lib/AnimatedExtrusionHandler.ts
@@ -5,7 +5,7 @@
  */
 
 import { TileKey } from "@here/harp-geoutils";
-import { ExtrusionFeature } from "@here/harp-materials";
+import { ExtrusionFeature, ExtrusionFeatureDefs } from "@here/harp-materials";
 import { MathUtils } from "@here/harp-utils";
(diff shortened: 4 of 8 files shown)


@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -4,15 +4,22 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+const MIN_BUILDING_HEIGHT_SQUARED = 0.001 * 0.001;
```
ninok: Why not use MIN_BUILDING_HEIGHT?

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -75,6 +75,7 @@ import {
 } from "@here/harp-datasource-protocol/lib/TechniqueAttr";
 // tslint:disable-next-line:max-line-length
 import { SphericalGeometrySubdivisionModifier } from "@here/harp-geometry/lib/SphericalGeometrySubdivisionModifier";
+import { ExtrusionFeature } from "@here/harp-materials/lib/MapMeshMaterials";
```
ninok: Can we move it somewhere else to avoid the harp-material dependency?

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
zbigg: Circular dependencies :/
This is first iteration, wil ltry to get proper place for definitions like these - it will end-up with "MaterialDefs.ts" or something that will contain only constants shared by decoders and renderer part.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
zbigg: Thanks for raising this, i also would like to have good place.
There is no good place in current architecture, however these are propositions
1) `harp-materials` - , because this feature and constant is actually property (limitation) of material ...still, this is more or less ok
2) `harp-datasource-protocol` ... .well, it's kitchen sink already ... but currently stands as "mapview-decoder" protocol, so also good place for all definitions shared by render&decode areas
3) `harp-omv-datasource` already depends on "materials" due to transitive dependency on `harp-lines`, so this is only minor harm ...

zbigg: OTOH, if we would given, we _already_ have dependency
harp-datasource-protocol -> harp-materials

We cannot move `ExtrusionFeature` nor `MIN_BUILDING_HEIGHT`  do `harp-datasource-protocol`, because we would end with loop.

I'll settle with specific `MapMeshMaterialDefs` containing only key constants in `harp-materials`.

ninok: I think the main question we have to answer is, what is the purpose of our packages?
1. Is it to separate decoding and rendering code or
2. to allow standalone usage of certain features like rendering of extruded geometry.

If it is 1, I would vote for `harp-datasource-protocol` b/c as you mentioned already it is the place where we put all shared code.
If it is 2, I would vote for `harp-materials` since a user that just want's to render some extruded geometry does not want to pull the whole `harp-datasource-protocol`. I guess having the `harp-materials` dependency for that usecase is also fine b/c you actually want to render at the end. Unfortunately it breaks our first usecase b/c we pull the whole `harp-materials` into the decoder bundle.

So to conclude: Since we have the `harp-material` dependency already it's fine for now to put it to `harp-materials` but going forward we should clean this up. Maybe we even need another package that contains all the shared stuff between geometry creation and materials. Also the extrusion should be moved out of OmvDecodedTile emitter into `harp-geometry`. And finally looking at the existing `harp-lines <- harp-materials` dependency: I think this should also be cleaned up b/c the material dependency is just there to have a fallback if no material is provided.

zbigg: > Unfortunately it breaks our first usecase b/c we pull the whole harp-materials into the decoder bundle.

If we're specific in imports i.e
> import { Foo} from "@here/harp-materials/lib/FooDefs"

instead of
> import { Foo } from "@here/harp-materials/"

Then we're ok and webpack (and possibly other bundlers) doesn't pick up whole library, but only needed symbols.

zbigg:
retest this please

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
zbigg: Resolved issue by moving `ExtrusionFeature` definitions to separate module, still in `harp-materials`. Now it can be used both in decoder and renderer context.

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -4,15 +4,25 @@
...
 uniform float extrusionRatio;
 varying vec4 vExtrusionAxis;
+varying float vExtrusionRatio;
+
 `,
     extrusion_vertex: `
-transformed = transformed + extrusionAxis.xyz * (extrusionRatio - 1.0);
+const float MIN_BUILDING_HEIGHT_SQUARED = ${MIN_BUILDING_HEIGHT_SQUARED};
+float extrusionAxisLenSquared = dot(extrusionAxis.xyz, extrusionAxis.xyz);
+vExtrusionRatio = mix(1.0, extrusionRatio, float(extrusionAxisLenSquared > MIN_BUILDING_HEIGHT_SQUARED));
```
kkostecki: Using float cast of bool expression with interpolation does not seem to be common and effective way, I would rather consider using OGLS step() function instead. Something like:
`vExtrusionRatio = max(1 - step(extrusionAxisLenSquared, MIN_BUILDING_HEIGHT_SQUARED), extrusionRatio)`
should give you same results but in more gentle manner.

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -50,10 +60,10 @@ vExtrusionAxis = vec4(normalMatrix * extrusionAxis.xyz, extrusionAxis.w);
 vec3 geometryNormal = normal;
 `,
     extrusion_pars_fragment: `
-uniform float extrusionRatio;
+varying float vExtrusionRatio;
```
kkostecki: I don't clearly understand the reasons for putting extrusion ratio into 'varying' scope, it will be interpolated between every fragment during rasterization (intentionally?). Maybe different mechanism should be used for that anyway to maintain the same (non-interpolated) value - `uniform` or  `flat` qualifier that ensures that there’s no interpolation.

kkostecki: As a matter of fact I feel that extrusion ratio should be pre-calculated (solved) per object and passed as uniform both in VS and FS. That seems to be most natural way.

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
zbigg: Fixed, `MIN_BUILDING_HEIGHT` is now defined once.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -1042,8 +1037,8 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
 
         // Prevent that extruded buildings are completely flat (can introduce errors in normal
         // computation and extrusion).
-        if (height === floorHeight) {
-            height += MIN_BUILDING_HEIGHT;
+        if (Math.abs(height - floorHeight) < ExtrusionFeatureDefs.MIN_BUILDING_HEIGHT) {
+            height += ExtrusionFeatureDefs.MIN_BUILDING_HEIGHT;
```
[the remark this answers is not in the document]
zbigg: `flat` modifier is not supported: https://stackoverflow.com/questions/40101023/flat-shading-in-webgl

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
zbigg: Done

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
zbigg: This patch in essence disables animation of extruded building for buildings with very small height.
There are two ways of implementing this:
 - creating two render batches - flat EB and hight EB, with different values of `extrusionRatio` unirorm
   - this can be implemented either in style - by defining two rules for `buildings` or ...
   - in "platform", i.e decoder emits two render batches for two types of buildings 
 - in shaders (current solution), so materials "just" work,

This patch choses:
 - "platform" approach, because MIN_BUILDING_HEIGHT and float precision issues are implementation detail (and limitation), so user shouldn't is not bothered with this and forced to write two rules for such simple feature.

 - shader approach to minimize change impact. Creating two render batches in decoder is much "more" code, shader approach introduces only minimal performance impact

@here/harp-materials/lib/MapMeshMaterialsDefs.ts
```diff
@@ -0,0 +1,22 @@
...
+     */
+    export const DEFAULT_RATIO_MIN: number = 0.0;
+    /**
+     * Maximum ratio value for extrusion effect
+     */
+    export const DEFAULT_RATIO_MAX: number = 1;
+
+    /**
+     * Height buildings take whenever no height-data is present. Used to avoid z-fighting and other
+     * graphics artifacts.
```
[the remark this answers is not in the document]
zbigg: done

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -1042,8 +1037,8 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
 
         // Prevent that extruded buildings are completely flat (can introduce errors in normal
         // computation and extrusion).
-        if (height === floorHeight) {
-            height += MIN_BUILDING_HEIGHT;
+        if (Math.abs(height - floorHeight) < ExtrusionFeatureDefs.MIN_BUILDING_HEIGHT) {
+            height = ExtrusionFeatureDefs.MIN_BUILDING_HEIGHT;
```
kkostecki: I guess it should rather be:
`height = floorHeight + ExtrusionFeatureDefs.MIN_BUILDING_HEIGHT`
or simply (without if):
'height = max(floorHeight + ExtrusionFeatureDefs.MIN_BUILDING_HEIGHT, height)'

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -4,15 +4,27 @@
...
 attribute vec4 extrusionAxis;
 uniform float extrusionRatio;
 varying vec4 vExtrusionAxis;
+varying float vExtrusionRatio;
+
 `,
     extrusion_vertex: `
-transformed = transformed + extrusionAxis.xyz * (extrusionRatio - 1.0);
+// Cancel extrusionRatio (meaning, force to 1) if extrusionAxisLen < MIN_BUILDING_HEIGHT.
+const float MIN_BUILDING_HEIGHT_SQUARED = ${MIN_BUILDING_HEIGHT_SQUARED};
```
kkostecki: I am a bit worried if min building height shouldn't be function of camera distance:
`const minHeight = calcMinHeight(modelViewMatrix)`
Could you check how it behaves in different geo projection setup with camera place far above the ground.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
zbigg: Fixed as suggested, thanks.

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
zbigg: For normal building - root cause of issue, current solution should be ok, as they are visible on levels where current value of MIN_BUILDING_HEIGHT is significant, let's deliver this one and implement fully robust one later if really needed.

kkostecki: Agreed!

PR #1181, opened 2020-01-06
HARP-8337: Fix small caps rendering.

atomicsulfate:
Small caps were being applied to measure text but not
to create the text buffer object. As a result small caps were
actually rendered in all caps, but with a smaller bounding box.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -10,6 +10,7 @@ import {
     FontCatalog,
     MeasurementParameters,
     TextBufferAdditionParameters,
+    TextBufferCreationParameters,
     TextCanvas,
     TextLayoutStyle,
     TextRenderStyle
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -144,6 +145,7 @@ const tempScreenPosition = new THREE.Vector2();
 const tempScreenPoints: THREE.Vector2[] = [];
 const tempPoiScreenPosition = new THREE.Vector2();
 const tempTextOffset = new THREE.Vector2();
+const tmpTextBufferCreationParams: TextBufferCreationParameters = {};
```
CraigBerlin: Could you use TempParams (passed to addPointLabel) instead?

atomicsulfate: I'm planning to get rid of that TempParams structure, it contains parameter structures with no relation to each other, some of them are unused. I'll remove it in my next changes.

PR #1184, opened 2020-01-07
HARP-6251: Clip the polygons.

robertoraggi:
Clip the polygon geometries when using fill and standard techniques.

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
+++ b/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
@@ -76,6 +76,7 @@ import {
 // tslint:disable-next-line:max-line-length
 import { SphericalGeometrySubdivisionModifier } from "@here/harp-geometry/lib/SphericalGeometrySubdivisionModifier";
 import { ExtrusionFeatureDefs } from "@here/harp-materials/lib/MapMeshMaterialsDefs";
+import { clipPolygon } from "./clipPolygon";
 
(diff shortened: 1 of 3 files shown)


robertoraggi:

robertoraggi:

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -631,13 +632,33 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
+                    if (!isExtrudedPolygonTechnique(technique)) {
+                        if (!clipRing) {
+                            clipRing = [
+                                new THREE.Vector2(0, 0),
+                                new THREE.Vector2(extents, 0),
+                                new THREE.Vector2(extents, extents),
+                                new THREE.Vector2(0, extents)
+                            ];
+                        }
+                        coords = clipPolygon(coords, clipRing);
```
ninok: Maybe I missed it but I cannot see any code handling the "pseudo" edges introduced by clipping. Will it work if we render the polygon with outline or will we see an outline on the tile border?
An image regression test for that case would also be highly appreciated.

robertoraggi: The way stroking is implemented is just not working reliably when tiles are clipped and we have sources where tiles are clipped. If we want to support stroking polygons we may need to rewrite that code, maybe using a similar approach as the one we use for outlines.

PR #1193, opened 2020-01-10
HARP-8361: Fix material's shader updates in line rendering materials.

kkostecki:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-materials/lib/SolidLineMaterial.ts
+++ b/@here/harp-materials/lib/SolidLineMaterial.ts
@@ -12,16 +12,30 @@ import {
     FadingFeature,
     FadingFeatureParameters
 } from "./MapMeshMaterials";
-import linesShaderChunk from "./ShaderChunks/LinesChunks";
-import { enforceBlending } from "./Utils";
+import linesShaderChunk, { LineCapsModes } from "./ShaderChunks/LinesChunks";
+import {
+    enforceBlending,
+    getShaderMaterialDefine,
+    setShaderDefine,
+    setShaderMaterialDefine
+} from "./Utils";
 
-export const LineCapsDefinitions: { [key in LineCaps]: string } = {
-    Square: "CAPS_SQUARE",
-    Round: "CAPS_ROUND",
-    None: "CAPS_NONE",
-    TriangleIn: "CAPS_TRIANGLE_IN",
-    TriangleOut: "CAPS_TRIANGLE_OUT"
+const LineCapsDefinesMapping: { [key in LineCaps]: number } = {
+    None: LineCapsModes.CAPS_NONE,
+    Square: LineCapsModes.CAPS_SQUARE,
+    Round: LineCapsModes.CAPS_ROUND,
+    TriangleIn: LineCapsModes.CAPS_TRIANGLE_IN,
+    TriangleOut: LineCapsModes.CAPS_TRIANGLE_OUT
 };
+
+const DefinesLineCapsMapping: { [key: number]: LineCaps } = Object.keys(
+    LineCapsDefinesMapping
+).reduce((r, lineCapsName) => {
+    const defineKey = lineCapsName as keyof typeof LineCapsDefinesMapping;
+    const defineValue: number = LineCapsDefinesMapping[defineKey];
+    r[defineValue] = defineKey;
+    return r;
+}, ({} as any) as { [key: number]: LineCaps });
 
 /**
  * The vLength contains the actual line length, it's needed for the creation of line caps by
@@ -34,6 +48,7 @@ export const LineCapsDefinitions: { [key in LineCaps]: string } = {
  * the full line length.
  */
 
+const tmpColor = new THREE.Color();
 const vertexSource: string = `
 #define SEGMENT_OFFSET 0.1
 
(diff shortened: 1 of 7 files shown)


kkostecki:

kkostecki:

kkostecki:

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -511,28 +556,32 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
...
+     * @note The property [[gapSize]] need to be set to enable dashed line.
      */
     get dashColor(): THREE.Color {
         return this.uniforms.dashColor.value as THREE.Color;
     }
     set dashColor(value: THREE.Color) {
-        this.uniforms.dashColor.value = value;
-        this.defines.USE_DASH_COLOR = 1.0;
+        this.uniforms.dashColor.value.copy(value);
+        setShaderMaterialDefine(this, "USE_DASH_COLOR", 1.0);
```
MPursche: Does this need to be a numeric define? Seems like it could also be implemented with a "boolean" define similar to `USE_COLOR`, `USE_OUTLINE`.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -568,40 +620,39 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
...
+     * @note You may also need to set [[dashSize]].
+     * @see dashSize.
      */
     get gapSize(): number {
         return this.uniforms.gapSize.value as number;
     }
     set gapSize(value: number) {
         this.uniforms.gapSize.value = value;
-        this.defines.DASHED_LINE = this.gapSize > 0.0 ? 1 : 0;
+        setShaderMaterialDefine(this, "DASHED_LINE", value > 0.0 ? 1 : 0);
```
MPursche: Use "boolean" define (See previous comment).

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -568,40 +620,39 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
...
 
     /**
      * Caps mode.
      */
     get caps(): LineCaps {
         let result: LineCaps = "Round";
-        if (this.defines.CAPS_SQUARE === 1) {
-            result = "Square";
-        } else if (this.defines.CAPS_NONE === 1) {
+        const capsMode = getShaderMaterialDefine(this, "CAPS_MODE");
```
MPursche: MINOR: If we introduce new caps modes we would need to keep this in sync with `LineCapsDefinitions`. Maybe introduce a second definition that does the reverse mapping: `string` -> `LineCapsModes` that is automatically generated from `LineCapsDefinitions`.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -652,4 +698,25 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
     set drawRangeEnd(value: number) {
         this.uniforms.drawRange.value.y = value;
     }
+
+    set clipTileSize(tileSize: THREE.Vector2) {
+        this.uniforms.tileSize.value.copy(tileSize);
+        const useTileClip = tileSize.x > 0 && tileSize.y > 0;
+        setShaderMaterialDefine(this, "TILE_CLIP", useTileClip ? 1 : 0);
```
MPursche: Could also use "boolean" define.

@here/harp-materials/lib/SolidLineMaterial.ts
kkostecki: Of course! I left is so to preserve entire shader code compatibility...
I did not wanted to make entire "world" better, but limit change to minimum, while in some places (see CAPS_MODE) I could not hold myself from cleaning up :). Probably you clearly see why I have done it.

@here/harp-materials/lib/SolidLineMaterial.ts
kkostecki: As before.... old code ... more cleaning required - not exactly related to my change, but could be done in some follow up.

@here/harp-materials/lib/SolidLineMaterial.ts
kkostecki: Sorry @MPursche I don't exactly understand your suggestion, `LineCapsDefinitions` actually does the `string` -> `enum (int)` mapping, using `LineCaps` type (string literals) as key and `LineCapsModes` (enum) as value.
Would you like me to perform this mapping automatically so whenever something is added to `LineCaps` then `LineCapsDefinitions` grow? It is possible, but I am not sure if makes any sense because support in shader need to be added manually - both with new shader const (define), so I doubt if it helps anyway.
On other side if you mean reverse mapping LineCapsModes -> LineCaps (string), I guess it should never be exposed because it is implementation detail of the shaders. Even making `LineCapsModes` exported enum is little bit mistake (should be private) but I needed this in second file where the shader is defined (**LinesChunks.ts**). To fix this I would need to move `LineCapsDefinitions` to **LinesChunks.ts** file, which is maybe even worse....

MPursche: What I meant is, that in `set caps()` the object `LineCapsDefinitions` is used to determine which `LineCaps` string value corresponds to which `LineCapesModes` enum value. For `get caps()` the reverse(`LineCapsModes` to `LineCaps`)is implemented manually with if/else but could also use an object.

kkostecki:
Merged as a part of #1236

PR #1201, opened 2020-01-13
HARP-7856 Background plane added only during fallback

nzjony:
The background plane is used to hide the details of parents that are shown at a lower render order.


nzjony:
Yes, it unfortunately grew in complexity, I thought that I could reuse the checkedTiles, and it was originally simple, but I didn't realise because of the IBC test that a tile can cover its grandparent, and that the parent isn't in the checkedTiles, because the child never falls back (if it is loaded). I introduced the while loop, but that made things much uglier.

I wasn't so keen on having some dependency between datasources, but I guess that is the best option to keep the fallback simple.

PR #1202, opened 2020-01-13
HARP-7856 Remove background tile when a covering tile is loaded.

nzjony:


@here/harp-mapview/lib/DataSource.ts
```diff
@@ -50,6 +50,12 @@ export abstract class DataSource extends THREE.EventDispatcher {
      */
     addGroundPlane: boolean = false;
 
+    /**
+     * Boolean which says whether a [[DataSource]] produces tiles that fully cover the tile, i.e.
+     * tiles underneath are completely hidden.
+     */
+    isFullyCovering: boolean = false;
```
ninok: This should be a method that could be overridden and the default implementation should just return `addGroundPlane`

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -837,6 +844,32 @@ export class VisibleTileSet {
         tile.dispose();
     }
 
+    /**
+     * Disposes of tiles that overlap twice.
+     **/
+    private checkDuplicateFullyCoveringTiles(dataSource: DataSource, tile: Tile) {
+        if (dataSource.isFullyCovering && tile.backgroundPlane !== undefined) {
```
ninok: Why checking `tile.backgroundPlane !== undefined`? For the WebTile datasource there is no backgroundPlane but the tiles are still fully covered

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -897,6 +930,9 @@ export class VisibleTileSet {
                     tile.offset
                 );
                 tile.levelOffset = 0;
+                if (tile.backgroundPlane !== undefined) {
```
ninok: Why is this needed?

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: I wasn't considering the webtile datasource, thanks for the reminder.

I use the backgroundPlane render order to determine which tile should be removed. If it doesn't exist, I will take it to mean that it should have the highest render order, and all existing / future tiles should be removed.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: I think a left over from a previous attempt, but not sure, will remove.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -837,6 +845,64 @@ export class VisibleTileSet {
         tile.dispose();
     }
 
+    /**
+     * Disposes of tiles that overlap twice.
+     **/
+    private checkDuplicateFullyCoveringTiles(dataSource: DataSource, tile: Tile) {
```
ninok: The name indicates there is no side effect, but there is. So `removeDuplicateFullyCobveringTiles` would be a better name

nzjony:

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -837,6 +845,65 @@ export class VisibleTileSet {
         tile.dispose();
     }
 
+    /**
+     * Disposes of tiles that overlap twice and that come from a [[DataSource]] where the [[Tiles]]s
+     * are fully covering.
+     **/
+    private removeDuplicateFullyCobveringTiles(dataSource: DataSource, tile: Tile) {
```
ninok: `Cobvering` ?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -837,6 +845,65 @@ export class VisibleTileSet {
...
+     **/
+    private removeDuplicateFullyCoveringTiles(dataSource: DataSource, tile: Tile) {
+        if (this.options.projection.type === ProjectionType.Spherical) {
+            // HARP-7899, currently the globe has no background planes in the tiles (it relies on
+            // the BackgroundDataSource), because the LOD mismatches, hence disabling for globe.
+            return;
+        }
+        if (dataSource.isFullyCovering()) {
+            const ts = dataSource.getTilingScheme();
+            let map = this.m_coveringMap.get(ts)!;
```
ninok: Why `!` ?

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -837,6 +845,65 @@ export class VisibleTileSet {
...
+            if (map === undefined) {
+                map = new Map<number, Tile>();
+                this.m_coveringMap.set(ts, map);
+            }
+            const key = TileOffsetUtils.getKeyForTileKeyAndOffset(tile.tileKey, tile.offset);
+            const entry = map.get(key);
+            if (entry === undefined) {
+                map.set(key, tile);
+            } else {
+                const disposeTile = (toDispose: Tile) => {
```
ninok: I don't like the idea that we modify tiles that are in the cache. Especially when thinking about the fallback logic it could happen that we fallback to that tile and suddenly it's empty.
Maybe it's good enough to skip the rendering of the covered tile.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: No idea, will remove.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: I don't see the difference, if you fallback to an empty tile (it is disposed), or you fall back to a tile which isn't rendered, the net (visual) result is the same.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -347,6 +347,13 @@ export class Tile implements CachedResource {
      */
     levelOffset: number = 0;
 
+    /**
+     * @hidden
+     *
+     * Background mesh used for the fallback logic.
+     */
+    backgroundPlane?: THREE.Mesh;
```
ninok: does this need to be public?

@here/harp-mapview/test/VisibleTileSetTest.ts
```diff
@@ -271,8 +371,136 @@ describe("VisibleTileSet", function() {
         }
     });
 
+    /**
```
ninok: There is no tests that checks if tiles below non-covering data-sources are actually not removed

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -534,6 +539,9 @@ export class VisibleTileSet {
                     numTilesLoading++;
                 } else {
                     tile.numFramesVisible++;
+                    // If this tile's data source is "covering", then remove all others from the
```
ninok: Is this comment still valid?

nzjony: done

@here/harp-mapview/lib/Tile.ts
nzjony: i changed it to be a get / set accessor, because I had the addBackgroundPlane and then this public, so now it should be a bit cleaner interface.

It is needed in the VisibleTileSet and the TileGeometryCreator

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -534,6 +539,9 @@ export class VisibleTileSet {
                     numTilesLoading++;
                 } else {
                     tile.numFramesVisible++;
+                    // If this tile's data source is "covering" then other tiles beneath it have
+                    // their rendering skipped, see [[Tile.willRender]].
+                    this.skipOverlappedTiles(dataSource, tile);
```
ninok: It looks like this code is assuming that the background data-source is processed first, so it will not work in other scenarios, right?!
I think it would also be good enough to only support the skipping of background tiles, i.e. have more concrete code like "if tile is covering, remove it from the visible background tiles'

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -837,6 +846,45 @@ export class VisibleTileSet {
...
+     * [[DataSource]] which is fully covering, i.e. there it is fully opaque.
+     **/
+    private skipOverlappedTiles(dataSource: DataSource, tile: Tile) {
+        if (this.options.projection.type === ProjectionType.Spherical) {
+            // HARP-7899, currently the globe has no background planes in the tiles (it relies on
+            // the BackgroundDataSource), because the LOD mismatches, hence disabling for globe.
+            return;
+        }
+        if (dataSource.isFullyCovering()) {
+            const ts = dataSource.getTilingScheme();
```
ninok: Isn't the backround-ds always using webmercator? So the map of maps is not needed

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1014,6 +1021,14 @@ export class Tile implements CachedResource {
         return this.projection.worldExtent(0, 0).max.x * this.offset;
     }
 
+    /**
+     * Adds the supplied mesh to the list of three.js objects
+     * @param plane Plane to add to the [[Tile]].
+     */
+    set backgroundPlane(plane: THREE.Mesh) {
```
ninok: Please remove this setter b/c it's really confusing, i.e. calling it multiple times will add the object multiple times instead of setting the plane

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -837,6 +846,39 @@ export class VisibleTileSet {
...
+        if (dataSource.isFullyCovering()) {
+            const key = TileOffsetUtils.getKeyForTileKeyAndOffset(tile.tileKey, tile.offset);
+            const entry = this.m_coveringMap.get(key);
+            if (entry === undefined) {
+                // We need to reset the flag so that if the covering datasource is disabled, that
+                // the tiles beneath then start to render.
+                tile.skipRender = false;
+                this.m_coveringMap.set(key, tile);
+            } else {
+                const setTileToSkip = (toDispose: Tile) => {
```
ninok: toDispose? I think you can remove this function completely

@here/harp-mapview/lib/Tile.ts
```diff
@@ -653,6 +656,10 @@ export class Tile implements CachedResource {
         // to be overridden by subclasses
     }
 
+    set skipRender(skipRender: boolean) {
```
ninok: `willRender` is a callback method whereas `skipRender` is a property. I suggest to rename it to `skipRendering`. Since there is no harm if people read this property you could also just rename `m_skipRendering` and make it public.

PR #1204, opened 2020-01-14
HARP-7831 Webapp optimization

zbigg:
harp.gl website refactor
    
Optimizing harp.gl wbsite for speed.
    
    * Moved `harp.gl` webpage to `www` folder
    * Refactored harp.gl app to be "proper" webpack app
    * Reduce number or requestd by bundling JS into one file and CSS into
    HTML.
    * <link rel="preload"> key resources

harp.gl www - cleanup theme
    
    * Remove 'text' techniques as we don't want text in this theme.
    * Migrate unsupported nested styles to flat form.

Related-to: HARP-7831

PR #1206, opened 2020-01-15
harp.gl www -  offscreen rendering

zbigg:
(don't merge - experiment)

Experiment that shows that
 - it's possible to render our map on Offscreen canvas in Web Worker
 - it's possible to share code between workers and decoders in weback (although in extremely ugly and unstable way).

Result:
We can achieve >90 both on mobile & destop on page speed index.

Depends on #1204.
Related-to: HARP-7831


www/src/WorkerMapViewService.ts
```diff
@@ -0,0 +1,79 @@
...
+    let map!: MapView;
+    const eventHandler = new EventEmitter();
+
+    self.addEventListener("message", (event: MessageEvent) => {
+        const message = event.data;
+        switch (message.type) {
+            case "mapview-worker-init":
+                {
+                    setAppBaseUrl(message.baseUrl);
+                    const canvas: OffscreenCanvas = message.canvas;
```
ninok: That's really cool. Would be awesome if you could make it an official harp.gl feature instead of implementing it in the webpage

zbigg: Thanks, glad someone likes this.

Frankly, i did it mostly to optimize "main thread" idle KPI of page speed index, but yeah.
The problem is this "offscreen" map is completely different beast that our main `MapView`.
You can't pass `OmvDataSource`, you can't register listener in efficient way, you can't access 3JS stuff inside.
The interface must be 100% async and plain-object based (if we want to keep it lean and mean).
We can can think about how to make good API from this in separate task.

And it means yet another worker bundle - and we love plaing with custom webpack configs.
And [it works only in Chrome](https://caniuse.com/#search=offscreencanvas) - ok not, only but key players like Safari/iOS and, Firefox don't even plan releasing it soon.

PR #1209, opened 2020-01-15
Harp-8819: Extending support for round and diamond shaped dashes

StefanDachwitz:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-materials/lib/SolidLineMaterial.ts
+++ b/@here/harp-materials/lib/SolidLineMaterial.ts
@@ -38,8 +38,7 @@ const DefinesLineCapsMapping: { [key: number]: LineCaps } = Object.keys(
 }, ({} as any) as { [key: number]: LineCaps });
 
 export enum LineDashesModes {
-    DASHES_NONE = 0,
-    DASHES_SQUARE,
+    DASHES_SQUARE = 0,
     DASHES_ROUND,
     DASHES_DIAMOND
 }
@@ -175,6 +174,10 @@ uniform vec2 drawRange;
 uniform float dashSize;
 uniform float gapSize;
 uniform vec3 dashColor;
+
+#define DASHES_SQUARE ${LineDashesModes.DASHES_SQUARE}
+#define DASHES_ROUND ${LineDashesModes.DASHES_ROUND}
+#define DASHES_DIAMOND ${LineDashesModes.DASHES_DIAMOND}
 #endif
 
 varying vec3 vPosition;
@@ -399,7 +402,8 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
 
         // Setup default defines.
         const defines: { [key: string]: any } = {
-            CAPS_MODE: LineCapsModes.CAPS_ROUND
+            CAPS_MODE: LineCapsModes.CAPS_ROUND,
+            DASHES_MODE: LineDashesModes.DASHES_SQUARE
         };
 
         // Prepare defines based on params passed in, before super class c-tor, this ensures
@@ -688,7 +692,7 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
     }
     set caps(value: LineCaps) {
         // Line caps mode may be set directly from theme, thus we need to check value
-        // correctness and if we provide string to define mapping.
+        // for correctness and provide string to define mapping in fragment shader.
         if (LineCapsDefinesMapping.hasOwnProperty(value)) {
             setShaderMaterialDefine(this, "CAPS_MODE", LineCapsDefinesMapping[value]);
         }
@@ -700,7 +704,7 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
     get dashes(): LineDashes {
         let result: LineDashes = "Square";
         const dashesMode = getShaderMaterialDefine(this, "DASHES_MODE");
-        // Sanity check if material define is numerical and has direct mapping to LineDashes)) {
+        // Sanity check if material define is numerical and has direct mapping to LineDashes type.
         if (typeof dashesMode === "number" && DefinesLineDashesMapping.hasOwnProperty(dashesMode)) {
             result = DefinesLineDashesMapping[dashesMode];
         }
@@ -708,8 +712,8 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
     }
     set dashes(value: LineDashes) {
         // Line dashes mode may be set directly from theme, thus we need to check value
-        // correctness and if we provide string to define mapping.
-        if (LineCapsDefinesMapping.hasOwnProperty(value)) {
+        // for correctness and provide string to define mapping in fragment shader.
+        if (LineDashesDefinesMapping.hasOwnProperty(value)) {
             setShaderMaterialDefine(this, "DASHES_MODE", LineDashesDefinesMapping[value]);
         }
     }
--- a/@here/harp-materials/lib/SolidLineMaterial.ts
+++ b/@here/harp-materials/lib/SolidLineMaterial.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { LineCaps } from "@here/harp-datasource-protocol";
+import { LineCaps, LineDashes } from "@here/harp-datasource-protocol";
 import * as THREE from "three";
 import {
     DisplacementFeature,
@@ -36,6 +36,28 @@ const DefinesLineCapsMapping: { [key: number]: LineCaps } = Object.keys(
     r[defineValue] = defineKey;
     return r;
 }, ({} as any) as { [key: number]: LineCaps });
+
+export enum LineDashesModes {
+    DASHES_NONE = 0,
+    DASHES_SQUARE,
+    DASHES_ROUND,
+    DASHES_DIAMOND
+}
+
+const LineDashesDefinesMapping: { [key in LineDashes]: number } = {
+    Square: LineDashesModes.DASHES_SQUARE,
+    Round: LineDashesModes.DASHES_ROUND,
+    Diamond: LineDashesModes.DASHES_DIAMOND
+};
+
+const DefinesLineDashesMapping: { [key: number]: LineDashes } = Object.keys(
+    LineDashesDefinesMapping
+).reduce((r, lineDashesName) => {
+    const defineKey = lineDashesName as keyof typeof LineDashesDefinesMapping;
+    const defineValue: number = LineDashesDefinesMapping[defineKey];
+    r[defineValue] = defineKey;
+    return r;
+}, ({} as any) as { [key: number]: LineDashes });
 
 /**
  * The vLength contains the actual line length, it's needed for the creation of line caps by
@@ -198,6 +220,11 @@ void main() {
     // Compute distance to dash edge (0.5: dashCenter, 0.0: dashEdge) and compute the
     // dashBlendFactor similarly on how we did it for the line opacity.
     float distToDashEdge = 0.5 - distance(distToDashOrigin, (d + g) / d * 0.5);
+    #if DASHES_MODE == DASHES_ROUND
+    distToDashEdge = 0.5 - distance(vec2(distToCenter * 0.5, distToDashEdge), vec2(0.0, 0.5));
+    #elif DASHES_MODE == DASHES_DIAMOND
+    distToDashEdge -= distToCenter * 0.5;
+    #endif
     float dashWidth = fwidth(distToDashEdge);
     float dashBlendFactor = 1.0 - smoothstep(-dashWidth, dashWidth, distToDashEdge);
 
@@ -316,6 +343,12 @@ export interface SolidLineMaterialParameters
      * Default is `1.0`.
      */
     drawRangeEnd?: number;
+
+    /**
+     * Describes line dash type (`"Round"`, `"Square"`, `"Diamond"`).
+     * Default is `"Square"`.
+     */
+    dashes?: LineDashes;
 
     /**
      * Line dashes color.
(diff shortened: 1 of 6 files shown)


@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { LineCaps } from "@here/harp-datasource-protocol";
```
ninok: I really don't like that harp-materials depends on harp-datasource-protocoll just b/c of the caps and dash enums. I would rather redefine the enums here and rely on the tests that they match

StefanDachwitz: Will do

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -1216,7 +1388,7 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
                             );
                             positionArray.push(worldPos.x, worldPos.y, 0);
                             if (texCoordType !== undefined) {
-                                uvArray.push(vertices[i + 2], vertices[i + 3]);
+                                uconstray.push(vertices[i + 2], vertices[i + 3]);
```
zbigg: Seems incidental.

StefanDachwitz: Thanks. Fixed.

@here/harp-materials/lib/SolidLineMaterial.ts
StefanDachwitz: Done

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -380,30 +392,111 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
+            // Add continuous line as individual segments to improve special dashes by overlapping
+            // their connecting vertices. The technique/style should defined round or rectangular
+            // caps.
+            if (hasIndividualLineSegments) {
+                localLineSegments = [];
+
+                // Compute length of whole line and offsets of individual segments.
+                let lineLength = 0;
+                if (polyline.positions.length > 1) {
+                    const pointCount = polyline.positions.length;
```
kkostecki: Move `pointCount` two line above so we may use it in `if` condition.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -380,30 +392,111 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
+                            lastSegmentOffset += segmentLength;
+                            segmentOffsets.push(lastSegmentOffset);
+                        }
+
+                        tmpV3.sub(this.m_decodeInfo.center);
+                        localLine.push(tmpV3.x, tmpV3.y, tmpV3.z);
+                        tmpV4.sub(this.m_decodeInfo.center);
+                        localLine.push(tmpV4.x, tmpV4.y, tmpV4.z);
+
+                        localLineSegments!.push(localLine);
```
kkostecki: Maybe I see something wrong (because of web editor), but for me it looks like the `worldLines.push(worldLine)`, etc.. should be out of loop, you are collecting lines for each two point..., but probably it is intended for `individualLineSegments` wright?

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -733,23 +889,39 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
                                 );
                                 line.push(tmpV3.x, tmpV3.y, tmpV3.z);
                             }
+
+                            if (needIndividualLineSegments && line.length > 0 && ringLength > 0) {
+                                // Scale down each individual segment to range [0..1] for the whole
+                                // line.
+                                segmentOffsets!.forEach((offs, index) => {
+                                    segmentOffsets![index] = offs / ringLength;
```
kkostecki: Maybe multiply with invRingLenght - this may be pre-computed, in theory should be faster...

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -380,30 +392,111 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
+                        const offset =
+                            this.findRelativePositionInLine(tmpV3, untiledLine) / lineDist;
+                        lineOffsets.push(offset);
+                    }
+                    tmpV3.sub(this.m_decodeInfo.center);
+                    localLine.push(tmpV3.x, tmpV3.y, tmpV3.z);
+                });
+                localLines.push(localLine);
+                worldLines.push(worldLine);
+                uvs.push(lineUvs);
```
kkostecki: Maybe check before if this is worth to collect uvs...
`this.getTextureCoordinateType(technique)`

Here we create array of empty arrays...

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -198,6 +220,11 @@ void main() {
     // Compute distance to dash edge (0.5: dashCenter, 0.0: dashEdge) and compute the
     // dashBlendFactor similarly on how we did it for the line opacity.
     float distToDashEdge = 0.5 - distance(distToDashOrigin, (d + g) / d * 0.5);
+    #if DASHES_MODE == DASHES_ROUND
+    distToDashEdge = 0.5 - distance(vec2(distToCenter * 0.5, distToDashEdge), vec2(0.0, 0.5));
+    #elif DASHES_MODE == DASHES_DIAMOND
+    distToDashEdge -= distToCenter * 0.5;
```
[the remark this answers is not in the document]
kkostecki: This is crucial ... rest is just minor stuff.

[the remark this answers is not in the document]
StefanDachwitz: Yeah, I saw that. Fixed.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -658,6 +694,26 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
         }
     }
 
+    /**
+     * Dashes mode.
+     */
+    get dashes(): LineDashes {
+        let result: LineDashes = "Square";
+        const dashesMode = getShaderMaterialDefine(this, "DASHES_MODE");
+        // Sanity check if material define is numerical and has direct mapping to LineDashes)) {
```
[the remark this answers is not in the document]
StefanDachwitz: oops....

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -356,6 +363,11 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
                 const otherTexCoordType = this.getTextureCoordinateType(technique);
                 assert(otherTexCoordType === undefined || texCoordinateType === otherTexCoordType);
             }
+
+            hasIndividualLineSegments =
+                hasIndividualLineSegments || isSpecialDashesLineTechnique(technique);
+
+            hasContinuousLineSegments = hasContinuousLineSegments || !hasIndividualLineSegments;
```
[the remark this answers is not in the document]
StefanDachwitz: Yes. The techniques may use either type of geometry (individual, continuous) or both.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
StefanDachwitz: Done.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
StefanDachwitz: Yes, this is only for individual lines, where every new point in the input generates a new segment (2 points) in the output.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
StefanDachwitz: I never saw a difference in JS/TS with regard to performance of multiplication vs. division. The theoretical difference is not visible anymore with all the overhead and the speed of accessing RAM.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
StefanDachwitz: That is old code, if I am not mistaken. I am not going to touch this.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -658,6 +694,26 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
...
+        if (typeof dashesMode === "number" && DefinesLineDashesMapping.hasOwnProperty(dashesMode)) {
+            result = DefinesLineDashesMapping[dashesMode];
+        }
+        return result;
+    }
+    set dashes(value: LineDashes) {
+        // Line dashes mode may be set directly from theme, thus we need to check value
+        // correctness and if we provide string to define mapping.
```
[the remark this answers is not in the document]
StefanDachwitz: Done

@here/harp-materials/lib/SolidLineMaterial.ts
StefanDachwitz: Since the latest (succesfully reviewed) changes still have a dependency on datasource-protocol, I reverted to use the existing scheme and keep that dependency.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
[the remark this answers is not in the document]
kkostecki: Ok!

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -153,6 +175,15 @@ uniform vec2 drawRange;
 uniform float dashSize;
 uniform float gapSize;
 uniform vec3 dashColor;
+
+#define DASHES_SQUARE ${LineDashesModes.DASHES_SQUARE}
+#define DASHES_ROUND ${LineDashesModes.DASHES_ROUND}
+#define DASHES_DIAMOND ${LineDashesModes.DASHES_DIAMOND}
+
+#ifndef DASHES_MODE
+#define DASHES_MODE ${LineDashesModes.DASHES_SQUARE}
```
kkostecki: I would rather prefer to setup (default) define via shader params, instead of injecting it into the shader code, this way it will be clear from TypeScript side which DASH_MODE is set by default.
Imagine somebody calling a getter (`get dashMode()`), without setting it in params, we have fallback there, but the ralation with shader code it is not that clear, if default will change we may get into regression. If we set it here:
```
       // Setup default defines.
        const defines: { [key: string]: any } = {
            CAPS_MODE: LineCapsModes.CAPS_ROUND,
            DASHES_MODE: LineDashesMode.DASHES_SQUARE
        }
```
then situation is clear, no need for `ifndef DASHES_MODE .... ` in shader, no need for special handling in getter.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -198,6 +229,11 @@ void main() {
     // Compute distance to dash edge (0.5: dashCenter, 0.0: dashEdge) and compute the
     // dashBlendFactor similarly on how we did it for the line opacity.
     float distToDashEdge = 0.5 - distance(distToDashOrigin, (d + g) / d * 0.5);
+    #if DASHES_MODE == DASHES_ROUND
```
kkostecki: No handling for DASHES_MODE == DASHES_NONE. How does it differ from DASHES_SQUARE?

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -37,6 +37,28 @@ const DefinesLineCapsMapping: { [key: number]: LineCaps } = Object.keys(
     return r;
 }, ({} as any) as { [key: number]: LineCaps });
 
+export enum LineDashesModes {
+    DASHES_NONE = 0,
```
kkostecki: I can not distinguish between DASHES_NONE and DASHES_SQUARE modes, I guess the first mode is just not needed.
See also my comments in the shader, regarding it.

kkostecki:
Mostly ready, but some subtle changes would be appreciated.

@here/harp-materials/lib/SolidLineMaterial.ts
StefanDachwitz: The idea was to not have the overhead of setting this #define for every line, not just the dashed ones. Will add the dashes to the defaults.

@here/harp-materials/lib/SolidLineMaterial.ts
StefanDachwitz: Removing DASHES_NONE, as it is not used anywhere. Historically the dashes are activated (#define USE_DASHED_LINE) by setting "gapSize" > 0.

@here/harp-materials/lib/SolidLineMaterial.ts
StefanDachwitz: Removing it.

@here/harp-materials/lib/SolidLineMaterial.ts
kkostecki: Ok got it!

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -422,17 +514,40 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
...
-                    techniqueIndex,
-                    lineType,
-                    env.entries,
-                    localLines,
-                    context,
-                    this.getTextureCoordinateType(technique) ? uvs : undefined,
-                    hasUntiledLines ? offsets : undefined
-                );
+                if (hasIndividualLineSegments) {
+                    assert(
```
kkostecki: This assert could be inside the `if(...)` block below... but it just detail.

StefanDachwitz: Stupid me... Fixed.

PR #1221, opened 2020-01-17
Improve resolving @here/harp-utils symbols in web target.

zbigg:
@here/harp-utils contains platform-dependent module - UrlPlatformUtils
that has to be resolved differently in node and web context.

Follow strategy from `@node/fetch` and use `package.json`s `browser`
field, so `.web` versions are default in `web` target.


zbigg:
With this, users will no longer need to add extra `.web.js` in `resolve.extensions` in web targed.

IOW, normal users can remove this:
```
resolve: {
    extensions: [ ...".web.js", ".js"]
                      ^^^^^^
},
```

However, one have to remember that bundler must be configurured to resolve main module using `browser` field, like webpack does by default, but only in `web` target.

PR #1224, opened 2020-01-20
Shadow test

ninok:


@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -377,6 +377,13 @@ export interface StandardTechniqueParams extends BaseTechniqueParams {
      */
     alphaMap?: string | TextureBuffer;
     alphaMapProperties?: TextureProperties;
+
+    /**
+     * Whether this object can receive a shadow. Note, shadows must be enabled on the [[MapView]]
+     * for this to take affect. Note also that this also only receives shadows, a standard technique
+     * can't cast because it is flat.
+     */
+    enableShadows?: boolean;
```
ninok: I would stick to the three.js name (i.e. castShadow and receiveShadow). This would allow us later to remove all hardcoded logic that maps technique properties to material or object properties

@here/harp-examples/src/threejs_shadows.ts
```diff
@@ -0,0 +1,373 @@
...
+
+    window.addEventListener("keypress", event => {
+        if (event.key === "s") {
+            swapCamera(mapControls, m_trackball, debugCamera);
+            map.update();
+        }
+    });
+};
+
+const computeShadowFrustum = () => {
```
ninok: This should be done inside MapView

@here/harp-mapview/lib/DepthPrePass.ts
```diff
@@ -133,6 +133,9 @@ export function createDepthPrePassMesh(mesh: THREE.Mesh): THREE.Mesh {
         mesh.material instanceof Array
             ? mesh.material.map(createDepthPrePassMaterial)
             : createDepthPrePassMaterial(mesh.material);
+    // if (!(depthPassMaterial instanceof Array)) {
+    //     mesh.customDepthMaterial = depthPassMaterial;
+    // }
 
     const depthPassMesh = new THREE.Mesh(depthPassGeometry, depthPassMaterial);
```
ninok: Do we also render the buildings twice when rendering the shadow map? This should be avoided

@here/harp-mapview/lib/MapView.ts
```diff
@@ -931,6 +938,11 @@ export class MapView extends THREE.EventDispatcher {
 
         this.setupRenderer();
 
+        if (this.m_options.enableShadows === true) {
+            this.m_renderer.shadowMap.enabled = true;
+            this.m_renderer.shadowMap.type = THREE.PCFSoftShadowMap;
```
ninok: I think we should expose the shadow map type in the MapViewOptions as well

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2517,6 +2528,10 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_elevationProvider;
     }
 
+    get shadowsEnabled(): boolean {
```
ninok: Is there a reason to not have a setter?

@here/harp-datasource-protocol/lib/TechniqueParams.ts
nzjony: Removed as discussed.

@here/harp-mapview/lib/MapView.ts
nzjony: @ninok , I added a setter, but there is a bug in three.js, enabling and then disabling doesn't stop rendering the shadows, I will create an issue in three.js

@here/harp-mapview/lib/DepthPrePass.ts
nzjony: This mesh has both cast / receive shadows flags false, so I don't think so.

@here/harp-mapview/lib/MapView.ts
nzjony: as discussed, I have added a call to `clearTileCache`

@here/harp-mapview/lib/MapView.ts
nzjony: as discussed, will be done in later ticket, because VSM support doesn't yet work

@here/harp-mapview/lib/DepthPrePass.ts
nzjony: fixed.

@here/harp-examples/src/threejs_shadows.ts
nzjony: done

nzjony:

@here/harp-examples/src/threejs_shadows.ts
```diff
@@ -0,0 +1,265 @@
...
+import { assert } from "@here/harp-utils";
+import { GUI } from "dat.gui";
+import * as THREE from "three";
+import "three/examples/js/controls/TrackballControls";
+import { ShadowMapViewer } from "three/examples/jsm/utils/ShadowMapViewer";
+import { apikey, copyrightInfo } from "../config";
+
+let map: MapView;
+
+const guiOptions = {
```
ninok: MINOR: Would be good to have some initial values that actually show shadows

@here/harp-examples/src/threejs_shadows.ts
```diff
@@ -0,0 +1,265 @@
...
+        }
+        const enableCameraHelpers = map.pointOfView !== undefined;
+        if (enableCameraHelpers) {
+            mapCameraHelper.update();
+        }
+        mapCameraHelper.visible = enableCameraHelpers;
+    });
+
+    window.addEventListener("keypress", event => {
+        if (event.key === "s") {
```
ninok: can we have sth. that also works on the phone?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -795,6 +808,9 @@ export class MapView extends THREE.EventDispatcher {
 
     private m_enableMixedLod: boolean | undefined;
 
+    private m_tmpVector1 = new THREE.Vector3();
```
ninok: there is already a `cache` for vectors at the top of the file

@here/harp-mapview/lib/MapView.ts
```diff
@@ -671,6 +679,11 @@ export class MapView extends THREE.EventDispatcher {
 
     private m_skyBackground?: SkyBackground;
     private m_createdLights?: THREE.Light[];
+    private m_directionalLight?: THREE.DirectionalLight;
```
ninok: Why do we have a dedicated member for directional light?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -671,6 +679,11 @@ export class MapView extends THREE.EventDispatcher {
...
+
+    get directionalLight(): THREE.DirectionalLight | undefined {
```
ninok: Why? If we really have to expose the lights in `MapView` we should expose them generic (i.e. `THREE.Light[]`).
Is this just for convenience, so that the user does not need to go through the scene?
Honestly right now I would not add this API at all and just search for the lights in the scene.
Eventually the lights should be updateable via. `MapView.setDynamicProperty` but right now this only works for `Technique` attributes. Nevertheless we should not add an API right now that we will deprecate later.

@here/harp-examples/src/threejs_shadows.ts
nzjony: Agreed, done.

@here/harp-examples/src/threejs_shadows.ts
nzjony: Good point, I will add to the gui.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2518,6 +2538,26 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_elevationProvider;
     }
 
+    get shadowsEnabled(): boolean {
+        return this.m_options.enableShadows === true;
+    }
+
+    set shadowsEnabled(enabled: boolean) {
+        if (
+            this.m_renderer.shadowMap === undefined ||
```
ninok: If 'this.m_renderer.shadowMap === undefined' and `enabled === true` we should create it and not just return ?!

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2620,6 +2660,92 @@ export class MapView extends THREE.EventDispatcher {
...
+        // TODO: HARP-9479Globe doesn't support shadows.
+        if (
+            !this.shadowsEnabled ||
+            this.directionalLight === undefined ||
+            this.projection.type === ProjectionType.Spherical
+        ) {
+            return;
+        }
+
+        const NDCToView = (vector: Vector3Like): THREE.Vector3 => {
```
ninok: Why local function?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2620,6 +2660,92 @@ export class MapView extends THREE.EventDispatcher {
...
+            return (
+                new THREE.Vector3(vector.x, vector.y, vector.z)
```
ninok: Use cached vector?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2620,6 +2660,92 @@ export class MapView extends THREE.EventDispatcher {
...
+                    .applyMatrix4(this.camera.projectionMatrixInverse)
+                    // Make sure to apply rotation, hence use the rte camera
+                    .applyMatrix4(this.m_rteCamera.matrixWorld)
+            );
+        };
+        const ViewToLightSpace = (worldPos: THREE.Vector3, camera: THREE.Camera): THREE.Vector3 => {
```
ninok: Why local function?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2620,6 +2660,92 @@ export class MapView extends THREE.EventDispatcher {
...
+            return worldPos.applyMatrix4(camera.matrixWorldInverse);
```
ninok: Is it expected that `worldPos` is modified?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2620,6 +2660,92 @@ export class MapView extends THREE.EventDispatcher {
...
+            { x: 1, y: 1, z: 1 }
+        ];
+        const transformedPoints = points
+            .map(p => NDCToView(p))
+            .map(p => ViewToLightSpace(p, this.directionalLight!.shadow.camera));
+        const box = new THREE.Box3();
+        transformedPoints.forEach(point => {
+            box.expandByPoint(point);
+        });
+        Object.assign(this.directionalLight.shadow.camera, {
```
ninok: Even though this looks nicer than setting the camera properties directly, it creates a temporary object

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2620,6 +2660,92 @@ export class MapView extends THREE.EventDispatcher {
...
     }
 
+    /**
+     * Update the directional light camera. Note, this requires the cameras to first be updated.
+     */
+    private updateLights() {
+        // TODO: HARP-9479Globe doesn't support shadows.
+        if (
+            !this.shadowsEnabled ||
+            this.directionalLight === undefined ||
```
ninok: There could be multiple directional lights, so you should iterate over all lights and check their type.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3313,6 +3440,29 @@ export class MapView extends THREE.EventDispatcher {
                     return;
                 }
                 this.m_scene.add(light);
+                if ((light as any).isDirectionalLight) {
+                    const directionalLight = light as THREE.DirectionalLight;
+                    // This is needed so that the target is updated automatically, see:
+                    // https://threejs.org/docs/#api/en/lights/DirectionalLight.target
+                    this.m_scene.add(directionalLight.target);
+
+                    this.m_directionalLight = directionalLight;
```
ninok: There could be multiple directional light. You just pick the last one but I don't see any reason why multiple directional lights would not work (except for performance)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3313,6 +3440,29 @@ export class MapView extends THREE.EventDispatcher {
...
+
+                    const enableDebugDirectionalLight = false;
+                    if (enableDebugDirectionalLight) {
```
ninok: This code should be moved to the example, since everything we need is in the public `MapView` API

@here/harp-mapview/lib/MapView.ts
nzjony: Didn't know this, thanks.

@here/harp-mapview/lib/MapView.ts
nzjony: I checked the three.js code, it is never undefined, see: https://github.com/mrdoob/three.js/blob/e42922ad88d0ab406fb48b30ccf271336c46336a/src/renderers/WebGLRenderer.js#L314

I will remove this check.

@here/harp-mapview/lib/MapView.ts
nzjony: Yes, I will add comment.

@here/harp-mapview/lib/MapView.ts
nzjony: I need this check, otherwise the tests fail.

ninok: which tests?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2613,6 +2655,133 @@ export class MapView extends THREE.EventDispatcher {
...
+                camera.near = -box.max.z;
+                camera.far = -box.min.z;
+                camera.updateProjectionMatrix();
+
+                const lightDirection = cache.vector3[0];
+                lightDirection.copy(directionalLight.target.position);
+                lightDirection.sub(directionalLight.position);
+                lightDirection.normalize();
+
+                const target = MapViewUtils.getWorldTargetFromCamera(this.camera, this.projection);
```
ninok: Shadows should also work if no target point exists. I recently merged my PR #1365 that ensures that `MapView.m_targetWorld` always exists, so maybe you should use that

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1033,6 +1042,9 @@ export class TileGeometryCreator {
                         object,
                         materialFeature: true
                     });
+                    object.customDepthMaterial = new MapMeshDepthMaterial({
+                        depthPacking: THREE.RGBADepthPacking
```
ninok: Why is that?

@here/harp-mapview/lib/MapView.ts
nzjony: the tests stub the GL renderer:

```
const webGlStub = sandbox
        .stub(THREE, "WebGLRenderer")
        .returns(TestUtils.getWebGLRendererStub(sandbox, clearColorStub));
```

@here/harp-mapview/lib/MapView.ts
nzjony: I see: `@deprecated This function is for internal use only and will be removed in the future.`, which doesn’t unfortunately give me any idea how to upgrade, some help would be greatly apprecated. 🙂

nzjony:

nzjony:

PR #1225, opened 2020-01-20
Make the type attribute of theme definitions optional

zbigg:
Currently theme definitions require a `type` attribute to classify values and selectors.
Introduce shortcut syntax, where literal string/number/boolean values:.

    "definitions": {
      "a": { "value": 123},
      "b": { "value", "#f00f00"},
      "foo": 123,
      "bar": "#f0f0f0",
      "spam": true
    }

Related-to: HARP-8438

### DIFF
--- a/@here/harp-datasource-protocol/test/ExprTest.ts
+++ b/@here/harp-datasource-protocol/test/ExprTest.ts
@@ -56,12 +56,18 @@ describe("Expr", function() {
     describe("#fromJson", function() {
         describe("ref operator support", function() {
             const baseDefinitions: Definitions = {
-                color: { type: "color", value: "#ff0" },
-                number: { type: "number", value: 123 }
+                color: "#ff0",
+                string: { value: "abc" },
+                number: { type: "number", value: 123 },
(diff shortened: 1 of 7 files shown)


@here/harp-datasource-protocol/test/ExprTest.ts
```diff
@@ -56,12 +56,14 @@ describe("Expr", function() {
     describe("#fromJson", function() {
         describe("ref operator support", function() {
             const baseDefinitions: Definitions = {
-                color: { type: "color", value: "#ff0" },
-                number: { type: "number", value: 123 }
+                color: "#ff0",
+                number: { type: "number", value: 123 },
+                boolean: true
```
robertoraggi: we need to support the case where type is optional, e.g.
                `number: { value: 123 },`            
                `width: { value: ["number", 123] },`
should also work.

zbigg: Ups. Fixed. Last patch contains support for `{value: 123}`

PR #1226, opened 2020-01-20
HARP-8427: Fix broken buildings on iOS due to bug in GLSL max impleme…

ninok:
…ntation

### DIFF
--- a/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
+++ b/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
@@ -17,12 +17,21 @@ uniform float extrusionRatio;
 varying vec4 vExtrusionAxis;
 varying float vExtrusionRatio;
 
+// max is broken on Safari for iOS
+float ios_safe_max(float x, float y) {
+    if (x < y) {
+        return y;
+    } else {
+        return x;
+    }
+}
+
 `,
     extrusion_vertex: `
 // Cancel extrusionRatio (meaning, force to 1) if extrusionAxisLen < MIN_BUILDING_HEIGHT.
 const float MIN_BUILDING_HEIGHT_SQUARED = ${MIN_BUILDING_HEIGHT_SQUARED};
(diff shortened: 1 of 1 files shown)


@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -17,12 +17,21 @@ uniform float extrusionRatio;
...
+        return x;
+    }
+}
+
 `,
     extrusion_vertex: `
 // Cancel extrusionRatio (meaning, force to 1) if extrusionAxisLen < MIN_BUILDING_HEIGHT.
 const float MIN_BUILDING_HEIGHT_SQUARED = ${MIN_BUILDING_HEIGHT_SQUARED};
 float extrusionAxisLenSquared = dot(extrusionAxis.xyz, extrusionAxis.xyz);
-vExtrusionRatio = max(1.0 - step(MIN_BUILDING_HEIGHT_SQUARED, extrusionAxisLenSquared), extrusionRatio);
```
zbigg: We use `max` also in `roundEdgesAndAddCaps` (and maybe in other paces), should this be fixed there too ?

Is `max` completly forbdden as for now ?

ninok: That's a good question. I did not see artifacts on the caps. Maybe it happens only in combination with varyings.

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -17,12 +17,21 @@ uniform float extrusionRatio;
...
+vExtrusionRatio = ios_safe_max(1.0 - step(MIN_BUILDING_HEIGHT_SQUARED, extrusionAxisLenSquared), extrusionRatio);
```
robertoraggi: maybe `transformed = transformed + extrusionAxis.xyz * clamp(vExtrusionRatio - 1.0, 0.0, 1.0);` ?

PR #1233, opened 2020-01-21
Fix potential crashes trying to get the color of value

robertoraggi:


zbigg:
This PR only scratches problem of unreliable expressions. Complete solution to problem only scratched here is in #1241

zbigg:
Obsoleted by #1351

PR #1238, opened 2020-01-22
HARP-8427 Fix ios building extrusion bug

MPursche:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
+++ b/@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
@@ -22,7 +22,7 @@ varying float vExtrusionRatio;
 // Cancel extrusionRatio (meaning, force to 1) if extrusionAxisLen < MIN_BUILDING_HEIGHT.
 const float MIN_BUILDING_HEIGHT_SQUARED = ${MIN_BUILDING_HEIGHT_SQUARED};
 float extrusionAxisLenSquared = dot(extrusionAxis.xyz, extrusionAxis.xyz);
-vExtrusionRatio = max(1.0 - step(MIN_BUILDING_HEIGHT_SQUARED, extrusionAxisLenSquared), extrusionRatio);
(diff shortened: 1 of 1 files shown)


MPursche:
GPU tests please

MPursche:
GPU tests please

MPursche:
GPU Tests please

@here/harp-materials/lib/ShaderChunks/ExtrusionChunks.ts
```diff
@@ -22,7 +22,7 @@ varying float vExtrusionRatio;
 // Cancel extrusionRatio (meaning, force to 1) if extrusionAxisLen < MIN_BUILDING_HEIGHT.
 const float MIN_BUILDING_HEIGHT_SQUARED = ${MIN_BUILDING_HEIGHT_SQUARED};
 float extrusionAxisLenSquared = dot(extrusionAxis.xyz, extrusionAxis.xyz);
-vExtrusionRatio = max(1.0 - step(MIN_BUILDING_HEIGHT_SQUARED, extrusionAxisLenSquared), extrusionRatio);
+vExtrusionRatio = (extrusionAxisLenSquared < MIN_BUILDING_HEIGHT_SQUARED) ? 1.0 : extrusionRatio;
```
zbigg: Interesting, this is more or less original implementation that was changed in favor of `step/max` solution which seemed more "GLSL-way".
Do we have any hints what caused the bug in this case?

MPursche: We use a depth prepass for the extruded buildings to fix some of the transparency artifacts. But this only works if the prepass gives the exact same depth result as the actual rendering pass. On iOS there seems to be some precision issue which causes the value of `vExtrusionRatio` to be slightly different in each pass causing the flickering.
Rounding `vExtrusionRatio` with something like `floor(vExtrusionRatio * 1024.0) / 1024.0` also worked.
I saw no significant performance difference with the fix above so I choose this one because it is also a bit easier to understand.

PR #1241, opened 2020-01-23
Support style expressions returning invalid values.

zbigg:
As dynamic expression may return invalid value (bad type, null, error),
we have to guard places when style expressions/values are used.

This patch protects against style / Expr yielding:
 - [x] null
 - [x] invalid type for numeric attributes
 - [x] invalid type or format for color attributes
 - [x] errors thrown while evaluating expressions

Also improves type safety around dynamic properties (by getting rid of `any`)

by falling back to default value for given attribute.

Related-to: HARP-8515

### DIFF
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -10,10 +10,8 @@ import {
     Env,
     Expr,
     getPropertyValue,
-    InterpolatedProperty,
     isExtrudedLineTechnique,
     isExtrudedPolygonTechnique,
-    isInterpolatedProperty,
     isJsonExpr,
     isShaderTechnique,
     isStandardTechnique,
@@ -232,7 +230,7 @@ export function createMaterial(
 
     if (isShaderTechnique(technique)) {
         // Special case for ShaderTechnique.
-        applyShaderTechniqueToMaterial(technique, material);
+        applyShaderTechniqueToMaterial(technique, material, options.env);
     } else {
         // Generic technique.
         applyTechniqueToMaterial(technique, material, options.env, options.skipExtraProps);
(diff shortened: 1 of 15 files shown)


@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -415,6 +414,103 @@ export function getMaterialConstructor(technique: Technique): MaterialConstructo
...
+    property: Value | Expr | undefined,
+    defaultValue: D,
+    envOrLevel: number | Env
+): number | D {
+    if (property === undefined || property === null) {
+        return defaultValue;
+    }
+    const evaluated = getPropertyValue(property, envOrLevel as number);
+
+    if (evaluated === undefined || property === null) {
```
MPursche: Shouldn't this be `evaluated === undefined || evaluated === null` ?

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -429,11 +525,11 @@ export function getMaterialConstructor(technique: Technique): MaterialConstructo
  */
 export function evaluateBaseColorProperty(
     technique: Technique,
-    zoomLevel?: number
+    zoomLevel: number
 ): number | undefined {
     const baseColorProp = getBaseColorProp(technique);
     if (baseColorProp !== undefined) {
-        return evaluateColorProperty(baseColorProp, zoomLevel);
+        return getColorPropertyValueSafe(baseColorProp, undefined, zoomLevel);
```
MPursche: Is the check `baseColorProp !== undefined` still necessary? I think `getColorPropertyValueSafe` already checks if the property is `null` or `undefined`.

@here/harp-mapview/lib/DecodedTileHelpers.ts
zbigg: Fixed

@here/harp-mapview/lib/DecodedTileHelpers.ts
zbigg: Agree, fixed.

zbigg:
@robertoraggi , i know you've opposed a little, because we don't fix our bad architecture and it's not generic enough - yes it, is

I'm not sure how big your concern is.

Do you think we can deliver this as is and have "slight improvement" in cway we currently handle inteprolated expressions ?

Or should we push harder and improve architecture when we will finally decide on direction.

One alternative is #1287 right direction?

zbigg:
Obsoleted by #1351

PR #1243, opened 2020-01-24
Improve object picking.

zbigg:
Picking returns also techniques is enabled.

Depends (and includes) on picking fix in #1244

Related-to: HARP-8584

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -457,6 +457,11 @@ export interface MapViewOptions extends TextElementsRendererOptions {
     enableRoadPicking?: boolean;
 
     /**
+     * Set to `true` to allow picking of technique information associated with objects.
+     */
+    enablePickTechnique?: boolean;
+
+    /**
      * An optional canvas element that renders 2D collision debug information.
      */
     collisionDebugCanvas?: HTMLCanvasElement;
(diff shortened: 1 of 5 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -456,6 +456,11 @@ export interface MapViewOptions extends TextElementsRendererOptions {
      */
     enableRoadPicking?: boolean;
 
+    /**
+     * Set to `true` to allow picking of technique information associated with objects.
+     */
+    enablePickTechnique?: boolean;
```
kkostecki: Very minor, but could be names:
` enableTechniquePick?: boolean;`
as other flags... or
` enableTechniquePicking?: boolean;`

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -165,6 +166,9 @@ export class PickHandler {
             }
 
             const featureData: TileFeatureData = intersect.object.userData.feature;
+            if (this.enablePickTechnique) {
```
kkostecki: Will this work for SolidLines techniques? As I see TileGeometry creator does not add technique to `userData` in this case?

kkostecki: Ok found it, it's OK.

PR #1244, opened 2020-01-24
Ensure style expressions are compiled for road picking.

zbigg:
Since #1135, we need to compile technique expressions before passing
to `getPropertyValue` and friends.

Related-to: HARP-8531


zbigg:
Merged as part of #1243

PR #1249, opened 2020-01-28
HARP-8521: Fix terrain overlay.

atomicsulfate:
TileDisplacementMap was not being set as user data
for terrain objects in TileGeometryCreator.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -839,8 +839,7 @@ export class TileGeometryCreator {
                     }
                 }
 
-                this.addFeatureData(srcGeometry, technique, object);
-                this.addGeometryObjInfos(tile, srcGeometry, technique, object);
+                this.addUserData(tile, srcGeometry, technique, object);
 
                 if (isExtrudedPolygonTechnique(technique) || isFillTechnique(technique)) {
                     // filled polygons are normal meshes, and need transparency only when fading or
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1368,48 +1367,29 @@ export class TileGeometryCreator {
...
+        const hasObjInfos = (srcGeometry.objInfos?.length ?? 0) > 0;
 
         if (isTerrainTechnique(technique)) {
+            assert(
+                Object.keys(object.userData).length === 0,
+                "Unexpected user data in terrain object"
+            );
+
             if (typeof srcGeometry.objInfos![0] === "number") {
                 assert(false, "Wrong attribute map type for terrain geometry");
```
kkostecki: I know it is not your code, but if you update this PR, please change it to simple assert as above. I guess it should do the same job, but cleaner. Maybe the author have intention for debug some stuff... 
I think that condition should be expressed more strictly, now type `string` may also pass the test.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1368,48 +1367,29 @@ export class TileGeometryCreator {
...
                 return;
             }
 
+            if (!hasObjInfos) {
```
kkostecki: I see that entire block has no sense to perform if hasObjInfos === false, otherwise it simply catches some assertions (but those are not extracted from JS code, so debugging will be always on).
Summarizing it would be the best just to move this:
```
if (!hasObjInfos) {
    return;
}
```
at the very beginning of the block.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1423,8 +1403,20 @@ export class TileGeometryCreator {
                 displacementMap
             };
             object.userData = tileDisplacementMap;
-        } else {
-            object.userData = srcGeometry.objInfos;
+        } else if (
+            (hasObjInfos || isCirclesTechnique(technique) || isSquaresTechnique(technique)) &&
```
kkostecki: I am wondering if there is no hidden bug here, because we may pass `undefined` srcGeometry.objInfos to `object.userData'. Of course I saw that is the part of original implementation. If I have right then entire function should return immediately if there is no object infos.

yarn.lock
```diff
@@ -254,7 +254,7 @@
   dependencies:
     "@types/chai" "*"
 
-"@types/chai@*":
+"@types/chai@*", "@types/chai@^4.2.7":
```
kkostecki: Is this needed for `expect` call?

atomicsulfate: expect belongs to chai yes, but if you ask why there's that new chai version there I don't know. Yarn install is adding that automatically.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
atomicsulfate: done

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
atomicsulfate: done

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
atomicsulfate: yes, you're probably right. I didn't want to change the logic to avoid breaking something, but I really cannot think of why would that special behavior be needed for circles and squares techniques. I guess I'll try and see what happens.

PR #1252, opened 2020-01-28
Remove typestring dependency

haraldF:
Typestring is unmaintained and brings along a super old TypeScript
version. Use the TypeScript Compiler API instead.

### DIFF
--- a/@here/harp-datasource-protocol/test/ThemeTypingsTest.ts
+++ b/@here/harp-datasource-protocol/test/ThemeTypingsTest.ts
@@ -9,33 +9,65 @@
 
 import { LoggerManager } from "@here/harp-utils";
 import { assert } from "chai";
-
-declare const require: any;
+import * as ts from "typescript";
 
 const logger = LoggerManager.instance.create("ThemeTypingsTest");
 
 const isNode = typeof window === "undefined";
 const describeOnlyNode = isNode ? describe : xdescribe;
 
-function validate(type: string, response: any, validationExpectedToFail: boolean = false) {
-    try {
-        const ts = require("typestring");
-        const repl = ts.compile(
-            `import { ${type} } from "./lib/Theme";\n` +
-                `const reply: ${type} = ` +
-                JSON.stringify(response) +
-                ";", // {String} input - TypeScript to compile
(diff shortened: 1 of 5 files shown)


@here/harp-datasource-protocol/test/ThemeTypingsTest.ts
```diff
@@ -9,33 +9,65 @@
 
 import { LoggerManager } from "@here/harp-utils";
 import { assert } from "chai";
-
-declare const require: any;
+import * as ts from "typescript";
```
zbigg: `webpack.tests.config`  contains externals: { ... typestring: "undefined" } so we don't load whole TS runtime for browser integration tests.

It may be viable to add this exception now to for `typescript`.





https://github.com/heremaps/harp.gl/blob/master/webpack.tests.config.js#L102

haraldF: @zbigg would it make sense to have a `.web.ts` suffix for tests as well instead of hacking around webpack config?

zbigg: @haraldF In this particular case we would need to invent new naming for node.js-only code (tests), like `.node.(js,ts)`... and i don't know if this is good idea.
Moreover, we wouldn't have any help from tools (correct me if i'm worng) and we would need to filter by this `.node.` component by ourselves, so it's still boilerplat on our side ... just other type of boilerplate.

So, yes, no, maybe from my side.

haraldF: @zbigg I was thinking the other way around - add an empty `ThemeTypingsTest.web.ts` test, and put the typestring dependency in the node-only `ThemeTypingsTest.ts` file. That way, webpack shouldn't bundle typestring for the web bundle as the `.web.ts` will override the test.

haraldF: Well - didn't work. It still took the nodejs into the web bundle, although we told webpack to prefer `.web.ts` :( Made `typescript` undefined now as you suggested

PR #1253, opened 2020-01-28
CI: Fix gh-pages deploy script.

zbigg:

# CI Check

Happy contributing!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,22 +23,22 @@ before_install:
 jobs:
   include:
     - name: "Test"
-      script: |
-        set -ex
-        yarn pre-test
-        yarn cov-test --forbid-only
-        yarn build-tests
-        yarn test-browser --headless-firefox
-        yarn test-browser --headless-chrome
-        ./scripts/test-npm-packages.sh
-        git status # just in case test-npm-packages leaves some garbage
+      script:
+        - yarn pre-test
+        - yarn cov-test --forbid-only
+        - yarn build-tests
+        - yarn test-browser --headless-firefox
+        - yarn test-browser --headless-chrome
+        - ./scripts/test-npm-packages.sh
+
     - name: "Build & Deploy"
-      script: |
-        set -ex
-        cross-env NODE_ENV=production yarn run build
-        yarn run typedoc
-        cross-env NODE_ENV=production yarn run build-www
-      before_deploy: yarn prepare-doc-deploy
+      script:
+        - cross-env NODE_ENV=production yarn run build
+        - yarn run typedoc
+        - cross-env NODE_ENV=production yarn run build-www
+        - yarn prepare-doc-deploy
+        # debug: lerna publish expects clean workspace
+        - ./scripts/git-check-clean-workspace.sh
       deploy:
         - provider: script
           script: ./scripts/publish-packages.sh
--- a/scripts/prepare_doc_deploy.ts
+++ b/scripts/prepare_doc_deploy.ts
@@ -6,7 +6,7 @@
 
 import { execSync } from "child_process";
 import { writeFileSync } from "fs";
-import { copySync, ensureDirSync, moveSync } from "fs-extra";
+import { copySync, ensureDirSync, removeSync } from "fs-extra";
 
 // tslint:disable-next-line:no-var-requires
 const fetch = require("node-fetch");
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -38,7 +38,9 @@ jobs:
         cross-env NODE_ENV=production yarn run build
         yarn run typedoc
         cross-env NODE_ENV=production yarn run build-www
-      before_deploy: yarn prepare-doc-deploy
+        git status # just in build leaves some garbage
+        yarn prepare-doc-deploy
+        find dist/
```
musculman: That is not a good test, as it can still fail even if dist/ is there. I would not add it.

scripts/prepare_doc_deploy.ts
```diff
@@ -30,13 +30,18 @@ const folderName = branch !== "master" ? commitHash : "master";
 // │   ├── index.html (and assets for minisite)
 // │   ├── releases.json (list all releases in order)
 
-ensureDirSync("dist/gh_deploy");
-copySync("www/dist/", "dist/gh_deploy/");
+const targetFodlerGh = `dist/gh_deploy`;
```
musculman: typo on folder/fodler

scripts/prepare_doc_deploy.ts
```diff
@@ -30,13 +30,18 @@ const folderName = branch !== "master" ? commitHash : "master";
...
 
-ensureDirSync(`dist/s3_deploy/${folderName}`);
-moveSync("dist/doc", `dist/s3_deploy/${folderName}/doc`);
-moveSync("dist/doc-snippets", `dist/s3_deploy/${folderName}/doc-snippets`);
-moveSync("dist/examples", `dist/s3_deploy/${folderName}/examples`);
+removeSync(targetFodlerGh);
+ensureDirSync(targetFodlerGh);
+copySync("www/dist/", `${targetFodlerGh}/`);
+
+removeSync(targetFodlerS3);
```
musculman: I think moving the whole `dist/` is risky.
Why don't we do like before? Explicit move of files/folders.
```
mkdir -p dist/gh_deploy
mv dist/index.html dist/gh_deploy
mv dist/css dist/gh_deploy
mv dist/resources dist/gh_deploy
mv dist/js dist/gh_deploy
mv dist/redirect_examples dist/gh_deploy/examples
mv dist/redirect_docs dist/gh_deploy/docs

mv dist/_config.yml dist/gh_deploy

mkdir -p dist/s3_deploy/${folder_name}
mv dist/doc* dist/s3_deploy/${folder_name}
mv dist/examples dist/s3_deploy/${folder_name}
```

.travis.yml
zbigg: Yes, i understand this is only `set -x`-like solution, so we see in logs what was actually produced.

scripts/prepare_doc_deploy.ts
zbigg: FThe folder structure has been changed, but anyway in my opinion fix this script by:

1) Don't move things at all now. 
2) Clean target folders, so we don't get old stuff
3) ... finally _copy_ stuff to respective dist folders. 

It's much easier to test and debug such script because it doesn't destroy build artifcats in their original place.
Ignoring these changes, new script does exactly what is required for S3 and GH dist ... let's see that CI says.

musculman: That is a good point, copying instead of moving. You have my vote for that :)

But, copying the whole `dist` folder is a bit of an overkill as in `gh-pages` we only need the documentation and webpage. So, adding docs and examples is perhaps too much? Well, probably is not that bad, but just bloats a bit the branch. I see that for s3 we only move doc, doc-snippets and examples (as it should be).
Don't want to be too obnoxious about it, just trying to avoid having a too generic rule :)

zbigg: > So, adding docs and examples is perhaps too much?

But examples and docs don't land in `dist/gh-pages`.

The structure is same as done with original script created by you:

* `dist/gh_deploy` -> only "website" and this goes to GH-pages
* `dist/s3_deploy/$ref` -> examples and docs and snippets and this goes to S3

This is how it looks: https://gist.github.com/zbigg/e3b0be3fec1d56f712408011514fbd25

musculman: Because I need to learn to read. I see you move www/dist to gh_deploy and not dist/ :)
I was stubbornly reading dist and I could not make sense of it.
My apologies, will resolve the conversation.

PR #1254, opened 2020-01-28
Sync harp-website with rest of project.

zbigg:
Sync harp-website (`www/`) with rest of project, so we've got consistent tags.


zbigg:
Merge _only_ after #1245

PR #1258, opened 2020-01-30
HARP-8344 Fixed CirclePointsMaterial and added tests.

MPursche:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-materials/lib/CirclePointsMaterial.ts
+++ b/@here/harp-materials/lib/CirclePointsMaterial.ts
@@ -21,9 +21,10 @@ void main() {
 
 const fragmentShader: string = `
 uniform vec3 diffuse;
+uniform float opacity;
 
 void main() {
-    float alpha = 1.0;
+    float alpha = opacity;
 
     float radius = 0.5;
     vec2 coords = gl_PointCoord.xy - vec2(0.5);
@@ -43,9 +44,12 @@ export interface CirclePointsMaterialParameters extends THREE.ShaderMaterialPara
      * Point size.
      */
     size?: number;
-}
 
-const DEFAULT_CIRCLE_SIZE = 1;
+    /**
+     * Point color.
+     */
+    color?: THREE.Color;
+}
 
 /**
  * Material designed to render circle points. Note that it is always transparent since the circle
@@ -53,13 +57,10 @@ const DEFAULT_CIRCLE_SIZE = 1;
  * not bring.
  */
 export class CirclePointsMaterial extends THREE.ShaderMaterial {
-    isCirclePointsMaterial: true;
-    uniforms: { [uniform: string]: THREE.IUniform };
-    vertexShader: string;
-    fragmentShader: string;
+    static readonly DEFAULT_CIRCLE_SIZE = 1;
 
-    private m_size: number;
     private m_color: THREE.Color;
+    private m_opacity: number;
 
     /**
      * Constructs a new `CirclePointsMaterial`.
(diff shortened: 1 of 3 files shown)


MPursche:
Rebuild please.

@here/harp-materials/lib/CirclePointsMaterial.ts
```diff
@@ -43,46 +45,47 @@ export interface CirclePointsMaterialParameters extends THREE.ShaderMaterialPara
...
 
-        this.isCirclePointsMaterial = true;
         this.type = "CirclePointsMaterial";
-        this.vertexShader = vertexShader;
-        this.fragmentShader = fragmentShader;
-
-        this.m_size = parameters.size || DEFAULT_CIRCLE_SIZE;
-        this.m_color = new THREE.Color();
+        this.m_color = getOptionValue(color, new THREE.Color());
+        this.m_opacity = getOptionValue(opacity, 1.0);
```
kkostecki: Just two side notes - make sure if it supports RGBA in technique (color assigned from technique should be already in RGB format) and opacity applied.... Maybe extend the test?
This should be OK, as generic algorithm is applied but could be good to double-check.

@here/harp-materials/lib/CirclePointsMaterial.ts
```diff
@@ -43,46 +45,47 @@ export interface CirclePointsMaterialParameters extends THREE.ShaderMaterialPara
...
 
         this.uniforms = {
+            size: new THREE.Uniform(getOptionValue(size, CirclePointsMaterial.DEFAULT_CIRCLE_SIZE)),
             diffuse: new THREE.Uniform(this.m_color),
-            size: new THREE.Uniform(this.m_size)
+            opacity: new THREE.Uniform(this.m_opacity)
         };
```
kkostecki: I know Marcel it is not your code, but we deal with quite gentle setup here, uniforms of material are re-assigned - changed and there is not information for WebGL about it.... Normally in such case you need to call:
```
this.uniformsNeedsUpdate = true;
```
but here you can make it even better, by passing uniforms already in `ShaderMaterial` params to super c-tor. It will be better performance wise... Please take care about it.

@here/harp-materials/lib/CirclePointsMaterial.ts
```diff
@@ -43,46 +45,47 @@ export interface CirclePointsMaterialParameters extends THREE.ShaderMaterialPara
...
-        parameters.depthTest = false;
+        const { size, color, opacity, ...shaderParams } = parameters;
+        shaderParams.name = "CirclePointsMaterial";
+        shaderParams.vertexShader = vertexShader;
+        shaderParams.fragmentShader = fragmentShader;
+        shaderParams.depthTest = false;
 
-        super(parameters);
+        super(shaderParams);
         enforceBlending(this);
```
kkostecki: Please also consider using blending only when opacity is below < 1.0.

@here/harp-materials/lib/CirclePointsMaterial.ts
MPursche: Ok i will change the implemenation. But I don't think it is really possible to run into a problem. The material is only compiled when it is used during rendering. The uniforms are changed in the constructor of the material so I don't see a chance that it will be compiled before the constructor code has completed.

@here/harp-materials/lib/CirclePointsMaterial.ts
MPursche: The rgba support is done by `DecodedTileHelper`. The Alpha component is removed from the color there and passed as opacity. So there is no need to handle it in the materials.

@here/harp-materials/lib/CirclePointsMaterial.ts
MPursche: We need top keep blending enabled even if opacity is 1.0. The circle material uses alpha blending to create a smooth edge. So the edge of the circle is always transparent.

@here/harp-materials/lib/CirclePointsMaterial.ts
```diff
@@ -43,78 +44,104 @@ export interface CirclePointsMaterialParameters extends THREE.ShaderMaterialPara
...
-
-        this.m_size = parameters.size || DEFAULT_CIRCLE_SIZE;
-        this.m_color = new THREE.Color();
-
-        this.uniforms = {
-            diffuse: new THREE.Uniform(this.m_color),
-            size: new THREE.Uniform(this.m_size)
-        };
-
-        this.extensions.derivatives = true;
```
kkostecki: Missing extension setup after update.

PR #1260, opened 2020-02-04
HARP-8737: Use projection to compute surface normal

ninok:


@here/harp-lines/test/rendering/RenderLines.ts
```diff
@@ -221,7 +222,7 @@ describe("Rendering lines: ", function() {
                 canvas.height / 2.0,
                 -canvas.height / 2.0
             );
-            orthoCamera.position.z = 1.0;
+            orthoCamera.position.z = 11.0;
```
kkostecki: What is the reason for moving camera 10 units in Z and moving lines the same distance? It should does not change anything especially in ortho projection...

@here/harp-lines/lib/Lines.ts
```diff
@@ -215,8 +215,10 @@ export function createLineGeometry(
         tmpTangent0.normalize();
         geometry.vertices.push(tmpTangent0.x, tmpTangent0.y, tmpTangent0.z);
         tmpVertices.push(tmpTangent0.x, tmpTangent0.y, tmpTangent0.z);
+        tmpNormal.add(center); // tmpNormal contains world position
+        projection.surfaceNormal(tmpNormal, tmpNormal);
```
kkostecki: This could be heavily optimized if we know the projection type before hand, wright? For planar projection the normal will be position independent (assuming the we do not consider terrain curvature - which is true for now, yes?).

ninok: I agree it could be optimized but at what cost? It is an internal property of the projection whether the normal is constant or not. Exposing this knowledge here could cause issues in the future (and was already causing the issue at hand).

@here/harp-lines/test/rendering/RenderLines.ts
ninok: it should not change anything but it actually did (due to the bug that is now fixed).

kkostecki: So if the bug was fixed could we stay with original values? I am afraid that offsetting this entire scene just hides some stuff (bug)...

@here/harp-lines/lib/Lines.ts
kkostecki: So can we assume that all Planar projections will have the constant normal? I saw it in many places in the code anyway... Taking into account that we create a lot of lines, this simplification may be worth some if-ology.

PR #1263, opened 2020-02-04
HARP-8521: Move text element overlay to TextElementsRenderer.

atomicsulfate:

# CI Check

Happy contributing!


@here/harp-mapview/test/stubElevationProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+export type ElevationProviderStub = sinon.SinonStubbedInstance<ElevationProvider>;
+
+/**
+ * Creates an ElevationProvider stub.
+ * @param sandbox Sinon sandbox to keep track of created stubs.
+ * @returns ElevationProvider stub.
+ */
+export function stubElevationProvider(sandbox: sinon.SinonSandbox): ElevationProviderStub {
+    const stub = sandbox.createStubInstance(ElevationProviderFake);
+    stub.getHeight.callsFake(geoPoint => {
```
nzjony: do you need this, why not provide a fake implementation in the ElevationProviderFake?

@here/harp-mapview/test/stubProjection.ts
```diff
@@ -0,0 +1,36 @@
...
+    ProjectionType
+} from "@here/harp-geoutils";
+import * as THREE from "three";
+
+/**
+ * Creates a stub of the Mercator projection.
+ * @param sandbox Sinon sandbox to keep track of created stubs.
+ * @returns MercatorProjection stub.
+ */
+export function stubMercatorProjection(sandbox: sinon.SinonSandbox): Projection {
```
nzjony: why do you stub the projection? Why not use mercator, or even just the identity projection?

@here/harp-mapview/test/stubElevationProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+        return geoPoint.altitude ?? 0;
+    });
+    stub.sampleHeight.callsFake(geoPoint => {
```
nzjony: as above

@here/harp-mapview/test/TextElementsRendererTestFixture.ts
```diff
@@ -316,6 +340,24 @@ export class TestFixture {
...
     }
 
+    private setTerrainTiles(indices: number[]) {
+        expect(this.m_viewState.elevationProvider).not.undefined;
+        this.m_elevationProviderStub.getDisplacementMap.resetBehavior();
+        this.m_elevationProviderStub.getDisplacementMap.returns(undefined);
+
+        for (const index of indices) {
+            const tileKey = this.m_allTiles[index].tileKey;
+            this.m_elevationProviderStub.getDisplacementMap
```
nzjony: do the arguments matter? Why not just .getDisplacementMap.returns(...) ?

@here/harp-mapview/test/stubElevationProvider.ts
atomicsulfate: Returning a stub is more flexible, it allows users to overwrite the behaviour with whatever they want.
Writing a fake with hardcoded behaviour would only be useful for my test.
In theory I wouldn't have to write the Fake class at all, but sadly sinon doesn't let you stub an interface directly, only classes.

@here/harp-mapview/test/stubProjection.ts
atomicsulfate: good point, the identity projection is enough for what I need.

@here/harp-mapview/test/TextElementsRendererTestFixture.ts
atomicsulfate: the test is simulating how the terrain tiles become available in the different frames.
At the beginning of a frame, that function "terrainTiles()" is called with the indices of the terrain tiles that should be available at that moment. For those available, getDisplacementMap must return the displacement map. For any other tiles that function returns undefined, as it's defined just above that loop.

@here/harp-mapview/test/stubElevationProvider.ts
atomicsulfate: besides, getDisplacementMap stub is configured in the test to return different results depending on the passed tile key (see my answer below).

@here/harp-mapview/test/stubElevationProvider.ts
atomicsulfate: see my answer above.

@here/harp-mapview/test/TextElementsRendererTestFixture.ts
nzjony: ok, makes sense.

@here/harp-mapview/test/stubElevationProvider.ts
nzjony: getDisplacementMap makes sense, and it makes sense to return a stub, I'm more interested to know why you need to have callsFake for getHeight and sampleHeight, couldn't they just be in the fake class?

atomicsulfate: createStubInstance() creates a fully stubbed instance of the ElevationProviderFake, so none of the method implementations are preserved. That's why I set the behaviour later with callsFake().

PR #1265, opened 2020-02-05
Sync chromedriver version with actual chrome in CI.

zbigg:
Travis CI always uses "current stable" version of chrome, and `chromedriver` has very limited compatibility (usually 1 or 2 major versions max).

Detect actual chrome version on system and upgrade chromedriver used for browser integration testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ branches:
     - master
     - release
 
+env:
+  - DETECT_CHROMEDRIVER_VERSION=true
+
 # upgrade yarn to a more recent version
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.13.0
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       script:
         - yarn pre-test
         - yarn cov-test --forbid-only
+        - ./scripts/install-chromedriver.sh
```
ninok: MINOR: installing chromedriver should be the first step b/c we might use it in pre-test or cov-test in the future

zbigg: @ninok Finally i've skipped this script and chosen to use `DETECT_CHROMEDRIVER_VERSION` introduced recently.
Tihs needs update too, so it's hard to tell if the feature works though - see new version of patch,

PR #1267, opened 2020-02-06
Optimize interpolants usage.

zbigg:
Optimize InteprolatedProperty to reuse Interpolant instances across
evaluations.

THREE.JS interpolants cache last interpolation range, so assuming that
we usually render value on similar zoom levels, we don't have search for
interpolation range on every frame.

Time impact: O(1) complexity of most of interpolations [1].

Mem impact: don't allocate interpolant + one array internally on each
evaluation.

Based on #1332, review only last commit.

[1] https://threejs.org/docs/#api/en/math/Interpolant


zbigg:
@heremaps-bot ok to test

zbigg:
> How about we remove all this code and simply add what’s needed to interpolate in 
> InterpolateOperators.ts?

Great idea, but that's just moving code around. 

This PR  is not about nor supporting non-zoom interpolation but about reducing waste in CPU and GC load.

PR #1268, opened 2020-02-06
Better handling of olp data

haraldF:
http/204 errors are now propagated as exceptions, not as replies.
Also handle the version in OlpCopyrightProvider as the license might
change between catalog versions.

### DIFF
--- a/@here/harp-olp-utils/lib/OlpDataProvider.ts
+++ b/@here/harp-olp-utils/lib/OlpDataProvider.ts
@@ -103,10 +103,17 @@ export class OlpDataProvider implements DataProvider {
                 abortSignal
             )
             .then(response => {
-                // 204 - NO CONTENT, no data exists at the given tile. Do nothing.
-                return response.status === 204 ? Promise.resolve({}) : response.arrayBuffer();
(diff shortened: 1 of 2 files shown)


@here/harp-olp-utils/lib/OlpDataProvider.ts
```diff
@@ -103,10 +103,17 @@ export class OlpDataProvider implements DataProvider {
                 abortSignal
             )
             .then(response => {
-                // 204 - NO CONTENT, no data exists at the given tile. Do nothing.
-                return response.status === 204 ? Promise.resolve({}) : response.arrayBuffer();
+                if (response.status !== 200) {
+                    throw new Error(response.statusText);
```
CraigBerlin: If response.status is 204 will this properly set error.status for the catch block below?

haraldF: @CraigBerlin http/get requests on Blob Store will never return 204, and if it does, it's an actual error that we should report (and stop processing the tile).

204 will only happen if the index indicates that the tile is "empty", e.g. there's no data for that tile.

PR #1271, opened 2020-02-06
Skip completly transparent objects.

zbigg:
Skip completly transparent objects, so then don't land in render
pipeline if opacity===0.

* Tile: add updateDynamicObjects callback that shall be called by
  MapView to update dynamic objects explicitly

* MapView: Call `updateDynamicObjects` on all rendered tiles and then don't add
  completly transparent objects to scene

* TileGeometryCreator Refactor most of `onBeforeRender` callbacks so they are executed
  in controlled (in scope of `Tile.updateDynamicObjects`)

* Unrelated - don't install fading render helper at all if fading is
  disabled

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2955,7 +2955,17 @@ export class MapView extends THREE.EventDispatcher {
     private renderTileObjects(tile: Tile, zoomLevel: number) {
         const worldOffsetX = tile.computeWorldOffsetX();
         if (tile.willRender(zoomLevel)) {
+            tile.updateDynamicObjects();
+
             for (const object of tile.objects) {
+                // Don't add completly transparent objects.
+                if ("material" in object && (object as any).material instanceof THREE.Material) {
+                    const material: THREE.Material = (object as any).material;
+                    if (material.opacity === 0) {
+                        continue;
+                    }
+                }
+
                 object.position.copy(tile.center);
                 if (object.displacement !== undefined) {
                     object.position.add(object.displacement);
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -257,6 +257,14 @@ export interface TextElementIndex {
 }
 
 /**
+ * Updater for dynamic objects (materials, scene objects, [[TextElement]]s).
+ */
+export type DynamicObjectUpdater = () => void;
+
+// WIP In future, it will accept all params needed for effective expression evaluation.
+//export type DynamicObjectUpdater = (env: MapEnv, cache?: Map<Expr, Value>) => void;
+
+/**
  * The class that holds the tiled data for a [[DataSource]].
  */
 export class Tile implements CachedResource {
@@ -383,6 +391,8 @@ export class Tile implements CachedResource {
     private m_ownedTextures: WeakSet<THREE.Texture> = new WeakSet();
 
     private m_animatedExtrusionTileHandler: AnimatedExtrusionTileHandler | undefined;
+
+    private m_dynamicObjectsUpdaters: DynamicObjectUpdater[] = [];
 
     /**
      * Creates a new [[Tile]].
@@ -651,6 +661,23 @@ export class Tile implements CachedResource {
      */
     didRender(): void {
         // to be overridden by subclasses
+    }
+
+    /**
+     * Update dynamic objects (materials, scene objects, [[TextElement]]s) on this tile.
+     */
+    updateDynamicObjects(): void {
+        for (const updater of this.m_dynamicObjectsUpdaters) {
+            updater();
+        }
+    }
+
+    /**
+     * Add `updater` callback, so it will be called before this tile is rendered within
+     * [[updateDynamicObjects]].
+     */
+    addUpdater(updater: DynamicObjectUpdater) {
+        this.m_dynamicObjectsUpdaters.push(updater);
     }
 
     /**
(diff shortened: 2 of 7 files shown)


@here/harp-mapview/lib/Tile.ts
```diff
@@ -639,6 +646,23 @@ export class Tile implements CachedResource {
         // to be overridden by subclasses
     }
 
+    /**
+     * Update dynamic objects on this tile.
+     */
+    updateDynamicObjects(): void {
+        for (const updater of this.m_dynamicObjectsUpdaters) {
```
ninok: Can we make this more explicit instead of registering some black box callbacks.
With callbacks it is hard to see how many are registered and what they are doing b/c they are registered in a completely different place.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -711,55 +711,56 @@ export class TileGeometryCreator {
...
-                                      technique.color,
-                                      mapView.zoomLevel
-                                  );
-                              }
-                            : undefined
+                        fadingParams.fadeFar
                     );
+
+                    if (Expr.isExpr(technique.color) || Expr.isExpr(technique.opacity)) {
+                        tile.addUpdater(() => {
```
ninok: As far as I can see we still register one callback per object, even if the material is the same for all objects.
I would prefer to have a function like `Tile.updateMaterials` and that one is explicitly iterating over the materials/techniques and updates them.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -711,55 +711,56 @@ export class TileGeometryCreator {
...
+                    }
                 }
 
                 if (isSolidLineTechnique(technique)) {
-                    const hasDynamicColor =
-                        Expr.isExpr(technique.color) || Expr.isExpr(technique.opacity);
+                    const lineMaterial = material as SolidLineMaterial;
+
                     const fadingParams = this.getFadingParams(displayZoomLevel, technique);
                     FadingFeature.addRenderHelper(
```
ninok: Can we also get rid of this render helper. It does not help anybody. Instead the material updated should also be part of an explicit method in Tile

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
zbigg: Agree, my bad i didn't noticed that we always call update per object, not material even if only one update is needed.
Will fix.

zbigg: I tried to be generic, as these updaters are meant to update not only materials, but also 
- `TextElement`s - at least `priority`, `color`, `size` has to be interpolated and it was reported as bug.
-  possibly object properties (like enable/disable depth pre-pass or even remove edge/outline objects if they become unnecessary when their size/width interpolate to 0 or null) 

In next followup PoC PR, i'll show my idea, in this one i'll try to focus on moving updates to controlled place (outside of onBeforeRender), so `object.material.opacity` & `textElement.renderStyle,opacity` is always updated in controlled manner.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
zbigg: Some kind of helper is still needed, but it will be moved to TIleGeometryCreator and gratly simplified.

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -536,22 +536,27 @@ export namespace FadingFeature {
      * actual camera setup.
      * @param fadeNear The fadeNear value to set in the material.
      * @param fadeFar The fadeFar value to set in the material.
-     * @param updateUniforms If `true`, the fading uniforms are set. Not required if material is
-     *          handling the uniforms already, like in a [[THREE.ShaderMaterial]].
-     * @param additionalCallback If defined, this function will be called before the function will
-     *          return.
      */
     export function addRenderHelper(
         object: THREE.Object3D,
```
nzjony: This chould be changed to take a TileObject, to simplify access to the material.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -256,6 +256,11 @@ export interface TextElementIndex {
     elementIndex: number;
 }
 
+export type DynamicObjectUpdater = () => void;
```
nzjony: Please add a comment what a dynamic object is, i.e. it is one where the material needs to update at runtime, might be obvious for us, but not so obvious for new people starting the project.

@here/harp-materials/lib/MapMeshMaterials.ts
zbigg: TIleObject is  in MapView so we would end with circular dependency ... or unrelated refactor.
Anyway, this code will be removed.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2913,7 +2913,17 @@ export class MapView extends THREE.EventDispatcher {
     private renderTileObjects(tile: Tile, zoomLevel: number) {
         const worldOffsetX = tile.computeWorldOffsetX();
         if (tile.willRender(zoomLevel)) {
+            tile.updateDynamicObjects();
+
             for (const object of tile.objects) {
+                // Don't add completly transparent objects.
+                if ("material" in object && (object as any).material instanceof THREE.Material) {
```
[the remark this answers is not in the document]
zbigg: Not needed anymore.

PR #1272, opened 2020-02-07
Fixes #1270. Construct correct Raycaster

### LINKED ISSUE #1270: map.intersectMapObjects(e.pageX, e.pageY) notWork
**Describe the bug**
after adding litener for click to [examle](https://www.harp.gl/docs/master/examples/#datasource_features_lines-and-points.html)
`
map.addDataSource(featuresDataSource).then(() => {
        map.canvas.addEventListener("click", (e: MouseEvent) => {
            const intersectionResults = map.intersectMapObjects(e.pageX, e.pageY);
            const usableResults = intersectionResults.filter(result => result.userData !== undefined);
            console.log(intersectionResults, usableResults)
        });
    });
`
has [error](http://tinyurl.com/yx89to8v) in console

nzjony:
Fixes #1270

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -757,7 +757,7 @@ export class MapView extends THREE.EventDispatcher {
     private m_enablePolarDataSource: boolean = true;
 
     // gestures
-    private readonly m_raycaster = new PickingRaycaster(this);
+    private readonly m_raycaster: PickingRaycaster;
     private readonly m_plane = new THREE.Plane(new THREE.Vector3(0, 0, 1));
     private readonly m_sphere = new THREE.Sphere(undefined, EarthConstants.EQUATORIAL_RADIUS);
 
@@ -966,6 +966,8 @@ export class MapView extends THREE.EventDispatcher {
         // setup camera with initial position
         this.setupCamera(options);
         this.m_targetDistance = this.m_camera.position.distanceTo(this.m_targetWorldPos);
+
+        this.m_raycaster = new PickingRaycaster(width, height);
 
         this.m_movementDetector = new CameraMovementDetector(
             this.m_options.movementThrottleTimeout,
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -51,6 +51,7 @@ import { SimpleTileGeometryManager, TileGeometryManager } from "./geometry/TileG
 import { MapViewImageCache } from "./image/MapViewImageCache";
 import { MapViewFog } from "./MapViewFog";
 import { PickHandler, PickResult } from "./PickHandler";
+import { PickingRaycaster } from "./PickingRaycaster";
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -749,7 +750,7 @@ export class MapView extends THREE.EventDispatcher {
     private m_enablePolarDataSource: boolean = true;
 
     // gestures
-    private readonly m_raycaster = new THREE.Raycaster();
+    private readonly m_raycaster = new PickingRaycaster(this);
```
ninok: I would prefer to get rid of this `PickingRaycaster` or at least not keep the whole `MapView` reference b/c as we know from history people will do bad things if the have access to the whole `MapView`.
As far as I can see, the `PickingRaycaster` is only used in `MapViewPoints` and there we use the camera and the canvas size. Would be cleaner to add these properties explicitly and we would also avoid to query the DOM for every pick event.
Long story short: I would prefer if you do the changes now but if you don't have time, please create a ticket.

nzjony: I will fix now, seems to be a small change.

@here/harp-mapview/lib/PickingRaycaster.ts
```diff
@@ -5,21 +5,21 @@
...
+ * space, used for example in the [[MapViewPoints]].
  */
 export class PickingRaycaster extends THREE.Raycaster {
     /**
-     * Constructs a `MapViewRaycaster`. It keeps a reference to [[MapView]] in order to access the
-     * active camera when calling the custom method `raycastPoints`.
+     * Constructor.
      *
-     * @param m_mapView the active [[MapView]].
+     * @param camera the camera used for projection to screen.
```
ninok: Maybe you can explain a bit more that this camera is the world camera (i.e. equal to `MapView.camera`) whereas the `Raycaster.camera` is our internal RTE camera (i.e. equal to `MapView.m_rteCamera`).
Maybe it also makes sense to rename camera to worldCamera b/c otherwise we have two properties with the same name `camera` (I wonder that typescript is not complaining abut this)

nzjony: I have fixed the PickingRaycaster to take the rteCamera.

@here/harp-mapview/lib/PickingRaycaster.ts
```diff
@@ -5,21 +5,22 @@
...
+     * Constructor.
      *
-     * @param m_mapView the active [[MapView]].
+     * @param camera the camera used for projection to screen. Note, this must use the camera used
+     * to finally render the scene and not the camera in world space.
+     * @param width the canvas width.
+     * @param height the canvas height.
      */
-    constructor(public mapView: MapView) {
+    constructor(public camera: THREE.Camera, public width: number, public height: number) {
```
ninok: Why do we still need the camera as a member? You should just forward the camera to the Raycaster constructor

nzjony:
@heremaps-bot  retest this please

PR #1275, opened 2020-02-07
HARP-8521: Fix extruded buildings overlay.

atomicsulfate:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-materials/lib/DisplacementFeature.ts
+++ b/@here/harp-materials/lib/DisplacementFeature.ts
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { HiddenThreeJSMaterialProperties } from "./MapMeshMaterials";
+
+/**
+ * Parameters used when constructing a new implementor of [[DisplacementFeature]].
+ */
+export interface DisplacementFeatureParameters {
+    /**
+     * Texture used for vertex displacement along their normals.
+     */
+    displacementMap?: THREE.Texture;
+}
+
+/**
+ * Interface to be implemented by materials that use displacement maps to overlay geometry
+ * on elevation data.
+ */
+export interface DisplacementFeature extends HiddenThreeJSMaterialProperties {
+    displacementMap: THREE.Texture | null;
+}
+
+/**
+ * Determines whether a given material supports displacement maps for elevation overlay.
+ * @param material The material to check.
+ * @returns Whether the given material supports displacement maps for elevation overlay.
+ */
+export function hasDisplacementFeature(material: any): material is DisplacementFeature {
+    return "displacementMap" in material;
+}
--- a/@here/harp-materials/lib/EdgeMaterial.ts
+++ b/@here/harp-materials/lib/EdgeMaterial.ts
@@ -6,9 +6,8 @@
 
 import * as THREE from "three";
 
+import { DisplacementFeature, DisplacementFeatureParameters } from "./DisplacementFeature";
 import {
-    DisplacementFeature,
-    DisplacementFeatureParameters,
     ExtrusionFeature,
     ExtrusionFeatureParameters,
     FadingFeature,
@@ -256,16 +255,16 @@ export class EdgeMaterial extends THREE.RawShaderMaterial
         setShaderMaterialDefine(this, "USE_EXTRUSION", useExtrusion);
     }
 
-    get displacementMap(): THREE.Texture | undefined {
+    get displacementMap(): THREE.Texture | null {
         return this.uniforms.displacementMap.value;
     }
 
-    set displacementMap(map: THREE.Texture | undefined) {
+    set displacementMap(map: THREE.Texture | null) {
         if (this.uniforms.displacementMap.value === map) {
             return;
         }
         this.uniforms.displacementMap.value = map;
-        const useDisplacementMap = map !== undefined;
+        const useDisplacementMap = map !== null;
         if (useDisplacementMap) {
             this.uniforms.displacementMap.value.needsUpdate = true;
         }
--- a/@here/harp-materials/lib/MapMeshMaterials.ts
+++ b/@here/harp-materials/lib/MapMeshMaterials.ts
@@ -10,6 +10,7 @@ import { disableBlending, enableBlending, insertShaderInclude, setShaderDefine }
 
 import * as THREE from "three";
 
+import { DisplacementFeature, DisplacementFeatureParameters } from "./DisplacementFeature";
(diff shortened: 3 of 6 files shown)


@here/harp-materials/lib/DisplacementFeature.ts
```diff
@@ -0,0 +1,34 @@
...
+     */
+    displacementMap?: THREE.Texture;
+}
+
+/**
+ * Interface to be implemented by materials that use displacement maps to overlay geometry
+ * on elevation data.
+ */
+export interface DisplacementFeature extends HiddenThreeJSMaterialProperties {
+    displacementMap: THREE.Texture | null;
```
atomicsulfate: I've changed the displacementMap type to match the one in THREE.js MeshStandardShader (THREE.Texture | null). That way the interface is homogeneous across all materials with the displacement feature and they can be handled in the same way (using displacementMap getter and setter).

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -1172,4 +1198,3 @@ applyMixinsWithoutProperties(MapMeshStandardMaterial, [FadingFeatureMixin]);
 applyMixinsWithoutProperties(MapMeshBasicMaterial, [ExtrusionFeatureMixin]);
 applyMixinsWithoutProperties(MapMeshStandardMaterial, [ExtrusionFeatureMixin]);
 applyMixinsWithoutProperties(MapMeshBasicMaterial, [DisplacementFeatureMixin]);
-applyMixinsWithoutProperties(MapMeshStandardMaterial, [DisplacementFeatureMixin]);
```
atomicsulfate: we were applying the DisplacementFeatureMixin to a material that already has a displacement map.

kkostecki: Wright... I did not spotted it before :+1: .

@here/harp-materials/lib/EdgeMaterial.ts
```diff
@@ -256,16 +255,16 @@ export class EdgeMaterial extends THREE.RawShaderMaterial
...
     }
 
-    set displacementMap(map: THREE.Texture | undefined) {
+    set displacementMap(map: THREE.Texture | null) {
         if (this.uniforms.displacementMap.value === map) {
             return;
         }
         this.uniforms.displacementMap.value = map;
-        const useDisplacementMap = map !== undefined;
+        const useDisplacementMap = map !== null;
```
kkostecki: Add similar change to c-tor:
```
const hasDisplacementMap = params !== undefined && params.displacementMap !== undefined;
```
Now is checking for undefined value.

Check also this line:
```
 displacementMap: new THREE.Uniform(
                    hasDisplacementMap ? params!.displacementMap : new THREE.Texture()
```

@here/harp-materials/lib/DisplacementFeature.ts
kkostecki: I am wondering if all c-tors are handling DisplacementFeatureParameters correctly now, cause we have mismatch here, undefined !== null.
Please verify it.

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -323,26 +312,33 @@ export namespace DisplacementFeature {
...
     uniformsNeedUpdate?: boolean;
     defines?: any;
     shaderDefines?: any;
     shaderUniforms?: UniformsType;
     onBeforeCompile?: CompileCallback;
-    private m_displacementMap?: THREE.Texture;
+    private m_displacementMap: THREE.Texture | null = null;
 
-    protected getDisplacementMap(): THREE.Texture | undefined {
+    get displacementMap(): THREE.Texture | null {
```
kkostecki: What about:
```
DisplacementFeatureMixin.addDisplacementProperties()
```
which already defines (adds) this getter/setter?

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -733,15 +729,15 @@ export class SolidLineMaterial extends THREE.RawShaderMaterial
         setShaderMaterialDefine(this, "USE_FADING", value > 0.0);
     }
 
-    get displacementMap(): THREE.Texture | undefined {
+    get displacementMap(): THREE.Texture | null {
```
kkostecki: This will never return null, we need to probably change c-tor section:
```
displacementMap: new THREE.Uniform(
                        displacementMap !== undefined ? displacementMap : new THREE.Texture()
                    ),
```

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -1101,6 +1102,27 @@ export class MapMeshStandardMaterial extends THREE.MeshStandardMaterial
         return this;
     }
 
+    /** @override */
+    get displacementMap(): THREE.Texture | null {
+        return super.displacementMap === undefined ? null : super.displacementMap;
```
kkostecki: `THREE.MeshStandardMaterial` defines `displacementMap` as `THREE.Texture` or `null`, not undefined, but ok it will work anyway. Have you found from THREE.js code that it may be undefined in some scenario?

@here/harp-materials/lib/DisplacementFeature.ts
atomicsulfate: I double checked, everything's fine.

@here/harp-materials/lib/EdgeMaterial.ts
atomicsulfate: That line is ok, there's no changes to the type of params.displacementMap.

@here/harp-materials/lib/MapMeshMaterials.ts
atomicsulfate: I needed to add it there so that tslint didn't complained about a missing property from the DisplacementFeature interface. 
I've added a comment to clarify that.

@here/harp-materials/lib/SolidLineMaterial.ts
atomicsulfate: That c-tor section doesn't need to change, the displacementMap checked there comes from the DisplacementParameters interface, which has the type "DisplacementMap|undefined".

The getter will return null if someone sets a null displacementMap with the setter.

@here/harp-materials/lib/MapMeshMaterials.ts
atomicsulfate: Yes, it's actually undefined by default. Threejs MeshStandardMaterialParameters has this type for the displacementMap:
displacementMap?: Texture | null;

If it's undefined in the parameters passed to the constructor (as it is right now in our code), the displacementMap member in the material is left with an undefined value.

kkostecki: Ok.

PR #1276, opened 2020-02-07
Remove bogus colorMap from TextStyleCache.

zbigg:
Removing bogus code.

### DIFF
--- a/@here/harp-mapview/lib/text/TextStyleCache.ts
+++ b/@here/harp-mapview/lib/text/TextStyleCache.ts
@@ -171,11 +171,6 @@ export interface TextElementStyle {
 export class TextStyleCache {
     private m_textRenderStyleCache = new TextRenderStyleCache();
     private m_textLayoutStyleCache = new TextLayoutStyleCache();
-    /**
-     * Cache for named colors.
-     */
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/text/TextStyleCache.ts
```diff
@@ -373,7 +370,7 @@ export class TextStyleCache {
                     backgroundOpacity = backgroundOpacity * alpha;
                     hexBgColor = ColorUtils.removeAlphaFromHex(hexBgColor);
                 }
-                this.m_colorMap.set(cacheId + "_bg", ColorCache.instance.getColor(hexBgColor));
```
nzjony: was there some reason for having the background colors separately cached?

zbigg: I don't know, this code looks like leftover as current this method is merge of duplicated functions from PoiRenderer/TileGeometryCreator.
Look, that this entries in `m_colorMap` always rewritten, so there is no caching at all. Effectively, it was obfuscated local variable. That's why i changed it to local variables.

nzjony: Ok

PR #1280, opened 2020-02-10
HARP-8521: Fix TileGeometryManager.hideKind()

atomicsulfate:
Some of the kinds in a set or array were not added due to the shortcircuited
evaluation of the or conditions.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/geometry/TileGeometryManager.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryManager.ts
@@ -193,23 +193,18 @@ export abstract class TileGeometryManagerBase implements TileGeometryManager {
         addOrRemoveToHiddenSet: boolean = true
     ): void {
         let visibilityHasChanged = false;
-        if (Array.isArray(kind)) {
-            for (const oneKind of kind as GeometryKind[]) {
-                visibilityHasChanged =
-                    visibilityHasChanged ||
-                    this.addRemove(this.hiddenKinds, oneKind, addOrRemoveToHiddenSet);
+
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/geometry/TileGeometryManager.ts
```diff
@@ -193,23 +193,15 @@ export abstract class TileGeometryManagerBase implements TileGeometryManager {
...
-            const kindSet = kind as GeometryKindSet;
-            for (const oneKind of kindSet) {
+
+        if (Array.isArray(kind) || kind instanceof Set) {
+            for (const oneKind of kind) {
                 visibilityHasChanged =
-                    visibilityHasChanged ||
-                    this.addRemove(this.hiddenKinds, oneKind, addOrRemoveToHiddenSet);
+                    this.addRemove(this.hiddenKinds, oneKind, addOrRemoveToHiddenSet) ||
+                    visibilityHasChanged;
```
kkostecki: I would even go further, create temp bool as result of `addRemove` and then make boolean alternative.
```
const visibilityChange = this.addRemove(this.hiddenKinds, oneKind, addOrRemoveToHiddenSet)
visibilityHasChanged = visibilityHasChanged || visibilityChange;
```
This is assuming that we want always call "addRemove". Previous code (and even) current, may be prone for optimization, I am not sure if any release of JIT compiler may ready it in revese order, as least I don't know if it is writtern in any standard, so we may do in cleaner and for sure.

atomicsulfate: done.

PR #1283, opened 2020-02-10
UrlUtils: URI reference resolution cross-platform

vega-jose:
In order to determine if a path of a referenced resource in a scheme (or child
URI) is absolute, the string is tested against a regular expression which only
covered non-Windows paths. This led to problems when actually working with
resource paths on Windows.

This change allows to detect absolute paths in a Windows-format as well

Signed-off-by: ESTIB <jose.vega@here.com>

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-utils/lib/UrlUtils.ts
+++ b/@here/harp-utils/lib/UrlUtils.ts
@@ -47,7 +47,7 @@ export function resolveReferenceUri(parentUri: string | undefined, childUri: str
     }
 }
 
-const absoluteUrlWithOriginRe = new RegExp("^(?:[a-z]+:)?//", "i");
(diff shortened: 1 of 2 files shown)


@here/harp-utils/lib/UrlUtils.ts
```diff
@@ -47,7 +47,7 @@ export function resolveReferenceUri(parentUri: string | undefined, childUri: str
     }
 }
 
-const absoluteUrlWithOriginRe = new RegExp("^(?:[a-z]+:)?//", "i");
+const absoluteUrlWithOriginRe = new RegExp(/^(?:[a-z]+:)?(\/\/?|\\\\?)/i);
```
ninok: you can remove the `new RegExp` otherwise you allocate two regexp objects

zbigg:
@heremaps-bot ok to test

PR #1288, opened 2020-02-12
HARP-8604: Fix for multiple time tile loading using setTimeout() call.

kkostecki:
The decoded tiles geometry was loaded/generated multiple times, both with
techniques processing in SimpleTileGeometryLoader.
This patchset fixes the flow so the request for tile load after decoding is
processed only once.
It also adds some sanity checks that ensures us that disposed or unvisible
tiles are no longer loaded/processed.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
@@ -155,6 +155,8 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
     private m_decodedTile?: DecodedTile;
     private m_isFinished: boolean = false;
     private m_availableGeometryKinds: GeometryKindSet | undefined;
+    private m_enabledKinds: GeometryKindSet | undefined;
+    private m_disabledKinds: GeometryKindSet | undefined;
     private m_timeout: any;
 
     constructor(private m_tile: Tile) {}
@@ -165,6 +167,13 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
 
     get isFinished(): boolean {
         return this.m_isFinished;
+    }
+
+    get geometryCreationPending(): boolean {
+        // Geometry loading not yet finished and timeout already set, but not yet processing
+        // (m_timeout !== undefined), or timeout callback already in progress
+        // (m_timeout === undefined), but decoded tile still not reset.
+        return !this.isFinished && this.m_decodedTile !== undefined;
     }
 
     get basicGeometryLoaded(): boolean {
@@ -204,10 +213,35 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
     ): void {
         const tile = this.tile;
 
-        // First time this tile is handled:
-        if (this.m_decodedTile === undefined && tile.decodedTile !== undefined) {
-            TileGeometryCreator.instance.processTechniques(tile, enabledKinds, disabledKinds);
+        // Cheap sanity check, do it first so no longer processing is needed.
+        if (this.isFinished) {
+            return;
+        }
 
+        // Check if tile should be already discarded (invisible, disposed).
+        // If the tile is not ready for display, or if it has become invisible while being loaded,
+        // for example by moving the camera, the tile is not finished and its geometry is not
+        // created. This is an optimization for fast camera movements and zooms.
+        if (this.discardNeedlessTile(tile)) {
+            return;
+        }
+
+        // Geometry kinds have changed when loading, if so reset entire loading because
+        // this geometry loader generates all geometry at once.
+        // TODO: Probably the update() interface will change soon, when phased loading support
+        // will be removed, then this code may be no longer necessary.
+        if (
+            this.geometryCreationPending &&
+            !this.compareGeometryKinds(enabledKinds, disabledKinds)
+        ) {
+            this.reset();
+        }
+
+        // First time this tile is handled, or reset has been requested.
+        // Note: Finished condition already checked above.
+        if (!this.geometryCreationPending && tile.decodedTile !== undefined) {
+            TileGeometryCreator.instance.processTechniques(tile, enabledKinds, disabledKinds);
+            this.setGeometryKinds(enabledKinds, disabledKinds);
             this.setDecodedTile(tile.decodedTile);
             this.prepareForRender(enabledKinds, disabledKinds);
         }
@@ -215,24 +249,34 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
 
     dispose(): void {
         this.m_decodedTile = undefined;
+        // TODO: Release other resource: availableGeometryKind, enabled/disabled sets, timeout?
     }
 
     reset(): void {
+        this.m_availableGeometryKinds?.clear();
+        this.m_enabledKinds?.clear();
+        this.m_disabledKinds?.clear();
+
         this.m_decodedTile = undefined;
         this.m_isFinished = false;
-        if (this.m_availableGeometryKinds !== undefined) {
-            this.m_availableGeometryKinds.clear();
-        }
+
         if (this.m_timeout !== undefined) {
             clearTimeout(this.m_timeout);
+            this.m_timeout = undefined;
         }
     }
 
     private finish() {
         this.m_tile.loadingFinished();
         this.m_tile.removeDecodedTile();
+
+        this.m_decodedTile = undefined;
         this.m_isFinished = true;
-        this.m_timeout = undefined;
+
+        if (this.m_timeout !== undefined) {
+            clearTimeout(this.m_timeout);
+            this.m_timeout = undefined;
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -246,16 +256,33 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
...
     ) {
+        if (this.isFinished) {
+            return;
+        }
         // If the tile is not ready for display, or if it has become invisible while being loaded,
         // for example by moving the camera, the tile is not finished and its geometry is not
         // created. This is an optimization for fast camera movements and zooms.
         const tile = this.tile;
         const decodedTile = this.m_decodedTile;
         this.m_decodedTile = undefined;
```
kkostecki: Anybody knows why this was set (garbage collector, rejections)? This is more or less the reason of entire problems, the isLoading method fixes it using different member, but ideally I would remove this line and change isLoading getter to:
```
get isLoading(): boolean {
return this.m_decodedTile !== undefined && this.m_timeout !== undefined;
}
```

nzjony: I also wondered about this, and even created a ticket for it: HARP-8742

Even though it is removed, it still exists in the TileLoader, so we aren't saving any memory.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -205,7 +209,7 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
         const tile = this.tile;
 
         // First time this tile is handled:
-        if (this.m_decodedTile === undefined && tile.decodedTile !== undefined) {
+        if (this.m_decodedTile === undefined && tile.decodedTile !== undefined && !this.isLoading) {
```
kkostecki: Using ```!this.isLoading``` check here stops processing both techniques and tile geometry load, but this way we will need to _wait_ until timeout to cancel invisible or disposed tile loading. Other method would be to check this condition in the update section (as in prepareForRender) and if tile is no longer visible cancel timeout.
This may give us some small improvements so less timeouts may be queued while panning.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -229,10 +236,13 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
         }
         if (this.m_timeout !== undefined) {
             clearTimeout(this.m_timeout);
+            this.m_timeout = undefined;
         }
     }
 
     private finish() {
+        // TODO: Would be easier to set m_decodedTile = undefined, then isLoading would check its
+        // value and if m_timeout is set.
```
kkostecki: In my opinion we should set:
```
this.m_decodedTile = undefined;
```
here.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -246,16 +256,33 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
...
         const decodedTile = this.m_decodedTile;
         this.m_decodedTile = undefined;
+        // TODO: Entire block could be moved to update method
         if (decodedTile === undefined || tile.disposed || !tile.isVisible) {
+            // TODO: Should we clearTimer here?
+            if (this.m_timeout !== undefined) {
+                clearTimeout(this.m_timeout);
+            }
+            // TODO: Should we dispose tile here?
+            //if (!tile.isVisible) {
```
kkostecki: This is the block that could be moved to update() function in order to cancel loading (timeout) unnecessary tiles immediately, otherwise once loading started it will happen when timeout is processed.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
zbigg: I confirm that originally this was to save memory, so decoded tile is wiped after closure passed to SetTimeout disapears ... then `*GeometryLoaders` were born and everything changed.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -246,17 +286,28 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
...
+            //    tile.mapView.visibleTileSet.disposeTile(tile);
+            //}
             this.finish();
             return;
         }
+
         this.m_timeout = setTimeout(() => {
+            this.m_timeout = undefined;
```
ninok: In my opinion we should not misuse the browser task queue as tile load queue.
This would all be much cleaner and more flexible if we would have a dedicated queue were we push new tiles to and only have the setTimeout fetching one tile from the queue.

In this way we would also be able to take the tile with the highest priority from the queue in the future.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -320,4 +378,73 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
...
+     *
+     * Method compares input sets with recently used geometry kinds in performance wise
+     * manner, taking special care of undefined and zero size sets.
+     *
+     * @param enabledKinds Set of geometry kinds to be displayed or undefined.
+     * @param disabledKinds Set of geometry kinds that won't be rendered.
+     * @return `true` only if sets are logically equal, meaning that undefined and empty sets
+     * may result in same geometry (techniques kind) beeing rendered.
+     */
+    private compareGeometryKinds(
```
ninok: How is this related to the task queue.
I think nobody is using geometry kind filtering so I would rather drop that feature than adding even more code b/c of a feature nobody uses.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
kkostecki: So the newest implementation will make it undefined only after TileGeometryLoaded has finished his job. Because decoded tile it stored elsewhere, currently it does not give us any optimization.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
kkostecki: I would agree, but this was introduced for Phased Loading and actually never supported in SimpleLoader! This provides support for it.
In simple words, I have a bad habit that whenever I touch something I am trying to at least comply with the interfaces somebody have prepared.
If we want to remove this we will need to drop PhasedLoading too, I guess and as far as I know there is separate ticket for this.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
kkostecki: I fully agree, just keep in mind that this PR have been introduced to fix multiple buildings extrusion animation that we spot on Firefox..., just look where it goes. Seems that in buildings fix I will need to provide entire new architecture, both with dropping some old features (Phased Loading?). Do we really want to go this way?
Maybe separate tickets should be created for all that steps..

kkostecki:
> I review when there is a test.

I will provide test, as soon as we agree on current architecture used...

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
kkostecki: Just a side note, maybe instead of trying to support kinds update, we could simply leave it as it was but simply add:
```
assert(enabledKinds == undefined || enabledKinds.size() === 0);
assert(disabledKinds == undefined || disabledKinds.size() === 0);
```
but this would assume that no one is using it... I heard that this is not purely true...

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
kkostecki: Separate ticket created for it.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
ninok: I like your second proposal to avoid adding more dead code.
Right now we anyways crash or assert when phased loading is enabled. So we should just ignore the `enablePhaseLoading` flag and never enable it.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -167,6 +169,13 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
         return this.m_isFinished;
     }
 
+    get isLoading(): boolean {
```
ninok: Remark for refactoring later (i.e. proper TileLoadQueue)
- `isLoading` is confusing `geometryCreationPending` would be more clear
- We should have a proper state enum in the Tile like:
```
enum TileState {
    Loading, // request was send to the server
    Decoding, // decoding in the worker
    GeometryCreation, // THREE.js object creation
    Finished,
    Failed
}
```
Right now it is really hard to get the tile state b/c there are so many components that need to be checked (`TileLoader`, `TileGeometryLoader`, ...)

kkostecki: Ok done.

PR #1289, opened 2020-02-12
HARP-8768: Fix cancellation handling in OlpDataProvider.

zbigg:
Don't handle `AbortError` in OlpDataProvider.getTile as this should
be loaded up in call chain by `TileLoader`.

Related-to: HARP-8768

### DIFF
--- a/@here/harp-olp-utils/lib/OlpDataProvider.ts
+++ b/@here/harp-olp-utils/lib/OlpDataProvider.ts
@@ -52,31 +52,27 @@ export class OlpDataProvider implements DataProvider {
      *
      * @returns A promise which is resolved when the connection has been established.
      */
-    connect(): Promise<void> {
+    async connect(): Promise<void> {
         const settings = new OlpClientSettings({
             environment: "here",
             getToken: this.params.getToken
         });
+
         if (this.params.version !== undefined && this.params.version >= 0) {
-            this.m_versionLayerClient = new VersionedLayerClient(
-                HRN.fromString(this.params.hrn),
-                this.params.layerId,
-                settings
-            );
             this.m_catalogVersion = this.params.version;
-            return Promise.resolve();
         } else {
-            return new CatalogClient(HRN.fromString(this.params.hrn), settings)
-                .getLatestVersion(new CatalogVersionRequest())
-                .then(response => {
-                    this.m_versionLayerClient = new VersionedLayerClient(
-                        HRN.fromString(this.params.hrn),
-                        this.params.layerId,
-                        settings
-                    );
-                    this.m_catalogVersion = response;
-                });
+            const latestVersion = await new CatalogClient(
+                HRN.fromString(this.params.hrn),
+                settings
+            ).getLatestVersion(new CatalogVersionRequest());
+
+            this.m_catalogVersion = latestVersion;
         }
+        this.m_versionLayerClient = new VersionedLayerClient(
+            HRN.fromString(this.params.hrn),
+            this.params.layerId,
+            settings
+        );
     }
 
     /**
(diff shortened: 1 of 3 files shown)


@here/harp-olp-utils/lib/OlpDataProvider.ts
```diff
@@ -93,33 +89,43 @@ export class OlpDataProvider implements DataProvider {
      * @param abortSignal Optional AbortSignal to cancel the request.
      * @returns A promise delivering the data as an [[ArrayBufferLike]], or any object.
      */
-    getTile(tileKey: TileKey, abortSignal?: AbortSignal): Promise<ArrayBufferLike | {}> {
+    async getTile(tileKey: TileKey, abortSignal?: AbortSignal): Promise<ArrayBufferLike | {}> {
```
nzjony: why do you need async?

Why can't you just go:

 return this.m_versionLayerClient.getData( ... ).then(...).catch(...)?

@here/harp-olp-utils/lib/OlpDataProvider.ts
```diff
@@ -52,31 +52,27 @@ export class OlpDataProvider implements DataProvider {
...
-                .getLatestVersion(new CatalogVersionRequest())
-                .then(response => {
-                    this.m_versionLayerClient = new VersionedLayerClient(
-                        HRN.fromString(this.params.hrn),
-                        this.params.layerId,
-                        settings
-                    );
-                    this.m_catalogVersion = response;
-                });
+            const latestVersion = await new CatalogClient(
```
nzjony: why await and not just use .then? The function anyway returns a promise.

@here/harp-olp-utils/test/OlpDataProviderTest.ts
```diff
@@ -0,0 +1,93 @@
...
+
+import * as sinon from "sinon";
+
+import "@here/harp-fetch";
+import { TileKey } from "@here/harp-geoutils";
+import { assertRejected } from "@here/harp-test-utils";
+import { VersionedLayerClient } from "@here/olp-sdk-dataservice-read";
+import { OlpDataProvider } from "../lib/OlpDataProvider";
+
+describe.only("OlpDataProvider", function() {
```
nzjony: 👍

@here/harp-olp-utils/lib/OlpDataProvider.ts
zbigg: As in previous comment - i find async/await cleaner and more robust.

Example of "more robust", now this method never throws "synchronously" (previously it could). Now it returns rejected promise even if "sync-like" precondition failed.
 
In this case, this change is only to be consistent with rest of module.

@here/harp-olp-utils/lib/OlpDataProvider.ts
zbigg: I've simplified original method, as it unnecessarily duplicated chunks of code. and duplication was caused by problems with conditionals when chaining promises with `.then.

As to question, why? Simple i find async/await more clean and robust than `.then` chains.

nzjony: Good point about reducing the duplicated code, and yes it is cleaner in this case.

@here/harp-olp-utils/lib/OlpDataProvider.ts
nzjony: Good point about it not throwing a sync error, though you could argue that in such a case, a sync error likely highlights some precondition missing, and is therefore a programmer error, as opposed to some network issue or some cancellation.

I'm not such a fan of the try / catch block to enable the async in this case. I will add a comment below.

@here/harp-olp-utils/lib/OlpDataProvider.ts
```diff
@@ -93,33 +89,43 @@ export class OlpDataProvider implements DataProvider {
...
+                throw err;
+            }
+            if (response.status !== 200) {
+                throw new Error(response.statusText);
+            }
+            return response.arrayBuffer();
+        } catch (error) {
+            if (error.name === "AbortError" || error.message === "AbortError: Aborted") {
+                // Rethrow abort errors as they shall be handled on higher level.
+                throw error;
```
nzjony: do you need to rethrow the abort error? why not take the if(abortSignal....) code above out of the try / catch?

The same for the response as well, or do I miss something?

zbigg: AbortError may also be thrown by proper implementation of `getData)_` or `arrayBuffer()`that just uses system fetch API.

AbortError should be rethrown because we cannot even handle it and it's not a point when we can decide what to do with this error.

> why not take the if(abortSignal....) code above out of the try / catch?

It can be moved ... but then it doesn't make sense. The real reason for this, is outlined in comment - if `getTile` ignored signal for whatever reason, just handle `abortSignal` API by ourselves and skip call to `arrayBuffer`.

zbigg: And this whole discussion is here only because someone decided to effectively add "catch(...) { log(...); return {}" only not to have duplicate log messages for tile load error (TileLoader would also log if we would rethrow).

So, good catch is no catch or catch that _rethrows_ exceptions it doesn't handle explicitly
Proper catch in this case should be
 - in this case, we handle explcitly HTTPError / 204
 - rethrow everything

But as i find value in verbosity of this logging message (includes catalog nrn), i left it as is with excplicit support for rethrowing `AbortError`s.

@here/harp-olp-utils/lib/OlpDataProvider.ts
zbigg: > though you could argue that in such a case, a sync error likely highlights some precondition missing, and is therefore a programmer error, as opposed to some network issue or some cancellation.

It's not like this. If this is programmer error, then it should be `assert` or but in this case it's null check, so assert is not good because it's removed in production. Uh, oh ... runtimeAssert !?

There is no rule (either in our project or in general) that sync errors are preconditions and async are not. There is nothing that stops programmers to check & throw preconditions asynchronously and it's very common in every JS code base.

What is bad, is mixing sync and async that becomes source of hard-to-find bugs very fast, if for example you write something like this (assuming original getTile):

```
 // not async scope
 getTile(...).then(...).catch(...)
```

Will this catch our precondition ? No.

You need to
```
try {
 getTile(...).then(...).catch(...)
} catch(...) {
}
```
or call it from `async` function with (or without await).

@here/harp-olp-utils/lib/OlpDataProvider.ts
nzjony: Opps, I commented on the wrong line, I mean why do we have the rethrow on line 106, but I guess that it is because the getData implementation as stated may not support it.

Good point about the skip to arrayBuffer, I thought it would throw an error, and so you would automatically skip it, but it doesn't.

@here/harp-olp-utils/lib/OlpDataProvider.ts
nzjony: I'm of the opinion that the precondition should not be "caught", i.e. the program should throw the error and crash as it would in the first `getTile` example above. That makes it easier to spot the bug and fix it. Otherwise the application ends up in some undefined state and you have a fun game of find the rabbit trying to locate the issue.

zbigg:

PR #1291, opened 2020-02-13
HARP-8835: Add support for symbolic render orders

robertoraggi:


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -690,6 +690,36 @@ export class TileGeometryCreator {
                 const hasSolidLinesOutlines: boolean =
                     isSolidLineTechnique(technique) && technique.secondaryWidth !== undefined;
 
+                const theme = mapView.theme;
```
zbigg: I think this logically belongs to decoder.
Even if it's quite cheap, there is no reason to this in this now.

But it requires `Theme` access in decoder ? 
Shall we create another PR that brings theme back to decoder ?

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -183,6 +183,11 @@ export interface BaseTechniqueParams {
      */
     renderOrder: number;
 
+    /**
+     * The category of this technique.
```
zbigg: Add comment that it references `Theme.priorities`, stg like this:

> Sets  `prority` (`renderOrder`) based on index of this category in [[Theme.priorities]].

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
robertoraggi: yep, we can't do that in the decoder because we don't have access to theme there.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
robertoraggi: i added some documentation explaining how to use priorities and labelPriorities. We will drop renderOrder so i'm not sure we should reference it in the documentation for `priorities`

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -690,6 +691,42 @@ export class TileGeometryCreator {
                 const hasSolidLinesOutlines: boolean =
                     isSolidLineTechnique(technique) && technique.secondaryWidth !== undefined;
 
+                const priorities = mapView.theme ? mapView.theme.priorities : undefined;
```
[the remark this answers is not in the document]
robertoraggi: it works already like this. The method is called `processPriorities`

PR #1296, opened 2020-02-17
HARP-8709 Implement strategy to fix mixed LOD cracks

MPursche:
Fixes mixed LOD cracks for webtile data source, background datasource and omv data source.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Copyright (C) 2017-2020 HERE Europe B.V.
  * Licensed under Apache 2.0, see full license in LICENSE
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -937,13 +937,15 @@ export class Tile implements CachedResource {
         };
 
         const disposeObject = (object: TileObject & DisposableObject) => {
-            if (object.geometry !== undefined && this.shouldDisposeObjectGeometry(object)) {
-                object.geometry.dispose();
-            }
+            if (this.shouldDisposeObjectGeometry(object)) {
+                if (object.geometry !== undefined) {
+                    object.geometry.dispose();
+                }
 
-            if (object.geometries !== undefined && this.shouldDisposeObjectGeometry(object)) {
-                for (const geometry of object.geometries) {
-                    geometry.dispose();
+                if (object.geometries !== undefined) {
+                    for (const geometry of object.geometries) {
+                        geometry.dispose();
+                    }
                 }
             }
 
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -16,6 +16,7 @@ import * as THREE from "three";
 import { AnimatedExtrusionTileHandler } from "./AnimatedExtrusionHandler";
 import { CopyrightInfo } from "./copyrights/CopyrightInfo";
 import { DataSource } from "./DataSource";
+import { LodMesh } from "./geometry/LodMesh";
 import { TileGeometryLoader } from "./geometry/TileGeometryLoader";
 import { MapView } from "./MapView";
 import { PathBlockingElement } from "./PathBlockingElement";
@@ -44,6 +45,7 @@ export type TileObject = THREE.Object3D & {
 
 interface DisposableObject {
     geometry?: THREE.BufferGeometry | THREE.Geometry;
+    geometries?: Array<THREE.BufferGeometry | THREE.Geometry>;
     material?: THREE.Material[] | THREE.Material;
 }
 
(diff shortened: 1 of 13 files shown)


@here/harp-mapview/lib/Tile.ts
```diff
@@ -995,6 +996,18 @@ export class Tile implements CachedResource {
...
     }
 
+    /**
+     * Update tile for current map view zoom level
+     * @param zoomLevel Zoom level of the map view
+     */
+    update(zoomLevel: number): void {
+        for (const object of this.objects) {
+            if (object instanceof LodMesh) {
+                object.setLevelOfDetail(zoomLevel - this.tileKey.level);
```
ninok: In which cases would `zoomLevel != this.tileKey.level`. During fallback phase `zoomLevel` can also be smaller than `this.tileKey.level`

@here/harp-mapview/lib/geometry/LodMesh.ts
```diff
@@ -0,0 +1,63 @@
...
+
+        this.geometries = geometries;
+    }
+
+    /**
+     * Update geometries of mesh
+     */
+    set geometries(geometries: Array<THREE.Geometry | THREE.BufferGeometry> | undefined) {
+        this.m_geometries = geometries;
+        this.geometry =
```
ninok: What happens to the previously used geometry? I mean will the GL resources be freed?

@here/harp-mapview/lib/geometry/LodMesh.ts
```diff
@@ -0,0 +1,63 @@
...
+import { MathUtils } from "@here/harp-geoutils";
+import * as THREE from "three";
+
+/**
+ * A mesh that can store multiple versions of a geometry with differents level of detail.
+ * The rendered level of detail can be adapted during runtime.
+ * @internal
+ * @hidden
+ */
+export class LodMesh extends THREE.Mesh {
```
ninok: Not sure if this is a good idea, b/c if the user modifies the mesh geometry, the changes could suddenly be overridden when the zoomlevel changes.
Can you do some measurements what's the actual difference in number of vertices/triangles for the more detailed mesh. If the difference is not to big, we could just always use the higher LOD

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1200,81 +1205,116 @@ export class TileGeometryCreator {
...
+            const sphericalModifier = new SphericalGeometrySubdivisionModifier(
                 THREE.Math.degToRad(10),
                 sourceProjection
             );
-            modifier.modify(g);
+            const enableMixedLod = mapView.enableMixedLod || mapView.enableMixedLod === undefined;
+
+            if (enableMixedLod) {
+                // Use a [[LodMesh]] to adapt tesselation of tile depending on zoom level
+                for (let zoomLevelOffset = 0; zoomLevelOffset < 4; ++zoomLevelOffset) {
```
ninok: Do we really need to do this for all zoomlevels and all features?
Especially in zoomed in cases the `LodMesh` will most likely contain 4 times the same geometry.
Especially for buildings this could be quite some memory overhead.

MPursche: Currently it is only used for `BackgroundDatasource` and `WebTileDatasource`. The fix for vector tiles will probably be slightly different and is not part of the current PR.

@here/harp-mapview/lib/Tile.ts
MPursche: When mixed LOD is enabled the tile key level can be smaller than the display zoom level, e.g., on the poles or high camera tilt. This PR basically fixes the cracks so that tiles of different zoom levels will use the same level of tessellation -> the edges of tile geometry will fit.
I can't simply increase tessellation of lower zoom levels because this will only change at which zoom level the cracks are visible and not remove them completly.

MPursche:
GPU tests please

MPursche:
Decoders tests please

@here/harp-mapview/lib/geometry/LodMesh.ts
MPursche: Do we have use-cases where the geometry created by harp.gl is changed later?
For mixed LOD to work without cracks in the globe it is important that tiles with different zoom level fit together at the edges, e.g., the tessellation of one level 3 tile needs to be the same as 2x2 level 4 tiles. This needs to be changed depending on display zoom level of the data source.

@here/harp-mapview/lib/geometry/LodMesh.ts
MPursche: Added `dispose` if geometry is changed.

MPursche:
Decoders tests please

MPursche:
GPU tests please

MPursche:
Rebuild please.

PR #1299, opened 2020-02-18
Remove the usages of Technique._keys.

robertoraggi:
This patch also modifies the cache policy of TextRenderStyle,
before this change TextRenderStyle entries where cached across
tiles created from the same datasource, with this change the
caching is limited to the tile.

### DIFF
--- a/@here/harp-mapview/lib/text/TextStyleCache.ts
+++ b/@here/harp-mapview/lib/text/TextStyleCache.ts
@@ -54,12 +54,14 @@ export const DEFAULT_TEXT_STYLE_CACHE_ID = "Default";
  *
  * @returns [[TextStyle]] id.
  */
-export function computeStyleCacheId(
-    datasourceName: string,
+function computeStyleCacheId(
+    tile: Tile,
     technique: Technique & Partial<IndexedTechniqueParams>,
     zoomLevel: number
 ): string {
-    return `${datasourceName}_${technique._key}_${zoomLevel}`;
+    const datasourceName = tile.dataSource.name;
+    const key = `${datasourceName}_${tile.tileKey.toHereTile()}_${technique._index}_${zoomLevel}`;
+    return key;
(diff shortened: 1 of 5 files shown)


robertoraggi:
@heremaps-bot GPU tests please

@here/harp-mapview/lib/text/TextStyleCache.ts
```diff
@@ -54,12 +54,14 @@ export const DEFAULT_TEXT_STYLE_CACHE_ID = "Default";
...
-export function computeStyleCacheId(
-    datasourceName: string,
+function computeStyleCacheId(
+    tile: Tile,
     technique: Technique & Partial<IndexedTechniqueParams>,
     zoomLevel: number
 ): string {
-    return `${datasourceName}_${technique._key}_${zoomLevel}`;
+    const datasourceName = tile.dataSource.name;
+    const key = `${datasourceName}_${tile.tileKey.toHereTile()}_${technique._index}_${zoomLevel}`;
```
zbigg: The cache in proposed this form doesn't make sense at all.

Currently, this is global cache that is reused across tiles in scope of one datasource. Apart from big cost of creating ana maintaining keys, we at least reuse styles across tiles.

With this patch, it's still global cache with global key, but there is mn reuse outside of one particular instance of Tile, so it's just hundreds of unnecessary string concatenations where, this cache could be just direct Tile member:

```
textRenderStyleCache: Map<technique index as number, Text*Style>.
```

(zoom level can be removed if cache is bound to Tile)

robertoraggi: Good point, I agree. I will update the change.

zbigg: Followup in #1338

zbigg:
Closing in favor of #1338

PR #1302, opened 2020-02-19
TEST: Added Berlin ZoomIn test

ninok:
* Add new Berlin zoom testcase that mimics HERE Maps for JavaScript testcase
* Add save functionality
* Allow configuration of themes via webpack (for internal themes)

### DIFF
--- a/@here/harp-examples/src/performance_benchmark.ts
+++ b/@here/harp-examples/src/performance_benchmark.ts
@@ -16,6 +16,12 @@ import { GUI, GUIController } from "dat.gui";
 import { PerformanceTestData } from "../lib/PerformanceConfig";
 import { PerformanceUtils } from "../lib/PerformanceUtils";
 
+interface ThemeList {
+    [name: string]: string;
+}
+// Theme list populated by webpack.config.js
+declare const THEMES: ThemeList;
+
 const logger = LoggerManager.instance.create("PerformanceUtils");
 
 const NAME_NUM_DECODER_OPTION = "Num Decoders";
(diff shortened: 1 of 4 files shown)


@here/harp-examples/src/performance_benchmark.ts
```diff
@@ -631,7 +648,7 @@ export namespace PerformanceBenchmark {
                 "6": 6,
                 "8": 8
             },
-            PhasedLoading: true,
+            PhasedLoading: false,
```
nzjony: This should have already been removed? Maybe make a ticket for this.

@here/harp-examples/src/performance_benchmark.ts
```diff
@@ -572,6 +588,7 @@ export namespace PerformanceBenchmark {
             },
             CanvasSize: {
                 default: undefined,
+                "1100×900": "1100×900",
```
nzjony: Why this resolution?

@here/harp-examples/webpack.config.js
```diff
@@ -19,6 +21,15 @@ const harpFontResourcesPath = path.dirname(require.resolve("@here/harp-fontcatal
 const isProduction = process.env.NODE_ENV === "production";
 const harpBundleSuffix = isProduction ? ".min" : "";
 
+themeList = {
```
nzjony: why is this moved to the webpack config? It would make sense if there are multiple files that need it, but it seems a bit strange to move it away from the performance test.

ninok: b/c the theme list is different for the internal build (i.e. it uses the internal styles).
Please see my internal change as well

@here/harp-examples/src/performance_benchmark.ts
ninok: JSAPI is using it for their tests.

@here/harp-examples/src/performance_benchmark.ts
ninok: true, will remove in a follow up commit

PR #1303, opened 2020-02-19
MINOR: Remove logging to console and check with spy instead

nzjony:
Removes the errors from the console, i.e.:

```
TileLoader: [anonymous-datasource#5]: failed to load tile 1 Error: No connection.
    at MockDataProvider.getTile (/home/travis/build/heremaps/harp.gl/@here/harp-mapview-decoder/test/TileLoaderTest.ts:50:35)
    at TileLoader.startLoading (/home/travis/build/heremaps/harp.gl/@here/harp-mapview-decoder/lib/TileLoader.ts:87:9)
    at TileLoader.loadAndDecode (/home/travis/build/heremaps/harp.gl/@here/harp-mapview-decoder/lib/TileLoader.ts:26:6)
    at Context.<anonymous> (/home/travis/build/heremaps/harp.gl/@here/harp-mapview-decoder/test/TileLoaderTest.ts:194:44)
```

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview-decoder/test/TileLoaderTest.ts
+++ b/@here/harp-mapview-decoder/test/TileLoaderTest.ts
@@ -20,7 +20,11 @@ import * as chai from "chai";
 import * as chaiAsPromised from "chai-as-promised";
 chai.use(chaiAsPromised);
 const { expect } = chai;
+import { LoggerManager } from "@here/harp-utils";
 import { DataProvider, TileLoader } from "../index";
+
+// Needed for using expect(...).true for example
+// tslint:disable: no-unused-expression
 
 class MockDataSource extends DataSource {
     /** @override */
@@ -128,7 +132,6 @@ describe("TileLoader", function() {
             );
 
             const loadPromise = tileLoader.loadAndDecode();
-            // tslint:disable-next-line: no-unused-expression
             expect(loadPromise).to.not.be.undefined;
 
             return expect(loadPromise).to.eventually.be.fulfilled;
@@ -144,11 +147,9 @@ describe("TileLoader", function() {
             );
 
             const loadPromise = tileLoader.loadAndDecode();
-            // tslint:disable-next-line: no-unused-expression
             expect(loadPromise).to.not.be.undefined;
 
             const secondLoadPromise = tileLoader.loadAndDecode();
-            // tslint:disable-next-line: no-unused-expression
             expect(secondLoadPromise).to.not.be.undefined;
             expect(secondLoadPromise).to.equal(loadPromise);
 
@@ -166,7 +167,6 @@ describe("TileLoader", function() {
 
             dataProvider.emptyPayload = true;
             let loadPromise = tileLoader.loadAndDecode();
-            // tslint:disable-next-line: no-unused-expression
             expect(loadPromise).to.not.be.undefined;
 
             return expect(loadPromise).to.eventually.be.fulfilled.then(() => {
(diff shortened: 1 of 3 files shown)


nzjony:
@zbigg , great idea, I will fix.

zbigg:
In one project i've done something like this:

```javascript
beforeEach(() => {
    LoggerManager.suspendLogs() // Suspends all log messages, so they are buffered until ...
})
afterEach(function() {
    if (this.currentTest.state !== "passed") {
        // eslint-disable-next-line no-console
        console.log("--- dumping logs from failed tests ---");
        // ... test finishes and failed, so we dump log buffer 
        LoggerManager.dumpSuspendedLogs();
    } 
    // ... discard them after dump or for passed tests
    LoggerManager.discardSuspenededLogs(); tests
});
```

@here/harp-mapview-decoder/test/TileLoaderTest.ts
```diff
@@ -201,9 +201,9 @@ describe("TileLoader", function() {
 
                 // tslint:disable-next-line: no-shadowed-variable
                 const loadPromise = tileLoader.loadAndDecode();
-                // tslint:disable-next-line: no-unused-expression
                 expect(loadPromise).to.not.be.undefined;
 
+                LoggerManager.instance.update("TileLoader", { enabled: true });
```
zbigg: I've specifically proposed to use `before` and `after` hooks, because now if test fail between "enable"/"disable" lines, loggger will remain disabled for rest of test execution.

While it is minor, test cases should leave environment unchanged even if it's only logger configuration.

... or, we can use async/await and good old `try catch` although we may discuss if it's elegant.

nzjony: good point, I missed that, sorry. Will fix.

nzjony: question, if I use the before / after hooks, the logger is disabled anyway for all tests, so it provides less cover.

I will use finally, that seems to be the best usage.

@here/harp-mapview-decoder/test/TileLoaderTest.ts
```diff
@@ -189,23 +188,31 @@ describe("TileLoader", function() {
                 new MockTileDecoder(),
                 0
             );
+            // This test writes an error to the console, so we disable it.
```
MPursche: It would be better to disable and enable logging with the test setup hooks(`before`, `after`). If the test fails, i.e., the `loadPromise` isn't rejected the logging is not enabled. This could have a side effect on other tests.
Something like:
```javascript
before(function () {
    LoggerManager.instance.update("TileLoader", { enabled: false });
}
after(function () {
    LoggerManager.instance.update("TileLoader", { enabled: true });
}
```

@here/harp-mapview-decoder/test/TileLoaderTest.ts
```diff
@@ -189,23 +188,31 @@ describe("TileLoader", function() {
...
+                    expect(tileLoader.state).to.equal(TileLoaderState.Failed);
+                    dataProvider.failsRequests = false;
+
+                    // tslint:disable-next-line: no-shadowed-variable
+                    const loadPromise = tileLoader.loadAndDecode();
+                    expect(loadPromise).to.not.be.undefined;
+
+                    return expect(loadPromise).to.eventually.be.fulfilled;
+                },
+                () => {
```
MPursche: This code might not be called if `expect(loadPromise).to.eventually.be.rejected` fails. See comment above.

@here/harp-mapview-decoder/test/TileLoaderTest.ts
nzjony: then you disable it for all tests, but I want to disable it just for these specific tests.

@here/harp-mapview-decoder/test/TileLoaderTest.ts
nzjony: why not? I tested it by removing the throw new Error and returning a valid object, and it works.

PR #1305, opened 2020-02-19
[WIP] Refactor technique handling to separate modules

zbigg:
Tear apart TileGeometryCreator into separate modules for handling particular
techniques.

It attempts to do two things:
  * separate logic of technique handling into  separate modules
  * provide coherent API for updating objects (materials, objects)
  * as PoC implements "skip rendering of invisible objects" #1271 

* Tile: add updateDynamicTechniques callback that shall be called by
  MapView to objects that depende on dynamic technique attributes

* MapView: Call `updateDynamicTechniques` on all rendered tiles and then don't add
  completly transparent objects to scene

* Extra - don't install fading render helper at all if fading is
  disabled

Follow-ups
  * Cross-tile material sharing where possible (textures example, most solid lines, fill areas etc) -> #1348
  * TechniqueHandlers concept applied to text elements and pois #1335 

What is missing
  * validation of parameters like in #1241 & #1287

### DIFF
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -4,38 +4,17 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import {
-    BaseTechniqueParams,
-    BufferAttribute,
     DecodedTile,
-    Env,
-    Expr,
-    ExtrudedPolygonTechnique,
-    FillTechnique,
-    Geometry,
     GeometryKind,
     GeometryKindSet,
-    getArrayConstructor,
     getFeatureId,
     getPropertyValue,
-    isCirclesTechnique,
-    isExtrudedLineTechnique,
-    isExtrudedPolygonTechnique,
-    isFillTechnique,
+    Group,
     isLineMarkerTechnique,
-    isLineTechnique,
     isPoiTechnique,
-    isSegmentsTechnique,
-    isSolidLineTechnique,
-    isSquaresTechnique,
-    isTerrainTechnique,
     isTextTechnique,
-    MakeTechniqueAttrs,
     MapEnv,
-    needsVertexNormals,
-    SolidLineTechnique,
-    StandardExtrudedLineTechnique,
     Technique,
-    TerrainTechnique,
     TextPathGeometry
 } from "@here/harp-datasource-protocol";
 // tslint:disable:max-line-length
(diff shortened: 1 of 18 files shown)


@here/harp-mapview/lib/techniques/ExtrudedPolygonTechniqueHandler.ts
```diff
@@ -0,0 +1,287 @@
...
+        super(technique);
+        this.mainMaterial = createMaterial({ technique, env: context.env }) as
+            | MapMeshBasicMaterial
+            | MapMeshStandardMaterial;
+        assert(
+            this.mainMaterial instanceof MapMeshBasicMaterial ||
+                this.mainMaterial instanceof MapMeshStandardMaterial
+        );
+
+        if (Expr.isExpr(technique.color) || Expr.isExpr(technique.opacity)) {
```
ninok: Just a generic question about performance since I'm not a JS perf expert:
What is the overhead (memory and speed) of all these callbacks?
We basically add a callback for every dynamic property in every technique.
The alternative would be to have one callback per technique and iterate through all properties.
That would lead to less callbacks but we would pay for the `isExpr` check every time.

So as I said I'm not an expert in JS perf so I cannot judge what would be the better aproach.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -620,99 +584,90 @@ export class TileGeometryCreator {
...
+                        technique as IndexedTechnique,
+                        tile,
+                        mapView.techniqueUpdateContext
                     );
-                }
 
-                if (srcGeometry.index) {
-                    bufferGeometry.setIndex(getBufferAttribute(srcGeometry.index));
-                }
+                    // TODO: text technique handler and world space technique handler doesn't
```
ninok: Maybe instead of splitting the technique handler into `WorldTechniqueHandler` and `ScreenTechniqueHandler` we could just have both methods:

- `createWorldObjects` and
- `createScreenObjects` in the interface.

For the current techniques one of them would just return empyt buffers, but maybe in the future we could even have a technique that produces geometry and labels at the same time.

@here/harp-mapview/lib/techniques/FillTechniqueHandler.ts
```diff
@@ -0,0 +1,137 @@
...
+                );
+            }
+
+            if (Expr.isExpr(technique.lineColor) || Expr.isExpr(technique.opacity)) {
+                this.updaters.push(this.updateLineColor.bind(this));
+            }
+        }
+    }
+
+    createObject(tile: Tile, srcGeometry: Geometry, group?: Group) {
```
ninok: Not sure if this method should add the objects to the tile. Maybe it should just return the created objects and we explicitely add them in the TileGeometryCreator.
Or we should at least rename this method to addObjectsToTile

@here/harp-mapview/lib/techniques/ExtrudedPolygonTechniqueHandler.ts
zbigg: TLDR: I expected this question and i will try to measure.

Yes, alternative would be one big "update" function that checks what is dynamic, what not.

We have 3 approaches here
1) super update function with "ifs"
2) dedicated callbacks for each property only if needed
3) completly generic updaters like in #1287

I like 2 only because engine will only call these updaters that are needed, albeit taking cost of most probably** dynamic dispatch.

"1" has advantage of static dispatch, but more work and boilerplate.

"3" is dynamic dispatch, correctness but no boilerplate.

Still i think dynamic dispatch can't be so expensive that i choose "2" or better "3" unless measured.

And of course, trying to measure this is really hard because this code (updates of materials) take up to 2-5% of render, so within error range it's really hard to prepare test that will will both isolate this particular functionality and behave like in full application.

@here/harp-mapview/lib/techniques/FillTechniqueHandler.ts
zbigg: Good point, i'll check if that simplify architecture.

@here/harp-mapview/lib/techniques/ExtrudedPolygonTechniqueHandler.ts
ninok: I agree this is really hard to measure.
Not sure how the JavaScript runtime is actually doing closures(i.e. to know what it should capture it needs to analyse the code when the callback is created). Compared to C++ Lambda where you can exactly specify what and how things should be captured JS closures are a black box to me that support you in shooting yourself into the foot.

Nevertheless I think 2 is a good compromise in code readability and performance.

ninok: Sorry I just looked at the code again and realised that you are not using closures but bind. So we know what get's captured and everything  is fine.

zbigg: > Not sure how the JavaScript runtime is actually doing closures(i.e. to know what it should capture it needs to analyse the code when the callback is created).

From what i see in V8/Chrome, in debugger at least, it looks like it captures sum of all captured variables in all closures from that scope:

image

Test:
 breakpoint in onBeforeRenderCallBack for solid lines

Look, that closure captured only `mapView`, `dataSource` and `Tile`. Looking at closures created in this scope, confirms suspicion.

On the other hand i'm not sure if debugger reflects real optimizations in real code emitted by JIT.

PR #1307, opened 2020-02-20
HARP-8971: Fix wrong computation of fullFrameTime (hence fps)

ninok:
Previously the fullFrameTime was computed by subtracting the last frame start time from the current frame start time.
This was done to make sure we also measure the time we spend outside the current animation frame (i.e. geometry creation time).
Unfortunately with this approach we also measure everything the application/test is doing.

With this change the fullFrameTime is explicitly computed by adding the geometry creation time to the current frame.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2709,34 +2709,36 @@ export class MapView extends THREE.EventDispatcher {
     /**
      * Renders the current frame.
      */
-    private render(time: number): void {
+    private render(frameStartTime: number): void {
         if (this.m_drawing) {
             return;
         }
-        ++this.m_frameNumber;
 
-        const stats = PerformanceStatistics.instance;
-        const gatherStatistics: boolean = stats.enabled;
-
-        const frameStartTime = time;
-
-        RENDER_EVENT.time = time;
+        RENDER_EVENT.time = frameStartTime;
         this.dispatchEvent(RENDER_EVENT);
 
-        let currentFrameEvent: FrameStats | undefined;
+        ++this.m_frameNumber;
 
+        let currentFrameEvent: FrameStats | undefined;
+        const stats = PerformanceStatistics.instance;
+        const gatherStatistics: boolean = stats.enabled;
         if (gatherStatistics) {
             currentFrameEvent = stats.currentFrame;
-            currentFrameEvent.setValue("renderCount.frameNumber", this.m_frameNumber);
 
             if (this.m_previousFrameTimeStamp !== undefined) {
+                // In contrast to fullFrameTime we also measure the application code
+                // for the FPS. This means FPS != 1000 / fullFrameTime.
                 const timeSincePreviousFrame = frameStartTime - this.m_previousFrameTimeStamp;
-                if (gatherStatistics) {
-                    currentFrameEvent.setValue("render.fullFrameTime", timeSincePreviousFrame);
-                    // For convenience and easy readability
-                    currentFrameEvent.setValue("render.fps", 1000 / timeSincePreviousFrame);
-                }
+                currentFrameEvent.setValue("render.fps", 1000 / timeSincePreviousFrame);
             }
+
+            // We store the last frame statistics at the beginning of the next frame b/c additional
+            // work (i.e. geometry creation) is done outside of the animation frame but still needs
+            // to be added to the `fullFrameTime` (see [[TileGeometryLoader]]).
+            stats.storeAndClearFrameInfo();
+
+            currentFrameEvent = currentFrameEvent as FrameStats;
+            currentFrameEvent.setValue("renderCount.frameNumber", this.m_frameNumber);
         }
 
         this.m_previousFrameTimeStamp = frameStartTime;
@@ -2847,7 +2849,7 @@ export class MapView extends THREE.EventDispatcher {
             });
         }
 
-        if (this.m_movementDetector.checkCameraMoved(this, time)) {
+        if (this.m_movementDetector.checkCameraMoved(this, frameStartTime)) {
             const { yaw, pitch, roll } = MapViewUtils.extractAttitude(this, this.camera);
             const { latitude, longitude, altitude } = this.geoCenter;
             this.dispatchEvent({
@@ -2866,7 +2868,7 @@ export class MapView extends THREE.EventDispatcher {
         const camera = this.m_pointOfView !== undefined ? this.m_pointOfView : this.m_rteCamera;
 
         if (this.renderLabels) {
-            this.prepareRenderTextElements(time);
+            this.prepareRenderTextElements(frameStartTime);
         }
 
         if (gatherStatistics) {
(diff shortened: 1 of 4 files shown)


@here/harp-mapview/lib/Statistics.ts
```diff
@@ -1017,55 +1017,51 @@ export class PerformanceStatistics {
...
+            this.currentFrame.setValue(
+                "gl.numPrograms",
+                webGlInfo.programs === null ? 0 : webGlInfo.programs.length
+            );
+        }
+    }
 
+    /**
+     * Add memory statistics to the current frame if available.
+     * @note Currently on supported on Chrome
```
kkostecki: Typo: 

> only supported

@here/harp-mapview/lib/Statistics.ts
```diff
@@ -1074,6 +1070,18 @@ export class PerformanceStatistics {
...
+    /**
+     * Stores the current frame events into the array of events and clears all values.
+     *
+     * @returns {boolean} Returns `false` if the maximum number of storable frames has been reached.
+     * @memberof PerformanceStatistics
+     */
+    storeAndClearFrameInfo(): boolean {
+        if (this.m_frameEvents.length >= this.maxNumFrames) {
+            return false;
+        }
```
kkostecki: I guess it would be much better to store frames info in the round buffer, this way we never loose last frames information... Maybe this is a cause why we got some unpredictable results in some cases.
BTW: I know, it's not your code :).

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -286,6 +286,10 @@ export class SimpleTileGeometryLoader implements TileGeometryLoader {
             if (stats.enabled) {
                 const geometryCreationTime = PerformanceTimer.now() - now;
                 const currentFrame = stats.currentFrame;
+
+                // Account for the geometry creation in the current frame.
+                currentFrame.addValue("render.fullFrameTime", geometryCreationTime);
+
```
kkostecki: I wonder how much details we would like to spot... but reasonably we would also need to add the time for processing techniques via `TileGeometryCreator`. You may find it in `prepareForRender()` method in the same class.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2709,30 +2709,32 @@ export class MapView extends THREE.EventDispatcher {
...
         const frameStartTime = time;
 
         RENDER_EVENT.time = time;
         this.dispatchEvent(RENDER_EVENT);
 
-        let currentFrameEvent: FrameStats | undefined;
-
         if (gatherStatistics) {
-            currentFrameEvent = stats.currentFrame;
+            currentFrameEvent = currentFrameEvent as FrameStats;
```
kkostecki: Although I understand this convention is look little bit awkward :).

ninok: Yeah, I was to lazy to add `!` everywhere

@here/harp-mapview/lib/Statistics.ts
ninok: true. for now I assume the buffer is big enough.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
ninok: Okay, I will check.

@here/harp-mapview/lib/Statistics.ts
ninok: done

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2917,15 +2922,25 @@ export class MapView extends THREE.EventDispatcher {
...
+
+            // Initialize the fullFrameTime with the frameRenderTime If we also create geometry in this frame
+            // this number will be increased in the TileGeometryLoader.
+            currentFrameEvent.setValue("render.fullFrameTime", frameRenderTime);
+
+            // Add THREE.js statistics
+            stats.addWebGLInfo(this.m_renderer.info);
 
-            PerformanceStatistics.instance.storeFrameInfo(this.m_renderer.info);
+            // Add memory statistics (FIXME: )
```
nzjony: What exactly needs to be fixed here? It is unclear to me.

ninok: ups, forgot to actually write it down.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2713,30 +2713,35 @@ export class MapView extends THREE.EventDispatcher {
...
         }
+
         ++this.m_frameNumber;
+        const frameStartTime = time;
 
+        let currentFrameEvent: FrameStats | undefined;
         const stats = PerformanceStatistics.instance;
         const gatherStatistics: boolean = stats.enabled;
+        if (gatherStatistics) {
+            currentFrameEvent = stats.currentFrame;
```
nzjony: why was this moved here, can't it be part of the if(gatherStat...) on line 2742?

@here/harp-mapview/lib/Statistics.ts
nzjony: this seems to be able to store 1000 frames, at 60 fps, that is roughly 15 seconds worth of data, for a flyover test, that may not be enough?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2713,30 +2713,35 @@ export class MapView extends THREE.EventDispatcher {
         if (this.m_drawing) {
             return;
         }
+
         ++this.m_frameNumber;
+        const frameStartTime = time;
```
nzjony: why do we have an alias for the time? Why not just rename the parameter ?

ninok: good point. I just copy pasted stuff around. Will just rename the param.

@here/harp-mapview/lib/Statistics.ts
ninok: I think the flyover test is fetching the statistics after every frame, so the buffer will not overflow but I will double check

@here/harp-mapview/lib/MapView.ts
ninok: I think the perf test gathers the statistics in the `RENDER_EVENT. Will also double check this.

@here/harp-mapview/lib/Statistics.ts
ninok: @nzjony : Actually your suspicion was correct. By default we store 1000 frames but flyover renders 5000 without fetching statistics in between. What is also interesting is, that m_frameEvents is actually a ring buffer but we just stop putting statistics into it.

@here/harp-mapview/lib/MapView.ts
ninok: the perftests are listening to the `AfterRender` event.So I move the dispatching of the render event to the very top (and also don't count it into the frame time b/c it's not rendering but application code)

@here/harp-mapview/lib/Statistics.ts
ninok: After looking again into our code and due to hints from @CraigBerlin I found out that the `PerformanceStatistics` are not really following the singleton pattern, i.e. `PerformanceStatistics.instance` will be recreated whenever a new `PerformanceStatistics` object is instantiated.

PR #1318, opened 2020-02-25
HARP-8048: PoC use dynamic styling to select roads and buildings

robertoraggi:


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2956,6 +2964,40 @@ export class MapView extends THREE.EventDispatcher {
...
             for (const object of tile.objects) {
+                if (object.userData.technique && object.userData.feature) {
+                    const technique: Technique = object.userData.technique;
+                    const objInfos = object.userData.feature.objInfos as ValueMap[];
+                    if (technique.hasOwnProperty("filter") && Array.isArray(objInfos)) {
+                        // ### TODO: evaluate only when properties references by the filter change.
+                        const geometry: THREE.BufferGeometry = (object as any).geometry;
+                        if (geometry && geometry.isBufferGeometry) {
+                            geometry.clearGroups();
+                            objInfos.forEach((properties, index) => {
```
robertoraggi: the representation of feature attributes (`objInfos`) needs to be way more efficient than what we have today.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2956,6 +2964,40 @@ export class MapView extends THREE.EventDispatcher {
         const worldOffsetX = tile.computeWorldOffsetX();
         if (tile.willRender(zoomLevel)) {
             for (const object of tile.objects) {
+                if (object.userData.technique && object.userData.feature) {
```
robertoraggi: We need to find the right place (and the right way) to get the techniques associated with the objects.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2956,6 +2964,40 @@ export class MapView extends THREE.EventDispatcher {
...
+                                    const start = object.userData.feature.starts[index] as number;
+                                    const end =
+                                        object.userData.feature.starts[index + 1] ??
+                                        geometry.getIndex()?.count;
+                                    if (end === undefined) {
+                                        return;
+                                    }
+                                    const count = end - start;
+                                    geometry.addGroup(start, count);
+                                }
```
robertoraggi: the dynamic filtering/creation of `geometry groups` is assuming that the `geometries` created to represent `geo features` have shared buffer attributes.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1738,6 +1742,10 @@ export class MapView extends THREE.EventDispatcher {
         return this.m_env;
     }
 
+    setDynamicProperty(name: string, value: Value) {
+        this.m_env.entries[name] = value;
+    }
+
```
robertoraggi: I would prefer to keep the keep the MapView environment read-only and maybe add APIs to DataSource to set/update envs.

@here/harp-map-theme/resources/object_selection.json
```diff
@@ -0,0 +1,104 @@
...
+            "when": ["==", ["get", "$geometryType"], "line"],
+            "technique": "solid-line",
+            "attr": {
+                "lineWidth": ["world-ppi-scale", 4],
+                "color": "#ee0",
+                "filter": [
+                    "~=",
+                    ["get", "name", ["dynamic-properties"]],
+                    ["get", "selected", ["dynamic-properties"]]
+                ]
```
robertoraggi: I believe I should move the `filter` property out of the `attr` block and have it at the same level as `when`

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1456,8 +1459,11 @@ export class TileGeometryCreator {
                 geoBox: tile.geoBox
             };
             object.userData = tileDisplacementMap;
-        } else if (isSolidLineTechnique(technique)) {
-            object.userData = srcGeometry.objInfos!;
+            // ### PoC this breaks the line picking
```
zbigg: I don't think it breaks line picking.
Line picking as implemented in `RoadPicker` is based purely on `decodedTile.tileInfo`.

On the other hand, i don't know origins of this specific `userData` overwrite commented here.

PR #1319, opened 2020-02-25
HARP-8523: Fix zoom level jitter and tiles flickering.

kkostecki:
Inaccuracies on the 13-th fractional digit of zoom level calculation
may be observed when doing small tilt changes, thus causing the zoom level
to be discretized to smaller value then real one, especially when
calculating tiles' storage or visibility level.
This causes displaying wrong tile set (with different zoom level) for
a certain camera arrangement (angles).

The change introduces zoom level rounding above 10-th fractional digit,
giving enough safety margin for inaccurate zoom level from distance
calculus and quite good precision for zoom level based interpolations.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/Utils.ts
+++ b/@here/harp-mapview/lib/Utils.ts
@@ -6,16 +6,10 @@
 
 import * as THREE from "three";
 
-import {
-    GeoCoordinates,
-    MathUtils,
-    Projection,
-    ProjectionType,
-    TileKey
-} from "@here/harp-geoutils";
+import { GeoCoordinates, Projection, ProjectionType, TileKey } from "@here/harp-geoutils";
 import { EarthConstants } from "@here/harp-geoutils/lib/projection/EarthConstants";
 import { MapMeshBasicMaterial, MapMeshStandardMaterial } from "@here/harp-materials";
-import { assert, LoggerManager } from "@here/harp-utils";
+import { assert, LoggerManager, MathUtils } from "@here/harp-utils";
 import { MapView, MAX_TILT_ANGLE } from "./MapView";
 import { getFeatureDataSize, TileFeatureData } from "./Tile";
 
@@ -450,7 +444,7 @@ export namespace MapViewUtils {
             mapView.projection.type === ProjectionType.Spherical
                 ? cache.vector3[0].copy(mapView.camera.position).normalize()
                 : cache.vector3[0].set(0, 0, 1),
-            MathUtils.degToRad(-deltaYawDeg)
+            THREE.MathUtils.degToRad(-deltaYawDeg)
         );
         mapView.camera.updateMatrixWorld();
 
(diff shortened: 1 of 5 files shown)


@here/harp-utils/lib/MathUtils.ts
```diff
@@ -163,4 +163,36 @@ export namespace MathUtils {
...
+    /**
+     * Round number to a specified number of fractional digits, using 'round to nearest' convention.
+     *
+     * Midpoint values of less significant fraction digit are promoted (increase previous digit), if
+     * they are greater then 5.
+     *
+     * @param value The floating point value to be rounded.
+     * @param digits The number of fractional digits to leave (accuracy after comma).
+     */
+    export function roundFraction(value: number, digits: number = 10) {
```
nzjony: please add some tests for this function

@here/harp-mapview/lib/Utils.ts
```diff
@@ -450,7 +444,7 @@ export namespace MapViewUtils {
             mapView.projection.type === ProjectionType.Spherical
                 ? cache.vector3[0].copy(mapView.camera.position).normalize()
                 : cache.vector3[0].set(0, 0, 1),
-            MathUtils.degToRad(-deltaYawDeg)
+            THREE.Math.degToRad(-deltaYawDeg)
```
ninok: `THREE.Math` got renamed to `THREE.MathUtils` with the three 0.113 update I just merged

@here/harp-utils/lib/MathUtils.ts
```diff
@@ -163,4 +165,44 @@ export namespace MathUtils {
...
+     * @note Function accepts negative values using correct arithmetic rounding, such as the correct
+     * numbers for `roundFraction(-0.5, 0)` will be `-1` and for `roundFraction(-1.55, 1)` will
+     * be `-1.6`.
+     * @note If you accidentally pass floating point number (with fraction) to @param digits
+     * this value will be _truncated_, but negative number of digits is not allowed.
+     *
+     * @param value The floating point value to be rounded.
+     * @param digits The number of fractional digits to leave (accuracy after comma).
+     */
+    export function roundFraction(value: number, digitsNum: number = 10) {
```
ninok: I think this code is a bit over engineered for what we want to do (i.e. `Math.round(zoomLevel * 10e10) / 10e10)`).
Do we gain anything compared to the old simple approach?

kkostecki: Just a precision and full control over number of digits + support to negative values.

@here/harp-mapview/lib/Utils.ts
kkostecki: Ok will change it.

@here/harp-utils/lib/MathUtils.ts
kkostecki: Tests added.

@here/harp-utils/lib/MathUtils.ts
ninok: Please keep things simple!

We should not implement sth. b/c maybe somebody might need this in the future.
I cannot imagine a scenario where somebody would need rounding of negative numbers.

PR #1323, opened 2020-02-26
MINOR: Fix commands in docs for creating harp.gl project

CraigBerlin:


DimitryRd:
Craig, are you sure that this command has to be changed? Because I can `npm init @here/harpgl-app` and it is works fine. 
And I checked it both - in @here/ and in coresdk/

CraigBerlin:
> Craig, are you sure that this command has to be changed? Because I can `npm init @here/harpgl-app` and it is works fine.
> And I checked it both - in @here/ and in coresdk/

Yes, it needs to be changed when pulling from the npm repo.
To verify create an empty directory and run the npm init command.

For testing locally you are correct, see the bottom of the @here/create-harp.gl-app/README.md.
Unfortunately newer versions of yarn started complaining about the "." in the bin target. After removing it the local package name now differs from name published to the npm repo.

PR #1328, opened 2020-02-27
HARP-8870 Fix cull frustum setup when terrain data source is used.

MPursche:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -688,6 +688,7 @@ export class MapView extends THREE.EventDispatcher {
 
     private m_visibleTiles: VisibleTileSet;
 
+    private m_elevationSource?: DataSource;
     private m_elevationRangeSource?: ElevationRangeSource;
     private m_elevationProvider?: ElevationProvider;
     private m_visibleTileSetLock: boolean = false;
@@ -2415,17 +2416,26 @@ export class MapView extends THREE.EventDispatcher {
      * @param elevationProvider Allows access to the elevation at a given location or a ray
      *      from the camera.
      */
-    setElevationSource(
+    async setElevationSource(
         elevationSource: DataSource,
         elevationRangeSource: ElevationRangeSource,
         elevationProvider: ElevationProvider
     ) {
-        // Add as datasource if it was not added before
-        const isPresent = this.m_tileDataSources.some(ds => ds === elevationSource);
-        if (!isPresent) {
-            this.addDataSource(elevationSource);
+        // Remove previous elevation source if present
+        if (this.m_elevationSource && this.m_elevationSource !== elevationSource) {
+            this.removeDataSource(this.m_elevationSource);
         }
+
+        // Add as datasource if it was not added before
+        const isPresent = this.m_tileDataSources.indexOf(elevationSource) !== -1;
+        if (!isPresent) {
+            await this.addDataSource(elevationSource);
+        }
+        this.m_elevationSource = elevationSource;
         this.m_elevationRangeSource = elevationRangeSource;
+        if (!this.m_elevationRangeSource.ready()) {
+            await this.m_elevationRangeSource.connect();
+        }
         this.m_elevationProvider = elevationProvider;
         this.dataSources.forEach(dataSource => {
             dataSource.setEnableElevationOverlay(true);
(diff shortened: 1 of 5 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2418,11 +2418,12 @@ export class MapView extends THREE.EventDispatcher {
...
-        // Try to remove incase this method was already called, will do nothing if it doesn't exist.
-        this.removeDataSource(elevationSource);
-        this.addDataSource(elevationSource);
+        // Add as datasource if it was not added before
+        const isPresent = this.m_tileDataSources.some((ds) => ds === elevationSource);
+        if (!isPresent) {
+            this.addDataSource(elevationSource);
+        }
         this.m_elevationRangeSource = elevationRangeSource;
-        this.m_elevationRangeSource.connect();
```
MPursche: This can be removed, because `dataSource.connect()` was already called by `addDataSource()`. Calling it twice creates some errors in the console.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2418,11 +2418,12 @@ export class MapView extends THREE.EventDispatcher {
         elevationRangeSource: ElevationRangeSource,
         elevationProvider: ElevationProvider
     ) {
-        // Try to remove incase this method was already called, will do nothing if it doesn't exist.
-        this.removeDataSource(elevationSource);
-        this.addDataSource(elevationSource);
+        // Add as datasource if it was not added before
+        const isPresent = this.m_tileDataSources.some(ds => ds === elevationSource);
```
nzjony: you are slightly modifying the behaviour of the API, because now the user can no longer swap between two elevation sources, once they call this method, the MapView can only use that specific ElevationSource. I don't know if there is a usecase to swap elevation sources, but because it is currently possible in the API, I think it should at least be documented that once an elevatoin source is set, it can't be changed.

I think this is

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -1190,22 +1190,21 @@ export class VisibleTileSet {
...
 
-        // If elevation is to be taken into account create extended frustum:
+        // If elevation is to be taken into account extend view frustum:
         // (near ~0, far: maxVisibilityRange) that allows to consider tiles that
         // are far below ground plane and high enough to intersect the frustum.
         if (elevationRangeSource !== undefined) {
-            const fp = MapViewUtils.getCameraFrustumPlanes(this.m_frustumIntersection.camera);
-            fp.near = this.m_viewRange.minimum;
-            fp.far = this.m_viewRange.maximum;
-            this.m_projectionMatrixOverride.makePerspective(
```
nzjony: why did you remove makePerspective? I don't see why you need to use a PerspectiveCamera, when makePerspective should be all you need (it seems that you just need to fix the near / far planes)?

@here/harp-mapview/lib/MapView.ts
MPursche: Ok, but that wouldn't have worked before either. The code would have removed the `elevationSource` that was given as a parameter and than add it again. For switching the source we may want to remove the previous `mapView.m_elevationRangeSource`.

@here/harp-mapview/lib/VisibleTileSet.ts
MPursche: Newest version of THREE.js uses `top, bottom, left, right, near, far` as parameters for `Matrix4.makePerspective()`. Calculating these parameters from `fov, aspect, near, far` is encapsuled in the `PerspectiveCamera`. I didn't want to reimplement the code to calculate these parameters in harp.gl as THREE.js may change the way it creates the `projectionMatrix` from the camera parameters or changes the API again.

@here/harp-mapview/lib/MapView.ts
nzjony: Oh you are right, I misread, resolving.

@here/harp-mapview/lib/VisibleTileSet.ts
nzjony: I understand, I made the same mistake as the author of this code, thinking that you can change the near / far without adjusting the sides. My bad.

PR #1331, opened 2020-02-27
Add support for passing multiple materials to ObjectConstructor.

robertoraggi:


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -636,39 +636,22 @@ export class TileGeometryCreator {
                 });
 
                 if (srcGeometry.interleavedVertexAttributes !== undefined) {
-                    srcGeometry.interleavedVertexAttributes.forEach(
```
ninok: We should print this code and put it on a wall of shame.

nzjony: Lol, I was just reading in Effective Typescript that such excessive typing is not very necessary, but they gave a very trivial example. This is a GREAT example.

PR #1336, opened 2020-02-28
MINOR: Remove 3D objects from root more efficiently.

atomicsulfate:
At the beginning of a frame, the previous frame objects are
removed from the scene calling MapView's mapTilesRoot.remove()
for each one of them. Each remove() call searches for the object
in an array and then splices it.
It's much simpler to just set the root's children array to length 0.

The average speedup is small, but it gets rid of some peaks where the
whole removal was taking ~1ms.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2765,9 +2765,7 @@ export class MapView extends THREE.EventDispatcher {
         this.m_renderer.clear();
 
         // clear the scene
-        while (this.m_mapTilesRoot.children.length > 0) {
-            this.m_mapTilesRoot.remove(this.m_mapTilesRoot.children[0]);
-        }
+        this.m_mapTilesRoot.children.length = 0;
 
         if (gatherStatistics) {
             setupTime = PerformanceTimer.now();


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2765,9 +2765,7 @@ export class MapView extends THREE.EventDispatcher {
         this.m_renderer.clear();
 
         // clear the scene
-        while (this.m_mapTilesRoot.children.length > 0) {
-            this.m_mapTilesRoot.remove(this.m_mapTilesRoot.children[0]);
-        }
+        this.m_mapTilesRoot.children.length = 0;
```
ninok: Looking at https://github.com/mrdoob/three.js/blob/dev/src/core/Object3D.js#L352
We slightly change the semantics, i.e.:
* we don't set the parent to null for the children
* The removed event will not be dispatched for the children anymore

Since these are our internal objects and I doubt that anybody cares about the add and remove event beeing fired basically every frame this change should be fine.

atomicsulfate: Agreed. In our code, once we remove an object from the root it's gone, there aren't any other references of it, so it doesn't matter if we don't clean its parent reference.
Also, these add/remove events make sense only if you actually keep a scene graph, and then you would have to handle it recursively (i.e. if you remove an object, you should also remove the whole subtree rooted in it).
I've double checked in THREEjs code and I haven't found any internal use of the 'removed' event.

ninok: okay, than I will approve

PR #1337, opened 2020-02-28
Remove special set up of Object.userData for solid lines

robertoraggi:


robertoraggi:
@atomicsulfate  i couldn't find anything requiring the solid-line to be handled differently.

PR #1341, opened 2020-02-28
Add support for filtering features based on their state

robertoraggi:
Add support for filtering features based on their state.

This change adds APIs to manipulate the state of a feature.
Note, that currently the feature state can only be used to
enable/disable features. That is, harp.gl themes can use
`["feature-state", property]` in expressions assigned to
the `enabled` attribute, for example:

    {
        when: ["==", ["geometry-type"], "LineString"],
        layer: "roads",
        technique: "solid-line",
        renderOrder: 10000,
        attr: {
            enabled: ["feature-state", "enabled"],
            lineWidth: "1px",
            color: "#c00"
        }
    }

Application code can programmatically update the state of a feature
using the DataSource API, for example:

    dataSource.setFeatureState(featureId, {
        enabled: true
    });
    mapView.update()

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -5,14 +5,19 @@
  */
 import {
     Env,
+    Expr,
     GeometryKind,
+    getFeatureId,
+    getPropertyValue,
     GradientSky,
     ImageTexture,
+    IndexedTechnique,
     Light,
     MapEnv,
     PostEffects,
     Sky,
-    Theme
+    Theme,
+    Value
 } from "@here/harp-datasource-protocol";
 import {
     EarthConstants,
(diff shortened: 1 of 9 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2001,6 +2006,35 @@ export class MapView extends THREE.EventDispatcher {
         this.update();
     }
 
+    /**
+     * Updates the value of a dynamic property.
+     *
+     * the style operator `["get", "property name", ["dynamic-properties"]]`.
+     *
+     * @param name The name of the property.
```
ninok: Please add a comment that name must not start with `$`

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2001,6 +2006,35 @@ export class MapView extends THREE.EventDispatcher {
...
+            throw new Error(`failed to update the value of the dynamic property '${name}'`);
+        }
+        this.m_env.entries[name] = value;
+        this.update();
+    }
+
+    /**
+     * Removes the given dynamic property from this [[MapView]].
+     *
+     * @param name The name of the property to remove.
```
ninok: same as above

@here/harp-mapview/lib/MapView.ts
robertoraggi: done

@here/harp-mapview/lib/MapView.ts
robertoraggi: done

PR #1343, opened 2020-02-29
HARP-7073: Use memoization to reduce the number of expr evaluations

robertoraggi:


@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -715,7 +722,10 @@ export function compileTechniques(techniques: Technique[]) {
             if (isJsonExpr(value) && propertyName !== "kind") {
                 // "kind" is reserved.
                 try {
-                    technique[propertyName] = Expr.fromJSON(value);
+                    const expr = Expr.fromJSON(value).intern(mapView.exprPool);
```
robertoraggi: we need APIs to shirk the expr pool.

robertoraggi:
GPU tests please

@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
```diff
@@ -101,7 +101,9 @@ export function getPropertyValue(property: Value | Expr | undefined, env: Env):
         return property.evaluate(env, ExprScope.Dynamic);
     }
 
-    if (typeof property !== "string") {
+    if (typeof property === "function") {
```
zbigg: Please update typings as well.

With this patch `Expr` cannot possibly exist in technique property in renderer (only! *), so first `if` is unnecessary ... but agin only in renderer.

Also it would be good to have same representation of dynamic attrs in renderer and decoder - possibly compile them same way in SSE - at least for "public" usage outside of SSE.

@here/harp-mapview/lib/DecodedTileHelpers.ts
zbigg: `MapView.set theme` is good enough IMO.
`Expr` instances are directly derived from `Theme`

@here/harp-datasource-protocol/lib/InterpolatedProperty.ts
robertoraggi: what do you mean with "Also it would be good to have same representation of dynamic attrs in renderer and decoder - possibly compile them same way in SSE - at least for "public" usage outside of SSE." ?

zbigg: I mean that in SSE compiles technique attribute to `Value | Expr` and decoder can observe these types.
In rendering phase technique attr after `compileTechniques` is `Value | function`.

Again 
 - 1) It breaks typings, we don't expect `function` there
 - 2) It breaks `Expr.isExpr` in many places to check if we really need to update these parameters

robertoraggi: all the Expr.isExpr should definitely go, we should change the type to `Value | Expr | function` (as you already mentioned) and maybe introduce a isDynamic(v) that should return true if v is an Expr or a function.

robertoraggi:
GPU tests please

PR #1349, opened 2020-03-04
HARP-7602: Add support for project references and out of source builds

robertoraggi:


@here/harp-datasource-protocol/tsconfig.json
```diff
@@ -2,13 +2,27 @@
...
+        "composite": true,
+        "strict": true,
         "strictNullChecks": true,
         "noImplicitAny": true,
         "noImplicitReturns": true,
         "noImplicitThis": true,
-        "noUnusedLocals": false,
-        "downlevelIteration": true
-    }
+        "noUnusedLocals": false
```
robertoraggi: I think I will add`tsconfig.common.json` in the root directory of harp.gl and included it in the the `@here/harp-*` packages

@here/harp-datasource-protocol/package.json
```diff
@@ -2,15 +2,15 @@
...
     "description": "Components used for the decoding and styling of data that is used by the Datasources.",
-    "main": "index.js",
-    "typings": "index",
+    "main": "dist/cjs/index.js",
+    "typings": "dist/cjs/index",
     "directories": {
         "test": "test"
     },
     "scripts": {
         "test": "cross-env mocha --require source-map-support/register $EXTRA_MOCHA_ARGS ./test/*.js",
```
zbigg: How about test files ?
Not sure, but lookg like they are generated by `tsconfig` if yes then probably not into `./test/..` but probably into `dist/cjs/test/` and ... possibly even land in final package.

@here/harp-datasource-protocol/tsconfig.json
```diff
@@ -2,13 +2,27 @@
     "compilerOptions": {
         "target": "es2017",
         "module": "commonjs",
+        "outDir": "dist/cjs",
```
zbigg: This changes layout of files so `@harp-mapview/lib/MapView.` will become `@harp-mapview/dist/*/lib/MapView` of course, people usually import from module `index`, but they need to be warned. 

#release-notes-needed ?

@here/harp-geojson-datasource/package.json
```diff
@@ -2,16 +2,16 @@
     "name": "@here/harp-geojson-datasource",
     "version": "0.13.0",
     "description": "Provides support for GeoJSON datasource",
-    "main": "index.js",
-    "typings": "index",
+    "main": "dist/cjs/index.js",
+    "typings": "dist/cjs/index",
     "worker": {
         "name": "@here/harp-geojson-datasource-worker",
         "main": "index-worker.js"
```
zbigg: `dist/cjs/index-worker.js` ... same in other occurences of `index-worker`.

@here/harp-debug-datasource/tsconfig.json
```diff
@@ -2,13 +2,30 @@
...
         "strictNullChecks": true,
         "noImplicitAny": true,
         "noImplicitReturns": true,
         "noImplicitThis": true,
-        "noUnusedLocals": false,
-        "downlevelIteration": true
-    }
+        "noUnusedLocals": false
+    },
+    "references": [
```
zbigg: You mentioned you have script to generate these references. It would be great to have it in this patch.

@here/harp-datasource-protocol/tsconfig.json
zbigg: I'm worried - not sure though that  that all code that imports directly from "@here/harp-*/lib/..." instead of index will fail.
And we do this internally:

```
$ grep -ER 'import .* from "@here/harp-.*/lib'. @here/harp-*/lib
@here/harp-datasource-protocol/lib/Theme.ts:import { Vector3Like } from "@here/harp-geoutils/lib/math/Vector3Like";
@here/harp-examples/lib/PerformanceUtils.ts:import { debugContext } from "@here/harp-mapview/lib/DebugContext";
@here/harp-geojson-datasource/lib/GeoJsonDecoder.ts:import { TileDecoderService } from "@here/harp-mapview-decoder/lib/TileDecoderService";
```

This will somehow work in our monorepo but my suspicion is that it will fail to find these files in workspace that contains only compiled installed by npm in node_modules.

PR #1352, opened 2020-03-06
HARP-7885: Support TypeScript project references.

robertoraggi:
This change adds explicit project references and the required
configuration options to build harp.gl modules using `tsc --build`.

Also, this change introduces a global tsconfig.base.json with
the common settings used by the harp.gl modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,7 @@ test-npm-packages-app/
 .DS_Store
 .nyc_output/
 rendering-test-results/
+*.tsbuildinfo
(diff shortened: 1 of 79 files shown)


.gitignore
```diff
@@ -6,5 +6,7 @@ test-npm-packages-app/
 .DS_Store
 .nyc_output/
 rendering-test-results/
+*.tsbuildinfo
+*.ts.map
```
zbigg: `js.map` ?

PR #1354, opened 2020-03-06
HARP-1026: Added api-extractor to Travis build

ninok:


@here/harp-geoutils/lib/projection/Projection.ts
```diff
@@ -218,7 +218,7 @@ export abstract class Projection {
         sourceProjection: Projection,
         worldPos: Vector3Like,
         result: WorldCoordinates
-    ): typeof result;
```
ninok: api-extractor cannot handle this

PR #1356, opened 2020-03-06
HARP-7942: Enable tslint deprecate check

ninok:


@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -338,7 +338,10 @@ export class OmvDecodedTileEmitter implements IOmvEmitter {
         const worldLines: number[][] = []; // lines in world space.
         const uvs: number[][] = [];
         const offsets: number[][] = [];
-        const { projectedTileBounds } = this.m_decodeInfo;
```
zbigg: @robertoraggi 

Not sure if {{projectedTileBounds}} should be deprecated - afair it was merely "cached" value of computation added below.

With this change, we're projecting bbox of tile for each line feature which is quite a waste ... not sure if significant though.

PR #1362, opened 2020-03-10
HARP-9344: WebTileDataSource - don't use const enums.

zbigg:
`const enum`s are not emitted to Javascript by default, so javascript
users would see `WebTileDataSource.ppiValue` as `undefined`.

Note `const enum`s are there only for performance reasons and those
contants are not intended to be used in performance sensitive code so
there is no reason to hack it around with `preserveConstEnums` compiler
flag.

### DIFF
--- a/@here/harp-webtile-datasource/lib/WebTileDataSource.ts
+++ b/@here/harp-webtile-datasource/lib/WebTileDataSource.ts
@@ -413,13 +413,13 @@ export class WebTileDataSource extends DataSource {
  * Definitions of variable values to be used with `WebTileDataSource`
  */
 export namespace WebTileDataSource {
-    export const enum ppiValue {
+    export enum ppiValue {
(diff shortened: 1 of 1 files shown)


@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -413,13 +413,13 @@ export class WebTileDataSource extends DataSource {
  * Definitions of variable values to be used with `WebTileDataSource`
  */
 export namespace WebTileDataSource {
```
zbigg: Not to to readers why we _still_ leave `namespace` and `class` merging.

We still didn't ban them and they are described in [Typescript Handbook](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) as legitimate language feature.

Let's remove them later and not only fix the bug.

PR #1365, opened 2020-03-10
HARP-7400: Added fallback for target and targetDistance

ninok:


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2574,18 +2571,63 @@ export class MapView extends THREE.EventDispatcher {
...
+            //For spherical projection we fallback to the tangent line distance
+            this.m_targetDistance =
+                this.projection.type === ProjectionType.Planar
+                    ? cameraPosZ / Math.cos(Math.min(cameraPitch, PITCH_LIMIT_RAD))
+                    : Math.sqrt(
+                          Math.pow(cameraPosZ + EarthConstants.EQUATORIAL_RADIUS, 2) -
+                              Math.pow(EarthConstants.EQUATORIAL_RADIUS, 2)
+                      );
+
+            const cameraDir = this.m_camera.getWorldDirection(cache.vector3[0]);
```
kkostecki: Shouldn't we use the same fallback (intersection point) for position calculation as for target distance calculus?
Here our target will fly in the void, I would rather assume that we lay back it on the surface:
TargetPos = CameraDir (with tilt constrained) * target distance

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2574,18 +2571,63 @@ export class MapView extends THREE.EventDispatcher {
         this.m_screenCollisions.update(width, height);
 
         this.m_pixelToWorld = undefined;
+        this.m_fog.update(this, this.m_viewRanges.maximum);
 
+        this.computeTargetAndDistance();
+        this.m_zoomLevel = MapViewUtils.calculateZoomLevelFromDistance(this, this.m_targetDistance);
```
kkostecki: Here I would strongly suggest to provide interface for zoom level calculation i.e.:
ZoomLevelEvaluator.
By default it would be the one based on target distance, but we could of course provide our own and customers will have free hand to define own implementation - for example ground distance based - as in other 3D engines.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2574,18 +2571,63 @@ export class MapView extends THREE.EventDispatcher {
...
+    }
+
+    private computeTargetAndDistance() {
         const cameraPitch = MapViewUtils.extractAttitude(this, this.m_camera).pitch;
```
kkostecki: This method is HIGHLY in-practical and over complicated for getting something like camera pitch, we may consider creating or using other method here (maybe something like extractCameraTilt()?).

ninok: Yes, this is just the first step. Eventually a lot of methods in MapViewUtils will be removed and either moved to MapView as private or completely removed.
My ultimate goal is to have one `MapViewState` object that is private in `MapView` and that wuill contain all necessary information (e.g. tilt, heading, target, targetDistance, ...). With this aproach we could also heavily reduce the public interface of `MapView` (e.g. nobody need `targetDistance` outside of `MapView`)

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2574,18 +2571,63 @@ export class MapView extends THREE.EventDispatcher {
...
-        const cameraPosZ = this.getCameraHeightAboveTerrain(TERRAIN_ZOOM_LEVEL);
-        const zoomLevelDistance = cameraPosZ / Math.cos(Math.min(cameraPitch, Math.PI / 3));
-        this.m_zoomLevel = MapViewUtils.calculateZoomLevelFromDistance(this, zoomLevelDistance);
-        this.m_fog.update(this, this.m_viewRanges.maximum);
 
-        const target = MapViewUtils.getWorldTargetFromCamera(this.m_camera, this.projection);
+        // Even for tilt 90 in flat projection getWorldTargetFromCamera is returning some point
```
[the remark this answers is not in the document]
ninok: true, I renamed the method after adding the comment

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2574,18 +2571,63 @@ export class MapView extends THREE.EventDispatcher {
...
+        const target = cameraPitch < PITCH_LIMIT_RAD ? this.raycastTargetFromCamera() : null;
```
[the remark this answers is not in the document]
ninok: The tilt limitation for the controls is a completely different thing and most likely it is much smaller (around 65 for navigation usecases).
The limit here is for the generic usecase (i.e. no map controls or custom map controls) where the user is modifying the camera directly.

@here/harp-mapview/lib/MapView.ts
ninok: The whole concept of target point makes not much sense if you have a free camera. We just do this here to always return a target point in our API. We could also change the API to not return a target point at all in these cases but I think also a lot of internal code is relying on a target point.
Currently there is at least one property that always holds for the target point: The (screen)projected target point is always in the middle of the screen. With your proposal that would not be true anymore, but as I said before what we choose does not really matter nobody cares about the target point in free camera mode.
The only important thing is the targetDistance b/c we use it to compute the zoomlevel.

@here/harp-mapview/lib/MapView.ts
ninok: Actually I think we need an interface for computing the distance b/c converting distance to zoomlevel will never change.
As I mentioned in another comment my goal is to create a proper `MapViewState` object and maybe we could have an interface that computes the whole `MapViewState`

ninok: Or giving it a second thought that might be a bad idea b/c the whole `MapViewState` exercise is to reduce our public API. If we have that interface `MapViewState` would be public again.
Anyways: I think this is out of scope for this task but we should keep it in mind.

ninok:

@here/harp-mapview/lib/MapView.ts
kkostecki: Ok, understood... although keep in mind that MapViewState may be still inaccessible in few components may be unlinked from MapView - dependency removed:
- ClipPlanesEvaluator(s),
- ZoomLevelEvaluator(s),
- new one - MapViewAtmosphere.

ninok: The MapViewState object will be a private member of MapView, but it can still be passed to the components that need it.
So instead of having the components pull what they need from MapView (hence the MapView API explodes) MapView would rather push the necessary properties to the components (like you did already for the ClipPlaneEvaluator).

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2592,18 +2585,77 @@ export class MapView extends THREE.EventDispatcher {
...
+        } else {
+            // We either reached the [[PITCH_LIMIT]] or we did not hit the ground surface.
+            // In this case we do the reverse, i.e. compute some fallback distance and
+            // use it to compute the tagret point by using the camera direction.
+            const cameraPosZ = this.getCameraHeightAboveTerrain(TERRAIN_ZOOM_LEVEL);
+
+            //For flat projection we fallback to the target distance at 89 degree pitch.
+            //For spherical projection we fallback to the tangent line distance
+            this.m_targetDistance =
+                this.projection.type === ProjectionType.Planar
```
kkostecki: Please see my corresponding patchset #1399. Two things I need to mention here:
- 89 degrees will not always return fine tuned target point, because angle to visual horizon highly depends on the camera altitude and at high attitudes it will end-up very far from the map itself.
- for spherical projections we may need some tolerance instead on tangent point, because the rotation (controls) math may break up quickly on the corner case.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2592,18 +2585,77 @@ export class MapView extends THREE.EventDispatcher {
...
+
+            const cameraDir = this.m_camera.getWorldDirection(cache.vector3[0]);
+            cameraDir.multiplyScalar(this.m_targetDistance);
+            this.m_targetWorldPos.copy(this.m_camera.position).add(cameraDir);
+            this.m_targetGeoPos = this.projection.unprojectPoint(this.m_targetWorldPos);
+        }
+        this.m_zoomLevel = MapViewUtils.calculateZoomLevelFromDistance(this, this.m_targetDistance);
+    }
+
+    private raycastTargetFromCamera(): THREE.Vector3 | null {
```
kkostecki: I far as I see, this function is not much different then `MapViewUtils.getWorldTargetFromCamera()`, why not use this one? Otherwise please move (reneme) it there, I will need it in ClipPlaneEvaluator(s) and Atmosphere component too.

ninok: I want to get rid of `MapViewUtils.getWorldTargetFromCamera()`. I don't see any usage outside of `MapView`. I also marked it as deprecated.

@here/harp-mapview/lib/MapView.ts
ninok: Maybe to go forward I could integrate the changes from #1399 int this PR?!

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2592,18 +2585,77 @@ export class MapView extends THREE.EventDispatcher {
...
+            // We have a valid target point. Use it to compute the distance.
             this.m_targetWorldPos.copy(target);
-            this.m_targetGeoPos = this.projection.unprojectPoint(target);
+            this.m_targetGeoPos = this.projection.unprojectPoint(this.m_targetWorldPos);
             this.m_targetDistance = this.camera.position.distanceTo(target);
+        } else {
+            // We either reached the [[PITCH_LIMIT]] or we did not hit the ground surface.
+            // In this case we do the reverse, i.e. compute some fallback distance and
+            // use it to compute the tagret point by using the camera direction.
+            const cameraPosZ = this.getCameraHeightAboveTerrain(TERRAIN_ZOOM_LEVEL);
```
kkostecki: We will certainly have the problem with this method (function) too. What I would suggest to move this to MapViewUtils as function taking THREE.Object3D and elevation provider. This way we will have handy interface also for landscape objects placement.
With this change we may pass elevation provider to ClipPlanesEvaluators so they will be also aware of that.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,6 +256,89 @@ export namespace MapViewUtils {
...
+    /**
+     * @internal
+     * @deprecated This method will be moved to MapView.
+     */
+    export function getTargetAndDistance(
+        projection: Projection,
+        camera: THREE.PerspectiveCamera,
+        elevationProvider?: ElevationProvider
+    ): { target: THREE.Vector3; distance: number } {
+        const cameraPitch = extractAttitude({ projection }, camera).pitch;
```
kkostecki: We just need to remember about improving pitch calculus - now is doing too much....

ninok: yeah, @zbigg also commented in my other PR, so I will remember :-)

PR #1375, opened 2020-03-12
HARP-9037 Refactored DataSource zoom level limits

MPursche:
* Added minDataLevel, maxDataLevel, minDisplayLevel, maxDisplayLevel.
* Deprecated minZoomLevel, maxZoomLevel.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/DataSource.ts
+++ b/@here/harp-mapview/lib/DataSource.ts
@@ -5,12 +5,59 @@
  */
 import { Definitions, StyleSet, Theme, ValueMap } from "@here/harp-datasource-protocol";
 import { Projection, TileKey, TilingScheme } from "@here/harp-geoutils";
-import { assert } from "@here/harp-utils";
+import { assert, LoggerManager } from "@here/harp-utils";
 import * as THREE from "three";
 import { MapView } from "./MapView";
 import { Tile } from "./Tile";
 
+const logger = LoggerManager.instance.create("DataSource");
 const UPDATE_EVENT = { type: "update" };
+
+/**
+ * Options for a [[DataSource]].
+ */
+export interface DataSourceOptions {
+    /**
+     * The unique name of a [[DataSource]] instance.
+     */
+    name?: string;
+    /**
+     * The name of the [[StyleSet]] to evaluate for the decoding.
+     */
+    styleSetName?: string;
+    /**
+     * The minimum zoom level at which data is available or displayed at
+     * (depending on [[DataSource]] subclass).
+     * @deprecated Use [[minDataLevel]] and [[minDisplayLevel]] instead.
+     */
+    minZoomLevel?: number;
+    /**
+     * The maximum zoom level at which data is available or displayed at
+     * (depending on [[DataSource]] subclass).
+     * @deprecated Use [[maxDataLevel]] and [[maxDisplayLevel]] instead.
+     */
+    maxZoomLevel?: number;
+    /**
+     * The minimum zoom level at which data is available.
+     */
+    minDataLevel?: number;
+    /**
+     * The maximum zoom level at which data is available.
+     */
+    maxDataLevel?: number;
+    /**
+     * The minimum zoom level at which [[DataSource]] is displayed.
+     */
+    minDisplayLevel?: number;
+    /**
+     * The maximum zoom level at which [[DataSource]] is displayed.
+     */
+    maxDisplayLevel?: number;
+    /**
+     * Storage level offset applied to this `DataSource`.
+     */
+    storageLevelOffset?: number;
+}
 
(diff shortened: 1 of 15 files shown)


@here/harp-mapview/lib/DataSource.ts
```diff
@@ -369,11 +502,19 @@ export abstract class DataSource extends THREE.EventDispatcher {
     getDisplayZoomLevel(zoomLevel: number): number {
         return THREE.MathUtils.clamp(
             zoomLevel + this.m_storageLevelOffset,
-            this.m_minZoomLevel,
-            this.m_maxZoomLevel
+            this.m_minDataLevel,
```
nzjony: Shouldn't this use the min/max display levels?

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -292,27 +365,87 @@ export abstract class DataSource extends THREE.EventDispatcher {
...
+            "DataSource.maxZoomLevel is deprecated. Use maxDataLevel and maxDisplayLevel instead."
+        );
+        this.m_maxDataLevel = level;
+        this.m_maxDisplayLevel = level;
+    }
+
+    /**
+     * The minimum zoom level at which data is available.
+     */
+    get minDataLevel(): number {
```
nzjony: why not just have a public property? I don't see the need to have get / set properties.

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -292,27 +365,87 @@ export abstract class DataSource extends THREE.EventDispatcher {
...
+        return this.m_maxDataLevel;
     }
 
     set maxZoomLevel(level: number) {
-        this.m_maxZoomLevel = level;
+        logger.warn(
+            "DataSource.maxZoomLevel is deprecated. Use maxDataLevel and maxDisplayLevel instead."
+        );
+        this.m_maxDataLevel = level;
+        this.m_maxDisplayLevel = level;
```
nzjony: this won't work for OLP, they have a maxZoomLevel of 1, but they want to display it up to level 9

nzjony: Nevermind, just read this is deprecated.

@here/harp-mapview/lib/DataSource.ts
MPursche: No the display zoom level controls what level of a `DataSource` should be displayed(taking the storage offset into account and clamping it to the data level range). It doesn't control if the `DataSource` should be displayed at a given zoom level. Before this PR it was done by `canGetTile` on a per tile level. I introduced `isVisible(zoomLevel)` to get around this issue because it can be done for the whole `DataSource` instead, which potentially makes culling more efficient(no need to cull each tile if we know the whole `DataSource` shouldn't be displayed).

@here/harp-mapview/lib/DataSource.ts
MPursche: The previous options `minZoomLevel` and `maxZoomLevel` also used properties getter and setter so I did the same for the new options. But true this can be simplified if getter and setter don't do anything special.

@here/harp-mapview/lib/DataSource.ts
nzjony: > It doesn't control if the DataSource should be displayed at a given zoom level.

You are saying that `getDisplayZoomLevel` shouldn't dictate if the Datasource can be displayed at supplied zoom level? This to me is very confusing. This should be renamed probably to `getDataZoomLevel`, because if I understand you correctly it gets the data level that should be displayed at the supplied zoom level and not the display level.

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -407,14 +494,22 @@ export abstract class DataSource extends THREE.EventDispatcher {
      * @param zoomLevel The zoom level of the [[MapView]].
      * @returns The zoom level to use for display.
```
nzjony: this comment needs to be updated

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -1226,7 +1226,7 @@ export class VisibleTileSet {
                 // For each data source check what tiles from the intersection should be rendered
                 // at this zoom level.
                 const visibleTileKeys: TileKeyEntry[] = [];
-                const displayZoomLevel = dataSource.getDisplayZoomLevel(zoomLevel);
+                const displayZoomLevel = dataSource.getDataZoomLevel(zoomLevel);
```
nzjony:
```suggestion
                const dataZoomLevel = dataSource.getDataZoomLevel(zoomLevel);
```

PR #1378, opened 2020-03-12
HARP-5925: Sky and ground atmospheric scattering for globe projection.

kkostecki:
Provide physical approximation of real life atmospheric scattering effect.
Patchset includes:
- MapViewAtmosphere - main atmosphere controller exposed for the users,
- GroundAtmosphereMaterial - material with a shader approximating rayleigh
and mia scattering effect visible on globe surface,
- SkyAtmosphereMaterial - material containing shader for simulation of
atmosphere around the globe,
- AtmosphereChunks - GLSL shader chunks with common functions used withing
the shaders,
- rendering_globe-atmosphere - example showing use case with spherical
projection,
- Atmosphere interface - base Theme interface for further theme based
configuration.

# CI Check

Happy contributing!


@here/harp-mapview/lib/MapViewAtmosphere.ts
```diff
@@ -0,0 +1,424 @@
...
+    /**
+     * Creates and adds `Atmosphere` effects.
+     *
+     * @note Currently works only with globe projection.
+     *
+     * @param m_mapView [[MapView]] instance where the effect will be added.
+     * @param m_atmosphere Atmosphere configuration parameters.
+     */
+    constructor(
+        private m_mapView: MapView,
```
ninok: Why do we need full MapView? I think we only need scene, projection and camera.
Passing MapView everywhere is like cancer in our codebase and at least for new code we should stop it.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -103,6 +103,10 @@ export type MapAnchor<T extends THREE.Object3D = THREE.Object3D> = T & {
      * The position of this [[MapObject]] in [[GeoCoordinates]].
      */
     geoPosition?: GeoCoordinates;
+    /**
+     * The position of this [[MapObject]] in [[WorldCoordinates]]
```
ninok: `WorldCoordinates` is not a type.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -50,7 +50,7 @@ export interface ClipPlanesEvaluator {
  * slightly different in each case. General algorithm sets near plane between camera and
  * ground level, while far plane is just calculated using scale and bias approach with far offset
  * and multiplier.
- * @deprecated Class contains the legacy (first and original) clip planes evaluation method, which
+ * @note Class contains the legacy (first and original) clip planes evaluation method, which
```
ninok: Please keep the deprecated. If you really need to use this class, disable the depreacation warning in the place where you use it

@here/harp-mapview/lib/MapViewAtmosphere.ts
kkostecki: Of course I know about that... and wasn't very happy to add it here, but unfortunately I need the access to MapView.anchors and MapView.scene. What I could do anyway is to replace map view with anchors list and scene reference, that should be enough to configure this component... What do you think?

@here/harp-mapview/lib/MapView.ts
kkostecki: Yeah... it was before "typedef" for Vec3, of course fixed that.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: That is something I would like to discuss. The depreciation flag in InterpolatedClipPlanesEvaluator was added by me personally :), cause we agreed that it will be no longer used / needed, but it seems that we actually have the use case for it - atmosphere needs different near/far planes then entire globe / map. So either I will disable tslint here, or remove depreciation or implement something similar inside MapViewAthmosphere, but I guess if we had it why not use it...

@here/harp-materials/lib/SkyAtmosphereMaterial.ts
```diff
@@ -0,0 +1,423 @@
...
+    ${AtmosphereShaderChunks.atmosphere_vertex_utils}
+
+    //
+    // Computes rayleight and mia atmosphere factors for sky.
+    //
+    // Code based on GPU Gems article.
+    //
+    // Author: Sean O'Neil
+    //
+    // Copyright (c) 2004 Sean O'Neil
```
MPursche: What kind of license does the shader from GPU gems use? I think we need to check if it is compatible with our Apache 2 license.

@here/harp-examples/src/rendering_globe-atmosphere.ts
```diff
@@ -0,0 +1,94 @@
...
+    enum DataSourceVariant {
+        Omv,
+        Satellite
+    }
+
+    // Create a new MapView for the HTMLCanvasElement of the given id.
+    function initializeMapView(id: string): MapView {
+        const canvas = document.getElementById(id) as HTMLCanvasElement;
+        // tslint:disable-next-line:deprecation
+        const clipPlanesEvaluator = new InterpolatedClipPlanesEvaluator(
```
MPursche: Is it possible to make this effect work with our other ClipPlanesEvaluators? Maybe we need to use a different camera(with different near/far plane setup) to render the atmospheric geometries.

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -1003,6 +1003,14 @@ export interface TextStyleDefinition {
     vAlignment?: "Above" | "Center" | "Below";
 }
 
+/**
+ * Interface that defines a procedural atmosphere overlay effect.
+ */
+export interface Atmosphere {
```
MPursche: Is this used anywhere?

@here/harp-examples/src/rendering_globe-atmosphere.ts
kkostecki: Yep this final plan is to integrate InterpolatedClipPlanesEvaluator into the MapViewAtmosphere, so near/far it is gonna be changed dynamically within onBeforeRender(), and restored onAfterRender(). This way all tiles geometry will use TiltViewClipPlanes... while the atmosphere geometry will have its own.

@here/harp-datasource-protocol/lib/Theme.ts
kkostecki: This will be interface for between Theme (definition) and MapViewAtmosphere, but you have wright I will remove it from this PR for now.... Because Theme atmosphere definition (and parsing) is not implemented yet.

@here/harp-materials/lib/SkyAtmosphereMaterial.ts
kkostecki: Yeah, already reported that to Stephan, so he will take care of that.

@here/harp-examples/src/rendering_globe-atmosphere.ts
kkostecki: I have resolved this problem so there is need to define _special_ clip planes evaluator for MapView component. Instead of creating new camera setup internally, we just update camera near/far just before rendering atmosphere object (sky), after render clip planes are restored for the rest of scene, so full integration is now added.

@here/harp-datasource-protocol/lib/Theme.ts
kkostecki: I have removed it from this PR.

@here/harp-mapview/lib/MapViewAtmosphere.ts
kkostecki: @ninok unfortunately when I implemented full integration with MapView component, so there is no need for special ClipPlanesEvaluator, I am even more constraint to this reference, cause I need to access current view ranges.... See my comments below.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
ninok: Okay, if you need it than remove the deprecation but please also adjust the comment (i.e. remove the `legacy (first and original)` and maybe also add that this is used internally for the atmosphere

@here/harp-mapview/lib/MapViewAtmosphere.ts
ninok: @kkostecki Still you could pass whatever you need e.g. in the update method.
I'm also wondering if we should really use the map anchor api for the atmosphere as well. We can do the world position update also in the atmosphere update.
Would be weird for the user to encounter the atmosphere when iterating through the map anchors. You would only expect your own objects.

@here/harp-mapview/lib/MapViewAtmosphere.ts
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    private get disposed() {
+        return this.m_skyMesh === undefined && this.m_groundMesh === undefined;
+    }
+
+    /**
+     * Handles atmosphere addition.
+     */
+    private addToScene(scene: THREE.Scene) {
```
ninok: I think this should be part of the interface and MapView could call it every frame, like we do for the Tile objects

@here/harp-mapview/lib/MapViewAtmosphere.ts
```diff
@@ -0,0 +1,481 @@
...
+        }
+        if (this.m_groundMesh !== undefined) {
+            scene.add(this.m_groundMesh);
+        }
+    }
+
+    /**
+     * Handles atmosphere removal.
+     */
+    private removeFromScene(scene: THREE.Scene) {
```
ninok: as above, if we do it like for the tile objects, the scene would be cleared every frame and the atmosphere would be readded

@here/harp-mapview/lib/MapViewAtmosphere.ts
```diff
@@ -0,0 +1,481 @@
...
+            if (this.m_skyMesh !== undefined) {
+                scene.remove(this.m_skyMesh);
+            }
+            if (this.m_groundMesh !== undefined) {
+                scene.remove(this.m_groundMesh);
+            }
+        }
+    }
+
+    private addToMapView(mapView: MapView) {
```
ninok: I'm confused, do we use map anchors or do we add the meshes to the scene?

@here/harp-mapview/lib/MapViewAtmosphere.ts
ninok: Just realized, there is no interface. So just make it public ad call it from mapview

@here/harp-mapview/lib/MapViewAtmosphere.ts
```diff
@@ -0,0 +1,481 @@
...
+
+    /**
+     * Updates the atmosphere at runtime, depending on the camera and projection settings.
+     *
+     * @param camera An instance of a `THREE.Camera`.
+     * @param projection An instance of Projection currently in use.
+     */
+    update(camera: THREE.PerspectiveCamera | THREE.OrthographicCamera, projection: Projection) {
+        assert(
+            projection.type === this.m_mapView.projection.type,
```
ninok: I guess this is called from MapView, so the check will always return true (i.e. projection and this.m_mapView.projection are the same).
One more reason not to keep map view

@here/harp-mapview/lib/MapViewAtmosphere.ts
kkostecki: @ninok regarding your last comment - of course we may not use anchors, but then we need to re-implement (duplicate) entire logic of objects handling - related to map position. You probably remember that on globe camera is in origin, but entire globe is transformed. Similar (but different transforms) situation we will have for planar projection (if ever supported with the same component).
Another question, how would you like to `update` MapViewAtmosphere` component? Do you want to create MapView - > Atmosphere dependency or create some generic container for "other" objects handled via MapView? Please clarify.

@here/harp-mapview/lib/MapViewAtmosphere.ts
kkostecki: Proposed new way of exposing worldAnchor(s) via MapView, see the second commit in this PR.

@here/harp-mapview/lib/MapViewAtmosphere.ts
kkostecki: See response above.

@here/harp-mapview/lib/MapViewAtmosphere.ts
kkostecki: Fixed now.

@here/harp-mapview/lib/MapViewAtmosphere.ts
kkostecki: This method will be gonne. It is not required and neither used, I though to find some way for MapViewAtmosphere (global) update, but this would probably require creating reversed dependency MapView -> MapViewAtmosphere, which I am trying to avoid.

@here/harp-mapview/lib/MapViewAtmosphere.ts
```diff
@@ -0,0 +1,478 @@
...
+            _group: THREE.Group
+        ) => {
+            assert(material instanceof GroundAtmosphereMaterial);
+            const mat = this.m_groundMaterial as GroundAtmosphereMaterial;
+            mat.updateUniforms(mat, this.m_groundMesh!, camera, this.m_lightDirection);
+        };
+    }
+
+    private overrideClipPlanes(rteCamera: THREE.Camera) {
+        const { camera, projection } = this.m_mapView;
```
kkostecki, line 448: This is the biggest problem, as you may see we use both "rendering" (so called relative to eye) camera as it goes to rendering pipeline and our globally positioned MapView available camera.
Please answer how could we possibly "update" MapViewAtmosphere with map view camera reference without creating MapView->MapViewAtmosphere dependency.
We may either leave MapViewAtmosphere -> MapView dependency or provide callback mechanism to update MapViewAtmosphere from MapView in generic way. My proposal is via MapView->MapAnchor that would contain optional callback method, so the interface would look like this:
```
export type MapAnchor<T extends THREE.Object3D = THREE.Object3D> = T & {
    /**
     * The position of this [[MapAnchor]] in [[GeoCoordinates]].
     */
    geoPosition?: GeoCoordinates;
    /**
     * Flag defining if the object may be picked.
     *
     * @note By default all objects are pick-able even if this flag is undefined.
     */
    pickable?: boolean;
    /**
     * Optional callback to get updates from MapView.
     */
    updateCallback?: (mapView: MapView, time: number) => void;
};
```

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -58,8 +58,8 @@ export interface ClipPlanesEvaluator {
  * slightly different in each case. General algorithm sets near plane between camera and
  * ground level, while far plane is just calculated using scale and bias approach with far offset
  * and multiplier.
- * @deprecated Class contains the legacy (first and original) clip planes evaluation method, which
- * is widelly used in examples thus is still kept for backward compatibility and comparisons.
+ * @depreciated Class contains the legacy (first and original) clip planes evaluation method, which
```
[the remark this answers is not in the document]
kkostecki: Fixed. Now I am using different evaluator.

@here/harp-mapview/lib/MapViewAtmosphere.ts
```diff
@@ -0,0 +1,478 @@
...
+        assert(!MapViewAtmosphere.isPresent(mapView.scene), "Atmosphere already added");
+        if (this.m_skyMesh !== undefined) {
+            mapView.worldAnchors.add(createWorldAnchor(this.m_skyMesh, Number.MIN_SAFE_INTEGER));
+        }
+        if (this.m_groundMesh !== undefined) {
+            mapView.worldAnchors.add(createWorldAnchor(this.m_groundMesh, Number.MAX_SAFE_INTEGER));
+        }
+
+        // Request an update once the anchor is added to [[MapView]].
+        mapView.update();
```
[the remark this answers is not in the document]
kkostecki: True. If we change dependency chain. Let's see what @ninok will say. For now I have provided update callback so MapViewAtmosphere is kind of independent from MapView -> does not have reference to it.

@here/harp-mapview/lib/MapViewAtmosphere.ts
kkostecki: Not a bad idea... The question is if we wont to grow this PR with all this changes? It would of course require MapViewFog refactoring.

@here/harp-mapview/lib/MapViewAtmosphere.ts
kkostecki: I think we could _integrate_ all so called MapView effects (or scene effects) in the same way, as you proposed in previous comments. Although I am not sure if it should be a part of this -> already big PR.

@here/harp-materials/lib/ShaderChunks/AtmosphereChunks.ts
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export default {
+    atmosphere_common_utils: `
+
+    float conditionalBranchFree(bool condition, float primary, float secondary) {
```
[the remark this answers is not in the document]
kkostecki: It is, but in more explanatory way... I can change it of course.

@here/harp-mapview/lib/MapViewAtmosphere.ts
ninok: I'm fine with merging this one and create a follow up ticket to align MapViewFog and MapViewAtmosphere

@here/harp-materials/lib/SkyAtmosphereMaterial.ts
```diff
@@ -0,0 +1,426 @@
...
+                0,
+                // Toggles the light modes:
+                // 0 - light always directly overhead,
+                // 1 - lighting uses light direction: uniform u_lightDirectionWorld
+                1
+            )
+        ),
+        u_hsvCorrection: new THREE.Uniform(new THREE.Vector3(0, 0, 0)),
+
+        topColor: new THREE.Uniform(new THREE.Color(0x0077ff)),
```
MPursche, line 40: See comment on other material.

MPursche:
I think the branch just needs to be rebased for IBCT to pass(a shadow rendering test is falining).

PR #1379, opened 2020-03-13
HARP-9275: Test intersection with tiles on picking.

atomicsulfate:
Only objects in intersected tiles are tested for intersection.

Tile bounding boxes are updated to go from minElevation to
maxElevation + maxGeometryHeight.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/PickHandler.ts
+++ b/@here/harp-mapview/lib/PickHandler.ts
@@ -7,6 +7,7 @@
 import { GeometryType, getFeatureId, Technique } from "@here/harp-datasource-protocol";
 import * as THREE from "three";
 
+import { OrientedBox3 } from "@here/harp-geoutils";
 import { MapView } from "./MapView";
 import { MapViewPoints } from "./MapViewPoints";
 import { RoadPicker } from "./RoadPicker";
@@ -97,6 +98,8 @@ export interface PickResult {
     userData?: any;
 }
 
+const tmpOBB = new OrientedBox3();
+
 /**
  * Handles the picking of scene geometry and roads.
  */
@@ -135,7 +138,6 @@ export class PickHandler {
      */
     intersectMapObjects(x: number, y: number): PickResult[] {
         const worldPos = this.mapView.getNormalizedScreenCoordinates(x, y);
-
         const rayCaster = this.mapView.raycasterFromScreenPoint(x, y);
         const pickResults: PickResult[] = [];
 
@@ -147,8 +149,19 @@ export class PickHandler {
             this.mapView.textElementsRenderer.pickTextElements(scenePosition, pickResults);
         }
 
-        // calculate objects intersecting the picking ray
-        const intersects = rayCaster.intersectObjects(this.mapView.worldRootObject.children, true);
+        const intersects: THREE.Intersection[] = [];
+        const tileList = this.mapView.visibleTileSet.dataSourceTileList;
+        tileList.forEach(dataSourceTileList => {
+            dataSourceTileList.renderedTiles.forEach(tile => {
+                tmpOBB.copy(tile.boundingBox);
+                tmpOBB.position.sub(this.mapView.worldCenter);
+
+                if (tmpOBB.intersectsRay(rayCaster.ray) !== undefined) {
+                    rayCaster.intersectObjects(tile.objects, true, intersects);
+                }
+            });
+        });
+
         for (const intersect of intersects) {
             const pickResult: PickResult = {
                 type: PickObjectType.Unspecified,
@@ -228,6 +241,8 @@ export class PickHandler {
         pickResults.sort((a: PickResult, b: PickResult) => {
             return a.distance - b.distance;
         });
+
+        this.mapView.update();
 
         return pickResults;
     }
--- a/@here/harp-mapview/lib/PickHandler.ts
+++ b/@here/harp-mapview/lib/PickHandler.ts
@@ -242,8 +242,6 @@ export class PickHandler {
             return a.distance - b.distance;
         });
 
-        this.mapView.update();
-
         return pickResults;
     }
 
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -285,11 +285,6 @@ export class Tile implements CachedResource {
     readonly boundingBox = new OrientedBox3();
 
     /**
-     * Maximum height of geometry on this tile above ground level.
-     */
-    maxGeometryHeight: number = 0;
-
-    /**
      * A record of road data that cannot be intersected with THREE.JS, because the geometry is
      * created in the vertex shader.
      */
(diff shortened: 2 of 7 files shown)


@here/harp-mapview/lib/Tile.ts
```diff
@@ -685,23 +687,52 @@ export class Tile implements CachedResource {
...
+     * @param maxElevation Maximum tile elevation in meters.
+     */
+    setElevation(minElevation: number, maxElevation: number) {
+        if (this.m_minElevation === minElevation && this.m_maxElevation === maxElevation) {
+            return;
+        }
+        this.m_minElevation = minElevation;
+        this.m_maxElevation = maxElevation;
+
+        if (this.m_maxGeometryHeight !== undefined) {
```
MPursche: Would be nice to extend the `Tile` unit test to also include the new methods that update the bounding box.

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -229,6 +242,8 @@ export class PickHandler {
             return a.distance - b.distance;
         });
 
+        this.mapView.update();
```
MPursche: Why is it necessary to update the mapview after the object picking?

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -147,8 +149,19 @@ export class PickHandler {
             this.mapView.textElementsRenderer.pickTextElements(scenePosition, pickResults);
         }
 
-        // calculate objects intersecting the picking ray
-        const intersects = rayCaster.intersectObjects(this.mapView.worldRootObject.children, true);
+        const intersects: THREE.Intersection[] = [];
```
MPursche: The `PickHandler` doesn't seem to have a unit test right now so we should probably add one.

@here/harp-mapview/lib/PickHandler.ts
atomicsulfate: It doesn't, that's part of some test code I left there unintentionally. good catch.

@here/harp-mapview/lib/PickHandler.ts
atomicsulfate: Agreed. But I'm gonna change this code further as part of the same ticket in next review and then I'm gonna delete PickHandler and move the code to PickingRaycaster. I'd rather add the tests then, otherwise I'd have to keep changing the tests on the way.
I've tested manually this change on planar/sphere and w/o terrain and it seems to be fine.

@here/harp-mapview/lib/Tile.ts
atomicsulfate: Will do.

atomicsulfate:

@here/harp-mapview/lib/PickHandler.ts
kkostecki: This new approach has this side effect that will not iterate objects that are not on tiles. I am not sure how much we want this, but I remember some examples adding geometry / objects as free scene entities. Would be good to ask @ninok if we would like this functionality. 
Probably the most obvious scenario is to add object via MapView anchors.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -378,9 +373,14 @@ export class Tile implements CachedResource {
     // It's `Undefined` when no text content has been added yet.
     private m_textElementsChanged: boolean | undefined;
 
+    // Center of the tile's unelevated bounding box world coordinates.
+    private readonly m_center = new THREE.Vector3();
```
kkostecki: Maybe m_centerWorld / worldCenter ?

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1057,6 +1061,38 @@ export class Tile implements CachedResource {
...
     }
 
+    /**
+     * Gets the tile's bounding box.
+     */
+    get boundingBox(): OrientedBox3 {
+        return this.m_boundingBox;
+    }
+
+    private setBoundingBox(newBoundingBox?: OrientedBox3) {
```
kkostecki: Sorry for being picky, but honestly I dislike this kind of setter (especially if we have "real" getter). Worth to note that property setter itself can be also private - like this method. On other side calling:
```
this.setBoundingBox()
```
without arguments seems to be strange... Maybe simply provide another method: `invalidateWorldBox()` that will convert geo to world bounding box, and leave the setter withing convention.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -598,11 +597,8 @@ export class VisibleTileSet {
             const tiles = renderListEntry.renderedTiles;
             tiles.forEach(tile => {
                 tile.update(renderListEntry.zoomLevel);
-                minElevation = MathUtils.min2(minElevation, tile.minElevation);
-                maxElevation = MathUtils.max2(
-                    maxElevation,
-                    tile.maxElevation + tile.maxGeometryHeight
-                );
+                minElevation = MathUtils.min2(minElevation, tile.geoBox.minAltitude);
+                maxElevation = MathUtils.max2(maxElevation, tile.geoBox.maxAltitude);
```
kkostecki: Please check if this is still compatible with datasource / polygons example.

@here/harp-mapview/lib/PickHandler.ts
ninok: Also previously we only iterated the tile objects (i.e. the ones in `m_worldRootObject`), so this is not changing.
Since `MapView.worldRootObject` is public for whatever reason, users could add objects there but we should forbid this and don't expose `worldRootObject`.
Nevertheless, you raised a good point, that picking is currently neither working for `mapAnchors` nor for any other object that is added to the scene.
The user could create an own raycaster with `MapView.raycasterFromScreenPoint` and intersect everything in the scene, but that would be quite slow b/c this approach would not do the early tile intersection.
I think we should have a `boolean` flag in `intersectMapObjects` to control whether only our objects or all objects are returned (even though the name `intersectMapObjects` is questionable then).
The bare minimum we should do is to test the `mapAnchor` objects as well b/c they are map objects.

atomicsulfate: @ninok @kkostecki Another option is to insert the tiles themselves as objects in the Scene graph, and override their intersect method to continue with their children only if their intersection test succeeds.
That way we handle both tile and non-tiled objects in the same way.

@here/harp-mapview/lib/Tile.ts
atomicsulfate: The problem is setter and getter must have same access specifier, otherwise you get a ts-lint error. The setter must be however private, code outside the class is not supposed to change the bounding box in the tile. 
I've added this method to keep in sync the bounding box with the worldCenter of the tile, their update must happen in the same single place.

However, I agree that with that name it's weird to call setBoundingBox without arguments. So I'll rename to updateBoundingBox and add a comment clarifying what happens when you don't pass a bounding box.

@here/harp-mapview/lib/Tile.ts
kkostecki: I saw it is done, so marking as resolved.

@here/harp-mapview/lib/PickHandler.ts
kkostecki: Ok good idea, let's proceed in different PR.

@here/harp-mapview/lib/Tile.ts
kkostecki: Nice, I like it too!

@here/harp-mapview/lib/PickHandler.ts
ninok: Yeah, I like that idea.

atomicsulfate:

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: It seems to work fine:
https://saeljira.it.here.com/secure/attachment/1217899/harp.gl%20_%20examples.webm

atomicsulfate:

@here/harp-mapview/lib/Tile.ts
atomicsulfate: done

PR #1391, opened 2020-03-17
MINOR: Fix WebTileDataSource API break due to removal of appId and appCode

ninok:


ninok:

PR #1392, opened 2020-03-17
HARP-7835: Improved lookAt API

ninok:
* MapView.lookAt takes an object parameter now
    * added getters for tilt and heading

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -292,7 +292,7 @@ export enum MapViewPowerPreference {
 /**
  * User configuration for the [[MapView]].
  */
-export interface MapViewOptions extends TextElementsRendererOptions, LookAtParams {
+export interface MapViewOptions extends TextElementsRendererOptions, Partial<LookAtParams> {
     /**
      * The canvas element used to render the scene.
      */
@@ -660,14 +660,14 @@ export interface LookAtParams {
      * @default `new GeoCoordinates(25, 0)` in [[MapView.constructor]] context.
      * @default [[MapView.target]] in [[MapView.lookAt]] context.
      */
-    target?: GeoCoordLike;
+    target: GeoCoordLike;
 
     /**
      * Zoomlevel of the MapView.
      * @default 5 in [[MapView.constructor]] context.
      * @default [[MapView.zoomLevel]] in [[MapView.lookAt]] context.
      */
-    zoomLevel?: number;
+    zoomLevel: number;
 
     /**
      * Tilt angle in degrees. 0 is top down view.
@@ -675,14 +675,14 @@ export interface LookAtParams {
      * @default [[MapView.tilt]] in [[MapView.lookAt]] context.
      * @note Maximum supported tilt is 89°
      */
-    tilt?: number;
+    tilt: number;
 
     /**
      * Heading angle in degrees and clockwise. 0 is north-up.
      * @default 0 in [[MapView.constructor]] context.
      * @default [[MapView.heading]] in [[MapView.lookAt]] context.
      */
-    heading?: number;
+    heading: number;
 }
 
 /**
@@ -2007,10 +2007,10 @@ export class MapView extends THREE.EventDispatcher {
      * Adjusts the camera to look at a given geo coordinate with tilt and heading angles.
      * @param params LookAtParams
      */
-    lookAt(params: LookAtParams): void;
+    lookAt(params: Partial<LookAtParams>): void;
 
     lookAt(
-        targetOrParams: GeoCoordLike | LookAtParams,
+        targetOrParams: GeoCoordLike | Partial<LookAtParams>,
         distance?: number,
         tiltDeg?: number,
         headingDeg?: number
@@ -2021,7 +2021,7 @@ export class MapView extends THREE.EventDispatcher {
                     ? MapViewUtils.calculateZoomLevelFromDistance(this, distance)
                     : undefined;
 
-            const params: LookAtParams = {
+            const params: Partial<LookAtParams> = {
                 target: targetOrParams,
                 zoomLevel,
                 tilt: tiltDeg,
@@ -2029,7 +2029,7 @@ export class MapView extends THREE.EventDispatcher {
             };
             this.lookAtImpl(params);
         } else if (typeof targetOrParams === "object") {
-            this.lookAtImpl(targetOrParams as LookAtParams);
+            this.lookAtImpl(targetOrParams as Partial<LookAtParams>);
         }
     }
 
@@ -2656,7 +2656,7 @@ export class MapView extends THREE.EventDispatcher {
         };
     }
 
-    private lookAtImpl(params: LookAtParams): void {
+    private lookAtImpl(params: Partial<LookAtParams>): void {
         const zoomLevel = THREE.MathUtils.clamp(
             getOptionValue(params.zoomLevel, this.zoomLevel),
             this.m_minZoomLevel,
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -652,22 +652,35 @@ const MapViewDefaults = {
  */
 export interface LookAtParams {
     /**
-     * Target/look at point of the MapView
+     * Target/look at point of the MapView.
+     * @note If the given point is not on the ground (altitude != 0) [[MapView]] will do a
+     * raycasting internally to find a target on the ground.
+     * As a consequence [[MapView.target]] and [[MapView.zoomLevel]] will not match the values
+     * that were passed into the [[MapView.lookAt]] method.
+     * @default `new GeoCoordinates(25, 0)` in [[MapView.constructor]] context.
+     * @default [[MapView.target]] in [[MapView.lookAt]] context.
      */
     target?: GeoCoordLike;
 
     /**
-     * Zoomlevel of the MapView. Overwrites [[distance]].
+     * Zoomlevel of the MapView.
+     * @default 5 in [[MapView.constructor]] context.
+     * @default [[MapView.zoomLevel]] in [[MapView.lookAt]] context.
      */
     zoomLevel?: number;
 
     /**
      * Tilt angle in degrees. 0 is top down view.
+     * @default 0 in [[MapView.constructor]] context.
+     * @default [[MapView.tilt]] in [[MapView.lookAt]] context.
+     * @note Maximum supported tilt is 89°
      */
     tilt?: number;
 
     /**
      * Heading angle in degrees and clockwise. 0 is north-up.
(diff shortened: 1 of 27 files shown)


ninok:
First commit is in a separate PR:
https://github.com/heremaps/harp.gl/pull/1365

ninok:

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1936,40 +1971,43 @@ export class MapView extends THREE.EventDispatcher {
...
+        if (isGeoCoordinatesLike(targetOrParams)) {
+            const zoomLevel =
+                distance !== undefined
+                    ? MapViewUtils.calculateZoomLevelFromDistance(this, distance)
+                    : undefined;
+
+            const params: LookAtParams = {
+                target: targetOrParams,
+                zoomLevel,
+                tilt: tiltDeg,
```
zbigg: Minor: doc says, default `tilt` and `heading` is `0`, but if they are not present, `undefined` here will be converted to "current" `tilt`/`heading` in `lookAtImpl`.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2518,6 +2556,94 @@ export class MapView extends THREE.EventDispatcher {
...
+        };
+    }
+
+    private lookAtImpl(params: LookAtParams): void {
+        const zoomLevel = THREE.MathUtils.clamp(
+            getOptionValue(params.zoomLevel, this.zoomLevel),
+            this.m_minZoomLevel,
+            this.m_maxZoomLevel
+        );
+        const tilt = Math.min(getOptionValue(params.tilt, this.tilt), PITCH_LIMIT_DEG);
```
zbigg: This will coerce custom pitch > 90 set by user manually to `PITCH_LIMIT_DEG` even if someone intended to change some other parameter like azimuth i. `lookAt({azimumuth: xxx})`.

Do we support camera with pitch > `PITCH_LIMIT_DEG` at all ? If not, then it's not documented.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2518,6 +2556,94 @@ export class MapView extends THREE.EventDispatcher {
...
+        const target = GeoCoordinates.fromObject(getOptionValue(params.target, this.target));
```
zbigg: Applies to `tilt` also.

We're calling `extractAttidute` 3 times for each call to `lookAt` (by always reading `this.tilt`, `this.heading` and `updateLookAtSettings`.

Can we
 1) cache `extractAttitude` output ?
 2) at least call read current values only if they are really needed: s/getOptionValue/good old ?: operator/

@here/harp-mapview/lib/MapView.ts
```diff
@@ -642,14 +629,42 @@ export const MapViewDefaults = {
...
 
+/**
+ * Parameters for MapView.lookAt
+ */
+export interface LookAtParams {
+    /**
+     * Target/look at point of the MapView.
+     * @note If the given point is not on the ground (altitude != 0) [[MapView]] will do a
+     * raycasting internally to find a target on the ground.
+     * As consequence [[MapView.target]] and [[MapView.zoomLevel]] will nit match the values
```
zbigg: - as a consequence 
- not match

@here/harp-mapview/lib/MapView.ts
```diff
@@ -642,14 +629,42 @@ export const MapViewDefaults = {
...
+     * that were passed into the [[MapView.lookAt]] method.
+     */
```
zbigg: Add notes on defaults to all params if used in context of `MapViewOptions` and when used as `lookAt`.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -642,14 +629,42 @@ export const MapViewDefaults = {
...
+
+    /**
+     * Zoomlevel of the MapView.
+     */
+    zoomLevel?: number;
+
+    /**
+     * Tilt angle in degrees. 0 is top down view.
+     */
+    tilt?: number;
```
zbigg: Note on max supported pitch/tilt ?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1454,6 +1467,16 @@ export class MapView extends THREE.EventDispatcher {
...
+     * @note When modifying the camera all derived properties like:
+     * - [[MapView.target]]
+     * - [[MapView.zoomLevel]]
+     * - [[MapView.tilt]]
+     * - [[MapView.heading]]
+     * could change.
+     * Some of the properties (like [[zoomLevel]]) are cached internaly
+     * and will only be updated in the next animation frame.
+     * FIXME: Unfortunatley THREE.js is not dispatching any events when camera properties change
+     * so we should have an API for enforcing update of cached values.
```
zbigg: I think that asks for api like `MapView.updateCamera` or `mapView.cameraChanged` that shall be called to sync internal cached params with camera when it was modified by user.

Having this, will also allow to cache `extractAttitude` result, so innocent people calls like `const { tilt, heading } = mapview` have expected complexity of just reading properties, not doing complex math twice.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1494,28 +1517,17 @@ export class MapView extends THREE.EventDispatcher {
...
-            target = this.worldTarget;
-        }
-        const targetCoordinates = this.projection.unprojectPoint(target);
-        const targetDistance = this.camera.position.distanceTo(target);
-        const attitude = MapViewUtils.extractAttitude(this, this.camera);
-        const pitchDeg = THREE.MathUtils.radToDeg(attitude.pitch);
-        const headingDeg = -THREE.MathUtils.radToDeg(attitude.yaw);
+        // Remember tilt and heading before setting the projection.
+        const tilt = this.tilt;
+        const heading = this.heading;
```
zbigg: ... like here

@here/harp-mapview/lib/MapView.ts
```diff
@@ -642,14 +629,42 @@ export const MapViewDefaults = {
...
+     * raycasting internally to find a target on the ground.
+     * As consequence [[MapView.target]] and [[MapView.zoomLevel]] will nit match the values
+     * that were passed into the [[MapView.lookAt]] method.
+     */
+    target?: GeoCoordLike;
+
+    /**
+     * Zoomlevel of the MapView.
+     */
+    zoomLevel?: number;
```
zbigg: I strongly propose that we support also `distance in meters` in this API.

1) we maintain some feature parity with old `lookAt` which supported `distance`
2) I think removing `distance` from API is wrong, as in _3D_ engine is much more natural for users than some abstract, concept like `zoomLevel` which means nothing to random users.

(Yes, distance in mercator at high latidutes is problematic, but that's another story)

This may be followup patch of course.

@here/harp-mapview/lib/MapView.ts
ninok: done

@here/harp-mapview/lib/MapView.ts
ninok: done

@here/harp-mapview/lib/MapView.ts
ninok: The `lookAt` API is mainly for users that are used to 2d engines like Tangram or Mapbox and these users don't know the concept of distance.
For 3D "power users" there are other ways, e.g. directly modifying the camera or `MapViewUtils.convertZoomLevelToDistance`

@here/harp-mapview/lib/MapView.ts
ninok: done

@here/harp-mapview/lib/MapView.ts
ninok: yes, this was my bigger plan. If you check https://github.com/heremaps/harp.gl/pull/1400 (sorry it is based on this PR and contains all these changes as well) or especially this commit:
https://github.com/heremaps/harp.gl/pull/1400/commits/ff2ae094af7aa6b3feaed6d6e72ae2bc754b0d82
I introduced a private `ViewState` in `MapView` and a method `MapView.updateViewState()`.

Things that I did not do yet are:
1. add tilt and heading to the `ViewState` so they are computed only once
2. Make `MapView.updateViewState` public

@here/harp-mapview/lib/MapView.ts
ninok: as above

@here/harp-mapview/lib/MapView.ts
ninok: done

@here/harp-mapview/lib/MapView.ts
ninok: As mentioned above: In a follow up change I will cache titl and heading (or yaw, pitch , roll) in the `ViewState`

@here/harp-mapview/lib/MapView.ts
ninok: You could only get into `pitch > PITCH_LIMIT_DEG` situation if you are not using the `lookAt` API but modify the camera directly.
So in general we support it but we don't support mixing of direct camera modifications and `MapView.lookAt`. I think this is a valid restriction.

ninok: Or let's put it differently: If you modify the camera directly there are no constraints, but when you use the lookAt API again the constraints for that API will be enforced.

@here/harp-mapview/lib/MapView.ts
zbigg: Hmm, i think there is potential misunderstanding.

I meant support both `distance` and `zoomLevel`, so we just use `MapViewUtils.convertZoomLevelToDistance` internally in `lookAt`

Anyway, that just means that many of our internal tests have to be rewritten to use zoom level, when they were originally created with distance and that's just more work in order to merge this PR..

ninok: Yeah, I could add it as well but I think it should be deprecated b/c it's hard to understand. It is not the actual distance of the camera to the target point but:
`the distance of the camera to the target point but clamped to min and max zoomlevel ?!?!`.
Also adding this as deprectated in the new API makes not much sense b/c you could also just use the old API.
Since there is an IBC test failing already I will look into switching them all to zoomLevel b/c this is also what you want to express. We have tests that want to test level 1 but instead of setting zoomLevel to 1 we set it to some weird distance.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -190,6 +190,7 @@ export namespace MapViewUtils {
             ),
             0
         );
+        // tslint:disable-next-line: deprecation
```
zbigg: You can use new API here.

PR #1395, opened 2020-03-18
MINOR: Refactor ClipPlanesEvaluator(s), remove MapView dependency.

kkostecki:
This change is required in order to provide MapViewAtmosphere support
without MapView dependency (see PR: #1378)
Cleanups the code and changes interfaces to not use MapView reference.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
+++ b/@here/harp-mapview/lib/ClipPlanesEvaluator.ts
@@ -8,7 +8,8 @@ import { ViewRanges } from "@here/harp-datasource-protocol/lib/ViewRanges";
 import { EarthConstants, Projection, ProjectionType } from "@here/harp-geoutils";
 import { assert } from "@here/harp-utils";
 import * as THREE from "three";
-import { MapView } from "./MapView";
+import { ElevationProvider } from "./ElevationProvider";
+import { MapViewUtils } from "./Utils";
 
 const epsilon = 0.000001;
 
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -873,9 +884,22 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
...
+        // in a same way the MapView component does.
+        // TODO: Target distance does not need to be cached when we provide full support for
+        // target point within all camera tilt angles.
+        const target = MapViewUtils.getWorldTargetFromCamera(camera, projection);
+        if (target !== null) {
+            this.m_targetDistance = camera.position.distanceTo(target);
+        }
+        // Fallback to camera altitude if target not found and not yet set.
+        // NOTE: May be inaccurate when launching map with camera looking to void.
+        else if (this.m_targetDistance < cameraAltitude) {
```
ninok: If `this.m_targetDistance` is  undefined this will always be false

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
```diff
@@ -873,9 +884,22 @@ export class TiltViewClipPlanesEvaluator extends TopViewClipPlanesEvaluator {
...
+            this.m_targetDistance = cameraAltitude;
+        }
         // Apply the constraints.
         const farMin = cameraAltitude - this.minElevation;
-        const farMax = mapView.targetDistance * this.farMaxRatio;
+        const farMax = this.m_targetDistance * this.farMaxRatio;
```
ninok: Can you please remove `targetDistance` from the `MapView` API before people start using it

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
kkostecki: ClipPlanesEvaluator.m_targetDistance is always defined, but may be equal zero, then (unfortunately) we need some fallback, here it is camera distance to ground, I could implement better fallback here, but I guess it is worthless because it is MapViewUtils.getWorldTargetFromCamera() which needs to be improved - it should always return some point - with fallback if required.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
ninok: I see, so leave it for now. I will do it when making ViewState a member of MapView instead of a proxy like it is right now.

@here/harp-mapview/lib/ClipPlanesEvaluator.ts
ninok: Okay, either `MapViewUtils.getWorldTargetFromCamera()` should always return sth. or we should always have a valid `targetDistance` in the `ViewState`.
Since the goal is to reduce MapViewUtils and maybe eventually remove it I would go for the latter, but as mentioned in the comment below I will do this later.

ninok:
btw. please also add this API change to the release notes draft:
https://github.com/heremaps/harp.gl/releases/tag/untagged-7c9c9f49cf72dfa42262

kkostecki:
Retest this please

kkostecki:
rebuild this please

kkostecki:
> btw. please also add this API change to the release notes draft:
> https://github.com/heremaps/harp.gl/releases/tag/untagged-7c9c9f49cf72dfa42262

Release notes updated.

PR #1397, opened 2020-03-19
HARP-9460: Fix Synchronous example.

atomicsulfate:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-examples/src/rendering_synchronous.ts
+++ b/@here/harp-examples/src/rendering_synchronous.ts
@@ -164,20 +164,20 @@ export namespace SynchronousRendering {
     const mapView = initializeMapView("mapCanvas");
 
     // snippet:harp_gl_rendering_synchronous_2.ts
-    mapView.addEventListener(MapViewEventNames.Update, update);
+    mapView.addEventListener(MapViewEventNames.Update, requestDraw);
     // end:harp_gl_rendering_synchronous_2.ts
 
     // snippet:harp_gl_rendering_synchronous_3.ts
-    let updatePending = false;
+    let requestDrawPending = false;
     let drawing = false;
 
     // Requests a redraw of the scene.
-    function update() {
+    function requestDraw() {
         // Cancel request for redrawing if already pending
-        if (updatePending) {
+        if (requestDrawPending) {
             return;
         }
-        updatePending = true;
+        requestDrawPending = true;
 
         requestAnimationFrame(draw);
     }
@@ -187,7 +187,7 @@ export namespace SynchronousRendering {
         if (drawing) {
             return;
         }
-        updatePending = false;
+        requestDrawPending = false;
         drawing = true;
 
         // Draw popup's connection line
--- a/@here/harp-examples/src/rendering_synchronous.ts
+++ b/@here/harp-examples/src/rendering_synchronous.ts
@@ -203,8 +203,8 @@ export namespace SynchronousRendering {
 
         // Draw map scene
         mapView.renderSync();
-
         drawing = false;
+        mapView.update();
     }
     // end:harp_gl_rendering_synchronous_3.ts
 }


@here/harp-examples/src/rendering_synchronous.ts
```diff
@@ -203,8 +203,8 @@ export namespace SynchronousRendering {
 
         // Draw map scene
         mapView.renderSync();
-
         drawing = false;
+        mapView.update();
```
nzjony: shouldn't this be done before the `renderSync` ?

Also, this won't "Requests a redraw of the scene." as the documentation says, so I guess you are doing this just to fire the Update event?

atomicsulfate: > shouldn't this be done before the `renderSync` ?
why requesting an update before even drawing the current frame?

> Also, this won't "Requests a redraw of the scene." as the documentation says, so I guess you are doing this just to fire the Update event?
Why not? Calling `mapView.update()` triggers MapView's UPDATE event, which executes the update() function installed as callback by the example. The `update()` function request a redraw with `requestAnimationFrame(draw);`.

nzjony: I don't know why you need to update, but conceptually, it makes sense to update something before rendering it, otherwise you render the old state of the object.

`update` calls `drawFrame`, which includes:

```
if (this.m_drawing || this.m_options.synchronousRendering) {
            return;
        }
```

I.e. it won't redraw.

nzjony: **Update**, I read the ticket, now I understand why you need the update call.

atomicsulfate:

@here/harp-examples/src/rendering_synchronous.ts
atomicsulfate: > I don't know why you need to update, but conceptually, it makes sense to update something before rendering it, otherwise you render the old state of the object.

The state of the scene is updated in the example's `draw()` function before calling renderSync:
`popup.drawConnectionLine();

        state.yawDeg += 0.1;
        // Set target and camera rotation
        const distance = MapViewUtils.calculateDistanceFromZoomLevel(mapView, state.zoomLevel);
        const tilt = state.pitchDeg;
        const heading = -state.yawDeg;

        mapView.lookAt(state.geoPos, distance, tilt, heading);`

`mapView.update()` is not called to update the state of any object , but to request a new frame.

 
> `update` calls `drawFrame`, which includes:
> 
> ```
> if (this.m_drawing || this.m_options.synchronousRendering) {
>             return;
>         }
> ```
> 
> I.e. it won't redraw.

`update` does also this : 
` update() {
        this.dispatchEvent(UPDATE);
      ....
        this.drawFrame();
    }`

It dispatches the UPDATE event, which will call the update() function installed as callback in the example, which in turn calls requestAnimationFrame(draw), and draw() will call renderSync().

nzjony: Now I see my confusion, I missed the link from the mapview update -> update -> draw function, thanks for explaining.

Why not just call `update` instead of `mapView.update`?

atomicsulfate: > Why not just call `update` instead of `mapView.update`?

`mapView.update()` dispatches the UPDATE event to all registered listeners. If we call `update()` directly any other potential listener of that event won't be called.

> And rename it from `update` to something more meaningful to help prevent others being confused like me.

Please, do tell me what name would you like for that function, I'm not sure what you do you expect.

nzjony: > mapView.update() dispatches the UPDATE event to all registered listeners.

What other listeners do you need to worry about, even the `mapView.addEventListener(MapViewEventNames.Update, update);` was probably a left over from the fact that ` mapView.setCameraGeolocationAndZoom(),` was calling `update` internally.

If there are any other internal updates you need to handle, it should be documented, this is a public example, and it should be as straight forward as possible.

The documentation says "Requests a redraw of the scene.", something like `requestRedraw` or equivalent. I understand this isn't your code, but it should be improved however.

atomicsulfate: > What other listeners do you need to worry about

Any potential listener, either internal in our code or installed by client code. MapView events are part of our API, clients may install any listener they like. Therefore we cannot make any assumptions on whether listeners are installed or not for a particular event.

> even the `mapView.addEventListener(MapViewEventNames.Update, update);` was probably a left over from the fact that ` mapView.setCameraGeolocationAndZoom(),` was calling `update` internally. (as you found out in your investigation).

It's actually necessary to add a listener for the Update event if you want the synchronous rendering to work. There's many places in our code where we call `mapView.update()`. If you don't add the update listener you won't redraw after a tile has been loaded for example.

> If there are any other internal updates you need to handle, it should be documented, this is a public example, and it should be as straight forward as possible.

No, you don't want to document all listeners that you may install for a particular event. Any internal listener is implementation detail and therefore shouldn't be the concern of client code. 
The code of an example should be as straightforward as possible yes, but the change you suggest breaks MapView's update events. Listeners will be called for update events triggered internally (e.g. tile reload), but not by update events coming from client code.

> The documentation says "Requests a redraw of the scene.", something like `requestRedraw` or equivalent. I understand this isn't your code, but it should be improved however.

Done.

nzjony: For the general purpose where we render only when necessary, yes, such a listener is necessary, but in this example, we render as much as possible, so any internal update that causes a new frame will anyway happen because of the `requestAnimationFrame` from the `requestFrame` function.

But this is a minor and I will approve.

atomicsulfate:

PR #1399, opened 2020-03-19
MINOR: MapUtils provides fallback mechanism for finding camera target.

kkostecki:
Camera target will be always found and laying on the ground.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2604,6 +2604,8 @@ export class MapView extends THREE.EventDispatcher {
             this.m_targetWorldPos.copy(target);
             this.m_targetGeoPos = this.projection.unprojectPoint(target);
             this.m_targetDistance = this.camera.position.distanceTo(target);
+        } else {
+            console.warn("Target not found");
         }
     }
 
--- a/@here/harp-mapview/lib/Utils.ts
+++ b/@here/harp-mapview/lib/Utils.ts
@@ -46,7 +46,7 @@ const tangentSpace = {
 };
 const cache = {
     quaternions: [new THREE.Quaternion(), new THREE.Quaternion()],
-    vector3: [new THREE.Vector3(), new THREE.Vector3()],
+    vector3: [new THREE.Vector3(), new THREE.Vector3(), new THREE.Vector3(), new THREE.Vector3()],
     matrix4: [new THREE.Matrix4(), new THREE.Matrix4()],
     transforms: [
         {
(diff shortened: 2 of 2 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2604,6 +2604,8 @@ export class MapView extends THREE.EventDispatcher {
             this.m_targetWorldPos.copy(target);
             this.m_targetGeoPos = this.projection.unprojectPoint(target);
             this.m_targetDistance = this.camera.position.distanceTo(target);
+        } else {
+            console.warn("Target not found");
```
ninok: When could this happen?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -239,9 +239,143 @@ export namespace MapViewUtils {
             groundPlane.constant = 0;
             groundSphere.radius = EarthConstants.EQUATORIAL_RADIUS;
         }
+        if (result !== null) {
```
ninok: Please change the signature (i.e. return value type) of the function. It should never return null now

@here/harp-mapview/lib/Utils.ts
```diff
@@ -239,9 +239,143 @@ export namespace MapViewUtils {
...
+            return result;
+        }
+        return getGazingTargetFromCamera(camera, projection, elevation);
+    }
+
+    function getGazingTargetFromCamera(
```
ninok: I still don't see why we need to spend so much effort for the target point. I  agree that we should do sth. like that for the `MapControls` but why for the target point?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -239,9 +239,143 @@ export namespace MapViewUtils {
...
+        elevation?: number
+    ): THREE.Vector3 | null {
+        const cameraPos = cache.vector3[0].copy(camera.position);
+        const cameraLookAt = camera.getWorldDirection(cache.vector3[1]).normalize();
+        let result: THREE.Vector3 | null = null;
+        // Find gazing point on the ground
+        if (projection.type === ProjectionType.Planar) {
+            // Tolerance may be adjusted to not exactly hit the horizon line.
+            //const planarTolerance = THREE.MathUtils.degToRad(1.0);
+            if (camera instanceof THREE.PerspectiveCamera) {
```
ninok: Please remove and expect `PerspecitveCamera` as parameter.
It looks like we are supporting other cameras but we dont't.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -239,9 +239,143 @@ export namespace MapViewUtils {
             groundPlane.constant = 0;
             groundSphere.radius = EarthConstants.EQUATORIAL_RADIUS;
         }
+        if (result !== null) {
+            return result;
+        }
+        return getGazingTargetFromCamera(camera, projection, elevation);
```
ninok: You should also do this fallback if the tilt is very high (i.e. >89°). Otherwise the raycasted point might be really far away.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -239,9 +239,143 @@ export namespace MapViewUtils {
...
+                const rotQuaternion = cache.quaternions[0].setFromAxisAngle(
+                    rotPlaneNorm,
+                    tangentAngle - sphericalTolerance
+                );
+                // Finally find the gazing (tangent based) look at vector.
+                const newLookAt = cameraToOrigin.applyQuaternion(rotQuaternion);
+                rayCaster.set(cameraPos, newLookAt);
+                result = rayCaster.ray.intersectSphere(groundSphere, new THREE.Vector3());
+            } else {
+                assert(false, "Orthogonal camera is not yet supported");
```
ninok: And most likely will never be. Please remove the if and change the input type.

@here/harp-mapview/lib/MapView.ts
kkostecki: This is just for debug purposes... Should be removed in the final commit, I was checking with this if everything is fine with the target calculation code.

@here/harp-mapview/lib/Utils.ts
kkostecki: As I mentioned above, just for debugging, normally I could put assert here, but wanted to test drive the stuff - of course not to merge yet.

@here/harp-mapview/lib/Utils.ts
kkostecki: The assumption here (and I guess easiest one) it that focus point based rotation should be based on the same target point as calculated in MapView. Of course we could introduce second 
_focus_ point just for MapControls, but I think it makes stuff over-complicated what do you think?

@here/harp-mapview/lib/Utils.ts
kkostecki: Hehe, so you don't believe in Ortho mode :).

@here/harp-mapview/lib/Utils.ts
kkostecki: True, that might be also good idea!

@here/harp-mapview/lib/Utils.ts
kkostecki: Hehhe... OK!

PR #1400, opened 2020-03-19
HARP-7835: Private MapViewState

ninok:
Based on https://github.com/heremaps/harp.gl/pull/1392


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2718,15 +2722,25 @@ export class MapView extends THREE.EventDispatcher {
         this.m_screenCollisions.update(width, height);
 
         this.m_pixelToWorld = undefined;
-        this.m_fog.update(this, this.m_viewRanges.maximum);
+        this.m_fog.update(this, this.m_viewState.viewRanges.maximum);
 
-        this.updateLookAtSettings();
+        this.updateViewState();
```
ninok: This is still a bit messy. We update the `ViewState` partially before we call `updateViewState`. Eventually everything should be updated in `updateViewState`

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2914,8 +2927,7 @@ export class MapView extends THREE.EventDispatcher {
                 // also the same. The position is then calculated based on the light direction and
                 // the height
                 // using basic trigonometry.
-                const tilt = this.m_pitch;
-                const cameraHeight = this.targetDistance * Math.cos(tilt);
+                const cameraHeight = this.targetDistance * Math.cos(this.m_viewState.pitch);
```
MPursche: Shouldn't this be computed with `projection.getGroundDistance(m_viewState.worldCenter)` instead? Using trigonometry only works for mercator projection.

ninok: true. Not sure what was the original intent of this code.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2704,7 +2708,7 @@ export class MapView extends THREE.EventDispatcher {
 
         // Make sure to update all properties that are accessable via API (e.g. zoomlevel) b/c
         // otherwise they would be updated as recently as in the next animation frame.
-        this.updateLookAtSettings();
+        this.updateViewState();
```
MPursche: When `mapView.lookAt()` is called target, zoom level, pitch and heading are used to compute targetDistance, camera position and camera attitude. But `mapView.updateViewState()` in turn will compute target, targetDistance and zoom level again from the camera position and attitude. So the value of `mapView.target` and `mapView.zoomLevel` can be slightly different than the value set with `mapView.lookAt({ target, zoomLevel })`. I understand that this is necessary if the camera position and attitude was directly modified from outside of MapView but it seems confusing if a user only uses the `mapView.lookAt()` API to change camera position. Maybe it would be better if users that directly manipulate `mapView.camera` need to explicitly call `mapView.updateViewState()`(or something similar) to update target and zoom level from the camera position.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -746,20 +764,10 @@ export class MapView extends THREE.EventDispatcher {
      */
     private readonly m_rteCamera = new THREE.PerspectiveCamera();
 
-    private m_yaw = 0;
-    private m_pitch = 0;
-    private m_roll = 0;
     private m_focalLength = 0;
-    private m_targetDistance = 0;
     private m_targetGeoPos = GeoCoordinates.fromObject(MapViewDefaults.target!);
```
MPursche: Should `m_targetGeoPos` and/or `m_targetWorldPos` be part of the `ViewState`? They are also related to the position and attitude of the camera.

PR #1402, opened 2020-03-20
HARP-9275: Implement VertexCache for mesh on the fly transformations.

atomicsulfate:
It'll be used for mesh displacements on picking. Terrain overlay is done
by displacing meshes in the vertex shader. To pick those meshes, they
are displaced in CPU. To avoid allocating the whole position buffer again,
intersection test is done on the fly for every triangle after it's displaced.

To avoid recomputations of the same vertex, a small LRU cache is used.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/geometry/VertexCache.ts
+++ b/@here/harp-mapview/lib/geometry/VertexCache.ts
@@ -34,7 +34,7 @@ export class VertexCache {
      * Creates a new cache with the specified maximum size.
      * @param maxVertexCount The maximum number of vertices the cache will store.
      */
-    constructor(public maxVertexCount: number) {
+    constructor(readonly maxVertexCount: number) {
         this.m_cache.length = this.maxVertexCount * Field.Count;
         this.clear();
     }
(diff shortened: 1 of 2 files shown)


atomicsulfate:
@harpgl-bot  retest this please

@here/harp-mapview/lib/geometry/VertexCache.ts
```diff
@@ -0,0 +1,146 @@
...
+    private m_cache: number[] = []; // Stores all fields for every cached vertex (see Field).
+    private m_vertexCount: number = 0;
+    private m_oldestIdx: number = 0;
+    private m_newestIdx: number = 0;
+
+    /**
+     * Creates a new cache with the specified maximum size.
+     * @param maxVertexCount The maximum number of vertices the cache will store.
+     */
+    constructor(public maxVertexCount: number) {
```
MPursche: Minor: Make maxVertexCount readonly(if cache does not need to be resized) or implement it with a property so that `m_cache` is also resized when `maxVertexCount` is changed. The array should be automatically resized when data is inserted out of bounds but that may be slightly slower. It will also not actually get smaller if the cache size is decreased.

atomicsulfate: I actually didn't plan to support variable size, so yes I'll make it readonly.

PR #1409, opened 2020-03-24
HARP-6487: Provide h/vPlacement theme attributes for point/line labels.

kkostecki:
Introduce new Theme style attributes for 'labeled-icon' technique:
- hPlacement : Left | Center | Right,
- vPlacement : Above | Center | Below.
The first one works in opposite way to hAlignment, while the second
has the same meaning. Both attributes takes precedence over the
h/vAlignment if set, otherwise alignment attributes may specify
text placement over the point anchor, although this attributes are
from now marked as deprecated.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-datasource-protocol/lib/TechniqueParams.ts
+++ b/@here/harp-datasource-protocol/lib/TechniqueParams.ts
@@ -691,12 +691,26 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
     wrappingMode?: DynamicProperty<"None" | "Character" | "Word">;
     /**
      * Text position regarding the baseline.
+     *
+     * @note Alternative to [[hPlacement]] attribute, which defines label text placement in exactly
+     * opposite way, text that were `Left` aligned are by deduction `Right` placed relative to
+     * the label anchor.
      */
     hAlignment?: DynamicProperty<"Left" | "Center" | "Right">;
     /**
      * Text position inside a line.
+     *
+     * @note Same as [[vPlacement]] attribute although the [[vPlacement]] takes precedence.
      */
     vAlignment?: DynamicProperty<"Above" | "Center" | "Below">;
+    /**
+     * Text horizontal position relating to the label's anchor point (icon, image).
+     */
+    hPlacement?: DynamicProperty<"Left" | "Center" | "Right">;
+    /**
+     * Text vertical position relating to the label anchor.
+     */
+    vPlacement?: DynamicProperty<"Top" | "Center" | "Bottom">;
 }
 
 export interface LineTechniqueParams extends BaseTechniqueParams {
--- a/@here/harp-datasource-protocol/lib/Theme.ts
+++ b/@here/harp-datasource-protocol/lib/Theme.ts
@@ -1001,6 +1001,8 @@ export interface TextStyleDefinition {
     wrappingMode?: "None" | "Character" | "Word";
     hAlignment?: "Left" | "Center" | "Right";
     vAlignment?: "Above" | "Center" | "Below";
+    hPlacement?: "Left" | "Center" | "Right";
+    vPlacement?: "Top" | "Center" | "Bottom";
 }
 
 /**
--- a/@here/harp-mapview/lib/text/TextStyleCache.ts
+++ b/@here/harp-mapview/lib/text/TextStyleCache.ts
@@ -20,12 +20,14 @@ import {
     FontUnit,
     FontVariant,
     HorizontalAlignment,
+    HorizontalPlacement,
     TextCanvas,
     TextLayoutParameters,
     TextLayoutStyle,
     TextRenderParameters,
     TextRenderStyle,
     VerticalAlignment,
+    VerticalPlacement,
     WrappingMode
 } from "@here/harp-text-canvas";
 import { getOptionValue, LoggerManager } from "@here/harp-utils";
@@ -317,7 +319,7 @@ export class TextStyleCache {
     }
 
     /**
-     * Create the appropriate [[TextRenderStyle]] to use for a label. Depends heavily on the label's
+     * Create the appropriate [[TextLayoutStyle]] to use for a label. Depends heavily on the label's
      * [[Technique]] and the current zoomLevel.
      *
      * @param tile The [[Tile]] to process.
@@ -334,25 +336,35 @@ export class TextStyleCache {
 
         const defaultLayoutParams = this.m_defaultStyle.layoutParams;
 
-        const hAlignment = getPropertyValue(technique.hAlignment, discreteZoomEnv) as
+        // Available only for PoiTechnique and LineMarkerTechnique, for TextTechnique
+        const hPlacement = getPropertyValue((technique as any).hPlacement, discreteZoomEnv) as
             | string
-            | undefined;
-        const vAlignment = getPropertyValue(technique.vAlignment, discreteZoomEnv) as
+            | null;
+        const vPlacement = getPropertyValue((technique as any).vPlacement, discreteZoomEnv) as
             | string
-            | undefined;
+            | null;
+
+        // Disable depreciation for PoiTechnique and LineMarkerTechnique
+        // tslint:disable-next-line deprecation
+        const hAlignment = getPropertyValue(technique.hAlignment, discreteZoomEnv) as string | null;
+        // tslint:disable-next-line deprecation
+        const vAlignment = getPropertyValue(technique.vAlignment, discreteZoomEnv) as string | null;
+
+        const { horizontalAlignment, verticalAlignment } = this.resolvePlacementAndAlignment(
+            hAlignment !== null ? hAlignment : undefined,
+            vAlignment !== null ? vAlignment : undefined,
+            hPlacement !== null ? hPlacement : undefined,
+            vPlacement !== null ? vPlacement : undefined
+        );
+
         const wrapping = getPropertyValue(technique.wrappingMode, discreteZoomEnv) as
             | string
             | undefined;
 
-        const horizontalAlignment: HorizontalAlignment | undefined =
-            hAlignment === "Left" || hAlignment === "Center" || hAlignment === "Right"
-                ? HorizontalAlignment[hAlignment]
-                : defaultLayoutParams.horizontalAlignment;
-
-        const verticalAlignment: VerticalAlignment | undefined =
-            vAlignment === "Above" || vAlignment === "Center" || vAlignment === "Below"
-                ? VerticalAlignment[vAlignment]
-                : defaultLayoutParams.verticalAlignment;
+        const wrappingMode =
+            wrapping === "None" || wrapping === "Character" || wrapping === "Word"
+                ? WrappingMode[wrapping]
+                : defaultLayoutParams.wrappingMode;
 
         const layoutParams = {
             tracking:
@@ -372,10 +384,7 @@ export class TextStyleCache {
             lineRotation:
                 getPropertyValue(technique.lineRotation, discreteZoomEnv) ??
                 defaultLayoutParams.lineRotation,
-            wrappingMode:
-                wrapping === "None" || wrapping === "Character" || wrapping === "Word"
-                    ? WrappingMode[wrapping]
-                    : defaultLayoutParams.wrappingMode,
+            wrappingMode,
             horizontalAlignment,
             verticalAlignment
         };
@@ -393,6 +402,12 @@ export class TextStyleCache {
         style: TextStyleDefinition,
         styleName: string
     ): TextElementStyle {
+        const { horizontalAlignment, verticalAlignment } = this.resolvePlacementAndAlignment(
+            style.hAlignment,
+            style.vAlignment,
+            style.hPlacement,
+            style.vPlacement
+        );
         return {
             name: styleName,
             fontCatalog: getOptionValue(style.fontCatalogName, this.m_defaultStyle.fontCatalog),
@@ -441,19 +456,48 @@ export class TextStyleCache {
                     style.wrappingMode === "Word"
                         ? WrappingMode[style.wrappingMode]
                         : WrappingMode.Word,
-                verticalAlignment:
-                    style.vAlignment === "Above" ||
-                    style.vAlignment === "Center" ||
-                    style.vAlignment === "Below"
-                        ? VerticalAlignment[style.vAlignment]
-                        : VerticalAlignment.Center,
-                horizontalAlignment:
-                    style.hAlignment === "Left" ||
-                    style.hAlignment === "Center" ||
-                    style.hAlignment === "Right"
-                        ? HorizontalAlignment[style.hAlignment]
-                        : HorizontalAlignment.Center
+                verticalAlignment,
+                horizontalAlignment
             }
         };
+    }
+
+    private resolvePlacementAndAlignment(
+        hAlignment?: string,
+        vAlignment?: string,
+        hPlacement?: string,
+        vPlacement?: string
+    ): {
+        horizontalAlignment: HorizontalAlignment;
+        verticalAlignment: VerticalAlignment;
+    } {
+        // Parse placement properties if available at they have higher priority in case of clash.
+        const horizontalPlacement =
+            hPlacement === "Left" || hPlacement === "Center" || hPlacement === "Right"
+                ? HorizontalPlacement[hPlacement]
+                : undefined;
+        const verticalPlacement =
+            vPlacement === "Above" || vPlacement === "Center" || vPlacement === "Below"
+                ? VerticalPlacement[vPlacement]
+                : undefined;
+
+        // Override alignment attributes with placement (if defined), otherwise parse alignment or
+        // assign default values.
+        const horizontalAlignment =
+            horizontalPlacement !== undefined
+                ? ((horizontalPlacement as unknown) as HorizontalAlignment)
+                : // Use alignment or default.
+                hAlignment === "Left" || hAlignment === "Center" || hAlignment === "Right"
+                ? HorizontalAlignment[hAlignment]
+                : defaultTextLayoutStyle.horizontalAlignment;
+
+        const verticalAlignment =
+            verticalPlacement !== undefined
+                ? ((verticalPlacement as unknown) as VerticalAlignment)
+                : vAlignment === "Above" || vAlignment === "Center" || vAlignment === "Below"
+                ? VerticalAlignment[vAlignment]
+                : defaultTextLayoutStyle.verticalAlignment;
+
+        return { horizontalAlignment, verticalAlignment };
     }
 }
--- a/@here/harp-text-canvas/lib/rendering/TextStyle.ts
+++ b/@here/harp-text-canvas/lib/rendering/TextStyle.ts
@@ -55,11 +55,33 @@ export enum VerticalAlignment {
 
 /**
  * Horizontal alignment to be used when placing text.
+ *
+ * @note Horizontal alignment value is exactly opposite to [[HorizontalPlacement]] value,
+ * cause when you place text on the right side of point (or icon) you normally want to have it
+ * left-aligned.
  */
 export enum HorizontalAlignment {
     Left = 0.0,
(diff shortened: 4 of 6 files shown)


kkostecki:
Provides minimal set of changes to support hPlacement / vPlacement attributes in label (point/line) techniques. Placement attributes will always override hAlignment / vAlignment (if available in theme style) such as internally we use only alignment as before.
More sophisticated approach would include support for both alignment and placement values, thus you could still place text on the right and align it to right (currently always opposite). PR that prepares engine for such support, which allows to extend functionality easily (4 line of code), but still solving original issue (HARP-6487) is presented here: #1409.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -691,12 +691,32 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
...
+     *
+     * @deprecated Use [[hPlacement]] attribute, which defines label text placement in exactly
+     * opposite way, text that were `Left` aligned are by deduction `Right` placed relative to
+     * the label icon.
+     */
+    hPlacement?: DynamicProperty<"Left" | "Center" | "Right">;
+    /**
+     * Text position inside a line.
+     *
+     * @deprecated Use [[vPlacement]] attribute instead.
```
ninok: users will enter an infinite loop in their brain

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -691,12 +691,32 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
...
      */
     vAlignment?: DynamicProperty<"Above" | "Center" | "Below">;
+    /**
+     * Text position regarding the baseline.
+     *
+     * @deprecated Use [[hPlacement]] attribute, which defines label text placement in exactly
+     * opposite way, text that were `Left` aligned are by deduction `Right` placed relative to
+     * the label icon.
+     */
+    hPlacement?: DynamicProperty<"Left" | "Center" | "Right">;
```
ninok: Not sure how harp-native is doing this but if we want to support multiple placements in the future this would become an array and it would be independent from `vPlacement`. Might be better to combine both placements into one property like
```
placement: {
    horizontal: "Left" | "Center" | "Right",
    vertical: "Top" | "Center" | "Bottom"}
```

ninok: not sure how well this works with `DynamicProperties` though.

@here/harp-mapview/lib/text/TextStyleCache.ts
```diff
@@ -334,25 +336,37 @@ export class TextStyleCache {
 
         const defaultLayoutParams = this.m_defaultStyle.layoutParams;
 
+        // Available only for PoiTechnique and LineMarkerTechnique, for TextTechnique
+        const hPlacement = getPropertyValue((technique as any).hPlacement, discreteZoomEnv) as
+            | string
+            | undefined;
+        const vPlacement = getPropertyValue((technique as any).vPlacement, discreteZoomEnv) as
+            | string
+            | undefined;
```
zbigg: Old code was wrong. Current `getPropertyValue` never returns `undefined`. If it's input was `undefined` it will return `null`. :/
Sorry, #1241 which corrected this typing got rejected.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -691,12 +691,32 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
...
+    /**
+     * Text position inside a line.
+     *
+     * @deprecated Use [[vPlacement]] attribute instead.
+     */
+    vPlacement?: DynamicProperty<"Top" | "Center" | "Bottom">;
```
zbigg: Top->Above

@here/harp-text-canvas/lib/rendering/TextStyle.ts
```diff
@@ -55,13 +55,38 @@ export enum VerticalAlignment {
 
 /**
  * Horizontal alignment to be used when placing text.
+ *
+ * @note Horizontal alignment value is exactly opposite to [[HorizontalPlacement]] value,
+ * cause when you place text on the right side of point (or icon) you normally want to have it
+ * left-aligned.
+ * @note Alignment attributes `hAlignment` and `vAlignment` are now depreciated in Theme
```
zbigg: Remove this note.
This module has no connection to Themes and technique.

zbigg: ... After reading whole patch i think whole change to TextStyle can be removed as you've already resolve "placement" to "alignment" in higher layers.

@here/harp-text-canvas/lib/rendering/TextStyle.ts
```diff
@@ -447,6 +472,28 @@ export class TextLayoutStyle {
...
+    get verticalPlacement(): VerticalPlacement {
+        return (this.m_params.verticalAlignment! as unknown) as VerticalPlacement;
+    }
+    set verticalPlacement(value: VerticalPlacement) {
+        // Update vertical alignment, both values works interchangeably.
+        this.m_params.verticalAlignment = (value as unknown) as VerticalAlignment;
+    }
+
+    /**
+     * Text element vertical position relating to the point of placement.
```
zbigg: And if you consider not removing this, add note that `*placement` overweites `*alignment`

@here/harp-datasource-protocol/lib/TechniqueParams.ts
zbigg: @ninok 
In case of dynamic properties, we cannot use "plain old array", but we need to use [ "array", ... ] expression and we're fine. It will be just a little more verbose.
I.e
```
"placement": ["array", { a: 1, b: 2 }, { a: 2, c: 2) ...]
```
instead of intuitive:
```
placement: [{ a: 1, b: 2 }, { a: 2, c: 2)]
```

But, i think this patch is far from having support for array of placement options.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -691,12 +691,32 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
...
     /**
      * Text position inside a line.
+     *
+     * @deprecated Use [[vPlacement]] attribute instead.
      */
     vAlignment?: DynamicProperty<"Above" | "Center" | "Below">;
+    /**
+     * Text position regarding the baseline.
+     *
+     * @deprecated Use [[hPlacement]] attribute, which defines label text placement in exactly
```
atomicsulfate: why do you mark this as deprecated?

@here/harp-datasource-protocol/lib/TechniqueParams.ts
atomicsulfate: As we've discussed by webex, one step further to make text styling the same as in native is to unify horizontal and vertical placement into one attribute.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -691,12 +691,32 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
     wrappingMode?: DynamicProperty<"None" | "Character" | "Word">;
     /**
      * Text position regarding the baseline.
+     *
+     * @deprecated Use [[hPlacement]] attribute, which defines label text placement in exactly
+     * opposite way, text that were `Left` aligned are by deduction `Right` placed relative to
```
atomicsulfate: text that **was**

@here/harp-datasource-protocol/lib/TechniqueParams.ts
ninok: okay, cool.

> But, i think this patch is far from having support for array of placement options.

I know but I just want to make sure that the style will not change again when we introduce array of placements. So I propose that the placement is a literal now, i.e.:
```
placement: ["literal", { v: "Center", h: "Center" }]
```
and in the future we could extend it to support both, literals and arrays as you proposed

@here/harp-datasource-protocol/lib/TechniqueParams.ts
ninok: Totally agree. Also see the discussion above about supporting arrays of placements in the future.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -691,12 +691,32 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
...
+     * the label icon.
      */
     hAlignment?: DynamicProperty<"Left" | "Center" | "Right">;
     /**
      * Text position inside a line.
+     *
+     * @deprecated Use [[vPlacement]] attribute instead.
```
ninok: I think we should not deprecate it. Ultimately we could support combinations of placement and alignment (or to go really crazy arrays of placements and alignments).
Not sure if anybody needs it though

@here/harp-datasource-protocol/lib/TechniqueParams.ts
atomicsulfate: Better call it hTextPlacement, otherwise you have to read the documentation to know if it refers to the label as whole, the icon or the text.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -691,12 +691,32 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
     wrappingMode?: DynamicProperty<"None" | "Character" | "Word">;
     /**
      * Text position regarding the baseline.
+     *
```
atomicsulfate: See the documentation for xOffset, yOffset. It's wrongly stating a hardcoded offset direction:
>  * Horizontal offset (to the right) in screen pixels.
> * Vertical offset (up) in screen pixels.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
atomicsulfate: what about TextTechniqueParams? We have to do the same change there.

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -1001,6 +1001,8 @@ export interface TextStyleDefinition {
     wrappingMode?: "None" | "Character" | "Word";
     hAlignment?: "Left" | "Center" | "Right";
     vAlignment?: "Above" | "Center" | "Below";
+    hPlacement?: "Left" | "Center" | "Right";
+    vPlacement?: "Above" | "Center" | "Below";
```
atomicsulfate: "Top" instead of "Above", "Bottom" instead of "Below"

@here/harp-text-canvas/lib/rendering/TextStyle.ts
atomicsulfate: >  @note Alignment attributes `hAlignment` and `vAlignment` are now **depreciated** in Theme
deprecated

@here/harp-mapview/lib/text/TextStyleCache.ts
```diff
@@ -441,19 +461,48 @@ export class TextStyleCache {
...
+        horizontalAlignment: HorizontalAlignment;
+        verticalAlignment: VerticalAlignment;
+    } {
+        // Parse placement properties if available at they have higher priority in case of clash.
+        const horizontalPlacement =
+            hPlacement === "Left" || hPlacement === "Center" || hPlacement === "Right"
+                ? HorizontalPlacement[hPlacement]
+                : undefined;
+        const verticalPlacement =
+            vPlacement === "Above" || vPlacement === "Center" || vPlacement === "Below"
```
atomicsulfate: "Above" to "Top" and "Below" to "Bottom"

@here/harp-text-canvas/lib/rendering/TextStyle.ts
```diff
@@ -55,13 +55,38 @@ export enum VerticalAlignment {
...
 export enum HorizontalAlignment {
     Left = 0.0,
     Center = -0.5,
     Right = -1.0
 }
 
+/**
+ * Vertical position of text area relative to the placement context (point, line).
+ */
+export enum VerticalPlacement {
```
atomicsulfate: Same as VerticalAlignment. if you could make an alias, it'd be ok (but i don't you think you can do that in typescript). Otherwise I don't think it's worth to duplicate the enum.

@here/harp-text-canvas/lib/rendering/TextStyle.ts
```diff
@@ -55,13 +55,38 @@ export enum VerticalAlignment {
 
 /**
  * Horizontal alignment to be used when placing text.
+ *
+ * @note Horizontal alignment value is exactly opposite to [[HorizontalPlacement]] value,
+ * cause when you place text on the right side of point (or icon) you normally want to have it
+ * left-aligned.
+ * @note Alignment attributes `hAlignment` and `vAlignment` are now depreciated in Theme
+ * definition for [[PoiTechinque]] and [[LineMarkerTechnique]] so use rather `hPlacement` and
+ * `vPlacement` for them.
```
atomicsulfate:
```suggestion
export enum HorizontalAlignment {
    Left = HorizontalPlacement.Right,
    Center = HorizontalPlacement.Center,
    Right = HorizontalPlacement.Left
}	}
```

@here/harp-text-canvas/lib/rendering/TextStyle.ts
```diff
@@ -55,13 +55,38 @@ export enum VerticalAlignment {
...
  */
 export enum HorizontalAlignment {
```
atomicsulfate:
```suggestion
export enum HorizontalAlignment {
    Left = HorizontalPlacement.Right,
    Center = HorizontalPlacement.Center,
    Right = HorizontalPlacement.Left
}	}
```

@here/harp-text-canvas/lib/rendering/TextStyle.ts
```diff
@@ -55,13 +55,38 @@ export enum VerticalAlignment {
...
     Left = 0.0,
     Center = -0.5,
     Right = -1.0
 }
```
atomicsulfate:
```suggestion
export enum HorizontalAlignment {
    Left = HorizontalPlacement.Right,
    Center = HorizontalPlacement.Center,
    Right = HorizontalPlacement.Left
}	}
```

@here/harp-datasource-protocol/lib/TechniqueParams.ts
kkostecki: See my comments in original ticket -> probably there is a handy solution how the placement should be defined, but... First of all we do not have (AT ALL) support for storing alignment/placement priorities, we do not have support for solving such labels allocation in screen space, we will perform much more computation to find good allocation spot, and last but not least it is not the scope of this ticket! Please read original description.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
kkostecki: I don't know either, but this is what we agreed on the last standup :). I can quickly get rid of depreciation of course, but the real problem is what should I actually implement in this ticket :).

@here/harp-datasource-protocol/lib/TechniqueParams.ts
kkostecki: I did not added it there, because as far as I know the change (placement) regards to the point based labels. I don't exactly know what the meaning would be for simple text... could you clarify.

@here/harp-mapview/lib/text/TextStyleCache.ts
kkostecki: Correct, sorry messed up during merge.

@here/harp-mapview/lib/text/TextStyleCache.ts
```diff
@@ -441,19 +461,48 @@ export class TextStyleCache {
...
-                    style.hAlignment === "Right"
-                        ? HorizontalAlignment[style.hAlignment]
-                        : HorizontalAlignment.Center
+                verticalAlignment,
+                horizontalAlignment
             }
         };
     }
+
+    private resolvePlacementAndAlignment(
```
atomicsulfate: This doesn't belong to TextStyleCache, it makes more sense to have it as a free function in TextStyle.

@here/harp-text-canvas/lib/rendering/TextStyle.ts
```diff
@@ -447,6 +472,28 @@ export class TextLayoutStyle {
         this.m_params.horizontalAlignment = value;
     }
 
+    /**
```
atomicsulfate: do you plan on using these methods? If no, better remove them.

@here/harp-text-canvas/lib/rendering/TextStyle.ts
```diff
@@ -447,6 +472,28 @@ export class TextLayoutStyle {
...
+     * Text element vertical position relating to the point of placement.
+     */
+    get horizontalPlacement(): HorizontalPlacement {
+        return (this.m_params.horizontalAlignment! as unknown) as HorizontalPlacement;
+    }
+    set horizontalPlacement(value: HorizontalPlacement) {
+        // Update horizontal alignment, which is exactly opposite to placement.
+        this.m_params.horizontalAlignment = (value as unknown) as HorizontalAlignment;
+    }
```
atomicsulfate: do you plan on using these new methods? If no, better remove them.

@here/harp-text-canvas/lib/rendering/TextStyle.ts
atomicsulfate:
```suggestion
export enum HorizontalAlignment {
    Left = HorizontalPlacement.Right,
    Center = HorizontalPlacement.Center,
    Right = HorizontalPlacement.Left
}	}
```

@here/harp-datasource-protocol/lib/TechniqueParams.ts
atomicsulfate: no, the ticket is actually asking to rename Above->Top , Below-> Bottom

atomicsulfate: For text it would be the placement relative to the text anchor.
As we've seen before in native the have "placement" attribute for all labels.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
atomicsulfate: @ninok @kkostecki Instead of deprecating alignment, we could just document the current limitation (that hAlignment is taken as the opposite value of hPlacement).
Then in the future we could add support for combinations.

But just to be on the same page, I only see the point of specifying alignment together with placement for multiline text right? Otherwise, a I don't see with respect to what would you align a left-align left-placed line of text.

kkostecki: I agree, maybe better to leave with both... not deprecated, and then we have the possibility to extend... actually the second PR #1407 was directed that way, but I wanted to still be on AC side so there are lines that override alignment with placement (well commented).
Anyway @atomicsulfate sorry for taking you time, but this approach will be probably all gone and replaced with placement: "TR" | "R" | "BR" ... attributes, possibly with priority array as proposed in ticket comments (see Jira).

@here/harp-datasource-protocol/lib/TechniqueParams.ts
kkostecki: Guys, nice to see the strong discussion here, but I guess the ticket description was in totally different area and now AC have changed significantly, probably this entire PR will be outdated, but need to wait for formal AC definition. I have also kind of "make it all" proposal in the Jira ticket comments:
```
placement: [
                "case",
                        ["==", ["get", "readDirection"], "left-right"],
                        ["array", "TR", "TL", "CC" ],
                        ["==", ["get", "readDirection"], "right-left"],
                        ["array", "TL", "TR", "CC" ]
                 ...
             ]
```

PR #1414, opened 2020-03-26
HARP-6488: Solve multiple label placements when checking collisions.

kkostecki:
POC code which checks different anchor point (placements) that comes
internally from hard-coded table. Final solution should read
possible placement points from the technique (theme) definition and
store them in TextElement.m_layoutStyle.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -429,6 +429,11 @@ function placePointLabelChoosingAnchor(
     // Store label state - persistent or new label.
     const persistent = labelState.visible;
 
+    // Text elements with multi-anchor placement require their own layout style.
+    // Provide private copy of label layout and attach it to current label and text canvas.
+    label.layoutStyle = label.layoutStyle!.clone();
+    textCanvas.textLayoutStyle = label.layoutStyle!;
+
     // The current implementation does not provide placements options via theme style yet,
     // so function tries the anchor placements from pre-defined placements arrays.
     // TODO: HARP-6487 Placements options should be loaded from the theme.
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -308,6 +308,71 @@ export function placeIcon(
 }
 
 /**
+ * Place a point label text using single or multiple optional placements.
+ *
+ * The current implementation does not provide placements options via style, so function
+ * tries to limit multiple placements test to scenarios where it is needed.
+ *
+ * @note This to be changed after HARP-6487 will be impemented and styles adapted.
+ *
+ * @param labelState State of the point label to place.
+ * @param screenPosition Position of the label in screen coordinates.
+ * @param scale Scale factor to be applied to label dimensions.
+ * @param isRejected Whether the label is already rejected (e.g. because its icon was rejected). If
+ * `true`, text won't be checked for collision, result will be either `PlacementResult.Invisible`
+ * for newly placed (upcoming) label or `PlacementResult.Rejected` if the label was persistent.
+ * @param textCanvas The text canvas where the label will be placed.
+ * @param screenCollisions Used to check collisions with other labels.
+ * @param outScreenPosition The final label screen position after applying any offsets.
+ * @returns `PlacementResult.Ok` if point __label can be placed__ at the base or any optional
+ * anchor point. `PlacementResult.Rejected` if there's a collision for all placements or it's
+ * __persistent label with icon rejected and text visible__. Finally `PlacementResult.Invisible`
+ * if it's text is not visible at any placement position or it's __new label with text or icon__
+ * __rejected__.
+ */
+export function placePointLabel(
+    labelState: TextElementState,
+    screenPosition: THREE.Vector2,
+    scale: number,
+    textCanvas: TextCanvas,
+    screenCollisions: ScreenCollisions,
+    isRejected: boolean,
+    outScreenPosition: THREE.Vector3
+): PlacementResult {
+    const layoutStyle = labelState.element.layoutStyle!;
+
+    // For centered point labels do only current anchor testing.
+    // TODO: Should be removed after placements are provided via style - see HARP-6487.
+    if (
+        layoutStyle.verticalAlignment === VerticalAlignment.Center &&
+        layoutStyle.horizontalAlignment === HorizontalAlignment.Center
+    ) {
+        return placePointLabelAtCurrentAnchor(
+            labelState,
+            screenPosition,
+            scale,
+            textCanvas,
+            screenCollisions,
+            isRejected,
+            false,
+            outScreenPosition
+        );
+    }
+    // Otherwise test also optional text placements.
+    else {
+        return placePointLabelChoosingAnchor(
+            labelState,
+            screenPosition,
+            scale,
+            textCanvas,
+            screenCollisions,
+            isRejected,
+            outScreenPosition
+        );
+    }
+}
+
+/**
  * Try to place a point label text using multiple optional placements.
  *
  * @param labelState State of the point label to place.
@@ -323,7 +388,7 @@ export function placeIcon(
  * `PlacementResult.Rejected` if there's a collision for all placements, `PlacementResult.Invisible`
  * if it's not visible any any placement position.
  */
-export function placePointLabelChoosingAnchor(
+function placePointLabelChoosingAnchor(
     labelState: TextElementState,
     screenPosition: THREE.Vector2,
     scale: number,
@@ -346,12 +411,14 @@ export function placePointLabelChoosingAnchor(
     };
     // Placements options will be read from label.layoutStyle.placements in final solution.
     const placements = possibleAnchorPlacementsCW;
-    // Find first anchor placement.
+    // Find current anchor placement on the optional placements list.
     const idx = placements.findIndex(val => {
         return val.h === basePlacement.h && val.v === basePlacement.v;
     });
-    // If not listed in placement options list (idx === -1) will start from current alignment setup.
+    // If not listed in placements list (idx === -1) will start from current anchor setup.
     const placementsNum = placements.length;
+    // Store label state - persistent or new label.
+    const persistent = labelState.visible;
     // Will be true if all text placements are invisible.
     let allInvisible: boolean = true;
     let i = 0;
@@ -362,23 +429,26 @@ export function placePointLabelChoosingAnchor(
         textCanvas.textLayoutStyle.horizontalAlignment = anchorPlacement.h;
         textCanvas.textLayoutStyle.verticalAlignment = anchorPlacement.v;
 
+        // Bounds will be already calculated for persistent label, when using base alignment.
+        const boundsReady = persistent && i === 0;
         // Compute label bounds, visibility or collision according to new layout settings.
-        const placementResult = placePointLabel(
+        const placementResult = placePointLabelAtCurrentAnchor(
             labelState,
             screenPosition,
             scale,
             textCanvas,
             screenCollisions,
             isRejected,
-            true,
+            !boundsReady,
             outScreenPosition
         );
 
-        // Check the text visibility.
+        // Check the text allocation
         if (placementResult === PlacementResult.Invisible) {
+            // Persistent label out of screen or the new label that is colliding - next iteration.
             continue;
-            // This placement if visible, check for collisions.
         } else {
+            // This placement is visible, but surely colliding.
             allInvisible = false;
         }
 
@@ -411,16 +481,15 @@ export function placePointLabelChoosingAnchor(
     return allInvisible
         ? // All text's placements out of the screen.
           PlacementResult.Invisible
-        : // Some text on screen:
-        labelState.visible
-        ? // both text and icon colliding have colliders.
+        : persistent
+        ? // All placements are either colliding or out of screen for persistent label.
           PlacementResult.Rejected
-        : // label's icon is already invisible.
+        : // No placement found for the new label.
           PlacementResult.Invisible;
 }
 
 /**
- * Places a point label on a specified text canvas.
+ * Places a point label on a specified text canvas using the alignment (anchor) currently set.
  *
  * @param labelState State of the point label to place.
  * @param screenPosition Position of the label in screen coordinates.
@@ -436,7 +505,7 @@ export function placePointLabelChoosingAnchor(
  * @returns `PlacementResult.Ok` if path label can be placed, `PlacementResult.Rejected` if there's
  * a collision, `PlacementResult.Invisible` if it's not visible.
  */
-export function placePointLabel(
+function placePointLabelAtCurrentAnchor(
     labelState: TextElementState,
     screenPosition: THREE.Vector2,
     scale: number,
@@ -481,18 +550,22 @@ export function placePointLabel(
         tmpBox.max.y - tmpBox.min.y
     );
 
-    // Check the text visibility.
+    // Check the text visibility if invisible finish immediately
+    // regardless of the persistence state - no fading required.
     if (!screenCollisions.isVisible(tmp2DBox)) {
         return PlacementResult.Invisible;
     }
 
+    const persistent = labelState.visible;
     // Check if icon's label was already rejected.
     if (isRejected) {
-        return labelState.visible ? PlacementResult.Rejected : PlacementResult.Invisible;
+        // Allows to fade out persistent label and simply ignore new one.
+        return persistent ? PlacementResult.Rejected : PlacementResult.Invisible;
     }
     // Check label's text collision.
     if (!label.textMayOverlap && screenCollisions.isAllocated(tmp2DBox)) {
-        return labelState.visible ? PlacementResult.Rejected : PlacementResult.Invisible;
+        // Allows to fade persistent and ignore new label.
+        return persistent ? PlacementResult.Rejected : PlacementResult.Invisible;
     }
 
     // Don't allocate space for rejected text. When zooming, this allows placement of a
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -422,23 +422,46 @@ function placePointLabelChoosingAnchor(
     };
     // Placements options will be read from label.layoutStyle.placements in final solution.
     const placements = possibleAnchorPlacementsCW;
-    // Find current anchor placement on the optional placements list.
-    const idx = placements.findIndex(val => {
-        return val.h === basePlacement.h && val.v === basePlacement.v;
-    });
-    // If not listed in placements list (idx === -1) will start from current anchor setup.
     const placementsNum = placements.length;
+    // Find current anchor placement on the optional placements list or best matching position.
+    // Index of exact match.
+    let matchIdx = -1;
+    // Index of closest placement to the base one.
+    let closeIdxOff = 0;
+    let closeDiff = Infinity;
+    for (let i = 0; i < placementsNum; ++i) {
+        const placement = placements[i];
+        const hDiff = placement.h - basePlacement.h;
+        const vDiff = placement.v - basePlacement.v;
+        const manhattanDist = Math.abs(hDiff) + Math.abs(vDiff);
+        if (manhattanDist === 0) {
+            matchIdx = i;
+            closeIdxOff = 0;
+            break;
+        } else if (manhattanDist < closeDiff) {
+            closeDiff = manhattanDist;
+            closeIdxOff = i;
+        }
+    }
+
     // Store label state - persistent or new label.
     const persistent = labelState.visible;
     // Will be true if all text placements are invisible.
     let allInvisible: boolean = true;
-    let i = 0;
-    for (i = 0; i < placementsNum; ++i) {
-        const anchorPlacement = idx + i < 0 ? basePlacement : placements[(idx + i) % placementsNum];
+    // If current anchor found on the list we start from it.
+    // If not listed in placements list (matchIdx === -1) we start from current anchor setup, but
+    // then immediately jump to the best matching option and iterate the over rest of placements.
+    // The number of iterations varies from `placementsNum` if perfect match found to
+    // `placementsNum + 1` otherwise (current placement + all optional).
+    for (let i = matchIdx; i < placementsNum; ++i) {
+        const anchorPlacement =
+            i < 0 ? basePlacement : placements[(i + closeIdxOff) % placementsNum];
         // Override label text alignment for measurements and leave it so if passed collisions test.
-        // NOTE: This actually writes to label.layoutStyle.verticalAlignment/horizontalAlignment.
-        textCanvas.textLayoutStyle.horizontalAlignment = anchorPlacement.h;
-        textCanvas.textLayoutStyle.verticalAlignment = anchorPlacement.v;
+        // NOTE: This actually writes also to:
+        // textCanvas.textLayoutStyle.verticalAlignment/horizontalAlignment.
+        // which is needed for text measurements being done on canvas.
+        label.layoutStyle!.horizontalAlignment = anchorPlacement.h;
+        label.layoutStyle!.verticalAlignment = anchorPlacement.v;
 
         // Bounds will be already calculated for persistent label, when using base alignment.
         const boundsReady = persistent && i === 0;
@@ -495,7 +518,6 @@ function placePointLabelChoosingAnchor(
 
         // Glyphs arrangement have been changed remove text buffer object.
         if (i !== 0) {
-            delete label.textBufferObject;
             label.textBufferObject = undefined;
         }
         // Proper placement found.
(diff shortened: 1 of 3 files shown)


kkostecki:
This is still in draft / work in progress state, but published as common PR to pass whole CI chain.

kkostecki:
restest this please

kkostecki:
retest this please

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -121,6 +121,27 @@ export enum PrePlacementResult {
...
+    v: VerticalAlignment;
+}
+
+/**
+ * @hidden
+ * Possible placement scenarios in clock-wise order.
+ *
+ * TODO: This array should be parsed from the style/theme definition of the technique.
+ */
+const possibleAnchorPlacementsCW = [
```
atomicsulfate: What about {Center,Above}, {Right,Center} ... ? Don't we need those relative positions for China?

I would expect to have here 8 combinations, all except {Center,Center}. If a label has centered placement, let's assume so far it doesn't want alternative placements. That way we avoid testing alternative placements for labels without icons (they are centered in our styles).

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -190,18 +211,29 @@ export function checkReadyForPlacement(
...
+ * @param anchorPlacement The optional text placement parameter that overrides layout style
+ * settings stored in [[textElement.layoutStyle.horizontalAlignment]] and
+ * [[textElement.layoutStyle.verticalAlignment]]
  */
 function computePointTextOffset(
     textElement: TextElement,
     scale: number,
-    offset: THREE.Vector2 = new THREE.Vector2()
+    offset: THREE.Vector2 = new THREE.Vector2(),
+    anchorPlacement?: AnchorPlacement
```
atomicsulfate: better remove this optional parameter if you're not using it.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -313,46 +345,118 @@ export function placePointLabel(
...
+    let i = 0;
+    for (i = 0; i < placementsNum; ++i) {
+        const anchorPlacement = idx + i < 0 ? basePlacement : placements[(idx + i) % placementsNum];
+        // Override label text alignment for measurements and leave it so if passed collisions test.
+        // NOTE: This actually writes to label.layoutStyle.verticalAlignment/horizontalAlignment.
+        textCanvas.textLayoutStyle.horizontalAlignment = anchorPlacement.h;
+        textCanvas.textLayoutStyle.verticalAlignment = anchorPlacement.v;
+
+        // Compute label bounds according to layout settings.
+        textCanvas.measureText(label.glyphs!, label.bounds, tmpMeasurementParams);
```
atomicsulfate: Why are you computing the label bounds for every possible placement? They are independent from the label position right? 
They could be computed once and cached in textElement just as it was done before.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -313,46 +345,118 @@ export function placePointLabel(
...
-
-    // Check the text visibility.
-    if (!screenCollisions.isVisible(tmp2DBox)) {
-        return PlacementResult.Invisible;
-    }
+    // Setup measurements parameters for textCanvas.measureText().
+    tmpMeasurementParams.outputCharacterBounds = undefined;
+    tmpMeasurementParams.path = undefined;
+    tmpMeasurementParams.pathOverflow = false;
+    tmpMeasurementParams.letterCaseArray = label.glyphCaseArray!;
```
atomicsulfate: For persistent labels you most probably won't have to change the placement and therefore won't have to measure the text again.
This function need to be split into two: the inner one is mostly like the placePointLabel and the new one that calls placePointLabel with every placement alternative.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -313,46 +345,118 @@ export function placePointLabel(
...
+    });
+    // If not listed in placement options list (idx === -1) will start from current alignment setup.
+    const placementsNum = placements.length;
+    let textOffset: THREE.Vector2;
+    // Will be true if all text placements are invisible.
+    let allInvisible: boolean = true;
+    let i = 0;
+    for (i = 0; i < placementsNum; ++i) {
+        const anchorPlacement = idx + i < 0 ? basePlacement : placements[(idx + i) % placementsNum];
+        // Override label text alignment for measurements and leave it so if passed collisions test.
```
atomicsulfate: Better not touching the layout style, just pass the placement alternative to the computePointTextOffset function.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -313,46 +345,118 @@ export function placePointLabel(
...
+        // If icon's is isRejected, should we test visibility of all placements, maybe better to
+        // interrupt tests now and save some CPU time?
+        if (isRejected) {
+            continue;
+            //return labelState.visible ? PlacementResult.Rejected : PlacementResult.Invisible;
+        }
+        // Check label's text collision.
+        // TODO: Same comment as above, maybe better to interrupt the tests here after checking
+        // first placement and save CPU time.
+        if (!label.textMayOverlap && screenCollisions.isAllocated(tmp2DBox)) {
```
atomicsulfate: if textMayOverlap you don't need to test further placements! It means the text will be placed wherever the style wants!

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -313,46 +345,118 @@ export function placePointLabel(
...
+        // TODO:
+        // Check what the old code actually gives for callers?
+        // If icon's is isRejected, should we test visibility of all placements, maybe better to
+        // interrupt tests now and save some CPU time?
+        if (isRejected) {
+            continue;
+            //return labelState.visible ? PlacementResult.Rejected : PlacementResult.Invisible;
+        }
+        // Check label's text collision.
+        // TODO: Same comment as above, maybe better to interrupt the tests here after checking
```
atomicsulfate: you can't interrupt the test here, this condition means the placement alternative failed, you need to test the rest.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -313,46 +345,118 @@ export function placePointLabel(
...
+        } else {
+            allInvisible = false;
+        }
+
+        // If icon rejected proceed to test further if all elements are invisible.
+        // TODO:
+        // Check what the old code actually gives for callers?
+        // If icon's is isRejected, should we test visibility of all placements, maybe better to
+        // interrupt tests now and save some CPU time?
+        if (isRejected) {
```
atomicsulfate: If the label isRejected, you only want to know if the text is invisible in its current placement, to decide whether it needs to be faded out or it can be removed directly.
So if isRejected is true, just check visibility for the current position and return either Invisible or Rejected.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -313,46 +345,118 @@ export function placePointLabel(
...
+    tmpMeasurementParams.pathOverflow = false;
+    tmpMeasurementParams.letterCaseArray = label.glyphCaseArray!;
+
+    // NOTE:
+    // We may save some CPU time by firstly checking visibility of all possible text placements,
+    // their bounding boxes merged altogether, this may save time for further calculations.
+    // Would be enough to include TL and BR placements.
+
+    // Start with currently set alignment settings.
+    const basePlacement = {
```
atomicsulfate: if the label is a persistent one, the basePlacement is not the one given in the style, but the one which is currently used by the label.
This means you're gonna have to store the current placement somewhere.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Good point of course! The problem is that with hard-coded anchors placement we would need to test 9 scenarios (8 on corners, and center-center), this is way too much I guess.

> That way we avoid testing alternative placements for labels without icons (they are centered in our styles).
Good catch I think we may use it as a pre-mature optimization which at least limits the computations, of course finally the best is to have it configured via theme .json.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: New version removes this function entirely so marking as solved.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: I have split it in the new commit as agreed, as for persistent labels do you suggest to check run only single (current) placement version? As far as I know this will "place" a label as new in the next run (frame) searching again for the new possibilities. Is this what we are aiming for?

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Yes, Andres and this is actually stored in label.layoutStyle!.horizontal/verticalAlignment, as far as I investigated it, this objects are not removed, so it will be hold until style is changed (zoom level change, etc..).  That is why I write directly to it... moreover textCanvas keeps shallow copy (reference) of label layout thus it is used internally everywhere during the measurement and text buffer creation. This was the main reason I couldn't pass new placement as the parameter, as tried in `computePointTextOffset` method. The change to fix that was massive and I have surrendered.... :).

@here/harp-mapview/lib/text/Placement.ts
kkostecki: See my comments above. I assume that last anchor placement (alignment) is stored in label object and this one in turn is persistent until label is removed or styling is changed (via theme rules - zoom level, etc.).

@here/harp-mapview/lib/text/Placement.ts
kkostecki: What I meant is that this logic is kind of awkward and I guess should be moved to caller (TextElementsRenderer). For me this function should simply return: Ok, Invisible (all invisible regardless of icon / label state), Rejected no solution found.
Currently (and in original version) there is a logic that complicates it a lot, because we are trying to make some deduction just for fading logic... This in turn results in unnecessary calculations for labels that are rejected because of icon, we would need to chat F2F to explain the details.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: You're losing here the preferred alignment coming from the style, overwriting it with the next succesful alignment option  that was found. 
This means next time you find that label (e.g. imagine you pan away from the label till it's not visible, rotate 90deg and pan back) you'll take that alignment and try to apply it first, even if the the original placement indicated by the style would be successful in this case.
To do this cleanly we should save the current alignment for persistent labels. I say the best place is in TextElementState. Even though we're saving lots of stuff in TextElement, i don't think it's the right place to do it.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -286,8 +307,121 @@ export function placeIcon(
...
+    const idx = placements.findIndex(val => {
+        return val.h === basePlacement.h && val.v === basePlacement.v;
+    });
+    // If not listed in placement options list (idx === -1) will start from current alignment setup.
+    const placementsNum = placements.length;
+    // Will be true if all text placements are invisible.
+    let allInvisible: boolean = true;
+    let i = 0;
+    for (i = 0; i < placementsNum; ++i) {
+        const anchorPlacement = idx + i < 0 ? basePlacement : placements[(idx + i) % placementsNum];
```
atomicsulfate: This loop isn't doing what's expected in several cases:
1) For centered labels we don't want to test other placement.
2) If the original placement is not in the placement options (idx ===-1) then you're only testing placementsNum-1 of the alternative placements (you will never test the last placement in the array, i.e. {Left, Below})!.
3) Since we're mising 4 placements in the array ({Top,Center}, {Bottom,Center}...), if the original placement is one of those, we won't start with the closest alternative placement, but always with {Left, Above}.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -477,6 +500,8 @@ function placePointLabelChoosingAnchor(
     // No placement found - revert back the original alignment.
     label.layoutStyle!.horizontalAlignment = basePlacement.h;
     label.layoutStyle!.verticalAlignment = basePlacement.v;
+    delete label.bounds;
```
atomicsulfate: Afaik this line is not needed, setting the bounds to undefined will suffice. I haven't seen this idiom used in our codebase, at least applied to typescript class members.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -341,11 +341,18 @@ export function placePointLabel(
...
+        return PlacementResult.Invisible;
+    }
+    // For centered point labels and labels with icon rejected, do only current anchor testing.
     // TODO: Should be removed after placements are provided via style - see HARP-6487.
     if (
-        layoutStyle.verticalAlignment === VerticalAlignment.Center &&
-        layoutStyle.horizontalAlignment === HorizontalAlignment.Center
+        isRejected ||
+        (layoutStyle.verticalAlignment === VerticalAlignment.Center &&
+            layoutStyle.horizontalAlignment === HorizontalAlignment.Center)
```
atomicsulfate: Since you're handling rejected labels here (both persistent and new ones), you don't need to pass the rejected flag to placePointLabelChoosingAnchor.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -443,6 +450,31 @@ function placePointLabelChoosingAnchor(
             outScreenPosition
         );
 
+        // Will be processed only once if the icon is rejected.
+        if (isRejected) {
```
atomicsulfate: With your current code we always have isRejected === false in this function, cause rejected labels are handled already in placePointLabel.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -452,16 +484,7 @@ function placePointLabelChoosingAnchor(
             allInvisible = false;
```
atomicsulfate: Comment is misleading, we don't know if label is colliding.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -286,8 +307,121 @@ export function placeIcon(
...
+        // Check label's text collision.
+        // TODO: Same comment as above, maybe better to interrupt the tests here after checking
+        // first placement and save CPU time.
+        if (placementResult === PlacementResult.Rejected) {
+            continue;
+        }
+
+        // Glyphs arrangement have been changed remove text buffer object.
+        if (i !== 0) {
+            delete label.textBufferObject;
```
atomicsulfate: Afaik this delete is not needed, setting the bounds to undefined will suffice. I haven't seen this idiom used in our codebase, at least applied to typescript class members.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Done.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: 1. Do you agree that 8 anchors is too much, may we stay with fixed 4 placements + eventually current one if not existent on placements list?
2. Done - centered labels are no longer processed with multi-anchor algorithm.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Please confirm that we don't want to search for alternatives if the label is persistent. See my comments above.
The rest is done.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: > You're losing here the preferred alignment coming from the style, overwriting it with the next succesful alignment option that was found.

That is intentional! Imagine different situation, when label is rejected for while by the more important label (usual when rotating the camera at some tilt) then it comes back after a sec, we should then keep layout consistent and reduce text jumping as much as possible. Another situation is when label goes out of screen but very close to it, when it returns back, should it be totally new layout or rather recently viewed one?

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Entire code has changed after function spiting.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: > if the label is a persistent one, the basePlacement is not the one given in the style, but the one which is currently used by the label.

As previously said - this is intentional. If we would like to store the new placement in the TextElementState, and leave TextElement.layoutStyle untouched, then we would need to override TextCanvas.textLayoutStyle anyway. Moreover for persistent labels this would need to be done all the time (with constant searching for better placement), because original layout will not fit in most cases (if camera does not change too much).

@here/harp-mapview/lib/text/Placement.ts
kkostecki: >1. For centered labels we don't want to test other placement.

And it will never do, but it is a part of different (and temporary logic) - see caller function.

> 2. If the original placement is not in the placement options (idx ===-1) then you're only testing placementsNum-1 of the alternative placements (you will never test the last placement in the array, i.e. {Left, Below})!.

Not true, it is exactly opposite, if original placement is in the options list then we test `placementsNum` iterations, if original is not on the list then we test `placementsNum + 1`, it is desired, and I don't want to add any artificial placement variant the the first scenario.

> 3. Since we're mising 4 placements in the array ({Top,Center}, {Bottom,Center}...), if the original placement is one of those, we won't start with the closest alternative placement, but always with {Left, Above}.

Good point for improvement, I have already changed that in the last update.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Done.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Probably you are thinking about caller function (`placePointLabel`). I agree that code there already excludes such scenario (rejected icon), but because I wanted it to be kind of independent function that may be used also for different scenarios I left rejected handling here. It is not that easy to understand that flow for further developers, so I am sure it is worth to leave this function flexible and compatible to its counterpart  `placePointLabelAtCurrentAnchor` as the name suggest.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Of course I know about that, but I would strongly recommend to leave this function compatible with other `placePointLabelAtCurrentAnchor` function. It may be used in some other scenarios (lineMarkers, etc...) too.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Done.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: If you really insist and make my PR green (approved) then I will remove it :).

kkostecki:
> placePointLabel() needs to be split, see my comments.

Done.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: So far `placePointLabelAtCurrentAnchor` is just an auxiliary function that is only used by placePointLabelChoosingAnchor. So better remove what's not needed, later it can be added if necessary.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: yes, I insist. I agree with you that handling the rejected state is confusing the further you get from TextElementRenderer. So, if we can avoid those 20 lines full of ifs, much better.
The way you handle it at the beginning of placePointLabelChoosingAnchor is clean and simple.

atomicsulfate: At this stage i would even rather remove isRejected from `placePointLabelAtCurrentAnchor ` as well, and just return Rejected for persistent labels with rejected icons at placePointLabel.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -422,23 +422,46 @@ function placePointLabelChoosingAnchor(
...
+        const manhattanDist = Math.abs(hDiff) + Math.abs(vDiff);
+        if (manhattanDist === 0) {
+            matchIdx = i;
+            closeIdxOff = 0;
+            break;
+        } else if (manhattanDist < closeDiff) {
+            closeDiff = manhattanDist;
+            closeIdxOff = i;
+        }
+    }
```
atomicsulfate: Frankly, this is overcomplicated. What if we just add the 8 placements in the placement list, but we just test 4 by jumping in increments of 2?
That way if base placement is {Top, Center}, we test  {Top, Center}, {Center,Right}. {Bottom,Center}, {Center,Left}
For base placement is on a corner like {Top,Right}, we test {T,R}, {B,R}, {B,L}, {T,L}.

And you can find always the match simply by doing findIndex(). No more handling of idx === -1 needed. Since {Center,Center} we handle it in placePointLabel.

kkostecki:
> needs to be split, see my comments.

Done.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Yes of course this is kind a nice workaround, but I will loose the flexibility and possibility to simply jump to Theme based placements (currently it enough to change array reference).
My proposal which should solve both problems is to provide two alternative tables:
- with centered anchors (North, South, East, West)
- with corner anchors (NW, NE, SE, SW).
Algorithm will check if theme placement is center based or corner and then do finding the closest match and iterate.
Although the current implementation may look over-complicated it is just finding the closest match and best match at the same time, I am doing so just to reduce number of searches (iterations) on placement array.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Done.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Done.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Current placement is now stored in TextElementState.LayoutState.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Done. We have two arrays for alternative placements:
- with centered positions (if that is the base one - in theme),
- with corner position (if theme has such layout).

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -429,6 +429,11 @@ function placePointLabelChoosingAnchor(
     // Store label state - persistent or new label.
     const persistent = labelState.visible;
 
+    // Text elements with multi-anchor placement require their own layout style.
+    // Provide private copy of label layout and attach it to current label and text canvas.
+    label.layoutStyle = label.layoutStyle!.clone();
```
atomicsulfate, line 435: so far let's do it like this, but after this PR we'll need to find a way to avoid cloning the style here, it can be a big performance impact.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -449,38 +450,15 @@ function placePointLabelChoosingAnchor(
...
-    // If not listed in placements list (matchIdx === -1) we start from current anchor setup, but
-    // then immediately jump to the best matching option and iterate the over rest of placements.
-    // The number of iterations varies from `placementsNum` if perfect match found to
-    // `placementsNum + 1` otherwise (current placement + all optional).
-    const triesNum = matchIdx < 0 ? placementsNum + 1 : placementsNum;
-    for (let i = matchIdx, tryIdx = 0; tryIdx < triesNum; ++i, ++tryIdx) {
-        const anchorPlacement =
-            i < 0 ? basePlacement : placements[(i + closeIdxOff) % placementsNum];
+    // Iterate all placements starting from current one.
+    for (let i = matchIdx, tryIdx = 0; tryIdx < placementsNum; ++i, ++tryIdx) {
```
atomicsulfate, line 460: minor: tryIdx is not needed. If you want to know if it's the base placement you can compare i === matchIdx.
You can fix this later.

kkostecki: Of course, but it is basically the same, `tryIdx` is not here for testing the base (first) run, but for counting the number of tests, so we are sure that we made a full round (circular iteration).

kkostecki:
gpu tests please

kkostecki:
@harpgl-bot aws tests please

kkostecki:
retest this please

kkostecki:

kkostecki:

kkostecki:
> Some note, can be done separate PR: there's no test for alternative placements when the original placement is centered (meaning Top,Center | Bottom,Center | Center,Left | Center.Right).

Yeah, good catch, I noticed it, but did not wanted to wait for another round of CI tests.

PR #1415, opened 2020-03-26
HARP-9449 Implement auth with token for webtile datasource.

MPursche:

# CI Check

Happy contributing!


@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -27,27 +33,41 @@ export interface WebTileRenderingOptions {
...
+ */
+export type AuthenticationProvider = () => Promise<string>;
+
 /**
  * An interface for the type of parameters that can be passed to the [[WebTileDataSource]].
  */
 export interface WebTileDataSourceParameters {
     /**
      * The `apikey` for the access of the Web Tile Data.
+     * @note Will not be used if authenticationCode is defined as well.
```
nzjony, line 47:
```suggestion
     * @note Will not be used if [[authenticationCode]] is defined as well.
```

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -27,27 +33,41 @@ export interface WebTileRenderingOptions {
...
      */
     apikey?: string;
 
     /**
      * The `appId` for the access of the Web Tile Data.
-     * @note Will not be used if apiKey is defined as well.
+     * @note Will not be used if apiKey or authenticationCode is defined as well.
```
nzjony, line 53:
```suggestion
     * @note Will not be used if apiKey or [[authenticationCode]] is defined as well.
```

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -27,27 +33,41 @@ export interface WebTileRenderingOptions {
...
      */
     appId?: string;
 
     /**
      * The `appCode` for the access of the Web Tile Data.
-     * @note Will not be used if apiKey is defined as well.
+     * @note Will not be used if apiKey or authenticationCode is defined as well.
```
nzjony, line 59:
```suggestion
     * @note Will not be used if apiKey or [[authenticationCode]] is defined as well.
```

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -398,6 +400,74 @@ export class WebTileDataSource extends DataSource {
...
+        }
+
+        if (requestParams.length > 0) {
+            return `?${requestParams.join("&")}`;
+        }
+
+        return "";
+    }
+
+    private async getRequestHeaders(): Promise<any> {
```
nzjony, line 452: can you give a bit more type information here? I guess any is "RequestHeaders"?

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -7,14 +7,20 @@
...
+} from "@here/harp-mapview";
 import { TileGeometryCreator } from "@here/harp-mapview/lib/geometry/TileGeometryCreator";
 import { getOptionValue, LoggerManager } from "@here/harp-utils";
 
 const logger = LoggerManager.instance.create("MapView");
 
-const textureLoader = new THREE.TextureLoader();
-textureLoader.crossOrigin = ""; // empty assignment required to support CORS
+const textureLoader = new TextureLoader();
+textureLoader.setCrossOrigin(""); // empty assignment required to support CORS
```
nzjony, line 23: can this be done in the constructor of the TextureLoader? I don't see the need to do it here...?

@here/harp-mapview/lib/copyrights/UrlCopyrightProvider.ts
```diff
@@ -22,11 +22,24 @@ export class UrlCopyrightProvider extends CopyrightCoverageProvider {
...
+        private m_requestHeaders?: any
+    ) {
         super();
     }
 
+    /**
+     * Sets request headers.
+     * @param headers
+     */
+    setRequestHeaders(headers: any) {
```
nzjony, line 39: why have `any` type?

@here/harp-mapview/lib/TextureLoader.ts
```diff
@@ -0,0 +1,98 @@
...
+
+        // Load image with fetch API if request header is set
+        const texture = new THREE.Texture();
+
+        this.manager.itemStart(url);
+        fetch(url, {
+            headers: this.requestHeaders,
+            mode: this.crossOrigin !== undefined ? "cors" : "no-cors"
+        })
+            .then(response => {
```
nzjony, line 53: is this formatted correctly? Seems weird.

MPursche: I also would like to format it differently but it's what prettier wants.
This would look much nicer:
```javascript
return textureLoader.load({
    //...
}).then(() => {
    //...
});
```

@here/harp-mapview/lib/TextureLoader.ts
```diff
@@ -0,0 +1,97 @@
...
+                this.manager.itemError(url);
+                this.manager.itemEnd(url);
+            });
+
+        return texture;
+    }
+
+    private loadImageFromBlob(blob: Blob): Promise<HTMLImageElement> {
+        return new Promise<HTMLImageElement>((resolve, reject) => {
+            try {
```
zbigg, line 84: This `try` catch is not needed: https://javascript.info/promise-error-handling#implicit-try-catch

@here/harp-mapview/test/TextureLoaderTest.ts
```diff
@@ -0,0 +1,60 @@
...
+const texturePath = getTestResourceUrl(
+    "@here/harp-mapview",
+    "test/resources/headshot.png"
+);
+
+describe("TextureLoader", function() {
+    describe("load", async function() {
+        if (!inNodeContext) {
+            it("loads images without request headers", async function() {
+                if (inNodeContext) {
```
zbigg, line 30: Will be always false

@here/harp-mapview/test/TextureLoaderTest.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const inNodeContext = typeof window === "undefined";
+const texturePath = getTestResourceUrl(
+    "@here/harp-mapview",
+    "test/resources/headshot.png"
+);
+
+describe("TextureLoader", function() {
+    describe("load", async function() {
+        if (!inNodeContext) {
```
zbigg, line 28: Early return ? Less indentation is better ?

@here/harp-mapview/lib/TextureLoader.ts
```diff
@@ -0,0 +1,97 @@
...
+        this.manager.itemStart(url);
+        fetch(url, {
+            headers: this.requestHeaders,
+            mode: this.crossOrigin !== undefined ? "cors" : "no-cors"
+        })
+            .then(response => {
+                return response.blob();
+            })
+            .then(blob => {
+                return this.loadImageFromBlob(blob);
```
zbigg, line 56: I don't think that we need to copy almost whole original TextureLoader here.

When you have blob, you can `super.load(URL.createObjectUrl(blob), ...).

See https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL

This should have exactly same complexity as code you've written here.

MPursche: I don't think this is possible if I want to keep the interface of `TextureLoader.load()` compatible. The function creates a texture that is immediately returned while starting the async loading of the image. But in order to call `super.load(URL.createObjectUrl(blob), ...)` the blob needs to be already loaded. The texture created by `super.load()` can't be returned immediately because the `fetch()` request needs to be completed before `super.load()` can be called. Additionally, the texture format(RGB vs RGBA) will not be detected correctly because the file format ending is only in the original url but not the object url.

@here/harp-mapview/lib/TextureLoader.ts
MPursche: Ok, will remove this.

@here/harp-mapview/lib/TextureLoader.ts
zbigg: You're right. Sorry for forcing you to write it, it's obvious.

Then why stick with "old", callback-era interface of TextureLoader. Do we really need to be compatible with original loader if we already reimplemented.

Why we can't have "our" texture loader with simple and robust promise-based API ?
``
const texture = await ourTextureLoader.load(...) // Proimise<texture>
```

zbigg: Answering potential question.

I know that original 3JS API returns empty texture synchronously, so we can use it, but i'm not sure if there is any use for texture that is not loaded, so API returning `Promise<Texture>` should be ok for us.

MPursche: I refactored the new Texture loaded to use a Promise based API.

PR #1417, opened 2020-03-27
HARP-9275 DisplacedMesh to support picking of elevated meshes.

atomicsulfate:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -307,6 +307,7 @@ export type ObjectConstructor = new (
  * Gets the default `three.js` object constructor associated with the given technique.
  *
  * @param technique The technique.
+ * @param tile The tile where the object is located.
  * @param elevationEnabled True if elevation is enabled, false otherwise.
  */
 export function getObjectConstructor(
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -39,8 +39,10 @@ import {
 } from "@here/harp-materials";
 import { LoggerManager } from "@here/harp-utils";
 import * as THREE from "three";
+import { DisplacedMesh } from "./geometry/DisplacedMesh";
 import { Circles, Squares } from "./MapViewPoints";
 import { toPixelFormat, toTextureDataType, toTextureFilter, toWrappingMode } from "./ThemeHelpers";
+import { Tile } from "./Tile";
 
 const logger = LoggerManager.instance.create("DecodedTileHelpers");
 
@@ -298,24 +300,37 @@ export function getBufferAttribute(attribute: BufferAttribute): THREE.BufferAttr
  * The default `three.js` object used with a specific technique.
  */
 export type ObjectConstructor = new (
-    geometry?: THREE.Geometry | THREE.BufferGeometry,
-    material?: THREE.Material | THREE.Material[]
+    geometry: THREE.Geometry | THREE.BufferGeometry,
+    material: THREE.Material | THREE.Material[]
 ) => THREE.Object3D;
 /**
  * Gets the default `three.js` object constructor associated with the given technique.
  *
  * @param technique The technique.
+ * @param elevationEnabled True if elevation is enabled, false otherwise.
  */
-export function getObjectConstructor(technique: Technique): ObjectConstructor | undefined {
+export function getObjectConstructor(
+    technique: Technique,
+    tile: Tile,
+    elevationEnabled: boolean
+): ObjectConstructor | undefined {
     if (technique.name === undefined) {
         return undefined;
     }
     switch (technique.name) {
         case "extruded-line":
         case "standard":
-        case "terrain":
         case "extruded-polygon":
         case "fill":
+            return elevationEnabled
+                ? DisplacedMesh.bind(undefined, () => {
+                      return {
+                          min: tile.elevationRange.minElevation,
+                          max: tile.elevationRange.maxElevation
+                      };
+                  })
+                : THREE.Mesh;
+        case "terrain":
         case "dashed-line":
         case "solid-line":
             return THREE.Mesh;
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -15,6 +15,7 @@ import * as THREE from "three";
 import { AnimatedExtrusionTileHandler } from "./AnimatedExtrusionHandler";
 import { CopyrightInfo } from "./copyrights/CopyrightInfo";
 import { DataSource } from "./DataSource";
+import { ElevationRange } from "./ElevationRangeSource";
 import { LodMesh } from "./geometry/LodMesh";
 import { TileGeometryLoader } from "./geometry/TileGeometryLoader";
 import { MapView } from "./MapView";
@@ -376,9 +377,8 @@ export class Tile implements CachedResource {
     // Center of the tile's unelevated bounding box world coordinates.
     private readonly m_worldCenter = new THREE.Vector3();
     private m_visibleArea: number = 0;
-    // Minimum and maximum tile elevation.
-    private m_minElevation: number = 0;
-    private m_maxElevation: number = 0;
+    // Tile elevation range in meters
+    private m_elevationRange: ElevationRange = { minElevation: 0, maxElevation: 0 };
     // Maximum height of geometry on this tile above ground level.
     private m_maxGeometryHeight?: number;
 
@@ -682,18 +682,28 @@ export class Tile implements CachedResource {
 
     /**
      * @internal
-     * Sets the tile's ground elevation range.
-     *
-     * @param minElevation Minimum tile elevation in meters.
-     * @param maxElevation Maximum tile elevation in meters.
+     * Gets the tile's ground elevation range in meters.
      */
-    setElevation(minElevation: number, maxElevation: number) {
-        if (minElevation === this.m_minElevation && maxElevation === this.m_maxElevation) {
+    get elevationRange(): ElevationRange {
+        return this.m_elevationRange;
+    }
+
+    /**
+     * @internal
+     * Sets the tile's ground elevation range in meters.
+     *
+     * @param elevationRange The elevation range.
+     */
+    set elevationRange(elevationRange: ElevationRange) {
+        if (
+            elevationRange.minElevation === this.m_elevationRange.minElevation &&
+            elevationRange.maxElevation === this.m_elevationRange.maxElevation &&
+            elevationRange.calculationStatus === this.m_elevationRange.calculationStatus
+        ) {
             return;
         }
 
-        this.m_minElevation = minElevation;
-        this.m_maxElevation = maxElevation;
+        this.m_elevationRange = elevationRange;
 
         // Only elevate bounding box if tile has already been decoded and a maximum geometry height
         // is provided by the data source.
@@ -1093,8 +1103,9 @@ export class Tile implements CachedResource {
         // Tile center remains unchanged.
         assert(this.m_maxGeometryHeight !== undefined);
 
-        this.geoBox.southWest.altitude = this.m_minElevation;
-        this.geoBox.northEast.altitude = this.m_maxElevation + this.m_maxGeometryHeight!;
+        this.geoBox.southWest.altitude = this.m_elevationRange.minElevation;
+        this.geoBox.northEast.altitude =
+            this.m_elevationRange.maxElevation + this.m_maxGeometryHeight!;
         this.mapView.projection.projectBox(this.geoBox, this.boundingBox);
     }
 
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -378,7 +378,7 @@ export class Tile implements CachedResource {
     private readonly m_worldCenter = new THREE.Vector3();
     private m_visibleArea: number = 0;
     // Tile elevation range in meters
-    private m_elevationRange: ElevationRange = { minElevation: 0, maxElevation: 0 };
+    private readonly m_elevationRange: ElevationRange = { minElevation: 0, maxElevation: 0 };
     // Maximum height of geometry on this tile above ground level.
     private m_maxGeometryHeight?: number;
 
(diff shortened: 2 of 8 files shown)


@here/harp-mapview/lib/geometry/DisplacedMesh.ts
```diff
@@ -0,0 +1,96 @@
...
+        const material: THREE.Material = Array.isArray(this.material)
+            ? this.material[0]
+            : this.material;
+
+        // Use default raycasting implementation if some type is unexpected.
+        if (
+            !(this.geometry instanceof THREE.BufferGeometry) ||
+            !hasDisplacementFeature(material) ||
+            !(material.displacementMap instanceof THREE.DataTexture)
+        ) {
```
kkostecki: Is this intended to use with different geometry or material (without displacement map). Wouldn't be easier to just assert in c-tor, all of that is already known during construction.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
```diff
@@ -0,0 +1,96 @@
...
+     * Creates an instance of displaced mesh.
+     * @param m_getDisplacementRange Displacement values range getter.
+     * @param [geometry] Original geometry to displace.
+     * @param [material] Material(s) to be used by the mesh. All must have the same displacement
+     * map.
+     */
+    constructor(
+        private m_getDisplacementRange: () => DisplacementRange,
+        geometry?: THREE.Geometry | THREE.BufferGeometry,
+        material?: THREE.Material | THREE.Material[]
```
kkostecki: Why not check the displacement map to be exactly equal (ref) or have equal properties? This may reduce the potential errors.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
```diff
@@ -0,0 +1,96 @@
...
+            !(material.displacementMap instanceof THREE.DataTexture)
+        ) {
+            super.raycast(raycaster, intersects);
+            return;
+        }
+        const displacementMap = material.displacementMap;
+        const displacementRange = this.m_getDisplacementRange();
+
+        if (this.m_displacedGeometry) {
+            this.m_displacedGeometry.reset(this.geometry, displacementMap, displacementRange);
```
kkostecki: Why it is needed? Do we expect displacementMap or displacementRange to change? I see that getter functor is const (without interface for change), so only displacementMap remains.. could you motivate it, just to make we are on the same path.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -552,7 +552,7 @@ export class VisibleTileSet {
                 // Update the visible area of the tile. This is used for those tiles that are
                 // currently loaded and are waiting to be decoded to sort the jobs by area.
                 tile.visibleArea = tileEntry.area;
-                tile.setElevation(tileEntry.minElevation, tileEntry.maxElevation);
+                tile.elevationRange = tileEntry;
```
kkostecki: This will mutate the elevationRange object, probably not intended. You want only min/maxElevation.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -682,18 +682,28 @@ export class Tile implements CachedResource {
...
+            elevationRange.minElevation === this.m_elevationRange.minElevation &&
+            elevationRange.maxElevation === this.m_elevationRange.maxElevation &&
+            elevationRange.calculationStatus === this.m_elevationRange.calculationStatus
+        ) {
             return;
         }
 
-        this.m_minElevation = minElevation;
-        this.m_maxElevation = maxElevation;
+        this.m_elevationRange = elevationRange;
```
kkostecki: Would be safer to make a copy of this object, elevationRange object probably (from caller) may changed out of this function scope.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
atomicsulfate: geometry and material in a THREE.Mesh can be changed at any moment, they are public writable properties, that's why we need to check on every call to raycast that everything is as expected.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
atomicsulfate: With what do you want to compare the displacement map?

@here/harp-mapview/lib/Tile.ts
atomicsulfate: True. I'm gonna let the caller do the copy though if it's needed. I'll change it in VisibleTileSet.

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: Yes, this is wrong, thanks for spotting it.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
atomicsulfate: Everything can change: displacementMap/Range due to terrain fallbacks and the geometry itself may also change since it's public in THREE.Mesh.

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -298,24 +300,37 @@ export function getBufferAttribute(attribute: BufferAttribute): THREE.BufferAttr
...
 export type ObjectConstructor = new (
-    geometry?: THREE.Geometry | THREE.BufferGeometry,
-    material?: THREE.Material | THREE.Material[]
+    geometry: THREE.Geometry | THREE.BufferGeometry,
+    material: THREE.Material | THREE.Material[]
 ) => THREE.Object3D;
 /**
  * Gets the default `three.js` object constructor associated with the given technique.
  *
  * @param technique The technique.
```
nzjony: please add Tile to the documentation

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -298,24 +300,37 @@ export function getBufferAttribute(attribute: BufferAttribute): THREE.BufferAttr
...
     if (technique.name === undefined) {
         return undefined;
     }
     switch (technique.name) {
         case "extruded-line":
         case "standard":
-        case "terrain":
         case "extruded-polygon":
         case "fill":
+            return elevationEnabled
```
nzjony: should we be able to enable the displaced mesh at run time? Why not always construct a DisplacedMesh, it would then remove the need for elevationEnabled

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -145,6 +145,14 @@ export function overlayOnElevation(tile: Tile): void {
         return;
     }
 
+    // Refine tile elevation range, which is initially set to the broader range given by
```
nzjony: because you are changing the Tile, please add this to the documentation, otherwise the side affect may not be apparent to users of this method.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
kkostecki: Ok.

@here/harp-mapview/lib/Tile.ts
kkostecki: I think it is good to do copy here, because the caller does not always knows what is the reference passed used for... This way single responsibility chain remains on the user (accessor class) to make sure object is not modified.
BTW we should already have some standards grown on such stuff... Good to ask on the team's channel.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
kkostecki: We could for example check if material has displacementMap defined, this is property appears for both build-on THREE.js materials and our feature mixins, keep in mind that displacementMap is Texture | null (not undefined). 
This is at least minimum requirement that gives the sense for using this class instead of standard mesh, otherwise it will be simply wrapper that works as any other mesh. 
Of course I see the test in raycast() method, but shouldn't we test before, if somebody is using the class with its designed purpose? I am leaving decision to you, but wanted to pinpoint that clearly.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -682,18 +682,28 @@ export class Tile implements CachedResource {
...
+    }
+
+    /**
+     * @internal
+     * Sets the tile's ground elevation range in meters.
+     *
+     * @param elevationRange The elevation range.
+     */
+    set elevationRange(elevationRange: ElevationRange) {
+        if (
```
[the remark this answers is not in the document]
atomicsulfate: if we don't check this, we'll recompute the tile's geoBox and boundingBox unnecessarily.

@here/harp-mapview/lib/DecodedTileHelpers.ts
atomicsulfate: done

@here/harp-mapview/lib/DecodedTileHelpers.ts
atomicsulfate: yes, we can do it at runtime. When we enable elevation, we clear the tile cache, and the newly constructed objects will use DisplaceMesh.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
atomicsulfate: ok, I'm gonna add the checks as assertions both in the constructor and raycast functions.

@here/harp-mapview/lib/Tile.ts
atomicsulfate: ok.

@here/harp-mapview/lib/DecodedTileHelpers.ts
nzjony: true, enabling elevation will anyway clear the cache, so it isn't like you can work around this.

@here/harp-mapview/lib/Tile.ts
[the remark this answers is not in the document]
nzjony: I see, sorry, I missed the lines beneath because of the conversation below and assumed there was a }, my bad.

@here/harp-mapview/lib/DecodedTileHelpers.ts
atomicsulfate: > true, enabling elevation will anyway clear the cache, so it isn't like you can work around this.
Sorry, I don't get it. What do you want to work around?

nzjony: My thinking was to enable the displacement at runtime, without having to redecode the tiles, but I agree, it was a dumb idea, because we anyway need to clear the tile cache (such runtime switching of decoding doesn't exist anyway).

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
```diff
@@ -0,0 +1,124 @@
...
+import { assert } from "@here/harp-utils";
+import * as THREE from "three";
+import { DisplacedBufferAttribute } from "./DisplacedBufferAttribute";
+import { DisplacedBufferGeometry, DisplacementRange } from "./DisplacedBufferGeometry";
+
+function isBufferGeometry(
+    geometry: THREE.Geometry | THREE.BufferGeometry
+): geometry is THREE.BufferGeometry {
+    const isBufferGeom = geometry instanceof THREE.BufferGeometry;
+    assert(isBufferGeom, "Unsupported geometry type.");
```
kkostecki: This assertions inside the check function looks little strange :). Either way you assert also from caller. Please remove all asserts out of isXXX function scope.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        // All materials in the object are expected to have the same displacement map.
+        const material = this.firstMaterial;
+
+        // Use default raycasting implementation if some type is unexpected.
+        if (
+            !isBufferGeometry(this.geometry) ||
+            !isDisplacementMaterial(material) ||
+            !isDataTextureMap(material.displacementMap)
```
kkostecki: Now we have assert inside isDataTextureMap()... so there is no other option here.

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: I guess you missed this within last PR update.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
```diff
@@ -0,0 +1,124 @@
...
+
+function isDisplacementMaterial(material: any): material is DisplacementFeature {
+    const isDisplacementFeature = hasDisplacementFeature(material);
+    assert(isDisplacementFeature, "Material does not support displacement maps.");
+    return isDisplacementFeature;
+}
+
+function isDataTextureMap(map: THREE.Texture | null): map is THREE.DataTexture {
+    const isDataTexture = map !== null && map instanceof THREE.DataTexture;
+    assert(isDataTexture, "Material does not support displacement maps.");
```
kkostecki: I guess this one may be checked in runtime (raycast), cause texture may change... Mixins (above) rather not.

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: i don't understand what you expect here now. I'm copying the elevation range inside the setter, what's the problem then?

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
atomicsulfate: I'm only asserting when calling from the constructor, not when calling from raycast. The issue there is that I need to keep an if so that tslint applies the casting correctly to the input parameters (material , geometry and displacementMap) so that they are recognized by tslint as having the appropiate types.
Anyway, I'll remove the assert from the constructor call.

@here/harp-mapview/lib/VisibleTileSet.ts
kkostecki: Oh... ok, I did not spotted that.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
atomicsulfate: i have to change something here, the map might be null in normal scenarios.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
atomicsulfate: done.

atomicsulfate:

atomicsulfate:

PR #1421, opened 2020-03-27
Speed up getMatchingTechniques.

robertoraggi:
Avoid evaluating filter conditions when the values of feature
properties used in conditions did not change.


robertoraggi:
GPU tests please

robertoraggi:
GPU tests please

PR #1430, opened 2020-03-31
HARP-8744 Fix lighting for shadows

nzjony:
This improves the lighting of flat objects, such that the final rendered pixel color isn't impacted by a diffuse component. Previously the color would be based on the direction (using a diffuse shading model), however this dramatically changes the color of the scene and would require a lot of rework from the designers / different theme files for shadow / non shadowed themes.

- "fill" techniques can now handle shadows, when shadows are enabled, they now
  use a tweaked version of the standard material.

### DIFF
--- a/@here/harp-examples/src/threejs_shadows.ts
+++ b/@here/harp-examples/src/threejs_shadows.ts
@@ -4,13 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {
-    isLiteralDefinition,
-    Style,
-    StyleDeclaration,
-    Theme
-} from "@here/harp-datasource-protocol";
-import { isJsonExpr } from "@here/harp-datasource-protocol/lib/Expr";
+import { Theme } from "@here/harp-datasource-protocol";
 import { GeoCoordinates } from "@here/harp-geoutils";
 import { MapControls, MapControlsUI } from "@here/harp-map-controls";
 import {
@@ -189,22 +183,11 @@ const addOmvDataSource = (): Promise<void> => {
     return map.addDataSource(omvDataSource);
 };
 
-const patchFillStyle = (styleDeclaration: StyleDeclaration) => {
-    if (!isJsonExpr(styleDeclaration)) {
-        const style = styleDeclaration as Style;
-        if (style.technique === "fill") {
-            (style as any).technique = "standard";
-        }
-    }
-};
-
 /**
- * Replace all occurences of "fill" technique in the theme with "standard" technique.
- * "standard" technique is using three.js MeshStandardMaterial and is needed to receive
- * shadows.
+ * Replaces the lights in the style to cast shadows.
  * @param theme The theme to patch
  */
-const patchTheme = (theme: Theme) => {
+const fixLights = (theme: Theme) => {
     theme.lights = [
         {
             type: "ambient",
@@ -214,36 +197,18 @@ const patchTheme = (theme: Theme) => {
         },
         {
             type: "directional",
-            color: "#ffcccc",
+            color: "#ffffff",
             name: "light1",
             intensity: 1,
+            // Will be overriden immediately, see `updateLight`
             direction: {
                 x: 0,
                 y: 0.01,
-                z: 1
+                z: -1
             },
             castShadow: true
         }
     ];
-    if (theme.styles === undefined || theme.styles.tilezen === undefined) {
-        throw Error("Theme has no tilezen styles");
-    }
-
-    if (theme.definitions !== undefined) {
-        for (const definitionName in theme.definitions) {
-            if (!theme.definitions.hasOwnProperty(definitionName)) {
-                continue;
-            }
-            const definition = theme.definitions[definitionName];
-            if (!isLiteralDefinition(definition)) {
-                const styleDeclaration = definition as StyleDeclaration;
-                patchFillStyle(styleDeclaration);
-            }
-        }
-    }
-    theme.styles.tilezen.forEach((styleDeclaration: StyleDeclaration) => {
-        patchFillStyle(styleDeclaration);
-    });
 };
 
 const addGuiElements = () => {
@@ -278,7 +243,7 @@ const addGuiElements = () => {
 
 export namespace ThreejsShadows {
     ThemeLoader.load("resources/berlin_tilezen_base.json").then((theme: Theme) => {
-        patchTheme(theme);
+        fixLights(theme);
         initializeMapView("mapCanvas", theme);
     });
 }
--- a/@here/harp-examples/src/threejs_shadows.ts
+++ b/@here/harp-examples/src/threejs_shadows.ts
@@ -7,12 +7,7 @@
 import { Theme } from "@here/harp-datasource-protocol";
 import { GeoCoordinates } from "@here/harp-geoutils";
 import { MapControls, MapControlsUI } from "@here/harp-map-controls";
-import {
-    CopyrightElementHandler,
-    MapView,
-    MapViewEventNames,
-    ThemeLoader
-} from "@here/harp-mapview";
+import { CopyrightElementHandler, MapView, MapViewEventNames } from "@here/harp-mapview";
 import { APIFormat, AuthenticationMethod, OmvDataSource } from "@here/harp-omv-datasource";
 import { assert } from "@here/harp-utils";
 import { GUI } from "dat.gui";
@@ -183,34 +178,6 @@ const addOmvDataSource = (): Promise<void> => {
     return map.addDataSource(omvDataSource);
 };
 
-/**
- * Replaces the lights in the style to cast shadows.
- * @param theme The theme to patch
- */
-const fixLights = (theme: Theme) => {
-    theme.lights = [
-        {
-            type: "ambient",
-            color: "#ffffff",
-            name: "ambientLight",
-            intensity: 0.9
-        },
-        {
-            type: "directional",
-            color: "#ffffff",
-            name: "light1",
-            intensity: 1,
-            // Will be overriden immediately, see `updateLight`
-            direction: {
-                x: 0,
-                y: 0.01,
-                z: -1
-            },
-            castShadow: true
-        }
-    ];
-};
-
 const addGuiElements = () => {
     // Instructions
     const message = document.createElement("div");
@@ -242,8 +209,29 @@ const addGuiElements = () => {
 };
 
 export namespace ThreejsShadows {
-    ThemeLoader.load("resources/berlin_tilezen_base.json").then((theme: Theme) => {
-        fixLights(theme);
-        initializeMapView("mapCanvas", theme);
-    });
+    const theme: Theme = {
+        extends: "resources/berlin_tilezen_base.json",
+        lights: [
+            {
+                type: "ambient",
+                color: "#ffffff",
+                name: "ambientLight",
+                intensity: 0.9
+            },
+            {
+                type: "directional",
+                color: "#ffffff",
+                name: "light1",
+                intensity: 1,
+                // Will be overriden immediately, see `updateLight`
+                direction: {
+                    x: 0,
+                    y: 0.01,
+                    z: -1
+                },
+                castShadow: true
+            }
+        ]
+    };
+    initializeMapView("mapCanvas", theme);
 }
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -86,6 +86,11 @@ export interface MaterialOptions {
      * recompiling them manually later (ThreeJS does not update fog for `RawShaderMaterial`s).
      */
     fog?: boolean;
+
+    /**
+     * Whether shadows are enabled or not, this is required because we change the material used.
+     */
+    shadowsEnabled?: boolean;
 }
 
 /**
@@ -102,7 +107,7 @@ export function createMaterial(
     textureReadyCallback?: (texture: THREE.Texture) => void
 ): THREE.Material | undefined {
     const technique = options.technique;
-    const Constructor = getMaterialConstructor(technique);
+    const Constructor = getMaterialConstructor(technique, options.shadowsEnabled === true);
 
     const settings: { [key: string]: any } = {};
 
@@ -116,6 +121,7 @@ export function createMaterial(
     ) {
         settings.fog = options.fog;
     }
+    settings.removeDiffuseLight = technique.name === "fill";
 
     const material = new Constructor(settings);
 
@@ -395,8 +401,13 @@ export type MaterialConstructor = new (params?: {}) => THREE.Material;
  * Returns a [[MaterialConstructor]] basing on provided technique object.
  *
  * @param technique [[Technique]] object which the material will be based on.
+ * @param shadowsEnabled Whether the material can accept shadows, this is required for some
+ * techniques to decide which material to create.
  */
-export function getMaterialConstructor(technique: Technique): MaterialConstructor | undefined {
+export function getMaterialConstructor(
+    technique: Technique,
+    shadowsEnabled: boolean
+): MaterialConstructor | undefined {
     if (technique.name === undefined) {
         return undefined;
     }
@@ -420,7 +431,7 @@ export function getMaterialConstructor(technique: Technique): MaterialConstructo
             return SolidLineMaterial;
 
         case "fill":
-            return MapMeshBasicMaterial;
+            return shadowsEnabled ? MapMeshStandardMaterial : MapMeshBasicMaterial;
 
         case "squares":
             return THREE.PointsMaterial;
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -121,7 +121,7 @@ export function createMaterial(
     ) {
         settings.fog = options.fog;
     }
-    settings.removeDiffuseLight = technique.name === "fill";
+    settings.removeDiffuseLight = options.shadowsEnabled === true && technique.name === "fill";
 
     const material = new Constructor(settings);
 
--- a/@here/harp-materials/lib/MapMeshMaterials.ts
+++ b/@here/harp-materials/lib/MapMeshMaterials.ts
@@ -14,6 +14,7 @@ import { DisplacementFeature, DisplacementFeatureParameters } from "./Displaceme
 import { ExtrusionFeatureDefs } from "./MapMeshMaterialsDefs";
 import extrusionShaderChunk from "./ShaderChunks/ExtrusionChunks";
 import fadingShaderChunk from "./ShaderChunks/FadingChunks";
+import { simpleLightingShadowChunk } from "./ShaderChunks/ShadowChunks";
 
 const emptyTexture = new THREE.Texture();
 
@@ -40,6 +41,18 @@ export interface FadingFeatureParameters {
      * Distance to the camera (range: `[0.0, 1.0]`) from which the objects are transparent.
      */
     fadeFar?: number;
+}
+
+/**
+ * Parameter used to control patching the standard material shader to ensure that the materials
+ * color isn't affected by the light direction, only valid for techniques that are "fill"
+ */
+export interface ShadowFeatureParameters {
+    /**
+     * Whether the diffuse light component is removed (i.e. the materials color is therefore just
+     * the ambient + shadow).
+     */
+    removeDiffuseLight?: boolean;
 }
 
 /**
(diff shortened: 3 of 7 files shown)


nzjony:

nzjony:

nzjony:

nzjony:

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -966,6 +966,12 @@ export interface DirectionalLight extends BaseLight {
     intensity: number;
     direction: Vector3Like;
     castShadow?: boolean;
+
+    /**
+     * Number between 0 and 1 which controls how strong the shadow is on flat surfaces.
+     * 0 means there is no shadow and 1 means the shadow appears as pure black.
+     */
+    shadowIntensity?: number;
```
ninok, line 974: Shouldn't this be in the material/technique?

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -60,6 +60,16 @@ export interface ExtrusionFeatureParameters {
...
 }
 
+/**
+ * Parameters used when constructing a [[MapMeshFlatStandardMaterial]].
+ */
+export interface ShadowParameters {
+    /**
+     * How strong the shadow is, 1 means complete black and 0 means no shadow shown.
+     */
+    shadowIntensity?: number;
```
ninok, line 70: So `shadowIntensity` is actually a material property. Why is it then not defined in the fill technique, but in the light when looking into the theme?
If the reason is, that you want to avoid patching the theme file we should add it to all styles Berlin theme so that you don't have to patch the theme

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -1230,6 +1240,129 @@ export class MapMeshStandardMaterial extends THREE.MeshStandardMaterial
...
+        this.extensions.derivatives = true;
+        this.lights = true;
+        this.shadowIntensity = params?.shadowIntensity ?? 0;
+        this.vertexShader = THREE.ShaderChunk.meshphysical_vert;
+        // There are three main changes, we add the shadowIntensity uniform, the
+        // RE_Direct_Physical function is changed to keep a consistent ambient color and the
+        // indirect diffuse / specular components are removed.
+        this.fragmentShader = THREE.ShaderChunk.meshphysical_frag.replace(
+            "#include <lights_physical_pars_fragment>",
+            `uniform float shadowIntensity;
```
ninok, line 1292: can you move this to a separate file to also be able to include it?

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -1230,6 +1240,129 @@ export class MapMeshStandardMaterial extends THREE.MeshStandardMaterial
     // Mixin declarations end -----------------------------------------------------------
 }
 
+/**
+ * Subclass of THREE.StandardMaterial. Uses the standard material (used in
```
ninok, line 1244: Code below sais it's derived from `ShaderMaterial`.

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -1230,6 +1240,129 @@ export class MapMeshStandardMaterial extends THREE.MeshStandardMaterial
...
+ * MapMeshStandardMaterial) as the base and removes the diffuse / specular components of the
+ * lighting to keep the colors consistent (that aren't in shadow).
+ *
+ * @see [[Tile#addRenderHelper]]
+ */
+export class MapMeshFlatStandardMaterial extends THREE.ShaderMaterial {
```
ninok, line 1250: Should't it derive from MapMeshBasicMaterial and just add the shadow chunk?

test/rendering/GeoJsonDataRendering.ts
```diff
@@ -29,12 +29,14 @@ describe("MapView + OmvDataSource + GeoJsonDataProvider rendering test", functio
...
             name: "light1",
-            intensity: 0.8,
+            intensity: 1.0,
             direction: {
                 x: 1,
                 y: 5,
                 z: 0.5
-            }
+            },
+            castShadow: true,
```
ninok, line 38: please create a dedicated testcase for shadows

@here/harp-materials/lib/MapMeshMaterials.ts
nzjony: I thought that as well, but there is much more similarity with the standard material than the basic one, because the standard material already has the correct structure, it just requires changing this one part. Otherwise I would need to pick out the correct #includes (they also depend on each other also), I would need to set the correct uniforms for the lights.

I can look into this if you think it is important, but it will take more time.

@here/harp-datasource-protocol/lib/Theme.ts
nzjony: Why would the designer want to customize it per material?

@here/harp-materials/lib/MapMeshMaterials.ts
nzjony: I don't see the need for such a level of customization, if the designers want to lighten the shadows, they would need to change it in 100 separate places, do we need this flexibility?

@here/harp-materials/lib/MapMeshMaterials.ts
nzjony: good catch

@here/harp-materials/lib/MapMeshMaterials.ts
ninok: you can still have a global variable in the style that is refered. So you still get the full flexibility and the ease of use.

@here/harp-datasource-protocol/lib/Theme.ts
ninok: Who knows. I don't see why we should restrict them.

@here/harp-materials/lib/MapMeshMaterials.ts
ninok: But why does it derive from ShaderMAterial. It should derive from MapStandardMaterial then (i.e. THREE.MeshStandardMaterial + Fading).

@here/harp-materials/lib/MapMeshMaterials.ts
nzjony: I'll answer on the previous comment, so will resolve this.

@here/harp-datasource-protocol/lib/Theme.ts
nzjony: Flexibility has its cost, we have to provide long term maintenance support, and removing it is difficult. Without a clear use case, I would prefer something simple that works. It would be easy to remove this later with minimal cost if there is a use case where a per material shadow is required.

I am also not so keen on mixing shadow properties with materials, to me it belongs with the light definition.

Maybe a good compromise, and one which would actually make sense, I remove the shadowIntensity and instead multiply the material color with the ambient light, that way, the ambient would actually make some sense and it would remove this parameter.

@here/harp-materials/lib/MapMeshMaterials.ts
nzjony: I tried this originally, it doesn't work because threejs has a shader cache internally, and if I modify the shader code, all MapMeshStandardMaterials are modified, see: https://github.com/mrdoob/three.js/blob/bcf0aca489b2a8e1da2b795fffcb36bb379ba665/src/renderers/webgl/WebGLPrograms.js#L55

@here/harp-datasource-protocol/lib/Theme.ts
ninok: I don't see why a per technique property is harder to maintain.
It's fine for me if we extend functionality later b/c it's to much effort to do it now but in this case we would have to deprecate the shadow intensity in the light source b/c some people might use it.
Or are you proposing to have both in the the future?

Not fully understand your last comment. How is the ambient light related to the shadows.

Maybe a third alternative to go forward is to just use the 0.46 intensity (where is this number coming from btw.?) and don't make it configurable for now.

nzjony: Every time a designer wants to style some thing with "fill" or "standard", they need to also copy across the shadow intensity.

Anyway, I still don't see how it is related to a material property.

Regarding the ambient light, we could say that the pixel color in shadow is material.color * ambientLight, because the material is lit just from the ambient light (and not the directional light).

nzjony: Yeah, it probably doesn't need to be configured, and when it does, we can find the best solution then, I will remove for now.

@here/harp-materials/lib/MapMeshMaterials.ts
zbigg: > I tried this originally, it doesn't work because threejs has a shader cache internally

@nzjony It _is_ possible to fix only this particular chunk. However it's tricky as hell. `getShaderObject` is followed by `onBeforeCompile` when you should "patch" shader like we do. We have many examples of this (grep onBeforeCompile in MapMeshMaterials).

Or ... you've found some fundamental problem and all of our "patching" machinery doesn't work.
@kkostecki did lots of work cleaning this up, he may help.

nzjony: Good point, I will investigate, I got some weird artifacts, and thought it wasn't possible,. because I noticed the shader was being changed for other shaders of the same type, but yes of course, it should be possible. I will retry.

nzjony: I looked into it, I wasn't differentiating between extruded-polygons and the fill technique (which both internally use the standard material from three.js, I read the three.js code wrong, so taking me down the wrong path, thanks @ninok  and @zbigg  for the comment in this regard!)

zhernovs:

nzjony:

nzjony:

nzjony:

@here/harp-materials/lib/ShaderChunks/ShadowChunks.ts
```diff
@@ -0,0 +1,51 @@
...
+    void RE_Direct_Physical( const in IncidentLight directLight,
+        const in GeometricContext geometry,
+        const in PhysicalMaterial material,
+        inout ReflectedLight reflectedLight ) {
+        // directLight.color is the light color * shadow, internally three.js uses a step function, so
+        // this value is either the light color or black. in order to lighten up the shadows, we
+        // subtract 1 from the light color and multiply that with 50% of the diffuse
+        // color.
+        #if defined(USE_SHADOWMAP)
+            reflectedLight.directDiffuse = (directLight.color +
```
MPursche, line 25: This formula can be simplified to
```GLSL
reflectedLight.directDiffuse = (0.5 * directLight.color + 0.5) * material.diffuseColor;
```

Explanation:
```
directLight = (directLight.color + (1.0 - directLight.color) * 0.5)
directLight = (directLight.color + 0.5 - 0.5*directLight.color)
directLight = (0.5 * directLight.color + 0.5)
```

@here/harp-materials/lib/ShaderChunks/ShadowChunks.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+export const correctShadowChunk = `
+    struct PhysicalMaterial {
+        vec3	diffuseColor;
+        float	specularRoughness;
+        vec3	specularColor;
+    };
+
+    #define DEFAULT_SPECULAR_COEFFICIENT 0.04
```
MPursche, line 14: Is this define still needed with the new dummy specular light calculation?

@here/harp-materials/lib/ShaderChunks/ShadowChunks.ts
```diff
@@ -0,0 +1,51 @@
...
+        #else
+            reflectedLight.directDiffuse = material.diffuseColor;
+        #endif
+    }
+
+    void RE_IndirectDiffuse_Physical( const in vec3 irradiance,
+        const in GeometricContext geometry,
+        const in PhysicalMaterial material,
+        inout ReflectedLight reflectedLight ) {
+            // Kept deliberately empty
```
MPursche, line 36: MINOR: `// Disable influence of indirect light`.

@here/harp-materials/lib/ShaderChunks/ShadowChunks.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    void RE_IndirectSpecular_Physical( const in vec3 radiance,
+        const in vec3 irradiance,
+        const in vec3 clearcoatRadiance,
+        const in GeometricContext geometry,
+        const in PhysicalMaterial material,
+        inout ReflectedLight reflectedLight) {
+            // Kept deliberately empty
```
MPursche, line 45: MINOR: `// Disable specular reflection of light`.

@here/harp-materials/lib/ShaderChunks/ShadowChunks.ts
nzjony: Yes, it is necessary in another shader chunk from threejs, I tried to remove, but was unsuccessful.

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -715,6 +728,41 @@ export class FadingFeatureMixin implements FadingFeature {
...
+
+    /**
+     * The mixin classes should call this method to register the property [[removeDiffuseLight]]
+     */
+    protected addShadowProperty(): void {
+        Object.defineProperty(this, "removeDiffuseLight", {
+            get: () => {
+                return this.m_removeDiffuseLight;
+            },
+            set: val => {
```
zbigg, line 759: I don't know if we this patch intends to support changing of `removeDiffuseLight`.

Because in this state `material.removeDiffuseLight = ...` completly doesn't work ...
If no, remove `set` and document

If it's indeed supported, one must ensure that changing value  of `m_removeDiffuseLight` triggers recompilation of shader (now it's only done in `applyShadowParameters` which is called only in constructor).

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -715,6 +728,41 @@ export class FadingFeatureMixin implements FadingFeature {
     }
 }
 
+/**
+ * Class to apply the code necessary to customize the standard shader to remove the diffuse light
+ * component.
+ */
+export class ShadowFeatureMixin {
```
zbigg, line 735: While current mixins live in kitchen sink `MapMeshMaterials` i would propose to move all stuff related to ShadowFeature to separate file ... (like `ShadowFeature.ts` ?)

@here/harp-materials/lib/ShaderChunks/ShadowChunks.ts
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const correctShadowChunk = `
```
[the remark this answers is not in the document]
nzjony: hi @zbigg , yes, the colors of the map are incorrect, at least from a design perspective, i.e. currently, setting a color to be some green, and adding a white light causes the light to be very light green, you can see it in the harp.gl demo:  https://www.harp.gl/docs/master/examples/#threejs_shadows.html

I will comment.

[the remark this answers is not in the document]
nzjony: i renamed to `simpleLightingShadowChunk` and added a comment.

@here/harp-materials/lib/MapMeshMaterials.ts
nzjony: These aren't needed in particular, however at least the MapMeshStandardMaterial class needs to have:
```
    get removeDiffuseLight(): boolean {
        // to be overriden
        return false;
    }
    set removeDiffuseLight(val: boolean) {
        // to be overriden
    }
```
Otherwise, if I remove them, then three.js will give a warning asking why the given parameter is passed into the material.

@here/harp-materials/lib/MapMeshMaterials.ts
nzjony: Agreed, this file is very hard to move about, I think however that all parts should be separated (i.e. the mixins separated from the materials).

@here/harp-materials/lib/MapMeshMaterials.ts
nzjony: Please see latest commit, I have added a comment to these properties.

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -1199,6 +1204,20 @@ export class MapMeshStandardMaterial extends THREE.MeshStandardMaterial
     set extrusionRatio(value: number) {
         // to be overridden
     }
+    /**
```
zbigg, line 1207: Not needed now

@here/harp-materials/lib/ShadowFeature.ts
```diff
@@ -0,0 +1,37 @@
...
+    removeDiffuseLight?: boolean;
+}
+
+/**
+ * Applies changes to THREE's MeshStandardMaterial which removes the diffuse component to ensure
+ * that colors are kept consistent.
+ */
+export namespace ShadowFeature {
+    export function onBeforeCompile(material: HiddenThreeJSMaterialProperties) {
+        material.onBeforeCompile = chainCallbacks(material.onBeforeCompile, shader => {
```
zbigg, line 30: Why not saving some bytes ?
This code can be just thrown into MapMeshStandardMaterial contstructor.

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -116,6 +121,7 @@ export function createMaterial(
     ) {
         settings.fog = options.fog;
     }
+    settings.removeDiffuseLight = technique.name === "fill";
```
zbigg, line 124: While extra param usually doesn't harm, i think it should be guarded with If `options.shadowsEnabled`.

@here/harp-materials/lib/ShadowFeature.ts
```diff
@@ -0,0 +1,37 @@
...
+import { chainCallbacks } from "@here/harp-utils";
+import * as THREE from "three";
+import { HiddenThreeJSMaterialProperties } from "./MapMeshMaterials";
+import { simpleLightingShadowChunk } from "./ShaderChunks/ShadowChunks";
+
+/**
+ * Parameter used to control patching the THREE.MeshStandardMaterial shader to ensure that the
+ * material's color isn't affected by the light direction.
+ */
+export interface ShadowFeatureParameters {
```
zbigg, line 16: W.r.t to removing "ShadingFeature"  i think i meand .... completly :)
I think now this interface is not needed at all.

All we need is just add this `removeDiffuseLight` to MapMeshMaterialParameters which should be named and look more or less like this:

```
interface MapMeshMaterialParameters extends THREE...., FirstMixin, SecondMixin, ... {
    removeDiffuseLight?: ...
}
```

Just to save some bytes and names.

@here/harp-examples/src/threejs_shadows.ts
```diff
@@ -189,22 +183,11 @@ const addOmvDataSource = (): Promise<void> => {
...
-
 /**
- * Replace all occurences of "fill" technique in the theme with "standard" technique.
- * "standard" technique is using three.js MeshStandardMaterial and is needed to receive
- * shadows.
+ * Replaces the lights in the style to cast shadows.
  * @param theme The theme to patch
  */
-const patchTheme = (theme: Theme) => {
+const fixLights = (theme: Theme) => {
```
zbigg, line 190: We don't have to patch loaded theme like this. We can simplify this by using builtin theme inheritance.

You can create right theme using `extends` clause right from beginning.

Just initialize MapView like thiscreate your theme like this:
```
const theme: Theme = { 
   extends: "resources/berlin_tilezen_base.json",
   lights: [... desired definitions of lights ]
}
initializeMapView(..., theme)
```

Much cleaners and encourages to use "preferred" methods for composing themes. 

Of course it wasn't completly possible in previous patch because we don't support "technique.name" rewrite.

@here/harp-materials/lib/ShadowFeature.ts
nzjony: Ok, will do.

@here/harp-mapview/lib/DecodedTileHelpers.ts
nzjony: Agreed.

@here/harp-materials/lib/ShadowFeature.ts
nzjony: Ok, will do.

@here/harp-examples/src/threejs_shadows.ts
nzjony: Thanks a lot! Good to know.

@here/harp-materials/lib/ShadowFeature.ts
nzjony: As discussed, will remove in favor of the approach of using the & { removeDiffuseLight }

@here/harp-materials/lib/MapMeshMaterials.ts
nzjony: as mentioned, it needs to stay unfortunately. I wasn't able to work around it.

PR #1431, opened 2020-04-01
Reduce the size of the index buffers.

robertoraggi:
Prefer to use uint16 buffers when possible.

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
+++ b/@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
@@ -137,6 +137,22 @@ const MAX_CORNER_ANGLE = Math.PI / 8;
  */
 const INVALID_ARRAY_INDEX = -1;
 
+function createIndexBufferAttribute(
+    elements: ArrayLike<number>,
+    maxValue: number,
+    name: string = "index"
+): BufferAttribute {
+    const type = maxValue > 65535 ? "uint32" : "uint16";
+    const storage = type === "uint32" ? new Uint32Array(elements) : new Uint16Array(elements);
+    const buffer = storage.buffer;
+    return {
+        itemCount: 1,
+        name,
+        buffer,
+        type
+    };
+}
+
 // for tilezen by default extrude all buildings even those without height data
 class MeshBuffers implements IMeshBuffers {
     readonly positions: number[] = [];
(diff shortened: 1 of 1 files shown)


@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -137,6 +137,38 @@ const MAX_CORNER_ANGLE = Math.PI / 8;
...
+
+function createIndexBuffer(elements: ArrayLike<number>) {
+    return maxValue(elements) > 65535 ? new Uint32Array(elements) : new Uint16Array(elements);
+}
+
+function createIndexBufferAttribute(
+    elements: ArrayLike<number>,
+    name: string = "index"
+): BufferAttribute {
+    const type = maxValue(elements) > 65535 ? "uint32" : "uint16";
```
nzjony, line 161: could you pass in an optional value which has the vertices.length (if available), if it is less than 65k, then you can avoid the maxValue function call?

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
```diff
@@ -137,6 +137,38 @@ const MAX_CORNER_ANGLE = Math.PI / 8;
  */
 const INVALID_ARRAY_INDEX = -1;
 
+function maxValue(array: ArrayLike<number>): number {
```
MPursche, line 140: Given that the index buffer does not contain indices that are out of range: the highest index can also be determined with `posAttr.count` which should be a bit faster than looping over the index buffer.

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
ninok, line 161: In native we even used 16 bit when exceeding 65K and did multiple draw calls with [glDrawElementsBaseVertex.](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDrawElementsBaseVertex.xhtml).
Not sure if sth. equivalent exists in WebGL.

robertoraggi: it's not vertexLength, it's the max value of the array.

MPursche: But shouldn't the highest value of the index buffer equal(or be smaller than) `posAttr.count - 1`?

nzjony: why isn't it vertextLength? (well actually vertextLength-1)

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
robertoraggi: ok

@here/harp-omv-datasource/lib/OmvDecodedTileEmitter.ts
MPursche: I think there is only this extension for WebGL 2: https://www.khronos.org/registry/webgl/extensions/WEBGL_draw_instanced_base_vertex_base_instance/. So not very likely that this is supported by many devices.

robertoraggi: sure, using vertex.length may be OK with the way we create geometries right now.

robertoraggi:
Retests this please

PR #1436, opened 2020-04-02
HARP-9664 Implement custom mip map generator

MPursche:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/image/MipMapGenerator.ts
+++ b/@here/harp-mapview/lib/image/MipMapGenerator.ts
@@ -1,0 +1,163 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as THREE from "three";
+
+import { ImageItem } from "./Image";
+
+const isNode = typeof window === "undefined";
+
+/**
+ * Mip map generator resizes textures to next bigger power-of-two size by adding padding
+ * and creates mip map levels.
+ * @internal
+ */
+export class MipMapGenerator {
+    /**
+     * Gets size of an image padded to the next bigger power-of-two size
+     * @param width Width of image
+     * @param height Height of image
+     */
+    static getPaddedSize(width: number, height: number): { width: number; height: number } {
+        return {
+            width: THREE.MathUtils.ceilPowerOfTwo(width),
+            height: THREE.MathUtils.ceilPowerOfTwo(height)
+        };
+    }
+
+    private m_paddingCanvas?: HTMLCanvasElement;
+    private m_paddingContext?: CanvasRenderingContext2D;
+    private m_resizeCanvas?: HTMLCanvasElement;
+    private m_resizeContext?: CanvasRenderingContext2D;
+
+    constructor() {
+        if (!isNode) {
+            this.m_paddingCanvas = document.createElement("canvas");
+            this.m_paddingContext = this.m_paddingCanvas.getContext("2d")!;
+            this.m_resizeCanvas = document.createElement("canvas");
+            this.m_resizeContext = this.m_resizeCanvas.getContext("2d")!;
+        }
+    }
+
+    /**
+     * Generate downsampled mip map levels from an image.
+     * If the input image is not power-of-two the image is padded to the
+     * next bigger power-of-two size.
+     * @param image Input image
+     * @returns A list of images with mip maps of the input image
+     */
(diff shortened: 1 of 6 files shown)


@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -119,13 +120,12 @@ class PoiRenderBufferBatch {
             trilinear ? THREE.LinearMipMapLinearFilter : THREE.NearestFilter,
             THREE.RGBAFormat
         );
-        texture.needsUpdate = true;
-        texture.premultiplyAlpha = premultipliedAlpha;
         // Generate mipmaps for distance scaling of icon
-        // TODO: Implement custom mip map generation if support for texture atlas
-        // with icons that are not power of two or have different sizes is necessary.
-        texture.generateMipmaps = true;
+        texture.mipmaps = mipMapGenerator.generateTextureAtlasMipMap(iconTexture.image);
```
ninok, line 124: I get the feeling this should all be optional. I.e. for plain top-down usecases you don't want to waste GPU resources and processing for the mip map generation. Especially b/c it's done in the main thread. Do you have any performance numbers comparing POIs w/o mipmaps vs. POIs w/ mipmaps. 
I know it's hard to get numbers with our perf. tests but maybe you can try locally using the benchmark example

nzjony: I agree, however do we have a means to enforce such a use case? If not, I think this is fine until such an API exists to have a top down use case, however.... if that is the use case, there are other map rendering engines which are much better in the 2D case, so it probably isn't worth doing.

MPursche: Even for top down views trilinear filtering/mipmaps can improve the image quality of the icons. The map theme allows to freely scale the size of the icon which can lead to some artifacts with just bilinear filtering enabled. And the memory overhead of mipmaps isn't that big only 1/3 more memory if the texture is already power-of-two. Slightly more if padding is added. Additionally, the current themes only use two relatively small textures for icons so the overall memory overhead should just be 1-2 MB in GPU memory. I will try locally how big the performance impact/memory overhead is.

MPursche: Just running the numbers of paper:
|Texture|size without mipmaps|size with mipmaps and padding|
| ------------- | ------------- | ------------- |
|icons_day.png|329×296×4 Byte = 389,54 kB|512x512x1,3333x4 Byte = 1,4 MB|
|city_center.png|38x38x4 Byte = 5,78 kB|64x64x1,3333x4 Byte = 21,84 kB|
|**Total**|395,32 kB|1,421 MB|

MPursche:
GPU tests please

@here/harp-mapview/lib/poi/PoiRenderer.ts
MPursche: I added links to the benchmarks in the JIRA ticket. Overall only very minor differences when it comes to performance and memory consumption.

MPursche:
GPU tests please.

MPursche:
GPU tests please

MPursche:
GPU tests please

@here/harp-mapview/lib/poi/Poi.ts
```diff
@@ -68,18 +68,18 @@ export interface ImageOptions {
...
-export class IconTexture {
-    /**
-     * The constructor of an `IconTexture`.
-     *
-     * @param image An [[ImageItem]] instance.
-     * @param options The rendering parameters.
-     */
-    constructor(public image: ImageItem) {}
-}
+// /**
```
nzjony, line 71: Please remove

@here/harp-mapview/lib/image/MipMapGenerator.ts
```diff
@@ -0,0 +1,159 @@
...
+                height - image.height
+            );
+        }
+
+        return this.m_paddingCanvas!;
+    }
+
+    /**
+     * Resize an image. Quality of resized image is best when
+     * width = image.width / 2 and height = image.height / 2.
```
nzjony, line 141: I guess you mean quality is best when it is image.width % 2 === 0 ?

MPursche: Yes, that and also that an image should be resized exactly by factor 0.5 or 2. Those are the only use cases where the canvas API uses "correct" down/upsampling on most browsers. Otherwise it uses a faster approximation that has a worse image quality.

nzjony: Yes!

PR #1437, opened 2020-04-02
HARP-8498: SolidLineMesh.

atomicsulfate:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/PickHandler.ts
+++ b/@here/harp-mapview/lib/PickHandler.ts
@@ -250,28 +250,39 @@ export class PickHandler {
         intersect: THREE.Intersection,
         pickResult: PickResult
     ) {
-        if (pickResult.intersection!.object instanceof MapViewPoints) {
-            pickResult.userData = featureData.objInfos![intersect.index!];
+        if (featureData.objInfos === undefined) {
             return;
-        } else if (
-            featureData.objInfos === undefined ||
+        }
+
+        if (pickResult.intersection!.object instanceof MapViewPoints) {
+            pickResult.userData = featureData.objInfos[intersect.index!];
+            return;
+        }
+
+        if (
             featureData.starts === undefined ||
-            intersect.faceIndex === undefined
+            featureData.starts.length === 0 ||
+            (intersect.faceIndex === undefined && intersect.index === undefined)
         ) {
             return;
         }
 
-        if (featureData.starts.length > 1) {
-            let objInfosIndex = 0;
-            for (const polygonStartFace of featureData.starts) {
-                if (polygonStartFace > intersect.faceIndex * 3) {
-                    break;
-                }
-                objInfosIndex++;
-            }
-            pickResult.userData = featureData.objInfos[objInfosIndex - 1];
-        } else {
+        if (featureData.starts.length === 1) {
             pickResult.userData = featureData.objInfos[0];
+            return;
         }
+
+        const intersectIndex =
+            intersect.faceIndex !== undefined ? intersect.faceIndex * 3 : intersect.index!;
+
+        // TODO: Implement binary search.
+        let objInfosIndex = 0;
+        for (const featureStartIndex of featureData.starts) {
+            if (featureStartIndex > intersectIndex) {
+                break;
+            }
+            objInfosIndex++;
+        }
+        pickResult.userData = featureData.objInfos[objInfosIndex - 1];
     }
 }
--- a/@here/harp-mapview/lib/geometry/DisplacedBufferGeometry.ts
+++ b/@here/harp-mapview/lib/geometry/DisplacedBufferGeometry.ts
@@ -7,9 +7,9 @@
 import * as THREE from "three";
 import { DisplacedBufferAttribute } from "./DisplacedBufferAttribute";
 
-const tmpNormalMin = new THREE.Vector3();
-const tmpNormalMax = new THREE.Vector3();
-const tmpBBox = new THREE.Box3();
+const tmpV1 = new THREE.Vector3();
+const tmpV2 = new THREE.Vector3();
+const tmpBox = new THREE.Box3();
 
 /**
  * @internal
@@ -18,6 +18,28 @@ const tmpBBox = new THREE.Box3();
 export interface DisplacementRange {
     min: number;
     max: number;
+}
+
+/**
+ * @internal
+ * Expands a box so that any point within the original box, displaced by any amount in a given
+ * range, is contained by the resulting box.
+ * @param box The original box to expand.
+ * @param displacementRange The range of possible displacement values for any point within the box.
+ * @param displacementDir Direction in which the displacement will be applied.
+ */
+export function expandBoxByDisplacementRange(
+    box: THREE.Box3,
+    displacementRange: DisplacementRange,
+    displacementDir: THREE.Vector3
+): THREE.Box3 {
+    tmpBox.copy(box);
+    const tmpNormalMin = tmpV1.copy(displacementDir);
+    const tmpNormalMax = tmpV2.copy(tmpNormalMin);
+    box.translate(tmpNormalMin.multiplyScalar(displacementRange.min)).union(
+        tmpBox.translate(tmpNormalMax.multiplyScalar(displacementRange.max))
+    );
+    return box;
 }
 
 /**
@@ -33,14 +55,14 @@ export class DisplacedBufferGeometry extends THREE.BufferGeometry {
      * Creates an instance of displaced buffer geometry.
      * @param originalGeometry The goeometry to be displaced.
      * @param displacementMap A texture with the displacement values.
-     * @param m_displacementRange The displacement value range found in the displacement map.
+     * @param displacementRange The displacement value range found in the displacement map.
      * @param displacedPositions Buffer attribute that will be used for displaced positions if
      * provided, otherwise a new buffer attribute will be created.
      */
     constructor(
         public originalGeometry: THREE.BufferGeometry,
         displacementMap: THREE.DataTexture,
-        private m_displacementRange: DisplacementRange,
+        public displacementRange: DisplacementRange,
         displacedPositions?: DisplacedBufferAttribute
     ) {
         super();
@@ -74,14 +96,15 @@ export class DisplacedBufferGeometry extends THREE.BufferGeometry {
         const uvs = geometry.attributes.uv;
         this.m_displacedPositions.reset(positions, normals, uvs, displacementMap);
         const displacementRangeChanged =
-            this.m_displacementRange.min !== displacementRange.min ||
-            this.m_displacementRange.max !== displacementRange.max;
-        this.m_displacementRange = displacementRange;
+            this.displacementRange.min !== displacementRange.min ||
+            this.displacementRange.max !== displacementRange.max;
+        this.displacementRange = displacementRange;
         this.resetAttributes();
         this.resetBoundingVolumes(displacementRangeChanged);
     }
 
     // HARP-9585: Override of base class method, however tslint doesn't recognize it as such.
+    // tslint:disable-next-line: explicit-override
     computeBoundingBox(): void {
         // Calculate a coarse approximation of the displaced geometry bbox by displacing the
         // original bbox and enlarging it to cover the whole displacement range.
@@ -96,15 +119,15 @@ export class DisplacedBufferGeometry extends THREE.BufferGeometry {
         } else {
             this.boundingBox.copy(origBBox);
         }
-        tmpBBox.copy(origBBox);
-        tmpNormalMin.fromBufferAttribute(this.attributes.normal as THREE.BufferAttribute, 0);
-        tmpNormalMax.copy(tmpNormalMin);
-        this.boundingBox
-            .translate(tmpNormalMin.multiplyScalar(this.m_displacementRange.min))
-            .union(tmpBBox.translate(tmpNormalMax.multiplyScalar(this.m_displacementRange.max)));
+        expandBoxByDisplacementRange(
+            this.boundingBox,
+            this.displacementRange,
+            tmpV1.fromBufferAttribute(this.attributes.normal as THREE.BufferAttribute, 0)
+        );
     }
 
     // HARP-9585: Override of base class method, however tslint doesn't recognize it as such.
+    // tslint:disable-next-line: explicit-override
     computeBoundingSphere(): void {
         // Use as coarse approximation the sphere bounding the bbox.
         if (this.boundingBox === null) {
--- a/@here/harp-mapview/lib/geometry/DisplacedBufferGeometry.ts
+++ b/@here/harp-mapview/lib/geometry/DisplacedBufferGeometry.ts
@@ -22,13 +22,13 @@ export interface DisplacementRange {
 
 /**
  * @internal
- * Expands a box so that any point within the original box, displaced by any amount in a given
- * range, is contained by the resulting box.
- * @param box The original box to expand.
- * @param displacementRange The range of possible displacement values for any point within the box.
+ * Displace a box in a given direction by a specified range. The original box min and max vertices
+ * are translated as a result by displacementRange.min and displacementRange.max respectively.
+ * @param box The original box to displace.
+ * @param displacementRange The minimum and maximum displacement values.
  * @param displacementDir Direction in which the displacement will be applied.
  */
-export function expandBoxByDisplacementRange(
+export function displaceBox(
     box: THREE.Box3,
     displacementRange: DisplacementRange,
     displacementDir: THREE.Vector3
@@ -119,7 +119,7 @@ export class DisplacedBufferGeometry extends THREE.BufferGeometry {
         } else {
             this.boundingBox.copy(origBBox);
         }
-        expandBoxByDisplacementRange(
+        displaceBox(
             this.boundingBox,
             this.displacementRange,
             tmpV1.fromBufferAttribute(this.attributes.normal as THREE.BufferAttribute, 0)
--- a/@here/harp-mapview/lib/geometry/DisplacedBufferGeometry.ts
+++ b/@here/harp-mapview/lib/geometry/DisplacedBufferGeometry.ts
@@ -27,6 +27,7 @@ export interface DisplacementRange {
  * @param box The original box to displace.
  * @param displacementRange The minimum and maximum displacement values.
  * @param displacementDir Direction in which the displacement will be applied.
+ * @return The displaced box.
  */
 export function displaceBox(
     box: THREE.Box3,
--- a/@here/harp-mapview/lib/geometry/DisplacedMesh.ts
+++ b/@here/harp-mapview/lib/geometry/DisplacedMesh.ts
@@ -10,14 +10,6 @@ import * as THREE from "three";
 import { DisplacedBufferAttribute } from "./DisplacedBufferAttribute";
 import { DisplacedBufferGeometry, DisplacementRange } from "./DisplacedBufferGeometry";
 
-function isBufferGeometry(
-    geometry: THREE.Geometry | THREE.BufferGeometry
-): geometry is THREE.BufferGeometry {
-    const isBufferGeom = geometry instanceof THREE.BufferGeometry;
-    assert(isBufferGeom, "Unsupported geometry type.");
-    return isBufferGeom;
-}
-
 function isDisplacementMaterial(material: any): material is DisplacementFeature {
     const isDisplacementFeature = hasDisplacementFeature(material);
     assert(isDisplacementFeature, "Material does not support displacement maps.");
@@ -66,42 +58,69 @@ export class DisplacedMesh extends THREE.Mesh {
     }
 
     m_displacedGeometry?: DisplacedBufferGeometry;
+    private m_displacementMaterial?: DisplacementFeature;
 
     /**
      * Creates an instance of displaced mesh.
+     * @param geometry Original geometry to displace.
+     * @param material Material(s) to be used by the mesh. All must have the same displacement map.
      * @param m_getDisplacementRange Displacement values range getter.
-     * @param [geometry] Original geometry to displace.
-     * @param [material] Material(s) to be used by the mesh. All must have the same displacement
-     * map.
+     * @param [m_raycastStrategy] Function that will be used to find ray intersections. If not
+     * provided, THREE.Mesh's raycast will be used.
      */
     constructor(
+        geometry: THREE.BufferGeometry,
+        material: THREE.Material | THREE.Material[],
         private m_getDisplacementRange: () => DisplacementRange,
-        geometry?: THREE.Geometry | THREE.BufferGeometry,
-        material?: THREE.Material | THREE.Material[]
+        private m_raycastStrategy?: (
+            mesh: THREE.Mesh,
+            raycaster: THREE.Raycaster,
+            intersects: THREE.Intersection[]
+        ) => void
     ) {
         super(geometry, material);
-        assert(!this.geometry || isBufferGeometry(this.geometry));
-        assert(!this.material || isDisplacementMaterial(this.firstMaterial));
     }
 
-    raycast(raycaster: THREE.Raycaster, intersects: THREE.Intersection[]): void {
-        if (!this.geometry || !this.material) {
-            return;
+    // HARP-9585: Override of base class method, however tslint doesn't recognize it as such.
+    // tslint:disable-next-line: explicit-override
+    get geometry(): THREE.BufferGeometry {
+        return super.geometry as THREE.BufferGeometry;
+    }
+
+    // HARP-9585: Override of base class method, however tslint doesn't recognize it as such.
+    // tslint:disable-next-line: explicit-override
+    set geometry(geometry: THREE.BufferGeometry) {
+        super.geometry = geometry;
+    }
+
+    // HARP-9585: Override of base class method, however tslint doesn't recognize it as such.
+    // tslint:disable-next-line: explicit-override
+    get material(): THREE.Material | THREE.Material[] {
+        return super.material;
+    }
+
+    // HARP-9585: Override of base class method, however tslint doesn't recognize it as such.
+    // tslint:disable-next-line: explicit-override
+    set material(material: THREE.Material | THREE.Material[]) {
+        super.material = material;
+        const firstMaterial = this.firstMaterial;
+        if (isDisplacementMaterial(firstMaterial)) {
+            this.m_displacementMaterial = firstMaterial;
         }
+    }
 
+    // HARP-9585: Override of base class method, however tslint doesn't recognize it as such.
+    // tslint:disable-next-line: explicit-override
+    raycast(raycaster: THREE.Raycaster, intersects: THREE.Intersection[]): void {
         // All materials in the object are expected to have the same displacement map.
-        const material = this.firstMaterial;
+        const displacementMap = this.m_displacementMaterial!.displacementMap;
 
-        // Use default raycasting implementation if some type is unexpected.
-        if (
-            !isBufferGeometry(this.geometry) ||
-            !isDisplacementMaterial(material) ||
-            !isDataTextureMap(material.displacementMap)
-        ) {
+        // Use default raycasting implementation if there's no displacement map or its type is not
+        // supported.
+        if (!isDataTextureMap(displacementMap)) {
             super.raycast(raycaster, intersects);
             return;
         }
-        const displacementMap = material.displacementMap;
         const displacementRange = this.m_getDisplacementRange();
 
         if (this.m_displacedGeometry) {
@@ -117,8 +136,12 @@ export class DisplacedMesh extends THREE.Mesh {
 
         // Replace the original geometry by the displaced one only during the intersection test.
         this.geometry = this.m_displacedGeometry;
-        super.raycast(raycaster, intersects);
-        this.geometry = this.m_displacedGeometry.originalGeometry;
+        if (this.m_raycastStrategy) {
+            this.m_raycastStrategy(this, raycaster, intersects);
+        } else {
+            super.raycast(raycaster, intersects);
+        }
+        super.geometry = this.m_displacedGeometry.originalGeometry;
     }
 
     private get firstMaterial(): THREE.Material {
--- a/@here/harp-mapview/lib/geometry/DisplacedMesh.ts
+++ b/@here/harp-mapview/lib/geometry/DisplacedMesh.ts
@@ -96,7 +96,7 @@ export class DisplacedMesh extends THREE.Mesh {
             return;
         }
         const displacementMap = firstMaterial.displacementMap;
-        const displacementRange = this.m_getDisplacementRange();
+        const displacementRange = { ...this.m_getDisplacementRange() };
 
         assert(this.geometry instanceof THREE.BufferGeometry, "Unsupported geometry type.");
         const geometry = this.geometry as THREE.BufferGeometry;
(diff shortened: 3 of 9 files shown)


atomicsulfate:

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+    hWidthSq: number,
+    plane: THREE.Plane,
+    interPlane: THREE.Vector3,
+    interLine: THREE.Vector3
+): boolean {
+    if (interPlane.equals(tmpRay.origin) && tmpRay.direction.dot(plane.normal) === 0) {
+        // Corner case: ray is coplanar to extruded line, find distance to extruded line sides
+        // and end caps.
+        const extrLineT = intersectExtrudedLine(ray, line, vExtrusion, plane.normal, hWidth);
+        const endCapT = intersectClosestEndCap(ray, line, hWidth);
```
kkostecki, line 181: There is a commonly used geometry type for such cases - Capsule. Maybe we should introduce this type of geometry too for further purposes (as follow up, or just write a comment), intersection algorithm is quite simple though, see these:
http://www.iquilezles.org/www/articles/intersectors/intersectors.htm
https://www.shadertoy.com/view/Xt3SzX

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+        ? 0
+        : ray.intersectSphere(sphere, tmpV4)
+        ? tmpV4.sub(ray.origin).length()
+        : Infinity;
+    return Math.min(startCapT, endCapT);
+}
+
+/**
+ * Intersects line
+ * @param ray Intersection ray in object local space.
```
kkostecki, line 157: object's

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+ * @param line The line to intersect.
+ * @param vExtrusion Line extrusion vector.
+ * @param hWidth The line's extrusion half width.
+ * @param hWidthSq The line's extrusion half width squared.
+ * @param plane The extrusion plane.
+ * @param interPlane The intersection of the ray with the extrusion plane.
+ * @param interLine The ray intersetion with the extruded line.
```
kkostecki, line 164: Please explain are they input/output params. How can I distinguish which of the params is actually computed?

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+    }
+    interLine.copy(interPlane);
+    return true;
+}
+
+/**
+ * Finds the intersections of a ray with a partition of a solid line mesh representing a feature.
+ * @param mesh The mesh whose intersections will be found.
+ * @param raycaster Contains the intersection ray.
+ * @param localRay Same ray as raycaster.ray but in object local space.
```
kkostecki, line 206: object's

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+ * @param intersects Array that will contain all intersections found between ray and feature.
```
kkostecki, line 207: @param intersections

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+    for (let i = beginIdx; i < endIdx; i += SEGMENT_STRIDE) {
+        const a = indices[i];
+        const b = indices[i + VERTEX_STRIDE];
+
+        // Find the plane containing the line segment, using the segment start, end and extrusion
+        // vector.
+        line.start.fromBufferAttribute(position, a);
+        line.end.fromBufferAttribute(position, b);
+        vExt.set(bitangent.getX(a), bitangent.getY(a), bitangent.getZ(a)).normalize();
+        plane.setFromCoplanarPoints(line.start, tmpV3.copy(line.start).add(vExt), line.end);
```
kkostecki, line 256: Plane equation requires simply its origin (line.start) and normal (attributes.normal). No need for tree points to locate and orient plane in 3d space. You may simply use:
```
plane.setFromNormalAndCoplanarPoint()
```

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+    ray: THREE.Ray,
+    line: THREE.Line3,
+    vExtrusion: THREE.Vector3,
+    hWidth: number,
+    hWidthSq: number,
+    plane: THREE.Plane,
+    interPlane: THREE.Vector3,
+    interLine: THREE.Vector3
+): boolean {
+    if (interPlane.equals(tmpRay.origin) && tmpRay.direction.dot(plane.normal) === 0) {
```
kkostecki, line 177: Using temp variables are an arguments here is quite fragile - where do you set it? - I don't see. Moreover we are dealing with floating point numbers, so dot product may be very sensitive to precision issues.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+        const b = indices[i + VERTEX_STRIDE];
+
+        // Find the plane containing the line segment, using the segment start, end and extrusion
+        // vector.
+        line.start.fromBufferAttribute(position, a);
+        line.end.fromBufferAttribute(position, b);
+        vExt.set(bitangent.getX(a), bitangent.getY(a), bitangent.getZ(a)).normalize();
+        plane.setFromCoplanarPoints(line.start, tmpV3.copy(line.start).add(vExt), line.end);
+
+        // The ray intersection if any, will be on the extrusion plane.
```
kkostecki, line 258: Based on that ... maybe we should not even use current intersectLine implementation which is more or less based on cylinder + caps (Capsule). If you want to search only point on the plane co-planar with ground / line surface then I would really simply entire test to check for plane -> ray intersection and then check intersection point distance from the line - this should be no bigger then halfWidth.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+    firstFeatureIdx: number,
+    groupEndIdx: number
+): number {
+    const bVolumes = mesh.userData.feature.boundingVolumes;
+    const geometry = mesh.geometry as THREE.BufferGeometry;
+    assert(isSolidLineMaterial(material), "Unsupported material type");
+    const solidLineMaterial = material as SolidLineMaterial;
+
+    const halfWidth = (solidLineMaterial.lineWidth + solidLineMaterial.outlineWidth) / 2;
+    const localHalfWidth = halfWidth / ((mesh.scale.x + mesh.scale.y + mesh.scale.z) / 3);
```
kkostecki, line 325: Shouldn't it be?
```
const localHalfWidth = halfWidth / mesh.scale.length()
```

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+
+const singleFeatureStart = [0];
+
+/**
+ * Finds the intersections of a ray with a group within a solid line mesh.
+ * @param mesh The mesh whose intersections will be found.
+ * @param material The material used by the group inside the mesh.
+ * @param raycaster  Contains the intersection ray.
+ * @param localRay Same ray as raycaster.ray but in object local space.
+ * @param intersects  Array that will contain all intersections found between ray and feature.
```
kkostecki, line 305: @param intersections
Move output params to the end of list and note they are output.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+ * @param beginIdx The index where the feature starts in the mesh geometry's indices attribute.
+ * @param endIdx The index where the feature end in the mesh geometry's indices attribute.
+ * @param bSphere The feature bounding sphere.
+ * @returns feature
+ */
+function intersectFeature(
+    mesh: THREE.Mesh,
+    raycaster: THREE.Raycaster,
+    localRay: THREE.Ray,
+    intersects: THREE.Intersection[],
```
kkostecki, line 220: Better to move output params to the end... and note in documentation that this is output.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+ * @param firstFeatureIdx Index of the first feature in the group.
+ * @param groupEndIdx Index of the last vertex in the group.
+ * @returns The next feature index after the group.
+ */
+function intersectGroup(
+    mesh: THREE.Mesh,
+    material: THREE.Material,
+    raycaster: THREE.Raycaster,
+    localRay: THREE.Ray,
+    intersects: THREE.Intersection[],
```
kkostecki, line 315: Same as above... output should be at the end and described (mostly done) as output.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+                    groupEndIdx
+                );
+            }
+        } else {
+            intersectGroup(mesh, mesh.material, raycaster, localRay, intersects, 0, indices.length);
+        }
+    }
+
+    // HARP-9585: Override of base class method, however tslint doesn't recognize it as such.
+    // tslint:disable-next-line: explicit-override
```
kkostecki, line 412: It is very dangerous to override the getters & setters in TS please consult with @zbigg  for insights, maybe he should also review just this part.

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -250,28 +250,39 @@ export class PickHandler {
...
-            }
-            pickResult.userData = featureData.objInfos[objInfosIndex - 1];
-        } else {
+        if (featureData.starts.length === 1) {
             pickResult.userData = featureData.objInfos[0];
+            return;
+        }
+
+        const intersectIndex =
+            intersect.faceIndex !== undefined ? intersect.faceIndex * 3 : intersect.index!;
```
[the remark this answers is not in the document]
atomicsulfate: We use a face index if given, otherwise we expect a vertex index. It can only get simpler if we always use the vertex index, but can't be sure it'll be provided since it's an optional member of  THREE.Intersection, so we need to check also faceIndex to be sure.

I don't see how checking the pick type would improve this. The PickObjectType doesn't necessarily tell you anything about the picked object. If the type is Line, the object can be represented as a Line or as a Mesh.

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -250,28 +250,39 @@ export class PickHandler {
...
+
+        // TODO: Implement binary search.
+        let objInfosIndex = 0;
+        for (const featureStartIndex of featureData.starts) {
+            if (featureStartIndex > intersectIndex) {
```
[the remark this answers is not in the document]
atomicsulfate: > This requires the feature data (TileFeatureData.starts array) to be sorted in ascending order is requirement always fulfilled?
Our decoders fulfill this requirement yes. In general these feature starts must be ordered otherwise they're useless. Every feature has multiple faces/vertices. Feature i starts at starts[i] and ends at starts[i+1]-1. If they weren't ordered you wouldn't know where a feature ends.
Anyway, I'll add a comment in Geometry.featureStarts and TileFeatureData.starts to make this clear.

> Anyway I see a lot of danger in the search function: try to find next face.. pretend that previous was the exact choice...
As I said, that's exactly how feature starts are arranged, the beginning of one feature marks the end of the previous one.

> Shouldn't it be:
> 
> ```
> const objInfoIndex = featureData.starts!.findIndex(faceVertexIdx => {
>             return faceVertexIdx === intersect.faceIndex! * 3;
>         });
> ```
No, that doesn't work, the vertex you're searching for might not be the start of any feature, but be somewhere in the middle.

@here/harp-mapview/lib/geometry/DisplacedBufferGeometry.ts
```diff
@@ -7,9 +7,9 @@
 import * as THREE from "three";
 import { DisplacedBufferAttribute } from "./DisplacedBufferAttribute";
 
-const tmpNormalMin = new THREE.Vector3();
-const tmpNormalMax = new THREE.Vector3();
-const tmpBBox = new THREE.Box3();
+const tmpV1 = new THREE.Vector3();
+const tmpV2 = new THREE.Vector3();
+const tmpBox = new THREE.Box3();
```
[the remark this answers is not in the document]
atomicsulfate: Please, if you want us to use that solution as a standard, suggest adding a rule in our [coding guidelines](https://github.com/heremaps/harp.gl/blob/master/CODINGSTYLE.md). The actual 'de facto' standard that I've seen used in our code is declaring separate tmp variables as I did here.
I have mixed feelings about that cache you suggest. It encourages extending the number of temporary variables indefinitely and we abuse them already.

@here/harp-mapview/lib/geometry/DisplacedBufferGeometry.ts
```diff
@@ -20,6 +20,28 @@ export interface DisplacementRange {
     max: number;
 }
 
+/**
+ * @internal
+ * Expands a box so that any point within the original box, displaced by any amount in a given
+ * range, is contained by the resulting box.
```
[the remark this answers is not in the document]
atomicsulfate: True. Done. Let me know if it's better now.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
```diff
@@ -66,42 +58,69 @@ export class DisplacedMesh extends THREE.Mesh {
     }
 
     m_displacedGeometry?: DisplacedBufferGeometry;
```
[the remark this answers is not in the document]
atomicsulfate: I'm actually needing this to be public for some changes I'm working on right now. So I'll remove the prefix.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
```diff
@@ -66,42 +58,69 @@ export class DisplacedMesh extends THREE.Mesh {
...
+    // tslint:disable-next-line: explicit-override
+    raycast(raycaster: THREE.Raycaster, intersects: THREE.Intersection[]): void {
         // All materials in the object are expected to have the same displacement map.
-        const material = this.firstMaterial;
-
-        // Use default raycasting implementation if some type is unexpected.
-        if (
-            !isBufferGeometry(this.geometry) ||
-            !isDisplacementMaterial(material) ||
-            !isDataTextureMap(material.displacementMap)
```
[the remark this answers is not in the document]
atomicsulfate: I've done changes that make checking those contraints here unnecessary:

- isBufferGeometry() is not needed anymore since now we only take THREE.BufferGeometry as parameter in the constructor and setter.
- isDisplacementMaterial() is checked in the constructor and setter, so no need to check it every time raycast is called.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+const tmpSphere = new THREE.Sphere();
+const tmpInverseMatrix = new THREE.Matrix4();
+const tmpRay = new THREE.Ray();
+const tmpLine1 = new THREE.Line3();
+const tmpBox = new THREE.Box3();
+const tmpPlane = new THREE.Plane();
+const tmpV1 = new THREE.Vector3();
+const tmpV2 = new THREE.Vector3();
+const tmpV3 = new THREE.Vector3();
+const tmpV4 = new THREE.Vector3();
```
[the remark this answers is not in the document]
atomicsulfate: Please check my reply in your previous comments.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+
+    const normal = tmpV2;
+    if (displacementRange) {
+        // If geometry is displaced, expand the bounding box to cover the whole displacement range,
+        // and return the sphere bounding the box. This is a coarse estimation, but avoids having
+        // to displace all vertices.
+        return expandBoxByDisplacementRange(
+            bbox,
+            displacementRange,
+            normal.fromBufferAttribute(geometry.attributes.normal as THREE.BufferAttribute, 0)
```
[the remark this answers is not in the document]
atomicsulfate: True. For sphere projection the normals are not equal. This is assuming though the they're similar enough to not matter in practice. 
But it's true that the resulting bbox can be potentially wrong for big enough objects.

To correct this, we can expand the box in the two directions corresponding to the min and max normals in the geometry. Computing these directions would imply though going through all normals and I'd like to avoid that if possible. Another option would be to pass to the mesh constructor the min/max normals corresponding to the tile where the object is.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+    const center = bbox.getCenter(sphere.center);
+    let maxRSq = 0;
+    for (let i = featureBeginIndex; i < featureEndIndex; i += SEGMENT_STRIDE) {
+        maxRSq = Math.max(
+            maxRSq,
+            center.distanceToSquared(vertex.fromBufferAttribute(pos, indices[i])),
+            center.distanceToSquared(vertex.fromBufferAttribute(pos, indices[i + VERTEX_STRIDE]))
+        );
+    }
+    sphere.radius = Math.sqrt(maxRSq);
```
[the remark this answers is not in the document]
atomicsulfate: I don't understand what you want to do here then. Do you say we should just take the sphere where the cube is inscribed instead of trying to find a smaller one?

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,451 @@
...
+    );
+    const obb = new OrientedBox3(
+        line.getCenter(tmpV4),
+        matrix,
+        new THREE.Vector3(line.distance() / 2, hWidth, hWidth)
+    );
+    if (obb.contains(ray.origin)) {
+        return 0;
+    }
+    return obb.intersectsRay(ray) ?? Infinity;
```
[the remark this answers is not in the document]
atomicsulfate: done.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: done

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: > Using temp variables are an arguments here is quite fragile - where do you set it? - I don't see. 

done, it's actually the ray passed as parameter.

>Moreover we are dealing with floating point numbers, so dot product may be very sensitive to precision issues.

If the plane intersection is equal to the ray origin it can be only for two reasons:
- The ray is not coplanar but the origin is actually on the plane.
- The ray is coplanar and ray.intersectPlane() that I called before actually returned the ray's origin as the plane intersection. See here the actual THREE.Ray.distanceToPlane code which is called from ray.intersectPlane(), it only sets the distance to 0 (and therefore the intersection is at ray Origin) if the dot product is exactly 0:

`
distanceToPlane: function ( plane ) {

		var denominator = plane.normal.dot( this.direction );

		if ( denominator === 0 ) {

			// line is coplanar, return origin
			if ( plane.distanceToPoint( this.origin ) === 0 ) {

				return 0;

			}
`
In practice this means that we will almost always consider the picking ray as not coplanar, but that's fine. We don't need to be super precise here. What's the problem if you try to pick a extruded line with a ray coplanar to the extrusion plane (I mean, who tries to do this in the first place) and we fail to detect it's coplanar and miss the intersection? You can try to pick again.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: We don't really need to implement a capsule intersection here, with a sphere is enough, that's what I use in intersectClosestEndCap. If the ray intersects first the half of the sphere that intersects the extruded center line is fine, that means that extrLineT === endCapT.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: done

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: done.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: done

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: done.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: That's exactly what I'm doing in case the ray is not coplanar to the extrusion plane. The cylinder+caps test is only used when the ray is coplanar to the extrusion plane.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: done

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: done.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: That's wrong, just see what would happen with no scaling (scale = [1,1,1]), the width would be divided by sqrt(3).

I took the formula above from THREE.js line picking. The result is only correct for uniform scaling (scale.x === scale.y === scale.z).  For non-uniform scaling we would need to compute the localHalfWidth depending on what direction is measured which would require more computations.
This here is an approximation assuming the scaling factor difference on each dimension is small enough. It holds now since we never scale at all (uniformly or not).
I'll add some clarification and an assert to catch any issue in the future.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: done, I checked with him.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: The surface normal and the extrusion plane normals are not the same, for example, when you have elevation.
If you imagine that the line end is elevated higher than the line start, the extrusion plane won't be parallel to the surface, it'll be tilted.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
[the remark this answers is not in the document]
atomicsulfate: I've been checking how bad can the approximation error be for different zoom levels by computing the dot product of the normals at the min max points of each tile. The results show that the error is very small starting from zoom level 4. See here some results:

TILE: (6,9,4), DOT: 0.9806892129880023
TILE: (12,17,5), DOT: 0.9946739445457075
TILE: (25,34,6), DOT: 0.9986326302953471
TILE: (50,68,7), DOT: 0.9996583822992287
TILE: (101,137,8), DOT: 0.9999152958483241
TILE: (202,274,9), DOT: 0.9999788289478592
TILE: (405,549,10), DOT: 0.9999947181695241
TILE: (810,1099,11), DOT: 0.9999986814690914
TILE: (1620,2199,12), DOT: 0.9999996706085572
...

Given this, I think the approximation is pretty valid. Trying to make it more accurate would make bbox computation slower unnecessarily.

@here/harp-mapview/test/SolidLineMeshTest.ts
```diff
@@ -0,0 +1,517 @@
...
+        meshBuilder.geometryBuilder.withUv([0, 0, 0, 0]).withNormal([0, 0, 1, 0, 0, 1]);
+        const mesh = meshBuilder.build();
+        mesh.geometry = new DisplacedBufferGeometry(
+            mesh.geometry as THREE.BufferGeometry,
+            new THREE.DataTexture(new Float32Array([0]), 1, 1),
+            { min: 10, max: 12 }
+        );
+        const intersects: THREE.Intersection[] = [];
+        const { raycaster } = buildRayTo([0, 1, 0]);
+        mesh.raycast(raycaster, intersects);
```
kkostecki, line 508: Could we check for intersection with displaced mesh, I haven't found any test case for that or I missed something?

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,439 @@
...
+    ray: THREE.Ray,
+    line: THREE.Line3,
+    vExtrusion: THREE.Vector3,
+    hWidth: number,
+    hWidthSq: number,
+    plane: THREE.Plane,
+    interPlane: THREE.Vector3,
+    outInterLine: THREE.Vector3
+): boolean {
+    if (interPlane.equals(ray.origin) && ray.direction.dot(plane.normal) === 0) {
```
kkostecki, line 177: Minor. I am little bit worried about lack of epsilon here, hitting exact ZERO may be hard to impossible in real life examples.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
kkostecki: Yeah, of course I saw it, but it simply complicates entire logic. Whatever... if tested let's leave it like that...

@here/harp-mapview/lib/geometry/DisplacedBufferGeometry.ts
```diff
@@ -20,6 +20,28 @@ export interface DisplacementRange {
...
 }
 
+/**
+ * @internal
+ * Displace a box in a given direction by a specified range. The original box min and max vertices
+ * are translated as a result by displacementRange.min and displacementRange.max respectively.
+ * @param box The original box to displace.
+ * @param displacementRange The minimum and maximum displacement values.
+ * @param displacementDir Direction in which the displacement will be applied.
+ */
```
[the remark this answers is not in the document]
atomicsulfate: done.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    if (displacementRange) {
+        // If geometry is displaced, expand the bounding box to cover the whole displacement range,
+        // and return the sphere bounding the box. This is a coarse estimation, but avoids having
+        // to displace all vertices.
+        // All normals in the geometry are assumed to be the same or close enough so that any of
+        // them can be used as displacement direction. For sphere projection, the surface normals
+        // within a tile are approximately the same from level 4 onwards. Here are some examples of
+        // the maximum dot product between normals in different tiles:
```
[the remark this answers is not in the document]
atomicsulfate: That's an error in my comment, I'll fix it. The sample values are the *minimum* dot products between normals in a tile (i.e. the normals with maximum angles between them).
There's no combination needed. The maximum angle between surface normals in a sphere's tile is the angle between the normal at the tile's bbox min and max corners.
I just added a log in the frustum intersection code were we compute the tile bboxes.

Given those numbers, there not much improvement in computing normal averages, and we want to keep the bbox test as fast as possible.

@here/harp-mapview/test/SolidLineMeshTest.ts
```diff
@@ -0,0 +1,517 @@
...
+}
+
+function checkIntersect(
+    actualIntersects: THREE.Intersection[],
+    expectedIntersect: THREE.Intersection
+) {
+    expect(actualIntersects).has.lengthOf(1);
+    const actualIntersect = actualIntersects[0];
+    expect(actualIntersect.distance).closeTo(expectedIntersect.distance, 1e-5);
+    expect(actualIntersect.index).deep.equals(expectedIntersect.index);
```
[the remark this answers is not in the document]
atomicsulfate: done

@here/harp-mapview/test/SolidLineMeshTest.ts
```diff
@@ -0,0 +1,517 @@
...
+        }
+        return { positions, bitangents };
+    }
+
+    private static triangulateLine(polyline: number[], indexOffset: number): number[] {
+        // triangulate lines.
+        const indices = [];
+        for (let i = 0; i < polyline.length / 3 - 1; ++i) {
+            const base = indexOffset + i * 4;
+            indices.push(base, base + 1, base + 2, base + 2, base + 1, base + 3);
```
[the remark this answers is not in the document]
atomicsulfate: The vertices in your diagram are in the wrong places. base and base+1 are the vertices corresponding to line start extruded in +/- direction respectively. base+2 and base+3 correspond to +/- extrusion of line end. 

// line start
// 1 ____0
// |  \     |
// |    \   |
// |____\ |
// 3       2
 // line end

Added the clarification in the code.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: attributes.normal are surface normals. See `createLineGeometry` in Lines.ts.

@here/harp-mapview/lib/geometry/DisplacedMesh.ts
```diff
@@ -65,60 +57,68 @@ export class DisplacedMesh extends THREE.Mesh {
...
-        const displacementMap = material.displacementMap;
+        const displacementMap = firstMaterial.displacementMap;
         const displacementRange = this.m_getDisplacementRange();
 
-        if (this.m_displacedGeometry) {
-            this.m_displacedGeometry.reset(this.geometry, displacementMap, displacementRange);
+        assert(this.geometry instanceof THREE.BufferGeometry, "Unsupported geometry type.");
+        const geometry = this.geometry as THREE.BufferGeometry;
+        if (this.displacedGeometry) {
+            this.displacedGeometry.reset(geometry, displacementMap, displacementRange);
```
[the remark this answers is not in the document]
atomicsulfate: done.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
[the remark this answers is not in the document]
atomicsulfate: Ok, i get it now. I'll remove the radius fitting step and use the box extent instead. 

> What I am trying to say that the approach used is neither accurate, nor fast. I simply see no point of calculating it, especially taking into account that in the same PR you are using `box.getBoundingSphere(sphere)` - very very rough approximation.

That very rough approximation is done when geometry is displaced. In that case you can't find a tight bounding volume unless you actually displace the whole geometry, vertex by vertex, that's what i try to avoid.

@here/harp-mapview/test/SolidLineMeshTest.ts
atomicsulfate: done.

@here/harp-mapview/lib/geometry/SolidLineMesh.ts
atomicsulfate: I think we better discuss this on a call.

atomicsulfate:

PR #1441, opened 2020-04-06
HARP-6488: Introduce Labels' LayoutState for storing alternative anchor.

kkostecki:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/LayoutState.ts
+++ b/@here/harp-mapview/lib/text/LayoutState.ts
@@ -1,0 +1,121 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+    DefaultTextStyle,
+    HorizontalAlignment,
+    TextLayoutStyle,
+    VerticalAlignment
+} from "@here/harp-text-canvas";
+
+/**
+ * Version of layout style stored.
+ */
+export enum LayoutUsed {
+    Undefined = 0,
+    ThemeBased,
+    Alternative
+}
+
+/**
+ * Defines possible text placement relative to anchor.
+ */
+export interface AnchorPlacement {
+    h: HorizontalAlignment;
+    v: VerticalAlignment;
+}
+
+/**
+ * Layout state of the text part of the `TextElement`.
+ *
+ * Used mainly for multi-anchor placement algorithm.
+ * @hidden
+ */
+export class LayoutState {
+    private m_state = LayoutUsed.Undefined;
+    private m_hAlign = DefaultTextStyle.DEFAULT_HORIZONTAL_ALIGNMENT;
+    private m_vAlign = DefaultTextStyle.DEFAULT_VERTICAL_ALIGNMENT;
+
+    /**
+     * Set layout based on theme style defined.
+     *
+     * @param layoutStyle The original theme label style.
+     */
+    setFromBase(layoutStyle: TextLayoutStyle) {
+        this.m_hAlign = layoutStyle.horizontalAlignment;
+        this.m_vAlign = layoutStyle.verticalAlignment;
+        this.m_state = LayoutUsed.ThemeBased;
+    }
+
+    /**
+     * Override label style using alternative anchor placement.
+     *
+     * @param placement The new anchor placement.
+     */
+    setFromPlacement(placement: AnchorPlacement) {
+        this.m_hAlign = placement.h;
+        this.m_vAlign = placement.v;
+        this.m_state = LayoutUsed.Alternative;
+    }
+
+    /**
+     * Reset existing `LayoutState` to appear like a fresh state.
+     */
+    reset() {
+        this.m_state = LayoutUsed.Undefined;
+    }
+
+    /**
+     * @returns `true` if element has not yet been placed and layout variant assigned.
+     */
+    isUndefined(): boolean {
+        return this.m_state === LayoutUsed.Undefined;
+    }
+
+    /**
+     * @returns `true` if element is using original style defined in theme.
+     */
+    isBase(): boolean {
+        return this.m_state === LayoutUsed.ThemeBased;
+    }
+
+    /**
+     * @returns `true` if element is fading in.
+     */
+    isAlternative(): boolean {
+        return this.m_state === LayoutUsed.Alternative;
+    }
+
+    get horizontalAlignment(): HorizontalAlignment {
+        return this.m_hAlign;
+    }
+
+    get verticalAlignment(): VerticalAlignment {
+        return this.m_vAlign;
+    }
+
+    /**
+     * Clone this `LayoutState`.
+     *
+     * @returns Cloned `LayoutState`.
+     */
+    clone(): LayoutState {
+        const res = new LayoutState();
+        res.copy(this);
+        return res;
+    }
+
+    /**
+     * Copies the values of `other` to this [[LayoutState]].
+     *
+     * @param other The other [[LayoutState]] to copy.
+     */
+    copy(other: LayoutState) {
+        this.m_hAlign = other.m_hAlign;
+        this.m_vAlign = other.m_vAlign;
+        this.m_state = other.m_state;
+    }
+}
--- a/@here/harp-mapview/lib/text/LayoutState.ts
+++ b/@here/harp-mapview/lib/text/LayoutState.ts
@@ -12,15 +12,6 @@ import {
 } from "@here/harp-text-canvas";
 
 /**
- * Version of layout style stored.
- */
-export enum LayoutUsed {
-    Undefined = 0,
-    ThemeBased,
-    Alternative
-}
-
-/**
  * Defines possible text placement relative to anchor.
  */
 export interface AnchorPlacement {
@@ -35,30 +26,20 @@ export interface AnchorPlacement {
  * @hidden
  */
 export class LayoutState {
-    private m_state = LayoutUsed.Undefined;
     private m_hAlign = DefaultTextStyle.DEFAULT_HORIZONTAL_ALIGNMENT;
     private m_vAlign = DefaultTextStyle.DEFAULT_VERTICAL_ALIGNMENT;
 
+    constructor(placement: AnchorPlacement) {
+        this.textPlacement = placement;
+    }
     /**
      * Set layout based on theme style defined and optional text placement.
      *
-     * @param layoutStyle The basic theme label style.
      * @param placement The optional new anchor placement.
      */
-    setPlacement(layoutStyle: TextLayoutStyle, placement?: AnchorPlacement) {
-        if (placement !== undefined) {
-            this.m_hAlign = placement.h;
-            this.m_vAlign = placement.v;
-            this.m_state =
-                placement.h === layoutStyle.horizontalAlignment &&
-                placement.v === layoutStyle.verticalAlignment
-                    ? LayoutUsed.ThemeBased
-                    : LayoutUsed.Alternative;
-        } else {
-            this.m_hAlign = layoutStyle.horizontalAlignment;
-            this.m_vAlign = layoutStyle.verticalAlignment;
-            this.m_state = LayoutUsed.ThemeBased;
-        }
+    set textPlacement(placement: AnchorPlacement) {
+        this.m_hAlign = placement.h;
+        this.m_vAlign = placement.v;
     }
 
     /**
@@ -66,43 +47,18 @@ export class LayoutState {
      *
      * Function returns alternative or base placement depending on layout state.
      *
-     * @param layoutStyle The label layout style.
      * @returns The current anchor placement.
      */
-    getPlacement(layoutStyle: TextLayoutStyle): AnchorPlacement {
-        if (this.isAlternative()) {
-            return { h: this.m_hAlign, v: this.m_vAlign };
-        } else {
-            return { h: layoutStyle.horizontalAlignment, v: layoutStyle.verticalAlignment };
-        }
+    get textPlacement(): AnchorPlacement {
+        return { h: this.m_hAlign, v: this.m_vAlign };
     }
 
     /**
-     * Reset existing `LayoutState` to appear like a fresh state.
+     * Reset existing `LayoutState` to contain values from style/theme layout.
      */
-    reset() {
-        this.m_state = LayoutUsed.Undefined;
-    }
-
-    /**
-     * @returns `true` if element has not yet been placed and layout variant assigned.
-     */
-    isUndefined(): boolean {
-        return this.m_state === LayoutUsed.Undefined;
-    }
-
-    /**
-     * @returns `true` if element is using original style defined in theme.
-     */
-    isBase(): boolean {
-        return this.m_state === LayoutUsed.ThemeBased;
-    }
-
-    /**
-     * @returns `true` if element is fading in.
-     */
-    isAlternative(): boolean {
-        return this.m_state === LayoutUsed.Alternative;
+    reset(layoutStyle: TextLayoutStyle) {
+        this.m_hAlign = layoutStyle.horizontalAlignment;
+        this.m_vAlign = layoutStyle.verticalAlignment;
     }
 
     get horizontalAlignment(): HorizontalAlignment {
@@ -111,27 +67,5 @@ export class LayoutState {
 
     get verticalAlignment(): VerticalAlignment {
         return this.m_vAlign;
-    }
-
-    /**
-     * Clone this `LayoutState`.
-     *
-     * @returns Cloned `LayoutState`.
-     */
-    clone(): LayoutState {
-        const res = new LayoutState();
-        res.copy(this);
-        return res;
-    }
-
-    /**
-     * Copies the values of `other` to this [[LayoutState]].
-     *
-     * @param other The other [[LayoutState]] to copy.
-     */
-    copy(other: LayoutState) {
-        this.m_hAlign = other.m_hAlign;
-        this.m_vAlign = other.m_vAlign;
-        this.m_state = other.m_state;
     }
 }
--- a/@here/harp-mapview/lib/text/LayoutState.ts
+++ b/@here/harp-mapview/lib/text/LayoutState.ts
@@ -40,25 +40,41 @@ export class LayoutState {
     private m_vAlign = DefaultTextStyle.DEFAULT_VERTICAL_ALIGNMENT;
 
     /**
-     * Set layout based on theme style defined.
+     * Set layout based on theme style defined and optional text placement.
      *
-     * @param layoutStyle The original theme label style.
+     * @param layoutStyle The basic theme label style.
+     * @param placement The optional new anchor placement.
      */
-    setFromBase(layoutStyle: TextLayoutStyle) {
-        this.m_hAlign = layoutStyle.horizontalAlignment;
-        this.m_vAlign = layoutStyle.verticalAlignment;
-        this.m_state = LayoutUsed.ThemeBased;
+    setPlacement(layoutStyle: TextLayoutStyle, placement?: AnchorPlacement) {
+        if (placement !== undefined) {
+            this.m_hAlign = placement.h;
+            this.m_vAlign = placement.v;
+            this.m_state =
+                placement.h === layoutStyle.horizontalAlignment &&
+                placement.v === layoutStyle.verticalAlignment
+                    ? LayoutUsed.ThemeBased
+                    : LayoutUsed.Alternative;
+        } else {
+            this.m_hAlign = layoutStyle.horizontalAlignment;
+            this.m_vAlign = layoutStyle.verticalAlignment;
+            this.m_state = LayoutUsed.ThemeBased;
+        }
     }
 
     /**
-     * Override label style using alternative anchor placement.
+     * Acquire current placement setup.
      *
-     * @param placement The new anchor placement.
+     * Function returns alternative or base placement depending on layout state.
+     *
+     * @param layoutStyle The label layout style.
+     * @returns The current anchor placement.
      */
-    setFromPlacement(placement: AnchorPlacement) {
-        this.m_hAlign = placement.h;
-        this.m_vAlign = placement.v;
-        this.m_state = LayoutUsed.Alternative;
+    getPlacement(layoutStyle: TextLayoutStyle): AnchorPlacement {
+        if (this.isAlternative()) {
+            return { h: this.m_hAlign, v: this.m_vAlign };
+        } else {
+            return { h: layoutStyle.horizontalAlignment, v: layoutStyle.verticalAlignment };
+        }
     }
 
     /**
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -18,6 +18,7 @@ import { PoiManager } from "../poi/PoiManager";
 import { PoiRenderer } from "../poi/PoiRenderer";
 import { CollisionBox, DetailedCollisionBox, IBox, ScreenCollisions } from "../ScreenCollisions";
 import { ScreenProjector } from "../ScreenProjector";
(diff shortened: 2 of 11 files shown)


@here/harp-mapview/lib/text/TextElementState.ts
```diff
@@ -225,6 +244,10 @@ export class TextElementState {
         }
 
         this.m_textRenderState = new RenderState();
+        this.m_textLayoutState = new LayoutState();
+        if (this.element.layoutStyle !== undefined) {
+            this.m_textLayoutState.setPlacement(this.element.layoutStyle);
```
atomicsulfate, line 249: Pass the layout style in the constructor and set the placement already there.

@here/harp-mapview/lib/text/TextElementState.ts
```diff
@@ -33,6 +34,11 @@ export class TextElementState {
      * Used during rendering.
      */
     private m_textRenderState?: RenderState;
+    /**
+     * @hidden
+     * Used to store recently used text layout.
+     */
+    private m_textLayoutState?: LayoutState;
```
atomicsulfate, line 41: A performance issue that we have in labeling is the allocation of several small objects for each label handled by the renderer. That leads to thousands of small allocations. This here adds yet another object, maybe we could combine it with RenderState or flatten the structure and put everything in TextElementState?

@here/harp-mapview/lib/text/TextElementState.ts
```diff
@@ -225,6 +244,10 @@ export class TextElementState {
         }
 
         this.m_textRenderState = new RenderState();
+        this.m_textLayoutState = new LayoutState();
```
atomicsulfate, line 247: Do we really need to initialize always the layout state? Can we avoid this initialization for labels that are placed in its default layout position and only initialize the layout state when they're moved to an alternative position?

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -430,62 +422,63 @@ function placePointLabelChoosingAnchor(
...
-    // Start with currently set alignment settings.
-    const basePlacement = {
-        h: label.layoutStyle!.horizontalAlignment,
-        v: label.layoutStyle!.verticalAlignment
-    };
+    // Start with recently set alignment settings if layout state is stored or
+    // simply begin from layout defined in theme.
+    const labelThemeLayout = label.layoutStyle!;
+    const labelStateLayout = labelState.textLayoutState;
+    const recentPlacement =
```
atomicsulfate, line 433: Better name: 'lastPlacement'. Shorter, and more precise.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -430,62 +422,63 @@ function placePointLabelChoosingAnchor(
...
-        // NOTE: This actually writes also to:
-        // textCanvas.textLayoutStyle.verticalAlignment/horizontalAlignment.
-        // which is needed for text measurements being done on canvas.
-        label.layoutStyle!.horizontalAlignment = anchorPlacement.h;
-        label.layoutStyle!.verticalAlignment = anchorPlacement.v;
-
-        // Bounds may be already calculated for persistent label, force re-calculation for
-        // for alternative placements.
-        const basePlacementTest = tryIdx === 0;
+
```
atomicsulfate, line 454: tryIdx is completely unneeded. tryIdx === i - matchIdx. It's just making the loop more complicated to understand than it should be.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -430,62 +422,63 @@ function placePointLabelChoosingAnchor(
...
+        // Bounds may be already calculated for persistent label, force re-calculation only
+        // for alternative (new) placements.
+        const recentPlacementTest = tryIdx === 0 && persistent;
```
atomicsulfate, line 457: do i === matchIdx instead of tryIdx === 0.

atomicsulfate: Also, 'isLastPlacement' is better name than 'recentPlacementTest'. More precise, shorter.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -430,62 +422,63 @@ function placePointLabelChoosingAnchor(
...
+        const basePlacementTest =
```
atomicsulfate, line 458: isBasePlacement

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -502,20 +495,31 @@ function placePointLabelChoosingAnchor(
 
         // Glyphs arrangement have been changed remove text buffer object and apply
         // new text layout to the label.
-        if (i !== 0) {
+        if (!recentPlacementTest) {
```
atomicsulfate, line 498: Move this invalidation of label.textBufferObject together with the code that makes the object invalid, i.e. next to the call to textCanvas.measureText.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -502,20 +495,31 @@ function placePointLabelChoosingAnchor(
...
             label.textBufferObject = undefined;
         }
+
+        if (!basePlacementTest) {
+            // Text elements with alternative placement require their own unique layout style.
+            // Provide private copy of label layout and attach it to current label's text canvas.
+            textCanvas.textLayoutStyle = label.layoutStyle!.clone();
```
atomicsulfate, line 505: We should only need to clone the layout style at most once per label. Instead, this code here is cloning the layout style  every time a label is not placed in the base placement. This must be hurting performance really badly, especially in the current code, where this clone is done always at the beginning of this function.

TextCanvas should have it's own textRenderStyle and textLayoutStyle objects. Whenever a label needs to be processed by text canvas, we copy the label's style by doing:
textCanvas.textLayoutStyle.copy(label.layoutStyle)

where we copy every parameter, without allocating new objects.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -623,6 +706,7 @@ function placePointLabelAtCurrentAnchor(
     // present in the new zoom level) before an even lower priority one takes the space.
     // Otherwise the lowest priority text will fade in and back out.
     // TODO: Add a unit test for this scenario.
+    // TODO!: Move this out of the function to make it full utility only!
```
atomicsulfate, line 709: Why do you want to move this out. Placement has two main responsibilities: find free screen space, and allocate it so that other labels can't use it. This code here is the second, so it belongs here.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -575,9 +656,11 @@ function placePointLabelAtCurrentAnchor(
         textCanvas.measureText(label.glyphs!, label.bounds, tmpMeasurementParams);
     }
 
+    // Revert back text canvas layout.
+    textCanvas.textLayoutStyle.horizontalAlignment = baseAlign.h;
+    textCanvas.textLayoutStyle.verticalAlignment = baseAlign.v;
```
atomicsulfate, line 661: no need to restore the base alignment. If the label is successfully placed, it will be with the last alignment set on the text canvas, if it's rejected, the last aligment set won't be used at all.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -502,20 +495,31 @@ function placePointLabelChoosingAnchor(
...
         }
+
+        if (!basePlacementTest) {
+            // Text elements with alternative placement require their own unique layout style.
+            // Provide private copy of label layout and attach it to current label's text canvas.
+            textCanvas.textLayoutStyle = label.layoutStyle!.clone();
+            // Setup TextCanvas layout settings of the new placement as it is required for further
+            // TextBufferObject allocations and measurements in addText().
+            textCanvas.textLayoutStyle.horizontalAlignment = anchorPlacement.h;
+            textCanvas.textLayoutStyle.verticalAlignment = anchorPlacement.v;
```
atomicsulfate, line 509: placePointLabelAtAnchor already sets these alignments and then reverts them back. Just leave the aligments set already in placePointLabelAtAnchor.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -553,6 +555,76 @@ function placePointLabelAtCurrentAnchor(
...
+
+    if (!basePlacementTest) {
+        // Text elements with different then base placement require their own unique layout style.
+        // Provide private copy of label layout and attach it to current label's text canvas.
+        textCanvas.textLayoutStyle = label.layoutStyle!.clone();
+        // Setup TextCanvas layout settings of the new placement as it is required for further
+        // TextBufferObject allocations and measurements in addText().
+        textCanvas.textLayoutStyle.horizontalAlignment = placement.h;
+        textCanvas.textLayoutStyle.verticalAlignment = placement.v;
+    }
```
atomicsulfate, line 597: Copypasta from placePointLabelChoosingAnchor(). Please read my comments there. This shouldn't be needed here.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -553,6 +555,76 @@ function placePointLabelAtCurrentAnchor(
...
+    const persistent = labelState.visible;
+    const labelThemeLayout = label.layoutStyle!;
+    const labelStateLayout = labelState.textLayoutState;
+    // Either not persistent or textLayoutState should be already initialized.
+    assert(!persistent || labelStateLayout !== undefined);
+    // Use recently rendered (state stored) layout if available, otherwise theme based style.
+    const placement =
+        persistent && labelStateLayout !== undefined
+            ? labelStateLayout.getPlacement(labelThemeLayout)
+            : { h: labelThemeLayout.horizontalAlignment, v: labelThemeLayout.verticalAlignment };
```
atomicsulfate, line 572: Copypasta from placePointLabelChoosingAnchor().

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -430,62 +422,63 @@ function placePointLabelChoosingAnchor(
...
-        v: label.layoutStyle!.verticalAlignment
-    };
+    // Start with recently set alignment settings if layout state is stored or
+    // simply begin from layout defined in theme.
+    const labelThemeLayout = label.layoutStyle!;
+    const labelStateLayout = labelState.textLayoutState;
+    const recentPlacement =
+        persistent && labelStateLayout !== undefined
+            ? labelStateLayout.getPlacement(labelThemeLayout)
+            : { h: labelThemeLayout.horizontalAlignment, v: labelThemeLayout.verticalAlignment };
```
atomicsulfate, line 436: Move this logic to a new member in TextElementState called getPlacement().

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -553,6 +555,76 @@ function placePointLabelAtCurrentAnchor(
     textCanvas: TextCanvas,
     screenCollisions: ScreenCollisions,
     isRejected: boolean,
+    outScreenPosition: THREE.Vector3
```
atomicsulfate, line 558: Please check my comments inside this function first. I would say, this function can be completely removed and replace by calling directly placePointLabelAtAnchor.

@here/harp-mapview/lib/text/LayoutState.ts
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Copies the values of `other` to this [[LayoutState]].
+     *
+     * @param other The other [[LayoutState]] to copy.
+     */
+    copy(other: LayoutState) {
+        this.m_hAlign = other.m_hAlign;
+        this.m_vAlign = other.m_vAlign;
+        this.m_state = other.m_state;
+    }
```
atomicsulfate, line 136: Remove this code if not used.
It's crucial that we keep our code as short as possible, even more important than completeness.
Our labeling code is already twice as large as MapBox one, and 5 times larger than Tangram's.

@here/harp-mapview/lib/text/LayoutState.ts
```diff
@@ -0,0 +1,137 @@
...
+    VerticalAlignment
+} from "@here/harp-text-canvas";
+
+/**
+ * Version of layout style stored.
+ */
+export enum LayoutUsed {
+    Undefined = 0,
+    ThemeBased,
+    Alternative
```
atomicsulfate, line 20: This distinction between ThemeBased and Alternative won't make sense once we have placement alternatives in the style.

@here/harp-mapview/lib/text/LayoutState.ts
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Acquire current placement setup.
+     *
+     * Function returns alternative or base placement depending on layout state.
+     *
+     * @param layoutStyle The label layout style.
+     * @returns The current anchor placement.
+     */
+    getPlacement(layoutStyle: TextLayoutStyle): AnchorPlacement {
+        if (this.isAlternative()) {
```
atomicsulfate, line 73: You're setting in setPlacement the layoutStyle aligment if no alternative placement is given. 
Then why you can't here just get whatever alignment is set in state? It might be an alternative aligment or the one coming from the layout style, you don't care.

@here/harp-mapview/lib/text/LayoutState.ts
```diff
@@ -0,0 +1,137 @@
...
+    isBase(): boolean {
+        return this.m_state === LayoutUsed.ThemeBased;
+    }
+
+    /**
+     * @returns `true` if element is fading in.
+     */
+    isAlternative(): boolean {
+        return this.m_state === LayoutUsed.Alternative;
+    }
```
atomicsulfate, line 106: I don't understand why you need this state.
You can just save here whatever was the last alignment used.

@here/harp-mapview/test/PlacementTest.ts
```diff
@@ -744,10 +753,316 @@ describe("Placement", function() {
...
+                    screenCollisions,
+                    false,
+                    outPosition,
+                    false
+                );
+                // Label out of screen and layout unchanged.
+                let layout = state.textLayoutState!;
+                expect(result).to.equal(PlacementResult.Invisible);
+                expect(layout.horizontalAlignment).to.be.equal(HorizontalAlignment.Right);
+                expect(layout.verticalAlignment).to.be.equal(VerticalAlignment.Center);
```
atomicsulfate, line 797: This code is repeated many many times along this file (the file is already almost 1500 lines!), making it hard to maintain. Any change in createTextElement, createTextElementState or placePointLabel parameters will have to be applied to all those places.

For these test cases where you use the same setup/check code over and over again, it's very useful to use a parameterized test pattern. You have a test that takes as input some configuration parameters (in this case the alignment alternatives and the labels anchor positions) and the expected outcome (in this case the placement result), and executes the same code.

See for example intersectsRay tests OrientedBox3Test.

@here/harp-mapview/lib/text/LayoutState.ts
kkostecki: In fact this things matters, especially when I provide the placement list from theme both with alignment settings. For now it is required even more because of this truly shitty initialization of TextElementState.... We have no guaranty that the layout there is proper, at a very beginning there is no state even... Moreover most of the unit tests (not mine) do not even care about the states! - so you have undefined values there... I need to hold some information about the state stored if it is valid already or anymore, so at least two states are needed -> Initialized / NotInitialized, having 3 states allows to provide more fine tuned logic that Nino has mentioned on the last call... Maybe I will need to reset the state on some scenarios:
- label mostly out of the screen,
- zoom level changed (less or more significantly).
My proposal would be even to make this LayoutState always available so it simplifies a code a lot, but as I saw (from other comments), you rather opt to make it even less commonly set to decrease allocations...

@here/harp-mapview/lib/text/TextElementState.ts
kkostecki: That is what I wanted to do :) - honestly I was dreaming about it ... but this does not comply with other states defined here (m_textRenderState...) and more importantly with your other comments about objects allocations - too many.
Generally I would opt to have this LayoutState as always defined -> this will simplify things A LOT!.
Then the question arises what we should do in TextElementState.replace() call, which moves states from predecessor? 
- copy by reference !!! No we don't want this,
- copy by ref and make undefined in predecessor - No we can not do it, because must be always defined (as said before),
- make a clone() from predecessor (that is what I have - thus clone() copy() methods in LayoutState),
- should we reset state to Theme Based?

I think that is the MOST IMPORTANT design issue in this PR - if we agree on LayoutState always defined (reference always set and never undefined), then the PR will change and simplify in many places.

@here/harp-mapview/lib/text/TextElementState.ts
kkostecki: How does it comply with you previous statement -> define - construct LayoutState in TextElementState c-tor?
We need to decide on something:
- either always allocated in c-tor - simplifies the code (also helps JIT),
- always allocated when TextElementState initialized - current code,
- allocated for features using it - for all point labels, requires some more code branches to current code,
- never allocated, but only on demand (point labels when placed alternatively) -> most difficult to hassle in the logic,

@here/harp-mapview/lib/text/TextElementState.ts
kkostecki: Again let's go back to previous comments.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Ok.

@here/harp-mapview/lib/text/LayoutState.ts
kkostecki: It is important to know because of label.bounds being cached from the last frame. Imagine two scenario:
- label bounds calculated for alternative placement, then got invisible/rejected, label placed again with bounds already stored (in label.bounds property) -> we should calculate placement for base layout but we have bounds there for alternative placement -> bounds should be re-calculated.
- secondly we need to know if the LayoutState is storing any reasonable values before using them.

@here/harp-mapview/lib/text/LayoutState.ts
kkostecki: True it is a part that I missed after changing implementation to hold LayoutState all the time.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Good point!

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Ok.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Ok.

@here/harp-mapview/lib/text/LayoutState.ts
kkostecki: It also depends from the decision about alignment, if we would like to have alignment working separately, i.e.: text aligned left and placed on the left side of icon. Then it might be useful.
I will see what do to with the states... For now I need at least two (initialized or undefined).

kkostecki: I will remove all state based behavior in update and use base / alternative checks on the fly.

@here/harp-mapview/lib/text/LayoutState.ts
kkostecki: In the next commit I will remove all states, let's see how it gonna look & work. Probably everybody should be satisfied.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Ok I will do so, but I am not sure if it's gonna be simpler to read... Keep in mind that we need to iterate as in round buffer.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: We already have that in previous version... but this way I need to revert them back every time the placement will fail (with rejected or invisible placement result). Keep in mind that rejected labels need to be rendered with the latest alignment that was successful (from previous frame) in order to fade out in the same position.
I have done it in new way just to simplify the logic -> override TextCanvas settings only if new & alternative placement have been found.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Please read my comments above, I am trying to do it in opposite way, extract as much as possible from `placePointLabelAtAnchor` to made it simple utility function that may ONLY check if placement is possible. I would even like to remove the screenCollisions handling from there.
This way I will be able to provide some optimizations inside. Moreover I have already some changes that would totally reduce the problem of TextCanvas state -> it's references to label.layoutStyle, but we need to make some smaller steps first - as already agreed - make PR as small as possible.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Wrong! It will be used... the last successful placement will be used even if label is rejected. I have a strong feeling that this is a basic problem of our misunderstanding:
**Label that are rejected will be rendered few more frames with FADE OUT effect and require to KEEP their layout consistent - no more multi-anchor placement **

@here/harp-mapview/lib/text/Placement.ts
kkostecki: I left this comment here to make a note that this function will be soon changed to simple utility that will only CHECK not PERFORM the placement, I would like to get rid of TextCanvas.measureText most of the time for alternative placement.
The idea is to perform it once! Then text bounds may be computed based on **base placement**, **base bounds** that will be computed only once for entire label live, and the **anchor point** difference between the base and current.

kkostecki: Almost forgot to mention with that change in mind I will get rid of reverting back out screen position, bounds etc... Utility will simply work on the cache objects and if placement will be found we will copy from cache to the output variables.

@here/harp-mapview/lib/text/LayoutState.ts
kkostecki: Ok, just to limit discussions, next update includes:
- removed enum state inside LayoutState,
- layout state created seldomly only when really needed.

@here/harp-mapview/test/PlacementTest.ts
kkostecki: Will do in the next update.

kkostecki: Done.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Done.

kkostecki: Fixed. TextCanvas holds its own copy of layout / render styles.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: But here you're reverting back to the base aligment, instead of the last aligment used. If placement fails and the label is rejected it will start fading out in the base aligment instead of using the last aligment?
I think we need to discuss this online, I'm really confused.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Ok moved them to placePointLabelAtAnchor() function, so the logic works the same and the flag `forceInvalidation` which equals `!isLastPlacement` is used to determine if buffer should be released. This will never happen if we use the last rendered text position.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Yep I see, that after refactoring I have created a bug here -> it should be `lastPlacement` instead of base one.
Anyway I have change the logic as you suggested, so TextCanvas layout state is reverted only on demand in `placePointLabelChoosingAnchor()`, this looks simpler. If I will need to optimize later as discussed previously I will figure out something better.

PR #1442, opened 2020-04-06
HARP-9711 Don't put tiles that are still loading into the renderedTil…

nzjony:
…es list

### DIFF
--- a/@here/harp-mapview/lib/VisibleTileSet.ts
+++ b/@here/harp-mapview/lib/VisibleTileSet.ts
@@ -936,7 +936,7 @@ export class VisibleTileSet {
                     tile.offset
                 );
                 tile.levelOffset = 0;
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -936,7 +936,7 @@ export class VisibleTileSet {
                     tile.offset
                 );
                 tile.levelOffset = 0;
-                if (tile.hasGeometry || defaultSearchDirection === SearchDirection.NONE) {
+                if (tile.hasGeometry) {
```
ninok, line 939: What was the reason why we did this?

nzjony: Good question, I don't know.

PR #1449, opened 2020-04-08
HARP-9576 Fix styling data-driven example

MPursche:
* Change to display of city population (countries do not contain population in new data source).
* Fix MapView.disableFadeIn option.
* Implement textFadeTime and iconFadeTime.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/TextElement.ts
+++ b/@here/harp-mapview/lib/text/TextElement.ts
@@ -327,6 +327,12 @@ export class TextElement {
 
     pathLengthSqr?: number;
 
+    /**
+     * Time to fade in text in milliseconds.
+     * @default [[DEFAULT_FADE_TIME]] 800
+     */
+    textFadeTime?: number;
+
     type: TextElementType;
 
     private m_poiInfo?: PoiInfo;
--- a/@here/harp-mapview/lib/text/TextElementState.ts
+++ b/@here/harp-mapview/lib/text/TextElementState.ts
@@ -117,11 +117,11 @@ export class TextElementState {
      * be out of view.
      */
     update(viewDistance: number | undefined) {
-        if (this.initialized) {
-            this.setViewDistance(viewDistance);
-        } else if (viewDistance !== undefined) {
-            this.initialize(viewDistance);
+        if (!this.initialized && viewDistance !== undefined) {
+            this.initializeRenderStates();
         }
+
+        this.setViewDistance(viewDistance);
     }
 
     /**
@@ -130,9 +130,6 @@ export class TextElementState {
      * be out of view.
      */
     setViewDistance(viewDistance: number | undefined) {
-        if (viewDistance === this.m_viewDistance) {
-            return;
-        }
         this.m_viewDistance = viewDistance;
     }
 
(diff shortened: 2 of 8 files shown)


MPursche:
GPU tests please

musculman:
Thanks for finding and fixing an error I introduced!

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1355,7 +1355,7 @@ export class TextElementsRenderer {
             placementStats.log();
         }
 
-        if (!this.m_options.disableFading && renderParams.fadeAnimationRunning) {
+        if (renderParams.fadeAnimationRunning) {
```
MPursche, line 1358: This was changed because with the current implementation of fading in `RenderState` it still takes 1 frame for the fading animation to complete when fading is disabled(`updateFading` needs to be called once). This caused some of the labels to not be visible with `MapView.disableFading=true`.

@here/harp-mapview/lib/text/TextElementState.ts
```diff
@@ -206,28 +203,37 @@ export class TextElementState {
...
+        const { textFadeTime } = this.element;
+        const iconFadeTime = this.element.poiInfo?.technique.iconFadeTime;
         if (this.element.type === TextElementType.LineMarker) {
             this.m_iconRenderStates = new Array<RenderState>();
             for (const _point of this.element.points as THREE.Vector3[]) {
                 const iconRenderStates = this.m_iconRenderStates as RenderState[];
                 const renderState = new RenderState();
+                if (iconFadeTime !== undefined) {
+                    renderState.fadeTime = iconFadeTime;
+                }
```
atomicsulfate, line 221: We should be able to do:
`const renderState = new RenderState(iconFadeTime);`

If undefined, it'll take the default parameter value. Right now we have 3 other parameters before it, which are never used (except for a couple of test here). I would move the fade time parameter to the first position and even remove the other constructor parameters. I don't see the use case of initializing the state with custom values for startTime,opacity etc...

@here/harp-mapview/lib/text/TextElementState.ts
```diff
@@ -206,28 +203,37 @@ export class TextElementState {
...
                 iconRenderStates.push(renderState);
             }
             return;
         }
 
         this.m_textRenderState = new RenderState();
+        if (textFadeTime !== undefined) {
+            this.m_textRenderState.fadeTime = textFadeTime;
+        }
```
atomicsulfate, line 230: idem

@here/harp-mapview/lib/text/TextElementState.ts
```diff
@@ -206,28 +203,37 @@ export class TextElementState {
...
 
         if (this.element.type === TextElementType.PoiLabel) {
             this.m_iconRenderStates = new RenderState();
+            if (iconFadeTime !== undefined) {
+                this.m_iconRenderStates.fadeTime = iconFadeTime;
+            }
```
atomicsulfate, line 236: idem

@here/harp-mapview/lib/text/TextElement.ts
```diff
@@ -327,6 +327,12 @@ export class TextElement {
 
     pathLengthSqr?: number;
 
+    /**
+     * Time to fade in text in milliseconds.
+     * @default [[DEFAULT_FADE_TIME]] 800
+     */
+    textFadeTime?: number;
+
```
atomicsulfate, line 335: Out of the scope of this change, but we should take all properties related to how a text element is rendered into a separate style object, much like TextRenderStyle & TextLayoutStyle.

PR #1451, opened 2020-04-09
HARP-9897: Add support for vector2/3/4 operators.

robertoraggi:
This change introduces new operator and type conversions to
support vector2/3/4.

### DIFF
--- a/@here/harp-datasource-protocol/StyleExpressions.md
+++ b/@here/harp-datasource-protocol/StyleExpressions.md
@@ -165,6 +165,33 @@ returns the value of the first `fallback` that is a number.
 ["to-number", value, fallback...]
 ```
 
+## to-vector2
+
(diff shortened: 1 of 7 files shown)


@here/harp-datasource-protocol/StyleExpressions.md
```diff
@@ -165,6 +165,33 @@ returns the value of the first `fallback` that is a number.
 ["to-number", value, fallback...]
 ```
 
+## to-vector2
+
+Converts the value to "vector2", if the value cannot be converted to a vector2
```
zbigg, line 170: It's not documented what can be successfully converted to `vector*` ?

robertoraggi: good point, I will write something

PR #1454, opened 2020-04-14
HARP-8067: Add @here/harp-vectortile-datasource

robertoraggi:
Introduce VectorTileDataSource sharing the common code of
2D vector tile data sources.


@here/harp-vectortile-datasource/index.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
```
tsteenbe, line 2: Please update the copyright statement to include the last modified year e.g. 2020 so it's "Copyright (C) 2017-2019 HERE Europe B.V."

PR #1457, opened 2020-04-14
HARP-6501: MapView.lookAt for GeoBox and arbitrary point support

zbigg:
MapView.lookAt for GeoBox and arbitrary point support

* MapView.lookAt: supports 'bounds' align view to fit GeoBox or arbitrary
  set of geo points

* new "getting-started / look-at" example to showing various usages of
  `lookAt`

* MapView: fix re-calculation of tilt/heading in lookAt() on sphere
  projection

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Copyright (C) 2017-2020 HERE Europe B.V.
  * Licensed under Apache 2.0, see full license in LICENSE
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -22,8 +22,11 @@ import {
 import { ViewRanges } from "@here/harp-datasource-protocol/lib/ViewRanges";
 import {
     EarthConstants,
+    GeoBox,
+    GeoBoxExtentLike,
     GeoCoordinates,
     GeoCoordLike,
+    isGeoBoxExtentLike,
     isGeoCoordinatesLike,
     mercatorProjection,
     Projection,
@@ -684,19 +687,44 @@ const MapViewDefaults = {
 };
 
 /**
- * Parameters for MapView.lookAt
+ * Parameters for [[MapView.lookAt]].
  */
 export interface LookAtParams {
     /**
      * Target/look at point of the MapView.
+     *
      * @note If the given point is not on the ground (altitude != 0) [[MapView]] will do a
      * raycasting internally to find a target on the ground.
+     *
      * As a consequence [[MapView.target]] and [[MapView.zoomLevel]] will not match the values
      * that were passed into the [[MapView.lookAt]] method.
      * @default `new GeoCoordinates(25, 0)` in [[MapView.constructor]] context.
      * @default [[MapView.target]] in [[MapView.lookAt]] context.
      */
     target: GeoCoordLike;
+
+    /**
+     * Fit MapView to these boundaries.
+     *
+     * If specified, `zoomLevel` and `distance` parameters are ignored and `lookAt` calculates best
+     * `zoomLevel` to fit given bounds.
+     *
+     * * if `bounds` is [[GeoBox]], then `lookAt` use [[LookAtParams.target]] or `bounds.target` and
+     *   ensure whole box is visible
+     *
+     * * if `bounds` is [[GeoBoxExtentLike]], then `lookAt` will use [[LookAtParams.target]] or
+     *   current [[MapView.target]] and ensure whole extents are visible
+     *
+     * * if `bounds` is [[GeoCoordLike]][], then `lookAt` will use [[LookAtParams.target]] or
+     *   calculated `target` as center of world box covering given points and ensure all points are
+     *   visible
+     *
+     * Note in sphere projection some points are not visible if you specify bounds that span more
+     * than 180 degreess in any direction.
+     *
+     * @see [[MapView.lookAt]] for defails how `bounds` interact with `target` parameter
+     */
+    bounds: GeoBox | GeoBoxExtentLike | GeoCoordLike[];
 
     /**
      * Camera distance to the target point in world units.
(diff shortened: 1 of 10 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -692,6 +693,38 @@ export interface LookAtParams {
...
+ */
+export interface FitBoundsOptions {
+    /**
+     * Margin added to view.
+     *
+     * Range:
+     *  - `0` - no margin,
+     *  - `1` - 100% of calculated distance if added
+     *
+     * @default `0.1` meaning 10% is added to calculated distance, so approximately 10% margin is
```
ninok, line 707: Maybe we should reach out to designers what they need. I can imagine they would expect margin to be in pixels.
Can you please check other renderers (including JSAPI and harp-native) what they have

zbigg: @ninok Any other option implementation adds huuge complexity, so if current `margin` is questionable i would
1) remove `margin` from this patch - it's minor
2) check what others and find optimal solution do in followup patch

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2036,6 +2069,43 @@ export class MapView extends THREE.EventDispatcher {
...
+     *
+     * Options
+     *  - `margin` - margin added to view, `0` - no margin, `1` - 100% of original size, defaults to
+     *     `0.1` meaning 10% is added to calculated distance, so approximately 10% margin is visible
+     *  - `tilt`   - optional, change tilt, defaults to current [[MapView]] tilt
+     *  - `heading` - optional, change heading, defaults to current [[MapView]] heading
+     *
+     * @param target [[GeoBox]] or list of points conforming to [[GeoCoordLike]]
+     */
+    fitBounds(target: GeoCoordLike[] | GeoBox, options?: FitBoundsOptions) {
```
ninok, line 2096: can we merge this method into the lookAt method? How are competitors doing this?

@here/harp-mapview/lib/MapView.ts
zbigg: After research, others usually support this as "padding in pixels".

* https://docs.mapbox.com/mapbox-gl-js/api/#map#fitbounds (google maps api does the same)
* https://leafletjs.com/reference-1.6.0.html#fitbounds-options

@here/harp-mapview/lib/MapView.ts
zbigg: @ninok Yes, we can merge it. Actually it's perfect fit for overload of `lookAt` if target is set of geocoords of GeoBox.

I would leave `fitBounds` as separate API and add overload for `lookAt`.

@here/harp-mapview/lib/MapView.ts
ninok: Okay, please remove margin for now and create a separate ticket,so we can get more input.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -410,6 +418,242 @@ export namespace MapViewUtils {
...
 
+    /**
+     * Add offset to geo points for minimal vieing box.
+     *
+     * In flat projection, with wrap around enabled, we should detect clusters of points around that
+     * wrap antimeridian.
+     *
+     * Here, we fit points into minimal geo box taking world wrapping into account.
+     */
+    export function wrapGeoPointsToScreen(points: GeoCoordLike[]): GeoCoordinates[] {
```
ninok, line 429: Do we need to expose this?
MapViewUtils is one of the worst parts of our public API b/c it constantly breaks.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -410,6 +418,242 @@ export namespace MapViewUtils {
...
+        }
+        return result;
+    }
+
+    /**
+     * Return [[GeoPoints]] bounding [[GeoBox]] applicable for [[getFitBoundsDistance]].
+     *
+     * @returns [[GeoCoordinates]] set that covers `box`
+     */
+    export function geoBoxToGeoPoints(box: GeoBox): GeoCoordinates[] {
```
ninok, line 471: Do we need to expose this?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -410,6 +418,242 @@ export namespace MapViewUtils {
...
+     *
+     * All dimensions belong to world space.
+     *
+     * @param points points which shall are to be covered by view
+     *
+     * @param worldTarget readonly, world target of [[MapView]]
+     * @param camera readonly, camera with proper `position` and rotation set
+     * @returns new distance to camera to be used with [[MapView.lookAt]]
+     */
+    export function getFitBoundsDistance(
```
ninok, line 497: Do we need to expose this?
Maybe the two method above are okay, b/c they are quite generic, but this one should be a private method in MapView b/c it is really specific and I don't see any use outside of MapView

@here/harp-mapview/lib/Utils.ts
zbigg: It's exported for unit-tests only. I can document it as @hidden but still i think it's worth testing.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -651,16 +654,31 @@ const MapViewDefaults = {
  */
 export interface LookAtParams {
     /**
-     * Target/look at point of the MapView.
+     * Look at point target (or box or set of points of the MapView.
```
ninok, line 657: This comment is not correct. target is still just a point

@here/harp-mapview/lib/MapView.ts
```diff
@@ -651,16 +654,31 @@ const MapViewDefaults = {
...
+     * Setup view, to look at box/ all points.
+     *
+     * If `target` is not specified, then geobox `center` or world box of `points` will be taken
+     * as new `target`.
+     *
+     * Note
+     *  * in sphere some points are not visible if you specify bounds that bigger than 180 degreess
+     *    in latitude
+     */
+    bounds: GeoBox | GeoBoxExtentLike | GeoCoordLike[];
```
ninok, line 680: What would be the center in case of `GeoBoxExtentLike` ?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -651,16 +654,31 @@ const MapViewDefaults = {
...
      * that were passed into the [[MapView.lookAt]] method.
      * @default `new GeoCoordinates(25, 0)` in [[MapView.constructor]] context.
      * @default [[MapView.target]] in [[MapView.lookAt]] context.
      */
     target: GeoCoordLike;
 
+    /**
+     * Setup view, to look at box/ all points.
+     *
+     * If `target` is not specified, then geobox `center` or world box of `points` will be taken
```
ninok, line 673: What do you mean with 

> world box of `points` will be taken as new `target`

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1997,6 +2015,58 @@ export class MapView extends THREE.EventDispatcher {
...
+     * #### Note on `target` and `bounds`
+     *
+     * If `bounds` are specified, `zoomLevel` and `distance` parameters are ignored and `lookAt`
+     * calculates best zoomLevel (and possibly target) to fit given bounds.
+     *
+     * Following table shows how relation between `bounds` and target.
+     *
+     * | `bounds`             | `target`    | actual `target`
+     * | ------               | ------      | --------
+     * | [[GeoBox]]           | _ignored_   | `bounds.center` is used as new `target`
```
ninok, line 2031: Why is the target ignored in this case?

ninok: I would expect the same behavior as for `GeoCoordLike[]` below, i.e. if target is undefined it will use the GeoBox.center if target is defined it will use it and zoom out so that the geobox is visible

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2042,6 +2106,40 @@ export class MapView extends THREE.EventDispatcher {
...
+     * find "minimal" box in projected/wrapped world instead showing whole world.
+     *
+     * Options
+     *
+     *  - `tilt`   - optional, change tilt, defaults to current [[MapView]] tilt
+     *  - `heading` - optional, change heading, defaults to current [[MapView]] heading
+     *
+     * @param target [[GeoBox]] or list of points conforming to [[GeoCoordLike]]
+     */
+    fitBoundsXX(
```
ninok, line 2132: Why do we need this method again? Just convenience for users of other APIs?

ninok: why XX?

@here/harp-mapview/lib/MapView.ts
zbigg: `target` if specified or `MapView.target` if not specified

@here/harp-mapview/lib/MapView.ts
zbigg: Leftover. I was sure i've finally removed it.

@here/harp-mapview/lib/MapView.ts
zbigg: Agree, that was unintuitive. Now `target` has priority and only if it's not specified, we derive it from  `bounds`.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2059,6 +2087,55 @@ export class MapView extends THREE.EventDispatcher {
...
+     * mapView.lookAt(bounds: { latitudeSpan: 10, longitudeSpan: 10})
+     *    // look at current `target`, but extending zoomLevel so we see 10 degrees of lat/long span
+     * ```
+     *
+     * @see More examples in [[LookAtExample]].
+     *
+     * @param params [[LookAtParams]]
+     */
+    lookAt(params: Partial<LookAtParams>): void;
+    // tslint:disable: max-line-length
```
ninok, line 2137: I guess this should be tslint:enable

@here/harp-mapview/lib/Utils.ts
```diff
@@ -410,6 +418,257 @@ export namespace MapViewUtils {
         return result;
     }
 
+    /**
+     * @hidden
```
ninok, line 422: AFAIK `@hidden` is only for typedoc.
Since we want to use api-extractor in the future I suggest to use to these tags:
https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags

I can also do this as part of my (never ending) API task

@here/harp-mapview/lib/MapView.ts
zbigg: Fixed

@here/harp-mapview/lib/Utils.ts
zbigg: Added both @hidden and @internal.

PR #1459, opened 2020-04-15
Add support for filtering features using expressions.

robertoraggi:
This change adds support for filtering data features using
expressions.

For example, the following expression can be used to filter
features in the 'buildings' and 'water' layers.
```typescript
  omvDataSource.setDataFilter([
      "match", ["get", "$layer"],
         ["buildings", "water"], true,
         false
  ]);
```

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvDataSource.ts
+++ b/@here/harp-omv-datasource/lib/OmvDataSource.ts
@@ -7,8 +7,11 @@
 import {
     AttributeMap,
     Definitions,
+    Expr,
     GeometryType,
+    isJsonExpr,
     ITileDecoder,
+    JsonArray,
     OptionsMap,
     StyleSet,
     WorkerServiceProtocol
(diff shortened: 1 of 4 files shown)


@here/harp-omv-datasource/lib/OmvDataSource.ts
```diff
@@ -261,14 +265,46 @@ export class OmvDataSource extends TileDataSource<OmvTile> {
...
-    setDataFilter(filterDescription: OmvFeatureFilterDescription): void {
-        this.m_decoderOptions.filterDescription =
-            filterDescription !== null ? filterDescription : undefined;
+    setDataFilter(filterDescription: OmvFeatureFilterDescription): void;
 
-        this.configureDecoder(undefined, undefined, undefined, {
-            filterDescription
-        });
+    /**
+     * Sets the expression used to filter features.
```
zbigg, line 274: I'm only concerned about meaning predicate of `filter`.
In this patch, we filter-out features that _match_ filter (albeit it's not documented)

I know that in common meaning `filter` is to "remove unneded stuff", so true should apply to stuff we don't need

On the other hand, filter in programming is commonly (Array.filter) ... used to "only accept those that match".

Please decide which version we have and document

Ref
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter
https://book.pythontips.com/en/latest/map_filter.html

PR #1466, opened 2020-04-16
HARP-9982 Disable line clipping by default

ninok:
We need to disable line clipping since HERE Vector Tiles will soon have a smaller margin.

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -6,7 +6,6 @@
 import {
     Env,
     Expr,
-    GeometryKind,
     getFeatureId,
     getPropertyValue,
     GradientSky,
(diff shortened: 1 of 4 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -3254,38 +3255,56 @@ export class MapView extends THREE.EventDispatcher {
...
+        // sure that nothing of the parent will be rendered on top of the children,
+        // as such, we shift using the FALLBACK_RENDER_ORDER_OFFSET.
+        // This does not apply to buildings b/c they are 3d and the overlaps
+        // are resolved with a depth prepass. Note we set this always to ensure that if
+        // the Tile is used as a fallback, and then used normally, that we have the correct
+        // renderOrder.
+
+        if (tile.levelOffset >= 0) {
+            if (object._backupRenderOrder !== undefined) {
+                // We meesed up the render order when this tile was used as fallback.
```
nzjony, line 3280:
```suggestion
                // We messed up the render order when this tile was used as fallback.
```

PR #1478, opened 2020-04-22
Initialize ColorCache before running the tests.

robertoraggi:
This change ensures the ColorCache is cleared before the test
starts.

### DIFF
--- a/@here/harp-mapview/test/ColorCacheTest.ts
+++ b/@here/harp-mapview/test/ColorCacheTest.ts
@@ -12,6 +12,10 @@ import { assert } from "chai";
 import { ColorCache } from "../lib/ColorCache";
 
 describe("ColorCache", function() {
+    this.beforeEach(() => {
+        ColorCache.instance.clear();
+    });
+
     it("empty", function() {
         assert.equal(ColorCache.instance.size, 0);
     });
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/test/ColorCacheTest.ts
```diff
@@ -12,6 +12,10 @@ import { assert } from "chai";
 import { ColorCache } from "../lib/ColorCache";
 
 describe("ColorCache", function() {
+    this.beforeAll(() => {
```
MPursche, line 15: I think it would be a good idea to also do `this.afterEach(() => ColorChache.instance.clear()`. Otherwise this test suite may have a side effect on other test suites.

robertoraggi: i think afterEach won't wrong, because the very 1st test is already failing for me

MPursche: Sorry, I meant both `beforeAll` and `afterEach`. The state should initially reseted and then after each test case. Otherwise the order of the test cases are important. Normally they should be independent from one another.

robertoraggi: well, but then we should clear the ColorCache before every single test suite we have in harp.gl (e.g interpolations and expressions, tile geometry creation) because pretty much everything that involves colors may change ColorCache

PR #1482, opened 2020-04-23
HARP-8308 Adding traffic line offsetting

nzjony:

### DIFF
--- a/@here/harp-materials/lib/SolidLineMaterial.ts
+++ b/@here/harp-materials/lib/SolidLineMaterial.ts
@@ -80,6 +80,7 @@ uniform mat4 modelViewMatrix;
 uniform mat4 projectionMatrix;
 uniform float extrusionWidth;
 uniform float outlineWidth;
+uniform float offset;
 uniform vec2 drawRange;
 
 #ifdef USE_DISPLACEMENTMAP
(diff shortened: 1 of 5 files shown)


nzjony:

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -138,6 +139,15 @@ void main() {
...
     pos += normalize( normal ) * texture2D( displacementMap, uv ).x;
     #endif
+
+    // Shift the line based on the offset, where the bitangent is the cross product of the average
+    // of the two direction vectors (the previous and next segment directions) and the normal of
+    // the line (facing into the sky). The w component is the angle between the two segments.
+    // Note, we need to take the angle into consideration, so we use trigonometry to calculate how
+    // much we need to extend the offset. Note, orthough this looks complicated we are doing this
+    // in the vertex shader, so it should not cause a performance issue.
+    pos += bitangent.xyz * offset * sqrt(1.0 + pow(abs(tan(bitangent.w / 2.0)), 2.0));
```
ninok, line 149: is the w component used somewhere else? If not we could do this computation on the CPU and store the result in w.

ninok: `tan(bitangent.w/2.0)` is already computed in line 116. Please reuse it.
Sorry to be picky here, but this code is affecting all rendered lines (even if offset is 0) and we we do 3 expensive operations(tan, pow and sqrt). Maybe we should `#ifdef` it, but that would lead to more shader compilations.
I still think we should store `tan(bitangent.w/2.0)` in the vertex attributes instead of computing it in the shader.

ninok: Just checked `extrudeLine` and it computes tan(bitangent.w/2.0) a third time. Please fix that.
Unfortunately it also computes cos(bitangent.w/2.0) so we cannot put tan(bitangent.w/2.0) into the vertex attribute

nzjony: @ninok  done

nzjony:

DimitryRd:
Rebuild please

nzjony:

PR #1484, opened 2020-04-23
HARP-7836: Add support for tiled GeoJson.

robertoraggi:
This change introduces an adapter that converts tiled GeoJson
to the internal respresentation used by omv-datasource.


robertoraggi:
Decoders tests please

PR #1486, opened 2020-04-23
HARP-10128: Use elevation at camera position to compute target.

atomicsulfate:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-map-controls/test/MapControlsTest.ts
+++ b/@here/harp-map-controls/test/MapControlsTest.ts
@@ -7,8 +7,8 @@
 // tslint:disable:only-arrow-functions
 //    Mocha discourages using arrow functions, see https://mochajs.org/#arrow-functions
 
-import { mercatorProjection } from "@here/harp-geoutils";
-import { MapView } from "@here/harp-mapview";
+import { GeoCoordinates, mercatorProjection, sphereProjection } from "@here/harp-geoutils";
+import { ElevationProvider, MapView, MapViewUtils } from "@here/harp-mapview";
 import { expect } from "chai";
 import * as sinon from "sinon";
 import * as THREE from "three";
@@ -20,9 +20,12 @@ const inNodeContext = typeof window === "undefined";
 
 describe("MapControls", function() {
     let sandbox: sinon.SinonSandbox;
+    let domElement: any;
+    let mapView: MapView;
+    let mapControls: MapControls;
+    let camera: THREE.Camera;
 
     beforeEach(function() {
-        sandbox = sinon.createSandbox();
         if (inNodeContext) {
             const theGlobal: any = global;
             // tslint:disable-next-line:no-empty
@@ -32,6 +35,13 @@ describe("MapControls", function() {
                 now: () => {}
             };
         }
+
+        sandbox = sinon.createSandbox();
+        domElement = { addEventListener: sandbox.stub() } as any;
+        mapView = sandbox.createStubInstance(MapView) as any;
+        sandbox.stub(mapView, "renderer").get(() => ({ domElement }));
+        camera = new THREE.PerspectiveCamera();
+        sandbox.stub(mapView, "camera").get(() => camera);
     });
 
     afterEach(function() {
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/Utils.ts
```diff
@@ -136,31 +147,26 @@ export namespace MapViewUtils {
...
         if (mapView.projection.type === ProjectionType.Planar) {
             // Calculate the difference and pan the map to maintain the map relative to the target
             // position.
-            targetPosition.sub(newTargetPosition);
-            panCameraAboveFlatMap(mapView, targetPosition.x, targetPosition.y);
+            worldTarget.sub(newWorldTarget);
+            panCameraAboveFlatMap(mapView, worldTarget.x, worldTarget.y);
         } else if (mapView.projection.type === ProjectionType.Spherical) {
-            panCameraAroundGlobe(mapView, targetPosition, newTargetPosition);
+            panCameraAroundGlobe(mapView, worldTarget, newWorldTarget);
```
kkostecki, line 169: I see slight problem here, although not introduced by your change... What happens when we encounter huge elevations in between? Camera may fly though the geometry? Although assuming that raycast is made with ndc and eleveation we should never aim to point which is "hidden" with some bigger mountain. Have you done such testing?

@here/harp-map-controls/test/MapControlsTest.ts
```diff
@@ -93,42 +91,95 @@ describe("MapControls", function() {
...
         expect(updateStub.callCount).to.be.equal(1);
     });
+
+    describe("zoomOnTargetPosition", function() {
+        const elevationProvider = ({} as any) as ElevationProvider;
+
+        function resetCamera() {
+            const target = GeoCoordinates.fromDegrees(0, 0);
+            const heading = 0;
+            const tilt = 0;
```
kkostecki, line 135: Could you add some test with tilt angle "enabled".

@here/harp-mapview/test/UtilsTest.ts
```diff
@@ -267,6 +275,81 @@ describe("map-view#Utils", function() {
...
+                        elevationProvider
+                    );
+
+                    expect(resultElevation.distance).equals(resultNoElevation.distance - elevation);
+                    const geoTargetNoElevation = projection.unprojectPoint(
+                        resultNoElevation.target
+                    );
+                    expect(geoTargetNoElevation).deep.equals(
+                        GeoCoordinates.fromDegrees(geoTarget.lat, geoTarget.lng, 0)
+                    );
```
kkostecki, line 344: For simplicity of reading the code I would just move "no elevation" expectation to little above - to section where it is calculated, so we have test section with and without elevation.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -102,31 +102,42 @@ export namespace MapViewUtils {
...
+              )
+            : undefined;
+
         // Get current target position in world space before we zoom.
-        const targetPosition = rayCastWorldCoordinates(
-            mapView,
-            targetPositionOnScreenXinNDC,
-            targetPositionOnScreenYinNDC
-        );
+        const worldTarget = rayCastWorldCoordinates(mapView, targetNDCx, targetNDCy, elevation);
```
kkostecki, line 128: I think we should really consider sampling terrain with approach similar to parallax mapping, this would give us much more accurate results of actual target. The function should then have eyeVector as an input.

@here/harp-mapview/lib/Utils.ts
atomicsulfate: There a terrain-camera interaction ticket where we can handle all that:
https://saeljira.it.here.com/browse/HARP-5430

This change is meant to be a small fix that makes globe with terrain at least reasonably usable (right now it's completely broken).

@here/harp-mapview/lib/Utils.ts
atomicsulfate: Might be worth a try, please add the idea here:
https://saeljira.it.here.com/browse/HARP-5430

@here/harp-mapview/test/UtilsTest.ts
atomicsulfate: done

@here/harp-map-controls/test/MapControlsTest.ts
atomicsulfate: done

@here/harp-mapview/lib/Utils.ts
kkostecki: Ok done.

PR #1488, opened 2020-04-24
Remove usage of bit operations when reading coords.

robertoraggi:
This change removes the usage of bit operations when decoding
the coordinates of OMV protobuf encoded vector tiles.

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvData.ts
+++ b/@here/harp-omv-datasource/lib/OmvData.ts
@@ -202,9 +202,9 @@ export class GeometryCommands {
                     const xx = geometry[cmdIndex++];
                     const yy = geometry[cmdIndex++];
 
-                    // tslint:disable:no-bitwise
-                    currX += (xx >> 1) ^ -(xx & 1);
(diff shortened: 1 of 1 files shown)


@here/harp-omv-datasource/lib/OmvData.ts
```diff
@@ -202,9 +202,9 @@ export class GeometryCommands {
                     const xx = geometry[cmdIndex++];
                     const yy = geometry[cmdIndex++];
 
-                    // tslint:disable:no-bitwise
-                    currX += (xx >> 1) ^ -(xx & 1);
-                    currY += (yy >> 1) ^ -(yy & 1);
+                    currX += xx % 2 ? (xx + 1) / -2 : xx / 2;
```
ninok, line 205: Wonder what's the performance impact of that since we have now 3 divisions instead of 3 bit operations.
I guess we will not see any difference in our perftests.
Here is some related testcase that shows that modulo and divison is much slower:
https://jsperf.com/modulo-vs-bitwise/19

PR #1490, opened 2020-04-24
Introduce ITransferManager

haraldF:


haraldF:
Depends on https://github.com/heremaps/harp.gl/pull/1489, please merge first

@here/harp-transfer-manager/src/TransferManager.ts
```diff
@@ -74,7 +105,7 @@ export class TransferManager {
     }
     private activeDownloadCount = 0;
     private downloadQueue = new Array<DeferredPromise<Response>>();
-    private activeDownloads = new Map<string, Promise<any>>();
+    private activeDownloads = new Map<RequestInfo, Promise<any>>();
```
zbigg, line 108: Minor,.

With new typings, this sentence found later in docs:

> This method merges multiple downloads of the same resource to (...)

Is becoming true only for if you pass `string` as first arg. Downloads initiated with `Request` instances which are typically temporary objects won't be shared.

haraldF: Thanks @zbigg, I've updated the docs

PR #1498, opened 2020-04-30
HARP-10138: Fix state handling when stopping zoom animation.

zbigg:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-map-controls/lib/MapControls.ts
+++ b/@here/harp-map-controls/lib/MapControls.ts
@@ -707,6 +707,7 @@ export class MapControls extends THREE.EventDispatcher {
     }
 
     private handleZoom() {
+        let resetZoomState = false;
         if (this.m_zoomDeltaRequested !== 0) {
             this.m_targetedZoom = Math.max(
                 Math.min(this.zoomLevelTargeted + this.m_zoomDeltaRequested, this.maxZoomLevel),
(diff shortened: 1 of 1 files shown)


@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -754,6 +753,10 @@ export class MapControls extends THREE.EventDispatcher {
             this.m_maxTiltAngle
         );
 
+        if (resetZoomState) {
+            this.m_targetedZoom = undefined;
+            this.m_currentZoom = undefined;
```
ninok, line 758: Why do we even have m_currentZoom? Are there situations when it will be different then mapview.zoomLevel?

nzjony: I'm not a MapControls expert, but I think it is used for animating / inertia.

ninok: Yeah, I would understand m_targetedZoom for the animation

ninok: Since this is unrelated to this change but a bigger cleanup for MapControls let's keep it for now.

nzjony: m_currentZoom is calculated as:

```
this.currentZoom =
            !this.inertiaEnabled || Math.abs(this.zoomLevelTargeted - this.m_startZoom) < EPSILON
                ? this.zoomLevelTargeted
                : this.easeOutCubic(
                      this.m_startZoom,
                      this.zoomLevelTargeted,
                      Math.min(1, this.m_zoomAnimationTime / this.zoomInertiaDampingDuration)
                  );

```

Do you want to store this directly in the mapView?

nzjony: Agreed it should be cleaned up!

zbigg: I can only say that i also wanted to understand internal state handling on MapControls and real meaning of `currentZoom` but i failed. That's why this patch is minimal so and doesn't try to change big picture of things.

PR #1500, opened 2020-04-30
HARP-10278: Adds Camera Key Track Animations

FraukeF:

# CI Check

Happy contributing!


@here/harp-examples/src/getting-started_camera-animations.ts
```diff
@@ -0,0 +1,220 @@
+/*
```
ninok, line 1: There is already a cmera animation example in the three.js section. I think we can remove it.

@here/harp-examples/src/getting-started_camera-animations.ts
```diff
@@ -0,0 +1,220 @@
...
+                    ...CameraAnimationBuilder.getPoseFromView(map),
+                    timestamp: 0
+                },
+                target,
+                5000,
+                10
+            );
+            cameraAnimation = new CameraKeyTrackAnimation(map, flyToOpts);
+            // end:snippet:harp_gl_camera_animation_1.ts
+            cameraAnimation.start();
```
ninok, line 162: I think this API is a bit complicated.
I would exect sth. like
```
MapView.lookAt({
    target,
    tilt,
    heading,
    zoomLevel,
    animation: MapView.BOW_ANIMATION
});
```
So there are two fundamental questions:
- Which module should contain the animations harp-mapview or harp-map-controls
- How does the API work. With a separate class (i.e CameraAnimation) or inside mapview

FraukeF: Regarding the API, I think they do not exclude each other and an extra parameter in the lookAt function could live in parallel with the CameraAnimationBuilder,and therefore give the user the choice to use what is needed. Although I am a bit hesitant to add the whole Animation logic to the MapView, as in many use cases it might not be needed, or users might want to add their own modules for the animation. 
Already in the harp.gl examples there is 3 ways on how animate, in the webpage and test another 2, so we might not want to limit users to this one way of adding animations.

Regarding the module, I think this is interconnected with the API question, if the animations should be tightly interwoven in the MapView (like a parameter in the lookAt) and will become a dependency anyway, then they might be better placed in the harp-mapview module. If we keep them as a separate addon like the current MapAnimations and Controls, there would be the place.

@here/harp-examples/src/getting-started_camera-animations.ts
FraukeF: It works differently and is basically also a valid way of animation, but maybe a better name and place for both would do, maybe an "animations" submenu, which contains the examples?

@here/harp-examples/src/camera-animations_key-track.ts
```diff
@@ -0,0 +1,220 @@
+/*
+ * Copyright (C) 2017-2020 HERE Europe B.V.
```
nzjony, line 2: Please remove 2017

@here/harp-examples/src/camera-animations_key-track.ts
```diff
@@ -0,0 +1,220 @@
...
+    ControlPoint
+} from "@here/harp-map-controls";
+import { CopyrightElementHandler, MapView } from "@here/harp-mapview";
+import { APIFormat, AuthenticationMethod, OmvDataSource } from "@here/harp-omv-datasource";
+import { GUI } from "dat.gui";
+
+import { apikey, copyrightInfo } from "../config";
+
+interface IData {
+    [key: string]: any;
```
nzjony, line 19: You can be a slight bit more type safe
```suggestion
    [key: string]: GeoCoordinates;
```

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,216 @@
...
+import { LoggerManager } from "@here/harp-utils";
+import THREE = require("three");
+
+import { CameraAnimation } from "./MapAnimations";
+import { CameraPose, CameraPoseOptions } from "./Utils";
+
+const logger = LoggerManager.instance.create("CameraKeyTrackAnimation");
+
+export interface ControlPointOptions extends CameraPoseOptions {
+    timestamp: number;
```
nzjony, line 16: what does the `timestamp` number mean / which units is it in?

@here/harp-examples/src/camera-animations_key-track.ts
```diff
@@ -0,0 +1,220 @@
...
+    // end:harp_gl_camera_camera_animation_3.ts
+
+    const gui = new GUI({ width: 300 });
+    gui.add(options, "globe").onChange(() => {
+        map.projection = options.globe ? sphereProjection : mercatorProjection;
+    });
+    gui.add(options, "orbit")
+        .onChange(enableOrbit)
+        .listen();
+    gui.add(options, "flyTo", ["", "BerlinTower", "BerlinReichstag", "Dubai", "BerlinStation"])
```
nzjony, line 109: please generate the options from the keys of the `geoLocations` if not too difficult

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,178 @@
+/*
+ * Copyright (C) 2017-2019 HERE Europe B.V.
```
nzjony, line 2: remove 2017

@here/harp-map-controls/lib/Utils.ts
```diff
@@ -86,3 +86,35 @@ export function getWidthAndHeightFromCanvas(
...
+    geoCoordinate: GeoCoordinates;
+    distance?: number;
+    azimuth?: number;
+    altitude?: number;
+    name?: string;
+}
+/**
+ * Defines a Camera Pose, meaning its target, direction and distance.
+ */
+export class CameraPose {
```
nzjony, line 100: please add units, i.e. degrees or radians for `azimuth`, `altitude` and `distance`

@here/harp-map-controls/lib/Utils.ts
```diff
@@ -86,3 +86,35 @@ export function getWidthAndHeightFromCanvas(
         height: safeParseDecimalInt(canvas.style.height, canvas.clientHeight)
     };
 }
+
+export interface CameraPoseOptions {
```
nzjony, line 90: You can use a partial type, instead of having to re type the properties, see: https://www.typescriptlang.org/docs/handbook/utility-types.html#partialt

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,216 @@
...
+                : THREE.InterpolateSmooth;
+
+        const loop = this.m_options.loop !== undefined ? this.m_options.loop : THREE.LoopOnce;
+        const repetitions =
+            this.m_options.repetitions !== undefined ? this.m_options.repetitions : 1;
+
+        this.m_options.rotateOnlyClockwise =
+            this.m_options.rotateOnlyClockwise !== undefined
+                ? this.m_options.rotateOnlyClockwise
+                : true; //default to true
```
nzjony, line 78: Please add the defaults to the type definition, so users of this class don't have to dig into the internals of its usage.

@here/harp-map-controls/lib/Utils.ts
```diff
@@ -86,3 +86,35 @@ export function getWidthAndHeightFromCanvas(
...
+    geoCoordinate: GeoCoordinates;
+    direction: {
+        azimuth: number;
+        altitude: number;
+    };
+    distance: number;
+    name?: string;
+
+    constructor(options: CameraPoseOptions) {
+        this.geoCoordinate = options.geoCoordinate
```
nzjony, line 110: truthy

You can instead use the ?? operator, i.e. 
```suggestion
        this.geoCoordinate = options.geoCoordinate ?? new GeoCoordinates(0,0);
```

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,178 @@
...
+            azimuth: THREE.MathUtils.radToDeg(Math.PI + sphericalCoordinates.azimuth),
+            altitude: THREE.MathUtils.radToDeg(sphericalCoordinates.tilt),
+            distance: mapView.targetDistance
+        });
+
+        return cameraPose;
+    }
+
+    /**
+     * Sets the view to the defiend [[CameraPose]]
```
nzjony, line 39:
```suggestion
     * Sets the view to the defined [[CameraPose]]
```

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /**
+     * Appends a [[ControlPoint]] to [[CameraKeyTrackAnimationOptions]]
+     *
+     * @param options
+     * @param point
+     */
+    static appendControlPoint(options: CameraKeyTrackAnimationOptions, point: ControlPoint) {
+        const timestamp = options.controlPoints[options.controlPoints.length - 1].timestamp + 10;
```
nzjony, line 57: what does the +10 mean? Is it ms or seconds? Should it be configurable? Or is the default satifisfactory?

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,178 @@
...
+        const midCoord0 = this.geoCoordLerp(
+            startControlPoint.geoCoordinate,
+            targetControlPoint.geoCoordinate,
+            0.5
+        );
+        const midPoint0 = new ControlPoint({
+            geoCoordinate: midCoord0,
+            distance: maxAltitude,
+            timestamp: duration / 2,
+            altitude: 0,
```
nzjony, line 122: What does this mean?

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,216 @@
...
+        this.m_mapView.removeEventListener(MapViewEventNames.Render, this.m_animateCb);
+        this.m_animationMixer.removeEventListener("finished", this.m_stopCb);
+        this.m_mapView.endAnimation();
+        if (this.onFinished !== undefined) {
+            this.onFinished();
+        }
+        this.running = false;
+    }
+
+    private updateCameraFromDummy() {
```
nzjony, line 184: what does the term "dummy" refer to?

@here/harp-examples/src/getting-started_camera-animations.ts
nzjony: I like that this doesn't add anything to the MapView, it could be done in future, but I think best keep it as separate for as long as possible.

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,216 @@
...
+        });
+        const posValues: [number?] = [];
+        const rotValues: [number?] = [];
+        this.m_options.controlPoints.map(point => {
+            const worldPos = this.m_mapView.projection.projectPoint(point.geoCoordinate);
+            posValues.push(worldPos.x);
+            posValues.push(worldPos.y);
+            posValues.push(worldPos.z);
+
+            const rot = new THREE.Quaternion();
```
zbigg, line 91: I have strong feeling that we don't need quaternions here, but only properly interpolate `LookAtParams.heading` and `LookAtParams.tilt` between particular animation frames. Only suggestion, ...

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,178 @@
...
+            azimuth: THREE.MathUtils.lerp(
```
zbigg, line 123: This is not proper mid angle. If start is 1 and end is 359, it will attempt to rotate almost full circle to get to mid point.

Please use `MathUtils.interpolateAnglesDeg` from geoutils

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,216 @@
...
+    timestamp: number = 0;
+    constructor(options: ControlPointOptions) {
+        super(options);
+        this.timestamp = options.timestamp;
+    }
+}
+
+export interface CameraKeyTrackAnimationOptions {
+    controlPoints: ControlPoint[];
+    repeat?: boolean;
```
zbigg, line 31: Clarify `repeat`, `loop` and `repetitions` ...

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,216 @@
...
+    constructor(name: string) {
+        super();
+        this.name = name;
+    }
+}
+
+/**
+ * A [[CameraAnimation]] along of a set of [[ControlPoint]]s
+ */
+export class CameraKeyTrackAnimation extends CameraAnimation {
```
zbigg, line 51: What is benefit of inheriting from this class ? It completly ignores original implementation and bring unnecessary code as dependency.

If interfacing with old animations and new animations in code is important consider creation of ICameraAnimation interface.
This for example brings whole tweenjs

@here/harp-examples/src/getting-started_camera-animations.ts
FraukeF: I moved both examples under Camera Animations

@here/harp-examples/src/camera-animations_key-track.ts
FraukeF: done

@here/harp-examples/src/camera-animations_key-track.ts
FraukeF: done

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: done

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
FraukeF: done

@here/harp-map-controls/lib/Utils.ts
FraukeF: Removed the CameraPose completly as it basically duplicates the LookAtParams from MapView

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: done

@here/harp-map-controls/lib/Utils.ts
FraukeF: This part of the code got removed completly

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
FraukeF: added a descriptive variable instead

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: I added a comment in the code

@here/harp-map-controls/lib/Utils.ts
```diff
@@ -86,3 +86,35 @@ export function getWidthAndHeightFromCanvas(
         height: safeParseDecimalInt(canvas.style.height, canvas.clientHeight)
     };
 }
+
+export interface CameraPoseOptions {
+    geoCoordinate: GeoCoordinates;
+    distance?: number;
+    azimuth?: number;
+    altitude?: number;
```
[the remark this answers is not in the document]
FraukeF: I removed this part and adjusted to the LookAtParams as you suggested, this code had been grown in a separate repository ,in the same time as the LookAtParams, therefor the duplication

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,216 @@
...
+
+            const rot = new THREE.Quaternion();
+            rot.setFromAxisAngle(
+                this.m_azimuthAxis,
+                THREE.MathUtils.degToRad(point.direction.azimuth)
+            );
+            rot.multiply(
+                new THREE.Quaternion().setFromAxisAngle(
+                    this.m_altitudeAxis,
+                    THREE.MathUtils.degToRad(point.direction.altitude)
```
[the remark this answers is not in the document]
FraukeF: I removed this part and adjusted to the LookAtParams as you suggested, this code had been grown in a separate repository ,in the same time as the LookAtParams, therefor the duplication

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: I have not tried, but I was hoping for a more constant animation speed and avoidance of gimbal lock this way

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: done

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: It actually implements the parts of the abstract class CameraAnimation, although the usage is minor, it still applies.But good point, that the CameraAnimation is actually carved for the usage with tween. I ll remove the dependency.

@here/harp-examples/src/camera-animations_key-track.ts
FraukeF: done

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
FraukeF: I added a comment in the code

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
FraukeF: done

@here/harp-map-controls/lib/Utils.ts
FraukeF: done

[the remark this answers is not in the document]
FraukeF: Good point, I changed this to use the LookAtParams, the animations "were grown" in another project on an older version of harp.gl, where the LookAtParams, did not yet exist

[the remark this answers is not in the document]
FraukeF: see other related comments, CameraPose is removed

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,289 @@
...
+    target: GeoCoordinates | GeoCoordLike;
+    /**
+     * A timestamp for the [[ControlPoint]], describes when during the animation the
+     * [[ControlPoint]] should be reached, in seconds
+     */
+    timestamp: number;
+    /**
+     * @override
+     */
+    tilt?: number;
```
ninok, line 26: Why are these ones here again? They are already part of LookAtParams.

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,289 @@
...
+    distance?: number;
+    /**
+     * A name for the ControlPoint
+     */
+    name?: string;
+}
+/**
+ * A ControlPoint used to defined the Track of the [[CameraKeyTrackAnimation]]
+ */
+export class ControlPoint
```
ninok, line 43: Why is this a class and why do we need this in addition to ControlPointOptions?
Or maybe the other way around, why do we need ControlPointOptions?

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,289 @@
...
+     * The time, when this CameraPose should be reached in the animation flow, in seconds.
+     */
+    timestamp: number = 0;
+    target: GeoCoordinates;
+    name?: string;
+    tilt: number;
+    heading: number;
+    distance: number;
+
+    constructor(options: ControlPointOptions) {
```
ninok, line 55: So as far as I can see, the difference between ControlPoint and ControlPointOptions is, that everything is defined, but not sure if this is even correct.
I would assume if I don't define the tilt for a control point the previous tilt is used in this control point. Below it is enforced to be 0

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,289 @@
...
+
+export interface CameraKeyTrackAnimationOptions {
+    /**
+     * The [[ControlPoint]]s the animation should pass
+     */
+    controlPoints: ControlPoint[];
+    /**
+     * If ´true` the animation restarts after finishing, therefore repeats. @default false.
+     */
+    repeat?: boolean;
```
ninok, line 75: Isn't this already doable with `loop`?

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -0,0 +1,289 @@
...
+    /**
+     * If `true`the Animation will interpolate rotations only clockwise.
+     * @default true
+     */
+    rotateOnlyClockwise?: boolean;
+    /**
+     * Defines how often the Animation should be repeated.
+     * @default 1
+     */
+    repetitions?: number;
```
ninok, line 99: How does this play together with `loop`. What happens for example if loop is `LoopOnce` and repetitions is 2?

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: Its a class to get sane defaults and the ControlPointOptions are there to not have to define everything at once. 
Basically new ControlPoint({target: <my-geo-coordinate>} gives you a valid ControlPoint

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: Yes, but LookAtParams is an interface it still needs implementation.... (typescript will complain otherwise)

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: In the current implementation the ControlPoints don't know of each other, the Animation expects a "complete" ControlPoint for each step.

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: Oh, good catch... the repetitions and loop actually go together in the THREE.AnimationAction         this.m_animationAction.setLoop(loop, repetitions);

But the "repeat" was a leftover, removed it

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: good catch, yes, removed that one, its not even used

@here/harp-examples/src/camera-animations_key-track.ts
```diff
@@ -0,0 +1,214 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
```
nzjony, line 2: Sorry frauke, advised us that this needs to be 2017-2020, see the message from Ignacio in the Core Team webex chat.

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Appends a [[ControlPoint]] to [[CameraKeyTrackAnimationOptions]]
+     *
+     * @param options The [[CameraKeyTrackAnimationOptions]] that the [[ControlPoint]] should be
+     * appended to
+     * @param point The [[ControlPoint]] to append.
+     * @param appendTime The time it should take from the former end of the animation to the
+     * appended [[ControlPoint]] in seconds, @default 10
+     */
+    static appendControlPoint(
```
nzjony, line 41: I don't see any usage of this function, can it be removed?

If not, please at least add a unit test to test it.

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,185 @@
...
+        options: CameraKeyTrackAnimationOptions,
+        point: ControlPoint,
+        appendTime?: number
+    ) {
+        appendTime = appendTime ?? 10;
+        const timestamp =
+            options.controlPoints[options.controlPoints.length - 1].timestamp + appendTime;
```
nzjony, line 48: this should probably check for an empty controlPoints array

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * Adds a [[ControlPoint]] to the beginning of an [[CameraKeyTrackAnimationOptions]]
+     *
+     * @param options
+     * @param point
+     * @param prependTime The time the animation from the inserted key to the next should take
+     */
+    static prependControlPoint(
```
nzjony, line 60: Can't find any usage of this, so please remove.

If not, please at least add a unit test to test it.

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,188 @@
...
+        startControlPoint: ControlPoint,
+        targetControlPoint: ControlPoint,
+        altitude?: number,
+        duration = 10
+    ): CameraKeyTrackAnimationOptions {
+        const controlPoints: ControlPoint[] = [startControlPoint];
+        const maxAltitude =
+            altitude ||
+            2 *
+                mapView.camera.position.distanceTo(
```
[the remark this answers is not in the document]
FraukeF: good catch, yes, this is some leftover, it should take the startPosition not assume its the same with the camera

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,188 @@
...
+     * Creates Options for a Bow Animation from the start to the target [[ControlPoint]]
+     *
+     * @param mapView
+     * @param startControlPoint
+     * @param targetControlPoint
+     * @param altitude The maximal altitude the bow should have, defaults to twice the start to
+     * target distance
+     * @param duration The duration of the Animation in seconds, @default 10
+     */
+    static createBowFlyToOptions(
```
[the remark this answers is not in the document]
FraukeF: Yes, its intentional there, and an adaption on the native implementation and used in the Tokyo App,  but I think there is way more options on how this could look or be configured, when and if there CameraKeyTrackAnimation gets used more, maybe we will find some more  useful AnimationOptions that might be nice to have in the Builder

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
```diff
@@ -0,0 +1,188 @@
...
+     */
+    static createOrbitOptions(
+        startControlPoint: ControlPoint,
+        duration = 10
+    ): CameraKeyTrackAnimationOptions {
+        const amountOfKeys = 4;
+        const controlPoints: ControlPoint[] = [startControlPoint];
+
+        const steps = amountOfKeys - 1;
+        const headingStep = (360 + 45) / steps;
```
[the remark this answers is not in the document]
FraukeF: Thats also a leftover from the tokyo app, sorry for that,  ill make it 360

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: I added these parameters to the example, so it can be tried out how it works together

@here/harp-map-controls/lib/CameraAnimationBuilder.ts
FraukeF: done

@here/harp-map-controls/test/CameraAnimationBuilderTest.ts
```diff
@@ -0,0 +1,138 @@
...
+            assert.equal(animationOptions.controlPoints[1].target.latitude, 0);
+            assert.equal(animationOptions.controlPoints[1].timestamp, 20);
+        });
+
+        it("appends a control point with default time", function() {
+            CameraAnimationBuilder.appendControlPoint(animationOptions, controlPoint_2);
+
+            assert.equal(animationOptions.controlPoints.length, 3);
+            assert.equal(animationOptions.controlPoints[2].target.latitude, 2);
+            assert.equal(animationOptions.controlPoints[2].timestamp, 110);
```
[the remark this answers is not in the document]
FraukeF: the time between the keyframes defaults to 10 if not set explicitly

@here/harp-map-controls/test/CameraAnimationBuilderTest.ts
```diff
@@ -0,0 +1,138 @@
...
+
+            assert.equal(animationOptions.controlPoints.length, 3);
+            assert.equal(animationOptions.controlPoints[0].timestamp, 0);
+            assert.equal(animationOptions.controlPoints[0].target.latitude, 2);
+            assert.equal(animationOptions.controlPoints[1].target.latitude, 0);
+            assert.equal(animationOptions.controlPoints[1].timestamp, 20);
+        });
+
+        it("appends a control point with default time", function() {
+            CameraAnimationBuilder.appendControlPoint(animationOptions, controlPoint_2);
```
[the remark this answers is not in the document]
FraukeF: done

@here/harp-map-controls/test/CameraAnimationBuilderTest.ts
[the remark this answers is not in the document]
nzjony: That makes sense, but from an API point of view I find it a bit strange, for example, if I setup the ControlPoints using some UI (say the x axis represents the time from the start of the animation), and I place two control points at 1 and 2 seconds respectively, then to append the 2nd to the 1st, I would need to call the function as:

CameraAnimationBuilder.appendControlPoint(animationOptions, controlPoint_2, controlPoint_2.timestamp);

I think it makes most sense not to have any side effects for the ControlPoints unless you reach a condition that makes no sense, i.e. when the 2nd timestamp is less than the first, in which case fixing it makes sense.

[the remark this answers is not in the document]
FraukeF: Yes, agree, that makes sense, I had another usecase, when I started it, where the ControlPoints were taken from the current camera poses, therefore there were no timestamps available, but this still could be done with your proposed API. I ll change that.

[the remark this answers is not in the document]
FraukeF: done

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
ninok: but `ControlPointOptions` is also just an interface

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
ninok: It's not so much about the control points knowing each other but to allow control points to be incomplete.
If the user for example just want's to animate the target and all other parameters should stay untouched it's a bit clumsy to use.
You have to get the current parameters like tilt and heading first and then assign it to every control points.
You also have to do this every time instead of just having one constant curve defined in the beginning.

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: so, found a solution, the types did not all match, but do now, and I used Pick over Partial, as it only adds the parameters needed.

@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
FraukeF: For the current use cases, the separate controlpoints worked fine, as they were basically recorded from an ongoing scene and then combined, but I get the point, it might be nice to modify the 3 keytracks for target, rotation and distance in a more independent manner. I propose to do this in a follow up and maybe add some timeline example using it.

PR #1510, opened 2020-05-06
.ort.yml: Exclude test dirs and replace deprecated reason

tsteenbe:
Updating .ort.yml to mark several dirs used for testing as not included in released artifacts to npmjs.com. Replaced scope exclude reason that has been deprecated in OSS Review Toolkit.

See also:
https://github.com/oss-review-toolkit/ort/blob/master/docs/config-file-ort-yml.md.

Signed-off-by: Thomas Steenbergen <thomas.steenbergen@here.com>

### DIFF
--- a/.ort.yml
+++ b/.ort.yml
@@ -1,6 +1,15 @@
 excludes:
+  paths:
+  - pattern: "test/**"
+    reason: "TEST_OF"
+    comment: "Directory is only used for testing. Not included in released artifacts."
+  - pattern: "@here/*/test/**"
+    reason: "TEST_OF"
+    comment: "Directory is only used for testing. Not included in released artifacts."
+  - pattern: "@here/harp-test-utils/**"
+    reason: "TEST_OF"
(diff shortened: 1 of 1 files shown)


.ort.yml
```diff
@@ -1,6 +1,15 @@
 excludes:
+  paths:
+  - pattern: "test/**"
+    reason: "TEST_OF"
+    comment: "Directory is only used for testing. Not included in released artifacts."
+  - pattern: "@here/*/test/**"
+    reason: "TEST_OF"
+    comment: "Directory is only used for testing. Not included in released artifacts."
+  - pattern: "@here/harp-test-utils/**"
```
zbigg, line 9: We release this package -> https://www.npmjs.com/package/@here/harp-test-utils
Maybe we shouldn't, but we do.
It's internal, but it's useful :)

tsteenbe: Thx @zbigg, created a fix-up PR in https://github.com/heremaps/harp.gl/pull/1512

PR #1512, opened 2020-05-06
.ort.yml: Remove @here/harp-test-utils as it's publisbed

tsteenbe:
@here/harp-test-utils is part of the released artifacts, see
https://www.npmjs.com/package/@here/harp-test-utils.

Signed-off-by: Thomas Steenbergen <thomas.steenbergen@here.com>


musculman:
@harpgl-bot ok to test

PR #1513, opened 2020-05-06
Fix CI on windows.

zbigg:
Fix CI on windows.
    
On windows, batch files are only recognized if you try to execute them via shell, so trampoline scripts like `npx` or `mocha-webdriver-runner` are not found we attempt to run them directly by system call.


zbigg:

PR #1515, opened 2020-05-06
HARP-10163: sunny coffee example

valerii-zinchenko:
The use case of this example is to show shadows for a location at a certain time of the day, so that a user can search around for a sunny place.


ninok:
@harpgl-bot ok to test

@here/harp-examples/src/real-time-shadows.ts
```diff
@@ -0,0 +1,324 @@
+/*
+ * Copyright (C) 2020 HERE Europe B.V.
```
nzjony, line 2: Should be 2017-2020

valerii-zinchenko: updated

PR #1518, opened 2020-05-06
HARP-10184: Fix tile's geo bbox computation.

atomicsulfate:
HARP-10184: Fix tile's geo bbox computation.
- Use always the geo box computed by the tiling scheme to ensure it covers the whole tile, since it's used later to generate uv coordinates for displacement maps.
- Elevate geo box using elevation range and maximum geometry height (if available).
- Remove wrong unproject of bounding box provided in decodedTile.

### DIFF
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -326,6 +326,7 @@ export class Tile implements CachedResource {
     ) {
         this.geoBox = this.dataSource.getTilingScheme().getGeoBox(this.tileKey);
         this.updateBoundingBox();
+        this.m_worldCenter.copy(this.boundingBox.position);
         this.m_localTangentSpace = localTangentSpace !== undefined ? localTangentSpace : false;
         this.m_textStyleCache = new TileTextStyleCache(this);
     }
@@ -607,11 +608,12 @@ export class Tile implements CachedResource {
         this.m_elevationRange.minElevation = elevationRange.minElevation;
         this.m_elevationRange.maxElevation = elevationRange.maxElevation;
         this.m_elevationRange.calculationStatus = elevationRange.calculationStatus;
+        this.elevateGeoBox();
 
         // Only elevate bounding box if tile has already been decoded and a maximum geometry height
         // is provided by the data source.
         if (this.m_maxGeometryHeight !== undefined) {
-            this.elevateBoundingBox();
+            this.updateBoundingBox();
         }
     }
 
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/Tile.ts
```diff
@@ -989,27 +988,19 @@ export class Tile implements CachedResource {
...
             this.m_boundingBox.copy(newBoundingBox);
-
-            // Update geo box to match the given bounding box.
-            const tmpPos = this.m_boundingBox.position.clone().add(this.m_boundingBox.extents);
-            this.geoBox.northEast.copy(this.mapView.projection.unprojectPoint(tmpPos));
-            tmpPos.copy(this.m_boundingBox.position).sub(this.m_boundingBox.extents);
-            this.geoBox.southWest.copy(this.mapView.projection.unprojectPoint(tmpPos));
+            this.m_worldCenter.copy(this.boundingBox.position);
         } else {
             this.projection.projectBox(this.geoBox, this.boundingBox);
```
kkostecki, line 993: I think that this.m_worldPos should be also updated here, after this.geoBox projection. Let's consider scenario:
1. Tile has been decoded.
2. Bounding box not provided via decoded tile.
3. GeoBox has been updated (elevated) based on decodedTile.maxGeometryHeight
4.  Method updateBoundingBox() gets called with undefined parameter, so this.boundingBox gets updated, but world position not.

Please check, is this intended?

@here/harp-mapview/lib/Tile.ts
```diff
@@ -607,11 +608,12 @@ export class Tile implements CachedResource {
         this.m_elevationRange.minElevation = elevationRange.minElevation;
         this.m_elevationRange.maxElevation = elevationRange.maxElevation;
         this.m_elevationRange.calculationStatus = elevationRange.calculationStatus;
+        this.elevateGeoBox();
 
         // Only elevate bounding box if tile has already been decoded and a maximum geometry height
```
kkostecki, line 613: Maybe better to say "update" bounding box, because it is the name of the method call.

@here/harp-mapview/lib/Tile.ts
```diff
@@ -607,11 +608,12 @@ export class Tile implements CachedResource {
...
         // is provided by the data source.
         if (this.m_maxGeometryHeight !== undefined) {
-            this.elevateBoundingBox();
+            this.updateBoundingBox();
```
kkostecki, line 616: I know that `this.maxGeometryHeight` should be undefined in the case decoded tile provides fine tuned bounding box, but for a sake of simplicity it could be:
```
this.updateBoundingBox(this.m_decodedTile?.boundingBox);
```
Otherwise is look like we are loosing accurate bbox calculated in decoding phase.

@here/harp-mapview/lib/Tile.ts
atomicsulfate: done

@here/harp-mapview/lib/Tile.ts
atomicsulfate: As you say, maxGeometryHeight and boundingBox are mutually exclusive, either you specify one or the other.
In decodedTile setter there's this code:
`        this.m_maxGeometryHeight = decodedTile.boundingBox
            ? undefined
            : decodedTile.maxGeometryHeight ?? 0;
`
Which ensures that if we have a bounding box, then the max geometry height is undefined. I've added an assertion inside this if to enforce that condition.

kkostecki: Cool!

@here/harp-mapview/lib/Tile.ts
atomicsulfate: The tile center used to compute local tile coordinates in the decoder is not elevated. If we update the tile center so that it matches the center of the elevated bounding box, the geometry will be rendered above of where it should actually be.

When a boudingbox is provided by the decoder though, we assume that the geometry is in local coordinates using the bounding box center as reference. So there we do update the tile center.

kkostecki: Ok, understood the reasons now... so that was probably the cause of bug. That's all fine, it is little pity that you did not added such explanation (maybe in updateBoundingBox) so everything is clear for the other devs.

PR #1521, opened 2020-05-11
HARP-8309 Get LTS at world pos

nzjony:

### DIFF
--- a/@here/harp-geoutils/lib/math/MathUtils.ts
+++ b/@here/harp-geoutils/lib/math/MathUtils.ts
@@ -64,6 +64,19 @@ export namespace MathUtils {
     }
 
     /**
+     * Copies the vector across.
+     *
+     * @param from The vector to copy from.
+     * @param to The resulting [Vector3Like] instance, with the contents copied from from
+     */
+    export function copyVector3<Vector extends Vector3Like>(from: Vector3Like, to: Vector): Vector {
+        to.x = from.x;
+        to.y = from.y;
+        to.z = from.z;
+        return to;
+    }
+
+    /**
      * Converts an angle measured in degrees to an equivalent value in radians.
      *
      * @param degrees Value in degrees.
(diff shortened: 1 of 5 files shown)


@here/harp-geoutils/lib/math/MathUtils.ts
```diff
@@ -63,6 +63,22 @@ export namespace MathUtils {
...
+    /**
+     * Copies the vector across.
+     *
+     * @param from The vector to copy from.
+     * @param to The resulting [Vector3Like] instance, with the contents copied from from
+     */
+    export function copyVector3<Vector extends Vector3Like>(from: Vector3Like, to: Vector): Vector {
+        if (from === undefined || to === undefined) {
+            throw new Error("from or to vector is/are empty");
+        }
```
robertoraggi, line 75: I don't think this checks are needed.

@here/harp-geoutils/lib/projection/MercatorProjection.ts
```diff
@@ -328,6 +328,15 @@ class WebMercatorProjection extends MercatorProjection {
         MathUtils.newVector3(0, 0, -1, result.zAxis);
         return result;
     }
+
+    /** @override */
+    localTangentSpaceFromWorld(worldPos: Vector3Like, result: TransformLike): TransformLike {
```
robertoraggi, line 333: `localTangentSpaceFromWorld` sounds strange to me, how about `localTangetSpaceAtPosition` ?
Alternatively you could overload the signature of localTangentSpace, e.g.
```
localTangentSpace(position: GeoCoordinatesLike | Vector3Like, result: TransformLike) {
    if (!isVector3Like(position)) {
        position = this.projectPoint(position);
   }
   // ...
```

nzjony:

@here/harp-geoutils/lib/projection/MercatorProjection.ts
ninok: I would also prefer to just add an overload for the existing method

ninok: just saw that you did this already. Should look at the code before commenting next time ;-)

PR #1528, opened 2020-05-13
HARP-8309 Remove unused variable

nzjony:
Removing featureId as parameter, because it is extracted from the context.

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvDecoder.ts
+++ b/@here/harp-omv-datasource/lib/OmvDecoder.ts
@@ -89,8 +89,7 @@ export interface IOmvEmitter {
         extents: number,
         geometry: THREE.Vector2[],
         context: AttrEvaluationContext,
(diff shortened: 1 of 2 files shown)


@here/harp-omv-datasource/lib/OmvDecoder.ts
```diff
@@ -248,16 +247,13 @@ export class OmvDecoder implements IGeometryProcessor {
...
-        const featureId = env.lookup("$id") as number | undefined;
-
         if (this.m_decodedTileEmitter) {
             this.m_decodedTileEmitter.processPointFeature(
                 layer,
                 extents,
                 geometry,
                 context,
-                techniques,
-                featureId
```
nzjony, line 260: this featureId was grayed out in VSCode, i.e. it is never used (the featureId is extracted from the context), hence removing it.

PR #1533, opened 2020-05-14
HARP-10452: Merge the style and the technique parameters.

robertoraggi:
Before this change the member "attr" was used to set the technique
parameters and the style properties where stored in the enclosing object.
This was inconsitent and confusing, for example `textMinZoomLevel`
was a member of "attr" but "minZoomLevel" was a member of style.

### DIFF
--- a/@here/harp-datasource-protocol/lib/TechniqueParams.ts
+++ b/@here/harp-datasource-protocol/lib/TechniqueParams.ts
@@ -185,7 +185,7 @@ export interface BaseTechniqueParams {
      * If not specified in style file, [[StyleSetEvaluator]] will assign monotonically increasing
      * values according to style position in file.
      */
-    renderOrder: number;
+    renderOrder?: DynamicProperty<number>;
 
     /**
      * The category of this technique.
@@ -193,7 +193,7 @@ export interface BaseTechniqueParams {
      * The category is used in conjunction with [[Theme.priorities]]
      * to assign render orders to the objects created by this [[Style]].
      */
-    category?: string;
+    category?: DynamicProperty<string>;
 
     /**
      * Optional. If `true`, no IDs will be saved for the geometry this technique creates.
@@ -487,7 +487,7 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
      *
      * See [[ExtendedTileInfo.getFeatureText]]
      */
-    text?: string;
+    text?: DynamicProperty<string>;
 
     /**
      * Field name of object containing the text to be rendered.
@@ -658,7 +658,7 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
     /**
      * Name of [[ImageTexture]] definition to use.
      */
-    imageTexture?: string;
+    imageTexture?: DynamicProperty<string>;
     /**
      * Field name to extract imageTexture content from.
      */
@@ -778,7 +778,7 @@ export interface MarkerTechniqueParams extends BaseTechniqueParams {
      * World space offset in meters applied to the icon. Valid only for icons which have the
      * "offset_direction" property as an attribute of the data.
      */
-    worldOffset?: number;
+    worldOffset?: DynamicProperty<number>;
 }
 
 export interface LineTechniqueParams extends BaseTechniqueParams {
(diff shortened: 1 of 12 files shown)


@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -465,7 +465,7 @@ export type StyleSet = StyleDeclaration[];
  * tile. [[Style]] is describing which features are shown on a map and in what way they are being
  * shown.
  */
-export interface BaseStyle {
+export type BaseStyle<Params> = Partial<Omit<Params, "technique">> & {
```
zbigg, line 468: Omit seems not needed, we're feeding BaseStyle with interfaces without `technique` attr.

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -801,121 +803,106 @@ export type Attr<T> = { [P in keyof T]?: T[P] | JsonExpr };
...
+export interface LineMarkerStyle extends BaseStyle<MarkerTechniqueParams> {
     technique: "line-marker";
-    attr?: Attr<MarkerTechniqueParams>;
 }
 
 /**
  * Render feature as line.
  */
-export interface LineStyle extends BaseStyle {
+export interface LineStyle extends BaseStyle<MarkerTechniqueParams> {
```
zbigg, line 840: `MarkerTechniqueParams` -> `LineTechniqueParams`

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -801,121 +803,106 @@ export type Attr<T> = { [P in keyof T]?: T[P] | JsonExpr };
...
     technique: "line";
     secondaryRenderOrder?: number;
```
zbigg, line 842: Remove secondary ... whey were not supported anyway

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -801,121 +803,106 @@ export type Attr<T> = { [P in keyof T]?: T[P] | JsonExpr };
...
-export interface SegmentsStyle extends BaseStyle {
+export interface SegmentsStyle extends BaseStyle<SegmentsTechniqueParams> {
     technique: "segments";
-    attr?: Attr<SegmentsTechniqueParams>;
 }
 
-export interface SolidLineStyle extends BaseStyle {
+export interface SolidLineStyle extends BaseStyle<SolidLineTechniqueParams> {
     technique: "solid-line" | "dashed-line";
     secondaryRenderOrder?: number;
```
zbigg, line 855: Remove secondary* it's already in SolidLineTechnqueParams ...

@here/harp-datasource-protocol/lib/Theme.ts
robertoraggi: removed

@here/harp-datasource-protocol/lib/Theme.ts
robertoraggi: fixed, i did this change with an automatic script. It looks like the current definition of LineStyle is wrong:
```
export interface LineStyle extends BaseStyle {
    technique: "line";
    secondaryRenderOrder?: number;
    secondaryCategory?: string;
    attr?: Attr<MarkerTechniqueParams>;
}
```

@here/harp-datasource-protocol/lib/Theme.ts
robertoraggi: done

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -185,7 +185,7 @@ export interface BaseTechniqueParams {
      * If not specified in style file, [[StyleSetEvaluator]] will assign monotonically increasing
      * values according to style position in file.
      */
-    renderOrder: number;
+    renderOrder?: DynamicProperty<number>;
```
zbigg, line 188: Slighly related.
I would remove `DynamicProperty` from these definitions and allow everything to be dynamic property - that's what we support though.
But that's another patch i guess.

robertoraggi: yes, i agree that's coming with an upcoming patch.

@here/harp-datasource-protocol/lib/Theme.ts
robertoraggi: this should be fixed.

PR #1535, opened 2020-05-15
HARP-10478: Apply text POIs margin to collision box and text offset.

kkostecki:
Additional text margin (used as padding between icon and text) was not
applied to the text position, resulting in growing collision box, but
without effect on relative distance between label icon and text.
This path fixes it and adds support to bounding box measurement via
TextCanvas that are not 0, 0 anchored. This means that change will work
even without margin applied.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -216,12 +216,26 @@ function computePointTextOffset(
     offset.x = textElement.xOffset;
     offset.y = textElement.yOffset;
 
+    switch (placement.h) {
+        case HorizontalPlacement.Left:
+            // Already accounts for any margin that is already applied to the text element bounds.
+            offset.x -= textElement.bounds!.max.x;
+            break;
+        case HorizontalPlacement.Right:
+            // Account for any margin applied as above.
+            offset.x -= textElement.bounds!.min.x;
+            break;
+    }
     switch (placement.v) {
         case VerticalPlacement.Top:
             offset.y -= textElement.bounds!.min.y;
             break;
         case VerticalPlacement.Center:
             offset.y -= 0.5 * (textElement.bounds!.max.y + textElement.bounds!.min.y);
+            break;
+        case VerticalPlacement.Bottom:
+            // Accounts for vertical margin that may be applied to the text bounds.
+            offset.y -= textElement.bounds!.max.y;
             break;
     }
 
@@ -292,7 +306,11 @@ const tmpCollisionBox = new CollisionBox();
 const tmpScreenPosition = new THREE.Vector2();
 const tmpTextOffset = new THREE.Vector2();
 const tmp2DBox = new Math2D.Box();
-const pointLabelMargin = new THREE.Vector2(4, 2);
+
+/**
+ * The margin applied to the text bounds of every point label.
+ */
+export const persistentPointLabelTextMargin = new THREE.Vector2(2, 2);
 
 export enum PlacementResult {
     Ok,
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -216,13 +216,27 @@ function computePointTextOffset(
...
+    switch (placement.h) {
+        case HorizontalPlacement.Left:
+            // In practice applies: -= pointLabelMargin.x
+            offset.x += textElement.bounds!.max.x;
+            break;
+        case HorizontalPlacement.Right:
+            // Applies: += pointLabelMargin.x
+            offset.x -= textElement.bounds!.min.x;
+            break;
+    }
```
kkostecki, line 228: This code should actually be there regardless of the margin applied, cause without it we assume that text bounding box starts or ends from (0,0), if TextCanvas measurement would do it differently then our offset may be incorrect, in the way that this old code does not support BB placed differently then from 0,0 point.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -216,13 +216,27 @@ function computePointTextOffset(
...
         case VerticalPlacement.Top:
             offset.y -= textElement.bounds!.min.y;
             break;
         case VerticalPlacement.Center:
             offset.y -= 0.5 * (textElement.bounds!.max.y + textElement.bounds!.min.y);
             break;
+        case VerticalPlacement.Bottom:
+            // Applies: -= pointLabelMargin.y
+            offset.y -= textElement.bounds!.max.y;
+            break;
```
kkostecki, line 239: The same comment as above applies.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -617,6 +631,8 @@ function placePointLabelAtAnchor(
         tmpMeasurementParams.letterCaseArray = label.glyphCaseArray!;
         // Compute label bounds according to layout settings.
         textCanvas.measureText(label.glyphs!, label.bounds, tmpMeasurementParams);
+        // TODO: Make the margin configurable
+        label.bounds.expandByVector(pointLabelMargin);
```
kkostecki, line 635: The biggest question mark is here. Should we apply the margin before or after scaling... Both cases has some strong and weak arguments.
- If we apply margin before scaling (margin is scaled together with other dimensions - font size, offsets) it is directly related to other sizes configured, but it is hard to say what is its final measure on the screen.
-If we apply margin after scaling is done on bounding box, then margin stays always the same (in screen space), but is not related to other measures and sizes. This approach will require performing more sophisticated logic, like with text offset, cause we need to know in which direction margin moves the text.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -625,10 +641,6 @@ function placePointLabelAtAnchor(
     tmpBox.copy(label.bounds!);
     tmpBox.min.multiplyScalar(scale);
     tmpBox.max.multiplyScalar(scale);
-    // Add margin after scaling, this ensures the margin is consistent across all
-    // labels - regardless of distance scaling (or any other) factor.
-    // TODO: Make the margin configurable
-    tmpBox.expandByVector(pointLabelMargin);
     tmpBox.translate(textOffset);
```
kkostecki, line 644: This part was a main problem, margin was applied to bounding box, but text was not moved by its (margin) amount from icon.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: with your change, the margin contributes to the text offset. And since we want the offset to scale, then it also makes sense that the margin does.
So it's ok as you have it here right now.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -216,13 +216,27 @@ function computePointTextOffset(
     offset.x = textElement.xOffset;
     offset.y = textElement.yOffset;
 
+    switch (placement.h) {
+        case HorizontalPlacement.Left:
+            // In practice applies: -= pointLabelMargin.x
```
atomicsulfate, line 221: I think this comment it's difficult to understand. I would just simply mention that it's expected that any margin is already applied to the text element bounds.

PR #1538, opened 2020-05-19
HARP-7073: Use memoization to reduce the number of expr evaluations.

robertoraggi:
This change introduces a simple `MemoCall` that caches
the previous execution of an Expr.

### DIFF
--- a/@here/harp-datasource-protocol/lib/Expr.ts
+++ b/@here/harp-datasource-protocol/lib/Expr.ts
@@ -53,6 +53,11 @@ export class ExprDependencies {
      * `true` if the expression depends on the feature state.
      */
     featureState?: boolean;
+
+    /**
+     * `true` if this expression cannot be cached.
+     */
+    volatile?: boolean;
 }
 
 class ComputeExprDependencies implements ExprVisitor<void, ExprDependencies> {
@@ -103,12 +108,21 @@ class ComputeExprDependencies implements ExprVisitor<void, ExprDependencies> {
         expr.args.forEach(childExpr => childExpr.accept(this, context));
 
         switch (expr.op) {
+            case "dynamic-properties":
+                context.volatile = true;
+                break;
             case "feature-state":
                 context.featureState = true;
                 context.properties.add("$state");
+                context.properties.add("$id");
                 break;
             case "id":
                 context.properties.add("$id");
+                break;
+            case "zoom":
+            case "world-ppi-scale":
+            case "world-discrete-ppi-scale":
+                context.properties.add("$zoom");
                 break;
             case "geometry-type":
                 context.properties.add("$geometryType");
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -7,6 +7,7 @@ import {
     Attachment,
     BaseTechniqueParams,
     BufferAttribute,
+    CallExpr,
     DecodedTile,
     Env,
     Expr,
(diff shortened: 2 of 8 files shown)


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -101,6 +109,41 @@ class AttachmentCache {
...
+    }
+
+    call(context: ExprEvaluatorContext): Value {
+        const expr = this.args[0];
+
+        let changed = false;
+
+        for (const d of this.deps) {
+            const newValue = context.env.lookup(d);
+            if (!changed && newValue !== this.cache.get(d)) {
```
zbigg, line 130: Know that's microoptimization, but `this.deps` is array.
So `this.cache` can be array too.
Only change indexing of this loop to `i`. We save some unnecessary lookups and unnecessary maintenance of small Maps.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -308,6 +348,29 @@ export class TileGeometryCreator {
...
+                if (isJsonExpr(value) && propertyName !== "kind") {
+                    // "kind" is reserved.
+                    try {
+                        let expr = Expr.fromJSON(value);
+                        if (expr.dependencies().volatile !== true) {
+                            expr = new MemoCall(Expr.fromJSON(value));
+                        }
+                        technique[propertyName] = expr.intern(exprPool);
+                    } catch (error) {
+                        logger.error("#TileGeometryCreator: Failed to compile expression:", error);
```
zbigg, line 369: This logger instance already prefixes all logs with "TileGeometryCreator: "...

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -53,6 +53,11 @@ export class ExprDependencies {
      * `true` if the expression depends on the feature state.
      */
     featureState?: boolean;
+
+    /**
+     * `true` if this expression cannot be cached.
+     */
+    volatile?: boolean;
```
zbigg, line 60: Hmm. I have problem with this particular definition of `volatile`.

It actually should read as <<results are dependent on ''dynamic-properties'>> .... but unless i'm missing something, they can be cached very effectively.

In particular `volatile` currenty only expressions with `dynamic-properties` are volatile now, but apparently they can be effectively cached as long as no one calls `MapView.setDynamicProperty` ... and that's easily handled by `MemoCall` already.

I have bigger problem with `$id` and `feature-state` (not marked as volatile?) this is changes so often that i would say it's not feasible to cache them ...

So, summarizing i think that defining `volatile` here is bogus. Only TileGeometryCreator and MapView khow envs will change (per frame / per feature) so only they can know if something is really volatile or not.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
robertoraggi: done

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
robertoraggi: removed

@here/harp-datasource-protocol/lib/Expr.ts
robertoraggi: well, `volatile` is general and it may be used for all the current and future `Expr` that cannot be cached.. e.g. say we add an operator `["debug", expr]` that evaluates expr, prints its value and then returns it. well, in that case `["debug", expr]` should be marked as `volatile`.

`volatile` can also be used in cases where the engine is not yet optimizing the fetching of feature properties.  The current implementation of `setDynamicProperty`/`["dynamic-properties"]` is modifying the `Env` directly and there is no way to detect changes there yet. So at least for now we don't optimize it and mark it as `volatile`, this may change very soon btw.

about `["feature-state"]`, feature-state is creating a completely new property `$state` for each feature and `MemoExpr` will catch that. same for `["id"]` it rely on $id and that changes for each feature.

PR #1540, opened 2020-05-19
Harp 9926: Fix labels flickering while panning on 2d view with cooplanar scale.

kkostecki:

# CI Check

Happy contributing!


@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -617,19 +649,31 @@ function placePointLabelAtAnchor(
...
+    tmpBox.getSize(tmpSize);
+    if (measureText) {
+        // New label padding/margin is applied only for the first time it is placed.
+        // This margin compensates the changes of label bounds due to distance scaling and
+        // it is designed to cover only size changes due to camera panning (not tilt).
+        // This ways label "reserves" enough space so it may "grow" without affecting other
+        // labels, no collisions occurs and thus limits labels flickering on pan.
+        const imprecisionFactor = 1.1;
+        // Account for imprecisions during calculus - increase the scale by 10%.
+        tmpSize.multiplyScalar(maxPlanarScale * imprecisionFactor);
```
kkostecki, line 670: Please note, that we could also simply use some Theme defined / hard-coded margin as above - see:
```
label.bounds.expandByVector(pointLabelMargin);
```
Saying that the margin above is `persistentLabelMargin` and the new one is `newLabelMaring`.
Although the effect will be the same the only difference with this PR solution is that we do not have any assumed values but calculate scaling on the fly... of course with some CPU cost.
I am leaving the decision to you dear reviewers :).

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -617,19 +649,31 @@ function placePointLabelAtAnchor(
...
+    } else {
+        tmpSize.multiplyScalar(scale);
+    }
+    tmpBox.setFromCenterAndSize(tmpCenter, tmpSize);
```
kkostecki, line 674: This is the only correct way to scale bounding box, min / max scaling (used before) is wrong.
Consider the box:
min: (-1, -1)
max: (10, 10).
Scaling 2x would result in:
min: (-2, -2)
max: (20, 20).

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1485,7 +1504,7 @@ export class TextElementsRenderer {
         // remains unchanged, the farther is label from that point the smaller size it is
         // rendered in screen space. This method is unaffected by near and far clipping planes
         // distances, but may be improved by taking FOV into equation or customizing the
-        // focus point screen position based on horizont, actual ground, tilt ets.
+        // focus point screen position based on horizon, actual ground, tilt ets.
         let factor = lookAtDistance / distance;
```
atomicsulfate, line 1508: This is what is wrong in distance scaling. The factor should be the coordinate z in view space.

PR #1554, opened 2020-05-25
HARP-9926: Suppress labels flickering while panning on 2d view.

kkostecki:
Some text labels do appear/disappear showing the current solution unstable
even when camera parameters are not changed and user performs top view (2D)
panning.
This problem is caused by labels distance scaling and changes in their
bounds. Because labels closer to screen center becomes bigger they need
more screen space and thus labels' collisions algorithm may drop them out.
The second problem was caused by limiting the number of labels being placed
from now this limit will be applied only to the new labels to be processed.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -306,11 +306,22 @@ const tmpCollisionBox = new CollisionBox();
 const tmpScreenPosition = new THREE.Vector2();
 const tmpTextOffset = new THREE.Vector2();
 const tmp2DBox = new Math2D.Box();
+const tmpCenter = new THREE.Vector2();
+const tmpSize = new THREE.Vector2();
 
 /**
  * The margin applied to the text bounds of every point label.
  */
 export const persistentPointLabelTextMargin = new THREE.Vector2(2, 2);
+/**
+ * Additional bounds scaling (described as percentage of full size) applied to the new labels.
+ *
+ * This additional scaling (margin) allows to account for slight camera position and
+ * orientation changes, so new labels are placed only if there is enough space around them.
+ * Such margin limits collisions with neighboring labels while doing small camera movements and
+ * thus reduces labels flickering.
+ */
+export const newPointLabelTextMarginPercent = 0.1;
 
 export enum PlacementResult {
     Ok,
@@ -645,15 +656,13 @@ function placePointLabelAtAnchor(
     const textOffset = computePointTextOffset(label, placement, scale, env, tmpTextOffset);
     textOffset.add(screenPosition);
     tmpBox.copy(label.bounds!);
-    tmpBox.min.multiplyScalar(scale);
-    tmpBox.max.multiplyScalar(scale);
     tmpBox.translate(textOffset);
-    tmp2DBox.set(
-        tmpBox.min.x,
-        tmpBox.min.y,
-        tmpBox.max.x - tmpBox.min.x,
-        tmpBox.max.y - tmpBox.min.y
-    );
+
+    tmpBox.getCenter(tmpCenter);
+    tmpBox.getSize(tmpSize);
+
+    tmpSize.multiplyScalar(scale);
+    tmp2DBox.set(tmpCenter.x - tmpSize.x / 2, tmpCenter.y - tmpSize.y / 2, tmpSize.x, tmpSize.y);
 
     // Update output screen position.
     outScreenPosition.set(textOffset.x, textOffset.y, labelState.renderDistance);
@@ -672,6 +681,21 @@ function placePointLabelAtAnchor(
         // It might be changed if we would like to render text without icon (at border, etc.).
         return persistent ? PlacementResult.Rejected : PlacementResult.Invisible;
     }
+
+    if (measureText) {
+        // Up-scaled label bounds are used only for new labels and after visibility check, this is
+        // intentional to avoid processing labels out of the screen due to increased bounds, such
+        // labels would be again invisible in the next frame.
+        tmpBox.getSize(tmpSize);
+        tmpSize.multiplyScalar(scale * (1 + newPointLabelTextMarginPercent));
+        tmp2DBox.set(
+            tmpCenter.x - tmpSize.x / 2,
+            tmpCenter.y - tmpSize.y / 2,
+            tmpSize.x,
+            tmpSize.y
+        );
+    }
+
     // Check label's text collision.
     if (!label.textMayOverlap && screenCollisions.isAllocated(tmp2DBox)) {
         // Allows to fade persistent and ignore new label.
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -41,26 +41,29 @@ const MIN_AVERAGE_CHAR_WIDTH = 5;
 
 const tmpPosition = new THREE.Vector3(0, 0, 0);
 const tmpCameraDir = new THREE.Vector3(0, 0, 0);
+const tmpPointDir = new THREE.Vector3(0, 0, 0);
 const COS_TEXT_ELEMENT_FALLOFF_ANGLE = 0.5877852522924731; // Math.cos(0.3 * Math.PI)
 
 /**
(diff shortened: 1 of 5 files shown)


@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -306,11 +306,22 @@ const tmpCollisionBox = new CollisionBox();
...
  * The margin applied to the text bounds of every point label.
  */
 export const persistentPointLabelTextMargin = new THREE.Vector2(2, 2);
+/**
+ * Additional scaling (margin described as percentage of full size) applied to the new labels.
+ *
+ * This margin allows to account for imprecisions and changes in labels' bounds calculus when
+ * labels are changing their screen positions while camera is moving in the same plane.
+ * Those imprecisions may cause the label to appear or disappear when their bounds are in close
+ * proximity duo to collisions.
```
atomicsulfate, line 322: That's only one use of the margin. In general the margin is there for any kind of camera movement. We want only to place new labels if they have a margin around them, so that slight camera changes in position/orientation doesn't case a  collision with neighboring labels.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -672,6 +681,23 @@ function placePointLabelAtAnchor(
...
     }
+
+    if (measureText) {
+        // Bigger label margin is applied only for the first time it is placed and its visible
+        // with its normal bounds. This margin compensates the changes of label bounds due to
+        // distance scaling and it is designed to cover size changes due to camera panning.
+        // This ways label "reserves" enough space so it may "grow" without affecting other
+        // labels, no collisions occurs and thus limits labels flickering on pan.
+        tmpBox.getSize(tmpSize);
+        tmpSize.multiplyScalar(scale * (1 + newPointLabelTextMarginPercent));
```
atomicsulfate, line 692: ok, so you went finally for applying the extra margin to the scaled bbox. it should be fine as soon as we fix how we compute the scaling factor so that it doesn't affect top down views.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -672,6 +681,23 @@ function placePointLabelAtAnchor(
...
+        tmp2DBox.set(
+            tmpCenter.x - tmpSize.x / 2,
+            tmpCenter.y - tmpSize.y / 2,
+            tmpSize.x,
+            tmpSize.y
+        );
```
atomicsulfate, line 698: Here you're recomputing the bbox for new labels again. I guess you're doing this because you want to use the bbox without extra margins to check the label visibility (line 672), is that right? 
I'd rather compute the enlarged bbox for new labels at the top once, and apply it even for visibility check.
What i mean is, just apply the proper scale in line 664 (scale for persistent labels, scale * (1 + newPointLabelTextMarginPercent) for new labels).

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Yep, I have already tried it and seems to work fine, would you like me to add this change to this PR?

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Believe me I have tried this one before, bigger margin for visibility makes new labels visible even if they are not, in the next frame they become invisible again - labels flickers, but behind the scene... It also affects a lot of unit test which are then failing - because of this change in logic.
It has much more sense to apply it after visibility checking, because we do not try to display unnecessary labels, it gives more consistency between first and next frames and the last but not least unit tests logic is preserved.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Would you like to change the comments then?

kkostecki:

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: yes better to give the whole picture, it doesn't mean to make the comment larger though. just mention a larger margin is used for new labels to ensure they'll stay on screen after small camera changes to avoid flickering.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -672,6 +681,23 @@ function placePointLabelAtAnchor(
         // It might be changed if we would like to render text without icon (at border, etc.).
         return persistent ? PlacementResult.Rejected : PlacementResult.Invisible;
     }
+
+    if (measureText) {
+        // Bigger label margin is applied only for the first time it is placed and its visible
+        // with its normal bounds. This margin compensates the changes of label bounds due to
+        // distance scaling and it is designed to cover size changes due to camera panning.
+        // This ways label "reserves" enough space so it may "grow" without affecting other
+        // labels, no collisions occurs and thus limits labels flickering on pan.
```
atomicsulfate, line 690: This comments tells basically the same as the comment for the margin percentage constant. So just leave one of the two, and specify that the extra margin is not only to solve flickering on panning with distance scaling, it's also useful for other scenarios where small camera changes would cause collisions between visible labels that are too close.

kkostecki:

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: ok.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1543,8 +1546,11 @@ export class TextElementsRenderer {
         tempScreenPosition.x = tempPoiScreenPosition.x = screenPosition.x;
         tempScreenPosition.y = tempPoiScreenPosition.y = screenPosition.y;
 
-        // Scale the text depending on the label's distance to the camera.
-        const textDistance = this.m_viewState.worldCenter.distanceTo(position);
+        // Scale the text depending on the label's distance to the camera "zero" plane - near
+        // plane at zero distance. The result is the same as taking the distance from "Z"
+        // coordinate in camera space. This way we avoid co-planar labels' scale variation.
+        const labelCamVec = this.m_tmpLabelCamVec.copy(position).sub(this.m_viewState.worldCenter);
```
atomicsulfate, line 1552: This should be done in computeViewDistance function in Placement.ts. That function should be call from here, as it is done for path labels in addPathLabel. Check that the calculations for path labels makes sense too.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Ok, I have removed that one, but replaced it with some explanation why it is after visibility check. I think it is crucial and may be not so clear to everybody.

@here/harp-mapview/lib/text/Placement.ts
kkostecki: Ok, fixed it. Hope will be better now.

kkostecki:

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1543,8 +1547,12 @@ export class TextElementsRenderer {
...
         tempScreenPosition.y = tempPoiScreenPosition.y = screenPosition.y;
 
-        // Scale the text depending on the label's distance to the camera.
-        const textDistance = this.m_viewState.worldCenter.distanceTo(position);
+        // Scale the text depending on the label's distance to the camera "zero" plane.
+        const textDistance = pointToPlaneDistance(
+            position,
+            this.m_viewState.worldCenter,
+            this.m_cameraLookAt
+        );
```
kkostecki, line 1555: You will probably ask me why I did not used exactly the same function:
```
computeViewDistance(pointLabel, ...)
``` 
with pointLabel (TextElement) as input parameter, unfortunately quick research shows that it is not that simple and sometimes the labels are created and positioned with slightly different position then stored in their state. So I need the position from `addPointLabel` function. See the calls to `addPointLabel` it will probably explain everything.

kkostecki:

kkostecki:

kkostecki:

PR #1566, opened 2020-05-27
Enable type checking in the webpack config files.

robertoraggi:
This change enables the type checking of webpack config files
using the `//@ts-check` directive.


webpack.tests.config.js
```diff
@@ -1,3 +1,11 @@
+/*
+ * Copyright (C) 2017-2018 HERE Europe B.V.
```
nzjony, line 2: 2017-2020

robertoraggi: done

PR #1578, opened 2020-06-02
HARP-10654 Fix incorrect return statement in...

nzjony:
... feature filtering

### DIFF
--- a/@here/harp-omv-datasource/lib/OmvDecoder.ts
+++ b/@here/harp-omv-datasource/lib/OmvDecoder.ts
@@ -223,13 +223,15 @@ export class OmvDecoder implements IGeometryProcessor {
         storageLevel: number
     ): void {
         if (this.m_featureModifiers !== undefined) {
-            this.m_featureModifiers.forEach(fm => {
-                // TODO: The logic of feature ignore should be actually in the feature filtering
-                // mechanism - see OmvFeatureFilter.
-                if (!fm.doProcessPointFeature(layer, env, storageLevel)) {
-                    return;
-                }
-            });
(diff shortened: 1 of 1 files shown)


@here/harp-omv-datasource/lib/OmvDecoder.ts
```diff
@@ -223,13 +223,15 @@ export class OmvDecoder implements IGeometryProcessor {
...
-                // mechanism - see OmvFeatureFilter.
-                if (!fm.doProcessPointFeature(layer, env, storageLevel)) {
-                    return;
-                }
-            });
+            if (
+                this.m_featureModifiers.find(fm => {
+                    // TODO: The logic of feature ignore should be actually in the feature filtering
+                    // mechanism - see OmvFeatureFilter.
+                    return !fm.doProcessPointFeature(layer, env, storageLevel);
```
ninok, line 230: Not sure I understand the logic here. We stop processing the point feature if there is one feature modifier that cannot process point features.
Shouldn't we instead check if there is no feature modifier that can process point features?

nzjony: See the original change: https://github.com/heremaps/harp.gl/commit/9a09fb17c12463dd4932e671184f4e2ad0dfcf82

Because @kkostecki changed it to an array, I am just fixing the the logic to work with arrays. But yes maybe the logic can be improved. But that should be a follow up ticket, this is just a hot fix.

ninok: okay, please create one

PR #1582, opened 2020-06-03
Set up package documentation.

robertoraggi:
Also remove deprecate vscode snippets and use @remarks to clean up
the Theme declarations.


@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -535,215 +512,7 @@ export type BaseStyle<Technique, Params> = Partial<Params> & {
     attr?: Partial<Params>;
 };
 
-/**
- *
- * @defaultSnippets [
```
ninok, line 540: Did this work before? Would be a nice feature to have theses snippets but I guess they could be in a separate file.

PR #1587, opened 2020-06-08
MINOR: Remove unneeded options duplicate

nzjony:


@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -392,9 +391,8 @@ export class VisibleTileSet {
     constructor(
         private readonly m_frustumIntersection: FrustumIntersection,
         private readonly m_tileGeometryManager: TileGeometryManager,
-        options: VisibleTileSetOptions
+        public options: VisibleTileSetOptions
```
ninok, line 394: Shouldn't we create a copy of the options like we do in MapView. Otherwise we could potentially modify the original options from the user.

nzjony: That would be a bit of refactor, the options are used to store the values, see `updateClipPlanes` for example.

The problems is that each class has their own idea of what some options actually mean.

nzjony: I have created HARP-10748 to do the bigger refactoring

PR #1588, opened 2020-06-08
HARP-10715: Use api-extractor to generate rollup d.ts typings.

robertoraggi:
This change introduces a simple api-extractor configuration and a npm
script to generate typings information for the bundled @here/harp.gl
package.

The typings are generated in `@here/harp.gl/dist/harp.gl.d.ts` by
the following command:

```
    yarn workspace @here/harp.gl run build-typings
```

### DIFF
--- a/@here/harp.gl/package.json
+++ b/@here/harp.gl/package.json
@@ -12,10 +12,13 @@
         "harpgl"
     ],
     "main": "dist/harp.js",
+    "types": "dist/harp.gl.d.ts",
     "scripts": {
         "prepare": "webpack && NODE_ENV=production webpack",
+        "prepack": "yarn build-typings",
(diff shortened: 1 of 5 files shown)


@here/harp.gl/package.json
```diff
@@ -12,10 +12,13 @@
         "harpgl"
     ],
     "main": "dist/harp.js",
+    "types": "dist/harp.gl.d.ts",
```
nzjony, line 15:
```suggestion
    "types": "dist/harp.d.ts",
```

nzjony: as mentioned, the type file should be named the same as the file it is covering, otherwise VSCode doesn't know that they belong together

PR #1589, opened 2020-06-08
MINOR: Added generated API docs to allow checking for API changes in CI

ninok:


input/harp-datasource-protocol.api.md
```diff
@@ -0,0 +1,1989 @@
+## API Report File for "@here/harp-datasource-protocol"
+
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
+
+```ts
+
+import { OrientedBox3 } from '@here/harp-geoutils';
```
ninok, line 7:
```suggestion
import { OrientedBox3 } from '@here/harp-geoutils';
import { suggestion } from "TestSuggestion";
```

PR #1592, opened 2020-06-09
HARP-10603: Fix outline rendering

StefanDachwitz:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
+++ b/@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
@@ -766,7 +766,16 @@ export class TileGeometryCreator {
                     bufferGeometry.setIndex(attachment.getBufferAttribute(index));
                 }
 
-                if (!bufferGeometry.getAttribute("normal") && needsVertexNormals(technique)) {
+                // Geometry with edges are typically buildings. If they are buildings, the outline
+                // effect may be applied to it, which needs the normals.
+                const isBuilding =
+                    isExtrudedPolygonTechnique(technique) &&
+                    attachment.info.edgeIndex !== undefined;
+
+                if (
+                    !bufferGeometry.getAttribute("normal") &&
+                    (isBuilding || needsVertexNormals(technique))
+                ) {
                     bufferGeometry.computeVertexNormals();
                 }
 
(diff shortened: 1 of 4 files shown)


@here/harp-materials/lib/EdgeMaterial.ts
```diff
@@ -19,12 +19,17 @@ import { enforceBlending, setShaderDefine, setShaderMaterialDefine } from "./Uti
 const vertexSource: string = `
 #define EDGE_DEPTH_OFFSET 0.0001
 
-attribute vec3 position;
 attribute vec4 color;
 
+// SHADER_NAME may be defined by THREE.JS own shaders in which case these attributes & uniforms are
```
nzjony, line 24: the class `EdgeMaterial` extends `THREE.RawShaderMaterial`, which isn't a THREE.js shader, so my guess is that this will be never defined?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -766,7 +766,14 @@ export class TileGeometryCreator {
                     bufferGeometry.setIndex(attachment.getBufferAttribute(index));
                 }
 
-                if (!bufferGeometry.getAttribute("normal") && needsVertexNormals(technique)) {
+                const isBuilding =
```
nzjony, line 769: This may not be immediately clear to other people reading this why a defined edgeIndex means that the normals should be computed, please add a comment here.

@here/harp-materials/lib/EdgeMaterial.ts
StefanDachwitz: There are cases when it will be defined, which then leads to the following runtime error:

three.min.js:83 THREE.WebGLProgram: shader error:  0 35715 false gl.getProgramInfoLog No compiled vertex shader when at least one graphics shader is attached.
 THREE.WebGLShader: gl.getShaderInfoLog() vertex
ERROR: 0:54: 'position' : redefinition
ERROR: 0:57: 'modelViewMatrix' : redefinition
ERROR: 0:58: 'projectionMatrix' : redefinition
ERROR: 0:59: 'normalMatrix' : redefinition
1: precision highp float;
2: precision highp int;
3: #define HIGH_PRECISION
4: #define SHADER_NAME ShaderMaterial
...

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
StefanDachwitz: Adding comment

StefanDachwitz: Added comment

@here/harp-materials/lib/EdgeMaterial.ts
nzjony: This is because the material is being reconstructed in Outline.ts, I see now. Ok, all good.

PR #1597, opened 2020-06-10
Tile upload throttling HARP-10363

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -265,6 +265,14 @@ export class Tile implements CachedResource {
     skipRendering: boolean = false;
 
     /**
+     * If the tile should not yet be rendered, this is used typically when the tile in question
+     * does not fit into the gpu upload limit of the current frame.
+     * Setting this value directly affects the [[willRender]] method, unless
+     * overriden by deriving classes.
+     */
+    delayRendering = false;
+
+    /**
      * @hidden
      *
      * Prepared text geometries optimized for display.
@@ -607,7 +615,7 @@ export class Tile implements CachedResource {
      * [[skipRendering]] property unless specifically overriden in deriving classes.
      */
     willRender(_zoomLevel: number): boolean {
-        return !this.skipRendering;
+        return !this.skipRendering && !this.delayRendering;
     }
 
     /**
--- a/@here/harp-mapview/lib/VisibleTileSet.ts
+++ b/@here/harp-mapview/lib/VisibleTileSet.ts
@@ -85,6 +85,15 @@ export interface VisibleTileSetOptions {
      * Number of levels to go down when searching for fallback tiles.
      */
     quadTreeSearchDistanceDown: number;
+
+    /**
+     * Maximal number of new tiles, that can be added to the scene per frame.
+     * if set to `0`the limit will be ignored and all available tiles be uploaded.
+     * @beta
+     * @internal
+     * @defaultValue 0
+     */
+    maxTilesPerFrame: number;
 }
 
 const MB_FACTOR = 1.0 / (1024.0 * 1024.0);
@@ -394,6 +403,8 @@ export class VisibleTileSet {
         private readonly m_tileGeometryManager: TileGeometryManager,
         public options: VisibleTileSetOptions
     ) {
+        this.options = options;
+        this.options.maxTilesPerFrame = Math.floor(this.options.maxTilesPerFrame ?? 0);
         this.m_resourceComputationType =
             options.resourceComputationType === undefined
                 ? ResourceComputationType.EstimationInMb
@@ -442,6 +453,28 @@ export class VisibleTileSet {
      */
     setNumberOfVisibleTiles(size: number) {
         this.options.maxVisibleDataSourceTiles = size;
+    }
+
+    /**
+     * Gets the maximum number of tiles that can be added to the scene per frame
+     * @beta
+     * @internal
+     */
+    get maxTilesPerFrame(): number {
+        return this.options.maxTilesPerFrame;
+    }
+
+    /**
+     * Gets the maximum number of tiles that can be added to the scene per frame
+     * @beta
+     * @internal
+     * @param value
+     */
+    set maxTilesPerFrame(value: number) {
+        if (value < 0) {
+            throw new Error("Invalid value, this will result in no tiles ever showing");
+        }
+        this.options.maxTilesPerFrame = Math.floor(value);
     }
 
     /**
@@ -500,6 +533,7 @@ export class VisibleTileSet {
         elevationRangeSource?: ElevationRangeSource
     ): { viewRanges: ViewRanges; viewRangesChanged: boolean } {
         let allVisibleTilesLoaded: boolean = true;
+        let newTilesPerFrame = 0;
 
         const visibleTileKeysResult = this.getVisibleTileKeysForDataSources(
             zoomLevel,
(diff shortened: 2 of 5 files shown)


@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -1023,7 +1065,7 @@ export class VisibleTileSet {
             );
 
             const nextLevelDiff = Math.abs(childTileKey.level - dataZoomLevel);
-            if (childTile !== undefined && childTile.hasGeometry) {
+            if (childTile !== undefined && childTile.hasGeometry && !childTile.delayRendering) {
                 // childTile has geometry, so can be reused as fallback
```
ninok, line 1069: Please adjust the comment:
`childTile has geometry and was/can be uploaded to the GPU, so we can use it as fallback`

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -84,6 +84,12 @@ export interface VisibleTileSetOptions {
      * Number of levels to go down when searching for fallback tiles.
      */
     quadTreeSearchDistanceDown: number;
+
+    /**
+     * Maximal number of new tiles, that can be added to the scene per frame
+     * @beta
```
ninok, line 90: `@defaultValue` ?

@here/harp-mapview/lib/Tile.ts
```diff
@@ -262,6 +262,14 @@ export class Tile implements CachedResource {
      */
     skipRendering = false;
 
+    /**
+     * If the tile should not yet be rendered, this is used typically when the tile in question
+     * does not fit into the gpu upload limit of the current frame.
+     * Setting this value directly affects the [[willRender]] method, unless
+     * overriden by deriving classes.
+     */
+    delayRendering = false;
```
ninok, line 271: I wonder if we really need a new flag. Couldn't we reuse `skipRendering`?

@here/harp-mapview/lib/VisibleTileSet.ts
ninok: I would also mark it as `@internal` b/c it is not expected that users set that flag

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -393,6 +399,8 @@ export class VisibleTileSet {
         private readonly m_tileGeometryManager: TileGeometryManager,
         public options: VisibleTileSetOptions
     ) {
+        this.options = options;
+        this.options.maxTilesPerFrame = Math.floor(this.options.maxTilesPerFrame ?? 100);
```
ninok, line 403: Instead of having an arbitrary number, I would rather use 0 by default and treat 0 as unlimited amount of tiles

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -552,14 +581,27 @@ export class VisibleTileSet {
...
                     this.skipOverlappedTiles(dataSource, tile);
 
                     if (tile.frameNumVisible < 0) {
-                        // Store the fist frame the tile became visible.
-                        tile.frameNumVisible = frameNumber;
+                        newTilesPerFrame++;
+                        //maxTilesPerFrame defaults in the constructor and therefore
+                        //cannot be undefined here
+                        if (
+                            newTilesPerFrame > (this.options.maxTilesPerFrame as number) &&
```
ninok, line 593: `this.options.maxTilesPerFrame!`

@here/harp-mapview/lib/VisibleTileSet.ts
ninok: Looking at all the other properties they are not optional b/c they are resolved in `MapView`.
We should do the same here and add the optional property to `MapViewOptions`

@here/harp-mapview/lib/Tile.ts
FraukeF: no, I tried :), this conflicts with the fallback handling, as the skipRendering is used for the fallbacks, and we then cannot differentiate and the gpu upload logig would occasionally override the skipRendering flag, that was set by the fallback handling and vice versa, then i.e. background tiles could be not skipped anymore because there is space to upload them etc.

@here/harp-mapview/lib/VisibleTileSet.ts
FraukeF: done

@here/harp-mapview/lib/Tile.ts
ninok: I see

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -84,6 +84,15 @@ export interface VisibleTileSetOptions {
      * Number of levels to go down when searching for fallback tiles.
      */
     quadTreeSearchDistanceDown: number;
+
+    /**
+     * Maximal number of new tiles, that can be added to the scene per frame.
+     * if set to `0`the limit will be ignored and all available tiles be uploaded.
+     * @beta
+     * @internal
+     * @defaults 0
```
ninok, line 93: Sorry if this was not communicated properly but it should be `@defaultValue`. See:
https://api-extractor.com/pages/tsdoc/tag_defaultvalue/

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -552,14 +586,27 @@ export class VisibleTileSet {
...
+                            // if set to 0, it will ignore the limit and upload all available
+                            this.options.maxTilesPerFrame !== 0 &&
+                            newTilesPerFrame > this.options.maxTilesPerFrame &&
+                            //if the tile was already visible last frame dont delay it
+                            !(tile.frameNumLastVisible === dataSource.mapView.frameNumber - 1)
+                        ) {
+                            tile.delayRendering = true;
+                            tile.mapView.update();
+                        } else {
+                            tile.numFramesVisible++;
```
nzjony, line 605: this will only ever be 1, because this code is only executed if `framNumVisible` is < 0. Should be fixed.

FraukeF: right, good catch, fixed it

FraukeF: right, good catch. fixed it

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -552,14 +586,27 @@ export class VisibleTileSet {
...
+                        newTilesPerFrame > this.options.maxTilesPerFrame &&
+                        //if the tile was already visible last frame dont delay it
+                        !(tile.frameNumLastVisible === dataSource.mapView.frameNumber - 1)
+                    ) {
+                        tile.delayRendering = true;
+                        tile.mapView.update();
+                    } else {
+                        if (tile.frameNumVisible < 0) {
+                            // Store the fist frame the tile became visible.
+                            tile.frameNumVisible = dataSource.mapView.frameNumber;
```
nzjony, line 605: Sorry, one more minor comment, you can use `frameNumber` rather than `dataSource.mapView.frameNumber`, @ninok  recently changed this from memory.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -552,14 +586,27 @@ export class VisibleTileSet {
...
 
-                    if (tile.frameNumVisible < 0) {
-                        // Store the fist frame the tile became visible.
-                        tile.frameNumVisible = frameNumber;
+                    if (
+                        // if set to 0, it will ignore the limit and upload all available
+                        this.options.maxTilesPerFrame !== 0 &&
+                        newTilesPerFrame > this.options.maxTilesPerFrame &&
+                        //if the tile was already visible last frame dont delay it
+                        !(tile.frameNumLastVisible === dataSource.mapView.frameNumber - 1)
```
ninok, line 598: As @nzjony mentioned, the frameNumer is already passed as parameter. No need to get it from `MapView` again

FraukeF: done

PR #1598, opened 2020-06-11
MINOR: Set codecoverage target to 60%

ninok:


codecov.yml
```diff
@@ -0,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 60%
```
musculman, line 5: Perhaps make sense to use `auto` for target? (auto compares coverage to the previous base commit)
So that we do not need to keep on updating the target?

Perhaps also add:
`require_ci_to_pass: true`
so that it is only done after CI passes?

codecov.yml
```diff
@@ -0,0 +1,6 @@
...
+        threshold: 0.5%
```
musculman, line 6: Does it make sense to add additional rules?
```
comment:
  require_changes: true
```
That would do the following
`Comments will now only post when coverage changes. Furthermore, if a comment already exists, and a newer commit results in no coverage change for the entire pull, the comment will be deleted.`

codecov.yml
ninok: Regarding `require_ci_to_pass` This will increase the total execution time. If there is no need to reduce the CI load I would not set it.

Regarding target: The per patch settings are auto by default. I'm not so sure why the bot was failing with the default settings for PRs that  increased the coverage

codecov.yml
ninok: Not sure if we need comments at all. If the coverage is to low you will see it in the check list

musculman: I think this refers to the comment the bot posts which at the moment we always get. So, it will only be posted when the coverage changes. But I am fine either way

codecov.yml
musculman: Interesting thing about the `auto`. Let's leave it like that then.

PR #1602, opened 2020-06-15
HARP-10797 Fix object-picking example

nzjony:


@here/harp-omv-datasource/lib/OmvDataSource.ts
```diff
@@ -276,20 +276,23 @@ function completeDataSourceParameters(
 
 export class OmvDataSource extends TileDataSource<OmvTile> {
     private readonly m_decoderOptions: OmvDecoderOptions;
+    private m_params: OmvWithRestClientParams | OmvWithCustomDataProvider;
 
-    constructor(private m_params: OmvWithRestClientParams | OmvWithCustomDataProvider) {
-        super(m_params.tileFactory || new TileFactory(OmvTile), {
-            styleSetName: m_params.styleSetName ?? "omv",
```
nzjony, line 282: this would set the incorrect `styleSetName`, because the merging of the defaults happens later

@here/harp-examples/src/object-picking.ts
```diff
@@ -221,7 +221,8 @@ export namespace PickingExample {
 
         const omvDataSource = new OmvDataSource({
             baseUrl: "https://vector.hereapi.com/v2/vectortiles/base/mc",
-            authenticationCode: apikey
+            authenticationCode: apikey,
+            gatherFeatureAttributes: true
```
nzjony, line 225: this was accidentally removed from the previous PR https://github.com/heremaps/harp.gl/pull/1591

PR #1603, opened 2020-06-15
HARP-10793: Add MapView.mapAnchorsOverlay.

robertoraggi:
Overlayed MapAnchor objects are used to overlay custom three.js
objects on top of the MapView scene. That is, any object added to
`mapView.overlayMapAnchors` is always rendered on top of labels, markers
and any object map object.

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -101,7 +101,6 @@ if (isProduction) {
  * mesh.geoPosition = new GeoCoordinates(latitude, longitude, altitude);
  * mapView.mapAnchors.add(mesh);
  * ```
- *
  */
 export type MapAnchor<T extends THREE.Object3D = THREE.Object3D> = T & {
     /**
@@ -114,6 +113,24 @@ export type MapAnchor<T extends THREE.Object3D = THREE.Object3D> = T & {
      * @note By default all objects are pickable even if this flag is undefined.
      */
     pickable?: boolean;
+
+    /**
+     * The styleSet that owns this anchor.
+     *
+     * @remarks
+     * This property is used together with [[Theme.priorities]] to compute the render
+     * order of this anchor.
+     */
+    styleSet?: string;
+
+    /**
+     * The category of this style.
+     *
+     * @remarks
+     * This property is used together with [[Theme.priorities]] to compute the render
+     * order of this anchor.
+     */
+    category?: string;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -3840,13 +3879,20 @@ export class MapView extends THREE.EventDispatcher {
...
                     // This is needed so that the target is updated automatically, see:
                     // https://threejs.org/docs/#api/en/lights/DirectionalLight.target
                     this.m_scene.add(directionalLight.target);
                 }
                 this.m_createdLights!.push(light);
+
+                const clonedLight: THREE.Light = light.clone();
+                this.m_overlayScene.add(clonedLight);
+                if (clonedLight instanceof THREE.DirectionalLight) {
+                    this.m_overlayScene.add(clonedLight.target.clone());
```
nzjony, line 3894: this isn't needed as far as I can tell, threejs already clones the target. when calling the clone method, see: https://github.com/mrdoob/three.js/blob/92ad8c5f4e8a4aa666343ab92b3fd6260fe9e4ce/src/lights/DirectionalLight.js#L35

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1812,6 +1817,20 @@ export class MapView extends THREE.EventDispatcher {
...
+    /**
+     * The node in this MapView's scene containing the overlayed user [[MapAnchor]]s.
+     *
+     * @remarks
+     * The `MapAnchor`s added to `mapAnchorsOverlay` are rendered on top of the MapView scene.
+     *
+     * All (first level) children of this node will be positioned in world space according to the
+     * [[MapAnchor.geoPosition]].
+     * Deeper level children can be used to position custom objects relative to the anchor node.
+     */
```
nzjony, line 1829: would it make sense to add a @beta? here

@here/harp-mapview/lib/MapView.ts
robertoraggi: okay, i removed it.

@here/harp-mapview/lib/MapView.ts
robertoraggi: done.

ninok: Now we have 3 anchor APIs: `mapAnchors`, `worldAnchors` and `mapAnchorsOverlay`.
Should be merged into one api (e.g. `mapObjects`) with following properties:
- anchor (world or geo space)
- overlay

Nevertheless it's fine for now to have it as a temporary API but we should clean it up asap.

robertoraggi: i would use just `MapAnchor` or `Anchor`. `MapObject` may not be the correct name, in fact `MapObject` was the initial name we used for this API but then we replaced it with MapAnchor.
https://github.com/heremaps/harp.gl/pull/530

PR #1605, opened 2020-06-15
HARP-10769: Fix frame complete event

ninok:


@here/harp-mapview/lib/MapView.ts
```diff
@@ -3503,15 +3503,16 @@ export class MapView extends THREE.EventDispatcher {
...
-            !this.isDynamicFrame &&
-            !this.textElementsRenderer.loading
+            !this.m_animatedExtrusionHandler.isAnimating
         ) {
-            this.m_firstFrameComplete = true;
-
-            if (gatherStatistics) {
-                stats.appResults.set("firstFrameComplete", frameStartTime);
+            if (this.m_firstFrameComplete) {
+                this.m_firstFrameComplete = true;
```
nzjony, line 3512: Not needed.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3503,15 +3503,16 @@ export class MapView extends THREE.EventDispatcher {
...
             this.m_initialTextPlacementDone &&
-            !this.isDynamicFrame &&
-            !this.textElementsRenderer.loading
+            !this.m_animatedExtrusionHandler.isAnimating
         ) {
-            this.m_firstFrameComplete = true;
-
-            if (gatherStatistics) {
-                stats.appResults.set("firstFrameComplete", frameStartTime);
+            if (this.m_firstFrameComplete) {
```
nzjony, line 3511: IMO this should be renamed, as it is no longer the first frame

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3503,15 +3503,16 @@ export class MapView extends THREE.EventDispatcher {
         // running. The initial placement of text in this render call may have changed the loading
```
nzjony, line 3503: This comment should be fixed, it mentions above "first complete frame"

@here/harp-mapview/lib/MapView.ts
ninok: For the performance test we capture the time it took until the first frame is complete. I kept that functionality even though it's questionable why this is done inside MapView and not in the test.

@here/harp-mapview/lib/MapView.ts
ninok: done

@here/harp-mapview/lib/MapView.ts
```diff
@@ -155,7 +155,10 @@ export enum MapViewEventNames {
     AfterRender = "didrender",
     /** Called after the first frame has been rendered. */
     FirstFrame = "first-render",
-    /** Called when the first view has all the necessary tiles loaded and rendered. */
+    /**
+     * Called when the rendered frame was complete, i.e. all the necessary tiles and resources
+     * where loaded and rendered.
```
nzjony, line 160:
```suggestion
     * are loaded and rendered.
```

@here/harp-mapview/lib/MapView.ts
nzjony: This is still missing, i.e. if m_firstFrameComplete is true, we don't need to set it true

ninok: Oh, should be `if (this.m_firstFrameComplete === false)` instead. Fixed it

ninok:

PR #1610, opened 2020-06-16
HARP-10793: Add styleSetName and category to MapAnchors

ninok:


@here/harp-mapview/lib/MapAnchors.ts
```diff
@@ -0,0 +1,151 @@
...
+     * Add a [[MapAnchor]].
+     * @param mapAnchor [[MapAnchor]] instance to add.
+     */
+    add(mapAnchor: MapAnchor) {
+        this.m_anchors.push(mapAnchor);
+    }
+
+    /**
+     * Remove a [[MapAnchor]].
+     * @param mapAnchor [[MapAnchor]] instance to remove.
```
robertoraggi, line 87: you are missing the - after `@param`

@here/harp-mapview/lib/MapAnchors.ts
```diff
@@ -0,0 +1,151 @@
...
+     */
+    category?: string;
+};
+
+export class MapAnchors {
+    private m_anchors: MapAnchor[] = [];
+
+    /**
+     *
+     */
```
robertoraggi, line 72: missing documentation.

@here/harp-mapview/lib/MapAnchors.ts
```diff
@@ -0,0 +1,151 @@
...
+     * The category of this style.
+     *
+     * @remarks
+     * This property is used together with [[Theme.priorities]] to compute the render
+     * order of this mao object.
+     */
+    category?: string;
+};
+
+export class MapAnchors {
```
robertoraggi, line 67: missing docs

@here/harp-mapview/lib/MapAnchors.ts
```diff
@@ -0,0 +1,151 @@
...
+    styleSet?: string;
+
+    /**
+     * The category of this style.
+     *
+     * @remarks
+     * This property is used together with [[Theme.priorities]] to compute the render
+     * order of this mao object.
+     */
+    category?: string;
```
robertoraggi, line 64: we could use a reserved styleSet/category combination to identify the objects that need to be overlayed on top of the map scene (after text, markers and extruded buildings are rendered), alternatively we may need to add a property to MapAnchor that can be used to specify the scene/pass.

ninok: yes, I wanted to discuss that as well. I will add the overlay scene in a separate PR though.

PR #1615, opened 2020-06-17
Tile Processing Queue

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview-decoder/lib/TileDataSource.ts
+++ b/@here/harp-mapview-decoder/lib/TileDataSource.ts
@@ -224,9 +224,11 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
      * this data source.
      *
      * @param tileKey - Quadtree address of the requested tile.
+     * @param delayLoad - If true, the Tile will be created, but Tile.load will not be called.
+     * @default false.
      * @override
      */
-    getTile(tileKey: TileKey): TileType | undefined {
+    getTile(tileKey: TileKey, delayLoad: boolean = false): TileType | undefined {
         const tile = this.m_tileFactory.create(this, tileKey);
         tile.tileLoader = new TileLoader(
             this,
@@ -247,7 +249,9 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
                     this.requestUpdate();
                 });
         }
-        tile.load();
+        if (!delayLoad) {
+            tile.load();
+        }
 
         return tile;
     }
--- a/@here/harp-mapview/lib/DataSource.ts
+++ b/@here/harp-mapview/lib/DataSource.ts
@@ -424,8 +424,10 @@ export abstract class DataSource extends THREE.EventDispatcher {
      * {@link @here/harp-geoutils#TileKey}.
      *
      * @param tileKey - The unique identifier for a map tile.
+     * @param delayLoad - If true, the Tile will be created, but Tile.load will not be called
+     * @default false.
      */
-    abstract getTile(tileKey: TileKey): Tile | undefined;
+    abstract getTile(tileKey: TileKey, delayLoad?: boolean): Tile | undefined;
 
     /**
      * This method is called by {@link MapView} before the
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -41,9 +41,11 @@ import {
     LoggerManager,
     LogLevel,
     PerformanceTimer,
+    TaskQueue,
     UriResolver
 } from "@here/harp-utils";
 import * as THREE from "three";
+
 import { AnimatedExtrusionHandler } from "./AnimatedExtrusionHandler";
 import { BackgroundDataSource } from "./BackgroundDataSource";
 import { CameraMovementDetector } from "./CameraMovementDetector";
@@ -61,6 +63,7 @@ import { MapViewImageCache } from "./image/MapViewImageCache";
 import { MapAnchors } from "./MapAnchors";
 import { MapObjectAdapter } from "./MapObjectAdapter";
 import { MapViewFog } from "./MapViewFog";
+import { MapViewTaskScheduler } from "./MapViewTaskScheduler";
 import { PickHandler, PickResult } from "./PickHandler";
 import { PickingRaycaster } from "./PickingRaycaster";
 import { PoiManager } from "./poi/PoiManager";
@@ -93,6 +96,12 @@ if (isProduction) {
 } else {
     // In dev: silence logging below log (silences "debug" and "trace" levels).
     LoggerManager.instance.setLogLevelForAll(LogLevel.Log);
+}
+export enum TileTaskGroups {
+    FETCH_AND_DECODE = "fetch",
+    //DECODE = "decode",
+    CREATE = "create"
+    //UPLOAD = "upload"
 }
 
 export enum MapViewEventNames {
@@ -633,6 +642,13 @@ export interface MapViewOptions extends TextElementsRendererOptions, Partial<Loo
      * @default false
      */
     enableShadows?: boolean;
+
+    /**
+     * Enable throttling for the TaskScheduler
+     * @default false
+     * @beta
+     */
+    throttlingEnabled?: boolean;
 }
 
 /**
@@ -742,17 +758,6 @@ export interface LookAtParams {
  */
 export class MapView extends THREE.EventDispatcher {
     /**
-     * Maximum FPS (Frames Per Second). If VSync in enabled, the specified number may not be
-     * reached, but instead the next smaller number than `maxFps` that is equal to the refresh rate
-     * divided by an integer number.
-     *
-     * E.g.: If the monitors refresh rate is set to 60hz, and if `maxFps` is set to a value of `40`
-     * (60hz/1.5), the actual used FPS may be 30 (60hz/2). For displays that have a refresh rate of
-     * 60hz, good values for `maxFps` are 30, 20, 15, 12, 10, 6, 3 and 1. A value of `0` is ignored.
-     */
-    maxFps: number;
-
-    /**
      * The instance of {@link MapRenderingManager} managing the rendering of the map. It is a public
      * property to allow access and modification of some parameters of the rendering process at
      * runtime.
@@ -842,6 +847,9 @@ export class MapView extends THREE.EventDispatcher {
 
     private m_forceCameraAspect: number | undefined = undefined;
 
+    // type any as it returns different types depending on the environment
+    private m_taskSchedulerTimeout: any = undefined;
+
     //
     // sources
     //
@@ -898,6 +906,7 @@ export class MapView extends THREE.EventDispatcher {
     private readonly m_renderOrderStencilValues = new Map<number, number>();
     // Valid values start at 1, because the screen is cleared to zero
     private m_stencilValue: number = DEFAULT_STENCIL_VALUE;
+    private m_taskScheduler: MapViewTaskScheduler;
 
     /**
      * Constructs a new `MapView` with the given options or canvas element.
@@ -978,7 +987,7 @@ export class MapView extends THREE.EventDispatcher {
         }
 
         this.m_pixelRatio = options.pixelRatio;
-        this.maxFps = options.maxFps === undefined ? 0 : options.maxFps;
+        this.m_options.maxFps = this.m_options.maxFps ?? 0;
 
         this.m_options.enableStatistics = this.m_options.enableStatistics === true;
 
@@ -1109,6 +1118,15 @@ export class MapView extends THREE.EventDispatcher {
             this.m_backgroundDataSource.setTilingScheme(this.m_options.backgroundTilingScheme);
         }
 
+        this.m_taskScheduler = new MapViewTaskScheduler(this.maxFps);
+        this.m_taskScheduler.addEventListener(MapViewEventNames.Update, () => {
+            this.update();
+        });
+
+        if (options.throttlingEnabled !== undefined) {
+            this.m_taskScheduler.throttlingEnabled = options.throttlingEnabled;
+        }
+
         this.initTheme();
 
         this.m_textElementsRenderer = this.createTextRenderer();
@@ -1122,6 +1140,10 @@ export class MapView extends THREE.EventDispatcher {
      */
     get lights(): THREE.Light[] {
         return this.m_createdLights ?? [];
+    }
+
+    get taskQueue(): TaskQueue {
+        return this.m_taskScheduler.taskQueue;
     }
 
     /**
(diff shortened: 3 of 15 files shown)


@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -0,0 +1,235 @@
...
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Task that can be added to the [[TaskQueue]]
+ */
+export interface Task {
+    /**
+     * The Function that will executed when the [[Task]] is processed
```
ninok, line 12: will be

@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * The group by which similar tasks in the TaskQueue are combined
+     */
+    group: string;
+
+    /**
+     * A function to retrieve the priority of the [[Task]], with 0 being
+     * the highest priority, and the first to be executed
+     */
+    getPrio: () => number;
```
ninok, line 25: rename to `getPriority`

@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -0,0 +1,235 @@
...
+     * Caution, the [[TaskQueue]] uses the last element in the Arrays first, so the
+     * highest priorities should be ordered to the end
+     */
+    prioSortFn?: (a: Task, b: Task) => number;
+}
+
+/**
+ * This is a helper interface to be able to index the TaskArrays via a string
+ */
+interface TaskArrays {
```
ninok, line 62: `Map<string, Task[]>` ?

@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -0,0 +1,235 @@
...
+ *  ```
+ */
+export class TaskQueue {
+    private m_taskArrays: TaskArrays = {};
+
+    constructor(private m_options: TaskQueueOptions) {
+        this.m_options.groups?.forEach(group => {
+            this.m_taskArrays[group] = [];
+        });
+        this.sort = this.m_options.prioSortFn ?? this.sort;
```
ninok, line 96: `this.sort` is `undefined`

@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -0,0 +1,235 @@
+/*
+ * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Task that can be added to the [[TaskQueue]]
+ */
+export interface Task {
```
ninok, line 10: This interface has a mixture of methods and properties. I think except `execute` all members should be properties

@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * Adds a Task to the Queue
+     *
+     * @param task
+     * @returns true if succesfully added, otherwise false
+     */
+    add(task: Task): boolean {
+        if (this.m_taskArrays[task.group]) {
+            this.m_taskArrays[task.group].push(task);
```
ninok, line 125: Should we check if the task is already inserted?

@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -0,0 +1,235 @@
...
+    constructor(private m_options: TaskQueueOptions) {
+        this.m_options.groups?.forEach(group => {
+            this.m_taskArrays[group] = [];
+        });
+        this.sort = this.m_options.prioSortFn ?? this.sort;
+    }
+
+    /**
+     * Updates the lists in the queue depending on their priority functions
+     * Needs to be called by using instance, as it would be to costly on every change
```
ninok, line 101: > Needs to be called by using instance, as it would be to costly on every change

I don't understand that sentence

@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -0,0 +1,235 @@
...
+    add(task: Task): boolean {
+        if (this.m_taskArrays[task.group]) {
+            this.m_taskArrays[task.group].push(task);
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Removes a Task to the Queue
```
ninok, line 132: from the queue

@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -0,0 +1,235 @@
...
+     *
+     * @param task
+     * @returns true if succesfully removed, otherwise false
+     */
+    remove(task: Task): boolean {
```
ninok, line 137: Why do we have this method? Don't we always pop the last one (i.e. the one with the highest priority)?

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -262,6 +262,32 @@ export class TileGeometryLoader {
...
     ) {
+        const decodedTile = this.m_decodedTile;
+
+        // Just a sanity check that satisfies compiler check below.
+        if (decodedTile === undefined) {
+            this.finish();
+            return;
+        }
+
+        this.tile.mapView.taskQueue.add({
```
ninok, line 273: I would rather pass the `TaskQueue` in the `TileGeometryLoader` `contructor` instead of getting it from `MapView`.
That makes the dependeny much clearer.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -262,6 +262,32 @@ export class TileGeometryLoader {
...
+            execute: this.prepare.bind(this, enabledKinds, disabledKinds),
+            group: TileTaskGroups.CREATE,
+            getPrio: this.getPriority.bind(this),
```
ninok, line 276: Will the priority change? If it changes, will `TaskQueue` handle it properly?
Maybe for simplicity just call `this.getPriority` here and assign it to the task?

@here/harp-mapview/lib/MapViewTaskScheduler.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+import { Task, TaskQueue } from "@here/harp-utils";
+
+import { MapView, TileTaskGroups } from "./MapView";
+
+export class MapViewTaskScheduler {
+    private m_taskQueue: TaskQueue;
+    private m_throttlingEnabled: boolean = false;
+
+    constructor(private m_mapView: MapView) {
```
ninok, line 14: Please don't add whole `MapView` as dependency. As far as I can see it is only used for calling update. That could be solved with a callback similar to `DataSource.requestUpdate`

@here/harp-mapview/lib/MapViewTaskScheduler.ts
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Processes the pending Tasks of the underlying [[TaskQueue]]
+     * !! This should run at the end of the renderLoop, so the calculations of the available
+     * frame time are better estimated
+     *
+     * @param frameStartTime the start time of the current frame, is used to calculate the
+     * still available time in the frame to process Tasks
+     *
+     */
+    processPending(frameStartTime: number) {
```
ninok, line 44: Please also add some numbers to the performance statistics, e.g.: Number of tasks processed & number of tasks pending

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3422,6 +3464,8 @@ export class MapView extends THREE.EventDispatcher {
             stats.addMemoryInfo();
         }
 
+        this.m_taskScheduler.processPending(frameStartTime);
```
ninok, line 3467: Previously at least the geometry creation was outside the animation frame by using setTimeout.
Now we are blocking the animation frame when we process tasks, so I think we should still use setTimeout to call `processPending`

@here/harp-utils/lib/TaskQueue.ts
FraukeF: this.sort is defined as its a member function   

 private sort(a: Task, b: Task): number {
        // the highest number in the beginning as the last in the array with
        // highest priority which equals 0 will start to be processed
        return b.getPriority() - a.getPriority();
    }

@here/harp-utils/lib/TaskQueue.ts
FraukeF: They are functions, as they might change until the Task is actually executed, so if i.e. the priority of the Task changes, as i.e. the camera moves and the regarding tile is further away the priority function will be able to generate the updated priority

@here/harp-utils/lib/TaskQueue.ts
FraukeF: yes, but you can remove the Task from the TaskQueue if you decide it should not be processed anymore

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
FraukeF: The Tiles could theoretically be from different MapViews, and therefore different TaskQueues

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
FraukeF: Yes, it will change, as the Camera moves, the priority will change based on the distance of the tile to the camera. 
I added a test to showcase the dynamic update of a Tasks priority

@here/harp-mapview/lib/MapViewTaskScheduler.ts
FraukeF: its also used for the maxFps property, but Ill pass that to the Scheduler, so the dependency is not needed.

@here/harp-utils/lib/TaskQueue.ts
ninok: ah okay, but to avoid confusion I would only overwrite `this.sort` with `prioSortFn` if it's not defined.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
ninok: wouldn't we have multiple TileGeometryLoader instances in that case?

@here/harp-utils/lib/TaskQueue.ts
FraukeF: Not sure I understand your comment, this.sort will always be defined, as its a member function, if the option.prioSortFn is defined it will overwrite the default this.sort.

@here/harp-utils/lib/TaskQueue.ts
FraukeF: updated it, hope its clearer now

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
FraukeF: yes, seems like there is one TileGeometryManager using one TileGeometryLoader for for one MapView :) gonna change that

@here/harp-mapview/lib/MapView.ts
FraukeF: ok

@here/harp-mapview/lib/MapViewTaskScheduler.ts
FraukeF: done

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -225,7 +225,7 @@ export class TileDataSource<TileType extends Tile> extends DataSource {
      * @param tileKey - Quadtree address of the requested tile.
      * @override
      */
-    getTile(tileKey: TileKey): TileType | undefined {
+    getTile(tileKey: TileKey, delayLoad: boolean = false): TileType | undefined {
```
ninok, line 228: document `delayLoad`

@here/harp-mapview/lib/DataSource.ts
```diff
@@ -425,7 +425,7 @@ export abstract class DataSource extends THREE.EventDispatcher {
      *
      * @param tileKey - The unique identifier for a map tile.
      */
-    abstract getTile(tileKey: TileKey): Tile | undefined;
+    abstract getTile(tileKey: TileKey, delayLoad?: boolean): Tile | undefined;
```
ninok, line 428: document `delayLoad`

@here/harp-mapview/lib/MapView.ts
```diff
@@ -739,18 +755,7 @@ export interface LookAtParams {
...
-     *
-     * E.g.: If the monitors refresh rate is set to 60hz, and if `maxFps` is set to a value of `40`
-     * (60hz/1.5), the actual used FPS may be 30 (60hz/2). For displays that have a refresh rate of
-     * 60hz, good values for `maxFps` are 30, 20, 15, 12, 10, 6, 3 and 1. A value of `0` is ignored.
-     */
-    maxFps: number;
-
-    /**
-     * The instance of {@link MapRenderingManager} managing the rendering of the map. It is a public
+     * The instance of [[MapRenderingManager]] managing the rendering of the map. It is a public
```
ninok, line 758: Please use `{@link}` instead of `[]` b/c [] is not supported by api-extractor

@here/harp-utils/lib/TaskQueue.ts
ninok: Yes, but if prioSortFn is undefined you do `this.sort = this.sort`. That is what confused me

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3439,6 +3492,10 @@ export class MapView extends THREE.EventDispatcher {
 
         this.checkCopyrightUpdates();
 
+        setTimeout(() => {
```
ninok, line 3495: Not sure if this can really happen, but I think we need to check whether the timeout from the last frame was even processed.

package-lock.json
```diff
@@ -0,0 +1,8724 @@
+{
```
ninok, line 1: Please remove

@here/harp-mapview/lib/MapView.ts
```diff
@@ -842,6 +847,8 @@ export class MapView extends THREE.EventDispatcher {
 
     private m_forceCameraAspect: number | undefined = undefined;
 
+    private m_taskSchedulerTimeout: NodeJS.Timeout | undefined = undefined;
```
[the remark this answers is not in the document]
FraukeF: I already changed that to "any" this seems the concurrent type for setTimout throughout harp.gl and "number" will fail in the test environment as it is NodeJS.Timeout type in node

package-lock.json
FraukeF: done

PR #1616, opened 2020-06-17
Add missing documentation for expressions and environments.

robertoraggi:
Add missing documentation for Expr and Env. Also, start to use reference
declaration using `@link` and hide some internal symbols.

### DIFF
--- a/@here/harp-datasource-protocol/lib/Expr.ts
+++ b/@here/harp-datasource-protocol/lib/Expr.ts
@@ -25,6 +25,9 @@ const exprEvaluator = new ExprEvaluator();
 
 const exprInstantiator = new ExprInstantiator();
 
+/**
+ * A visitor for {@link Expr} nodes.
+ */
 export interface ExprVisitor<Result, Context> {
     visitNullLiteralExpr(expr: NullLiteralExpr, context: Context): Result;
     visitBooleanLiteralExpr(expr: BooleanLiteralExpr, context: Context): Result;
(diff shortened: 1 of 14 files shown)


@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -195,31 +197,34 @@ interface ReferenceResolverState {
...
+     * The scope of an {@link Expr} used as dynamic property attribute value.
      */
     Dynamic
 }
 
+// tslint:disable:max-line-length
 /**
- * Abstract class defining a shape of a [[Theme]]'s expression
+ * Abstract class representing the
+ * {@link https://github.com/heremaps/harp.gl/blob/master/%40here/harp-datasource-protocol/StyleExpressions.md | style expressions}
```
nzjony, line 222: can this link to a local file? You are anyway in the `harp-datasource-protocol` lib

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -195,31 +197,34 @@ interface ReferenceResolverState {
...
     Condition,
 
     /**
-     * The scope of an [[Expr]] used as dynamic property attribute value.
+     * The scope of an {@link Expr} used as dynamic property attribute value.
      */
     Dynamic
 }
 
+// tslint:disable:max-line-length
```
nzjony, line 219: this disables it for the rest of the file, is that what you want?

@here/harp-datasource-protocol/lib/Expr.ts
```diff
@@ -626,7 +642,7 @@ export class MatchExpr extends Expr {
 }
 
 /**
- * @hidden
+ * A node representing a `case` expression.
```
nzjony, line 645: would it make sense to link to the syntax of this expression?

robertoraggi: all the expressions follow the same syntax and the link to the style expression is in the dcoumentation of the base Expr class.

@here/harp-datasource-protocol/lib/Expr.ts
robertoraggi: oops, how can i disable the max-line-length only for the comment?

PR #1636, opened 2020-06-24
Minor camera animation

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
+++ b/@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
@@ -15,6 +15,7 @@ import { LoggerManager } from "@here/harp-utils";
 import THREE = require("three");
 
 const logger = LoggerManager.instance.create("CameraKeyTrackAnimation");
+const MIN_DISTANCE = 0;
 
 /**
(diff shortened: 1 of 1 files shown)


@here/harp-map-controls/lib/CameraKeyTrackAnimation.ts
```diff
@@ -243,6 +244,7 @@ export class CameraKeyTrackAnimation {
         this.m_mapView.addEventListener(MapViewEventNames.Render, this.m_animateCb);
         this.m_mapView.beginAnimation();
         this.m_running = true;
+        this.m_animationAction.reset();
```
nzjony, line 247: you call `.play()` above, and then `.reset()`, that sounds a bit strange..?

FraukeF: yes, but still its ok, as the .play() will not start the animation, but by definition of the AnimationAction  is activating the Animation to be able to be played (sounds weired, but see here: https://threejs.org/docs/#api/en/animation/AnimationAction) .. 
Therefore .play() will activate and .reset() make sure it starts from the beginning, the order doesnt matter at this point

PR #1642, opened 2020-06-25
HARP-10648: Add a buffer of about 5% around the clipped tile.

robertoraggi:
This change configures geojson-vt to use a buffer of about 5% of the
extents when tiling. The introduction of the buffer allows us to
better classify the edges introduced when tiling feature geometries.

### DIFF
--- a/@here/harp-mapview-decoder/lib/GeoJsonTiler.ts
+++ b/@here/harp-mapview-decoder/lib/GeoJsonTiler.ts
@@ -7,10 +7,20 @@
 import { GeoJson, ITiler } from "@here/harp-datasource-protocol";
 import { TileKey } from "@here/harp-geoutils";
 
-// tslint:disable-next-line:no-var-requires
-const geojsonvtExport = require("geojson-vt");
+// @ts-ignore
+import * as geojsonvtExport from "geojson-vt";
+
 // to be able to run tests on nodejs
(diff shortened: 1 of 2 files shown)


@here/harp-mapview-decoder/lib/GeoJsonTiler.ts
```diff
@@ -7,10 +7,19 @@
...
-const geojsonvt = geojsonvtExport.default || geojsonvtExport;
+const geojsonvt = geojsonvtExport.default ?? geojsonvtExport;
+
+const EXTENT = 4096;
+
+// the factor used to compute the size of the buffer.
+const BUFFER_FACTOR = 0.05;
+
+// tslint:disable-next-line: no-bitwise
+const BUFFER = -(-Math.ceil(EXTENT * BUFFER_FACTOR) & -2);
```
nzjony, line 22: what does -2 mean?

robertoraggi: yes, it's a round up with alignment of 2.

robertoraggi:
```
> const RoundUp = (x,align=2) => -((-Math.ceil(x)) & -align)
undefined
> RoundUp(4 * 1024 * 0.05)
206
```

PR #1648, opened 2020-06-25
HARP-8943 Minor refactorings

nzjony:
Remove the dependency on the MapView from the VisibleTileSet

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -1064,8 +1064,12 @@ export class MapView extends THREE.EventDispatcher {
         this.projection.projectPoint(this.m_targetGeoPos, this.m_targetWorldPos);
         this.m_scene.add(this.m_camera); // ensure the camera is added to the scene.
         this.m_screenProjector = new ScreenProjector(this.m_camera);
-        // setup camera with initial position
 
+        // Must be initialized before setupCamera, because the VisibleTileSet is created as part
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -3384,6 +3384,7 @@ export class MapView extends THREE.EventDispatcher {
                 Math.floor(this.zoomLevel),
                 this.getEnabledTileDataSources(),
                 this.m_frameNumber,
+                this.taskQueue,
```
FraukeF, line 3387: As the visibleTileSet is basically not working without a taskQueue and it will always be the same taskQueue passed through here, I would prefer adding it as a hard dependency in the constructor of VisibleTileSet, instead of passing it here every frame.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -577,7 +578,8 @@ export class VisibleTileSet {
                     dataSource,
                     tileEntry.tileKey,
                     tileEntry.offset,
-                    frameNumber
+                    frameNumber,
+                    taskQueue
```
FraukeF, line 582: see comment in MapView, make it a member, then it does not have to be passed around and can be accessed via this...

@here/harp-mapview/lib/MapView.ts
nzjony: Good point! Have fixed it.

PR #1655, opened 2020-06-30
HARP-11015 Upgrade to three@0.118

nzjony:


@here/harp-examples/src/textcanvas.ts
```diff
@@ -335,7 +335,7 @@ export namespace TextCanvasDynamicExample {
 
     function main() {
         // Init Three.JS
-        webglRenderer = new THREE.WebGLRenderer({
+        webglRenderer = new THREE.WebGL1Renderer({
```
robertoraggi, line 338: with this change we require all our users to update to 0.118. I'm okay with this but if we want to support users on older versions of three (117?) we could replace this test with something like
```
webglRenderer = new ((THREE as any).WebGL1Renderer ?? THREE.WebGLRenderer)(…)
```

@here/harp-lines/test/rendering/RenderLines.ts
```diff
@@ -281,7 +281,7 @@ describe("Rendering lines: ", function() {
         }
 
         scene.position.set(-canvas.width / 2, -canvas.height / 2, 0);
-        webglRenderer = new THREE.WebGLRenderer({ canvas });
+        webglRenderer = new THREE.WebGL1Renderer({ canvas });
```
robertoraggi, line 284: with this change we require all our users to update to 0.118. I'm okay with this but if we want to support users on older versions of three (117?) we could replace this test with something like

webglRenderer = new ((THREE as any).WebGL1Renderer ?? THREE.WebGLRenderer)(…)

PR #1658, opened 2020-07-01
Specialize the clipping algorithm for the vector-tile use case.

robertoraggi:
This change specializes the Sutherland-Hodgman algorithm to
clip a polygon against rectangular bounds. This should improve
precision and speed up the clipping.

### DIFF
--- a/@here/harp-geometry/lib/ClipPolygon.ts
+++ b/@here/harp-geometry/lib/ClipPolygon.ts
@@ -6,63 +6,192 @@
 
 import { Vector2 } from "three";
 
-const tmpBA = new Vector2();
-const tmpQP = new Vector2();
-const tmpA = new Vector2();
-const tmpB = new Vector2();
+/**
+ * Abstract helper class used to implement the Sutherland-Hodgman clipping algorithm.
+ *
+ * @remarks
+ * Concrete implementation of this class are used to clip a polygon
+ * against one edge of a bounding box.
+ *
+ * @internal
+ */
+abstract class ClippingEdge {
+    /**
+     * Tests if the given point is inside this clipping edge.
+     *
+     * @param point A point of the polygon.
+     * @param extent The extent of the bounding box.
+     */
+    abstract inside(point: Vector2, extent: number): boolean;
+
+    /**
+     * Computes the intersection of a line and this clipping edge.
+     *
+     * @remarks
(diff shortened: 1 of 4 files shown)


@here/harp-geometry/test/ClipPolygonTest.ts
```diff
@@ -178,7 +179,7 @@ describe("ClipPolygon", () => {
         assert.isTrue(polygon.some(vert => vert.y < 0));
         assert.isTrue(polygon.some(vert => vert.y > extents));
 
-        const clippedPolygon = clipPolygon(polygon, tileBounds);
+        const clippedPolygon = clipPolygon(polygon, extents);
 
         assert.isNotEmpty(clippedPolygon);
```
germanz, line 185: I would add a test with concave polygon resulting into multiple parts after clipping.

```
    it("Concave polygon resulting into 2 parts after clipping", () => {
        const polygon: Vector2[] = [
            new Vector2(-100, 0),
            new Vector2(4096, 0),
            new Vector2(-50, 2048),
            new Vector2(4096, 4096),
            new Vector2(-100, 4096)
        ];

        const expectedPolygon: Vector2[] = [
            new Vector2(0, 0),
            new Vector2(4096, 0),
            new Vector2(0, 2023),
            new Vector2(0, 2073),
            new Vector2(4096, 4096),
            new Vector2(0, 4096)
        ];

        const clippedPolygon = clipPolygon(polygon, extents);
        assert.strictEqual(JSON.stringify(clippedPolygon), JSON.stringify(expectedPolygon));
    });
```

robertoraggi: done, after clipping it results in two triangles.

@here/harp-geometry/lib/ClipPolygon.ts
```diff
@@ -6,63 +6,192 @@
...
-    for (let e = 0; e < clip.length; ++e) {
-        const p = clip[e];
-        const q = clip[(e + 1) % clip.length];
-        const inputList = outputList;
-        outputList = [];
+abstract class ClippingEdge {
+    /**
+     * Test on which side of the polygon the edge belongs..
+     *
+     * @param point The point to te
```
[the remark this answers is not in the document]
robertoraggi: oops, fixed.

PR #1670, opened 2020-07-08
HARP-11145 Fix adding rotating callback to completion of first frame

nzjony:


www/src/index.ts
```diff
@@ -117,7 +120,8 @@ function main() {
         setTimeout(() => {
             map.beginAnimation();
         }, 0.5);
-    });
+    };
+    map.addEventListener(MapViewEventNames.FrameComplete, onFrameComplete);
```
robertoraggi, line 124: the fix look strange. Why are we using FrameComplete here? if it is for driving the animation we should use `AfterRender` or some other event

robertoraggi, line 124: this fix is strange, why do we need to use FrameComplete here? can't we just start the animation without waiting for FrameComplete?

PR #1677, opened 2020-07-09
HARP-11196 Expand getScreenPosition...

nzjony:
It can now take a world position also

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2547,14 +2547,19 @@ export class MapView extends THREE.EventDispatcher {
     }
 
     /**
-     * Returns the screen position of the given geo coordinates.
+     * Returns the screen position of the given position.
      *
-     * @param geoPos - The geo coordinates.
+     * @param geoPos - The position as a {@link @here/harp-geoutils#GeoCoordinates} or
+     * {@link https://threejs.org/docs/#api/en/math/Vector3 | THREE.Vector3} world position.
      * @returns The screen position in CSS/client coordinates (no pixel ratio applied) or
      * `undefined`.
      */
-    getScreenPosition(geoPos: GeoCoordinates): THREE.Vector2 | undefined {
-        this.projection.projectPoint(geoPos, cache.vector3[0]);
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2547,14 +2547,18 @@ export class MapView extends THREE.EventDispatcher {
     }
 
     /**
-     * Returns the screen position of the given geo coordinates.
+     * Returns the screen position of the given position.
      *
-     * @param geoPos - The geo coordinates.
+     * @param geoPos - The position as a {@link GeoCoordinate} or `Vector3` world position.
```
robertoraggi, line 2552: declarations from other harp modules must be prefixed with the module name, e.g.
`{@link @here/harp-geoutils#GeoCoordinates}`

in case you want to reference three.js symbols you can use the following syntax:
`{@link https://threejs.org/docs/#api/en/math/Vector3 | THREE.Vector3}`

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2547,14 +2547,19 @@ export class MapView extends THREE.EventDispatcher {
...
      *
-     * @param geoPos - The geo coordinates.
+     * @param geoPos - The position as a {@link @here/harp-geoutils#GeoCoordinates} or
+     * {@link https://threejs.org/docs/#api/en/math/Vector3 | THREE.Vector3} world position.
      * @returns The screen position in CSS/client coordinates (no pixel ratio applied) or
      * `undefined`.
      */
-    getScreenPosition(geoPos: GeoCoordinates): THREE.Vector2 | undefined {
-        this.projection.projectPoint(geoPos, cache.vector3[0]);
+    getScreenPosition(pos: GeoCoordinates | THREE.Vector3): THREE.Vector2 | undefined {
```
robertoraggi, line 2557: not strictly related to this change but you're touching this code so it may be worth mentioning that there is a new type in geoutils that is used to represent the different ways coordinates in geo space can be defined in harp.gl (`GeoCoordLike`). We should consider to use `GeooCoordsLike` instead of `GeoCoordinates` in methods like this.

PR #1684, opened 2020-07-14
MINOR: Trigger release only if tag is added

ninok:


ninok:

ninok:
> This change may work not like you expect. Documentation says:
> 
> "Setting tags to true causes the branch condition to be ignored, otherwise $TRAVIS_TAG is ignored, and the branch condition is considered". (https://docs.travis-ci.com/user/deployment-v2/conditional#tag)
> 
> As far as I understood, after this change job will also start if you'll put a tag to some feature branch, not only release one.

Yeah, that's indeed strange. Since nobody will/should add tags somewhere else I will remove the branch condition

PR #1685, opened 2020-07-14
Harp 10772 limit picking results

atomicsulfate:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/PickListener.ts
+++ b/@here/harp-mapview/lib/PickListener.ts
@@ -1,0 +1,127 @@
+/*
+ * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { IntersectParams } from "./IntersectParams";
+import { PickResult } from "./PickHandler";
+
+// Default sorting by distance first and then by reversed render order.
+function defaultSort(lhs: PickResult, rhs: PickResult) {
+    const distanceDiff = lhs.distance - rhs.distance;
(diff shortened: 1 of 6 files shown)


@here/harp-mapview/lib/PickListener.ts
```diff
@@ -0,0 +1,127 @@
...
+        return this.maxResults ? this.m_results.length >= this.maxResults : false;
+    }
+
+    /**
+     * Returns the collected results, ordered by distance first, then by reversed render order
+     * (topmost/highest render order first).
+     * @returns The pick results.
+     */
+    get results(): PickResult[] {
+        this.finish();
```
ninok, line 84: I would rather do an explicit call to finish in the pick handler instead of having side effects in the getter

ninok:
As far as I can see, this PR also removes the duplicates from the pick result if they have ids. Why do we have duplicates in the results in the first place?

atomicsulfate:
> As far as I can see, this PR also removes the duplicates from the pick result if they have ids. Why do we have duplicates in the results in the first place?

Usually because there's geometry associated with the same feature id in multiple meshes (e.g. line and outline meshes for roads) or as part of different groups in the same mesh. It's not easy to avoid cause you need to check what's the closest intersection for a feature to then be able to determine which feature goes first.

Another less usual possibility is that the same mesh feature geometry is intersected more than once. Picking in THREE.js doesn't stop at the first intersection (otherwise you cannot determine the picking distance).

@here/harp-mapview/lib/PickListener.ts
atomicsulfate: yes, makes sense, will fix it in a separate PR.

PR #1688, opened 2020-07-15
HARP-11205 Adding dependencies to DecodedTile/Tile

nzjony:
These have dependencies which allow a tile which has
content that exceeds its border to be visible when it is
within another Tile

### DIFF
--- a/@here/harp-datasource-protocol/lib/DecodedTile.ts
+++ b/@here/harp-datasource-protocol/lib/DecodedTile.ts
@@ -62,6 +62,12 @@ export interface DecodedTile {
      * @see [[CopyrightInfo]]
      */
     copyrightHolderIds?: string[];
+
+    /**
+     * List of {@link @here/harp-geoutils#TileKey}s stored as mortonCodes representing
+     * {@link @here/harp-mapview#Tile}s that have geometry covering this `Tile`.
+     */
+    dependencies?: number[];
 }
 
 /**
--- a/@here/harp-examples/src/tile_dependencies.ts
+++ b/@here/harp-examples/src/tile_dependencies.ts
@@ -1,0 +1,288 @@
+/*
+ * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Theme } from "@here/harp-datasource-protocol";
+import { DebugTileDataSource } from "@here/harp-debug-datasource";
+import { TileKey, webMercatorTilingScheme } from "@here/harp-geoutils";
+import { MapControls, MapControlsUI } from "@here/harp-map-controls";
+import { CopyrightElementHandler, MapView, PickResult, Tile } from "@here/harp-mapview";
+import {
+    DataProvider,
+    TileDataSource,
+    TileDataSourceOptions,
+    TileFactory
+} from "@here/harp-mapview-decoder";
+import { GUI } from "dat.gui";
+import { CUSTOM_DECODER_SERVICE_TYPE } from "../decoder/custom_decoder_defs";
+
+/**
+ * This example shows how to use the {@link @here/harp-mapview#Tile.dependencies} property to ensure
+ * that tiles which have geometry overlapping other.
+ *
+ * If you pan further enough left / right, you will see that the tile disappears.
+ *
+ * It combines part of the https://www.harp.gl/docs/master/examples/#object-picking.html and
+ * https://www.harp.gl/docs/master/examples/#datasource_custom.html examples.
+ * ```
+ * {@link @here/harp-mapview#Tile}s that contain the geometry from another Tile need to have a
+ * reference to the Tile containing the overlapping geometry, this is achieved using the
+ * `dependencies` property of the {@link @here/harp-datasource-protocol#DecodedTile}
+ * ```typescript
+ * [[include:tile_dependencies.ts]]
+ * ```
+ **/
+
+export namespace TileDependenciesExample {
+    const guiOptions = { tileDependencies: false };
+    document.body.innerHTML += `
+        <style>
+            #mouse-picked-result{
+                position:absolute;
+                bottom:5px;
+                border-radius: 5px;
+                margin-left:10px;
+                padding: 9px 12px;
+                background: #37afaa;
+                display: inline-block;
+                visibility: hidden;
+                text-align: left;
+                right:50px;
+            }
+            #mapCanvas {
+              top: 0;
+            }
+            #info{
+                color: #fff;
+                width: 80%;
+                left: 50%;
+                position: relative;
+                margin: 10px 0 0 -40%;
+                font-size: 15px;
+            }
+            @media screen and (max-width: 700px) {
+                #info{
+                    font-size:11px;
+                }
+            }
+        </style>
+        <p id=info>Click/touch a feature on the map to read its data (Land masses are not features).
+        </p>
+        <pre id="mouse-picked-result"></pre>
+    `;
+    class CustomDataProvider implements DataProvider {
+        enableTileDependencies: boolean = false;
+        connect() {
+            // Here you could connect to the service.
+            return Promise.resolve();
+        }
+
+        ready() {
+            // Return true if connect was successful.
+            return true;
+        }
+
+        getTile(tileKey: TileKey, abortSignal?: AbortSignal): Promise<ArrayBufferLike | {}> {
+            // Generate some artifical data. Normally you would do a fetch here.
+            // In this example we create some geometry in geo space that will be converted to
+            // local world space by [[CustomDecoder.convertToLocalWorldCoordinates]]
+
+            const data = new Array<number>();
+            // Do some scaling so that the data fits into the tile.
+            // tslint:disable-next-line: no-bitwise
+            const scale = 10.0 / (1 << tileKey.level);
+            data.push(0.0, 0.0);
+            for (let t = 0.0; t < Math.PI * 4; t += 0.1) {
+                const x = Math.cos(t) * t * scale;
+                const y = Math.sin(t) * t * scale * 10;
+                data.push(x, y);
+            }
+
+            const mainTileKey = new TileKey(16384, 16384, 15);
+            if (tileKey.mortonCode() === mainTileKey.mortonCode()) {
+                return Promise.resolve(new Float32Array([data.length, ...data, 0]));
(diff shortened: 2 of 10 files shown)


@here/harp-datasource-protocol/lib/DecodedTile.ts
```diff
@@ -62,6 +63,12 @@ export interface DecodedTile {
      * @see [[CopyrightInfo]]
      */
     copyrightHolderIds?: string[];
+
+    /**
+     * List of {@link @here/harp-geoutils#TileKey}s representing {@link @here/harp-mapview#Tile}s
+     * that have geometry covering this `Tile`.
+     */
+    dependencies?: TileKey[];
```
robertoraggi, line 71: you cannot have TIleKey in DecodedTile because DecodedTile may be transferred from the webworker to the main app so it cannot contain object instances. You should serialize the list of tile keys to morton codes or strings here,
and then convert them to TIleKey when initializing the `Tile`.

@here/harp-mapview/test/VisibleTileSetTest.ts
```diff
@@ -268,6 +268,46 @@ describe("VisibleTileSet", function() {
...
+                }
+            ],
+            allBoundingBoxesFinal: true
+        });
+
+        // Adding the dependency to make sure it is visible
+        const tile1 = fixture.vts.getTile(fixture.ds[0], tileKey1, 0, 0);
+        assert.notEqual(tile1, undefined);
+        const tileKey3 = TileKey.fromMortonCode(371506852);
+        tile1!.dependencies.push(tileKey3);
```
nzjony, line 292: this is where the dependency is added from another tile to tile 1, i.e. when tile 1 is visible, tile with tileKey 3 will also be shown

ninok:
One general question for that change:
Tile dependencies also affect the caching or more precisely the cache eviction.
If tile `X` depends on tile `Y` and tile `X` is visible but tile `Y` is not, we cannot remove tile `Y` from cache.

Will this implicitly be handled b/c we mark tile `Y` as visible. Would be good to add a dedicated test for the caching behavior.

nzjony:
> One general question for that change:
> Tile dependencies also affect the caching or more precisely the cache eviction.
> If tile `X` depends on tile `Y` and tile `X` is visible but tile `Y` is not, we cannot remove tile `Y` from cache.
> 
> Will this implicitly be handled b/c we mark tile `Y` as visible. Would be good to add a dedicated test for the caching behavior.

@ninok , yes, tile Y is declared to be part of the visible tile key set, so it will be requested and the `frameNumLastRequested` will be updated accordingly.

I will add a test to check this case explicitly though.

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -617,6 +616,10 @@ export class VisibleTileSet {
                 tile.elevationRange = tileEntry;
 
                 actuallyVisibleTiles.push(tile);
+
+                visibleTileKeys.push(
+                    ...tile.dependencies.map(tileKey => new TileKeyEntry(tileKey, 0, tile.offset))
+                );
```
robertoraggi, line 622: please make sure to remove duplicates

nzjony:

@here/harp-examples/src/tile_dependencies.ts
```diff
@@ -0,0 +1,288 @@
...
+ * ```
+ * {@link @here/harp-mapview#Tile}s that contain the geometry from another Tile need to have a
+ * reference to the Tile containing the overlapping geometry, this is achieved using the
+ * `dependencies` property of the {@link @here/harp-datasource-protocol#DecodedTile}
+ * ```typescript
+ * [[include:tile_dependencies.ts]]
+ * ```
+ **/
+
+export namespace CustomDatasourceExample {
```
ninok, line 38: Choose proper name

@here/harp-examples/src/tile_dependencies.ts
```diff
@@ -0,0 +1,288 @@
+/*
```
ninok, line 1: I wonder if we should/could simplify this example a bit.
It might be enough to have a synchronous datasource that has tile dependencies, even though that would not cover the decode tile depency propagation part.

nzjony: I discussed this with @robertoraggi , he wasn't so keen to have an example that shows how to use the tile.dependencies property, because it could be copy / pasted by others (@robertoraggi , correct me if I misunderstood) and hence misused.

Hence adapting the example to use the dependencies in the `DecodedTile`.

@here/harp-examples/src/tile_dependencies.ts
nzjony: Done

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -146,20 +146,52 @@ export class PickHandler {
...
+                checkedDependencies.add(mortonCode);
+                const otherTile = this.mapView.visibleTileSet.getCachedTile(
+                    tile.dataSource,
+                    tileKey,
+                    tile.offset,
+                    this.mapView.frameNumber
+                );
+                if (otherTile !== undefined) {
+                    rayCaster.intersectObjects(otherTile.objects, true, intersects);
+                    addIntersectResults(intersects);
```
atomicsulfate, line 183: you're not clearing intersects array beforehand, so you're adding the results from previous dependencies again. 
Better just call addIntersectResults once for each iteration of the loop below (once you've collected all intersections for a tile and its dependencies).

atomicsulfate, line 183: This could be moved to a separate private method to improve readability.

nzjony: Opps! Thanks for spotting. It seems that the `PickListener.ts` rejects duplicates so I didn't notice it in the test. But good to remove.

nzjony: Agreed, will do.

@here/harp-omv-datasource/test/MapViewPickingTest.ts
```diff
@@ -391,8 +404,30 @@ describe("MapView Picking", async function() {
...
+                                TileKey.fromRowColumnLevel(
+                                    (_tileKey.row + 1) % _tileKey.rowCount(),
+                                    _tileKey.column,
+                                    _tileKey.level
+                                )
+                            );
+                        }
+                        return tile;
+                    };
+                }
```
atomicsulfate, line 425: you can use sinon to simplify a little bit:
sinon.stub(geoJsonDataSource, "getTile").callsFake( yourFunction);

later to restore tje original function:
geoJsonDataSource.getTile.restore()

nzjony: Good point, will do.

nzjony:

@here/harp-omv-datasource/test/MapViewPickingTest.ts
```diff
@@ -391,8 +404,42 @@ describe("MapView Picking", async function() {
                           testCase.lookAt.length > 2 ? testCase.lookAt[2] : undefined
                       )
                     : rayOrigin;
+                // typeof required because the bind returns any
+                const getTile: typeof geoJsonDataSource.getTile = geoJsonDataSource.getTile.bind(
+                    geoJsonDataSource
+                );
```
atomicsulfate, line 410: actually this was the part you could get rid of now that you use sinon, since sinon saves for you the original function. anyway, it's minor.

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -149,23 +149,80 @@ export class PickHandler {
...
     }
 
+    // Adds the list of intersection results to the pick listener. The intersection results are
+    // cleared after being added.
+    private addIntersectResults(intersects: THREE.Intersection[], pickListener: PickListener) {
+        for (const intersect of intersects) {
+            pickListener.addResult(this.createResult(intersect));
+        }
+        intersects.length = 0;
+    }
```
atomicsulfate, line 183: make this one better a free function, it doesn't need anything from the class.

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -149,23 +149,80 @@ export class PickHandler {
...
+    // to the PickListener
+    private intersectObjects(
+        tile: Tile,
+        intersects: THREE.Intersection[],
+        pickListener: PickListener,
+        rayCaster: THREE.Raycaster
+    ) {
+        rayCaster.intersectObjects(tile.objects, true, intersects);
+        this.addIntersectResults(intersects, pickListener);
+    }
```
atomicsulfate, line 195: you can get rid of this method, just call rayCaster.intersectObjects directly inside the loop at intersectMapObjects, then call this.intersectDependentObjects, and then call addIntersectResults just once.

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -149,23 +149,80 @@ export class PickHandler {
...
+                tile.dataSource,
+                tileKey,
+                tile.offset,
+                this.mapView.frameNumber
+            );
+            if (otherTile !== undefined) {
+                rayCaster.intersectObjects(otherTile.objects, true, intersects);
+            }
+        }
+        this.addIntersectResults(intersects, pickListener);
```
atomicsulfate, line 223: better do this once per iteration of the loop at intersectMapObjects(). It's enough to do it just before processing the next tile.

@here/harp-omv-datasource/test/MapViewPickingTest.ts
nzjony: you mean using `callThrough()`, awesome, didn't know this was possible, thanks.

nzjony: I tried `const tile = stub.callThrough()(_tileKey, delayLoad);`, however this gives infinite recursion, how do I get access to the original method? I couldn't find anything here: https://sinonjs.org/releases/latest/stubs/

nzjony: Thanks, didn't know this!

@here/harp-mapview/lib/PickHandler.ts
nzjony: it calls `this.createResult`

nzjony: I removed this function, it was only called once anyway.

PR #1689, opened 2020-07-16
HARP-10772: Finish PickListener results explicitely.

atomicsulfate:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/PickHandler.ts
+++ b/@here/harp-mapview/lib/PickHandler.ts
@@ -162,6 +162,7 @@ export class PickHandler {
             }
         }
 
+        pickListener.finish();
         return pickListener.results;
     }
 
--- a/@here/harp-mapview/lib/PickListener.ts
+++ b/@here/harp-mapview/lib/PickListener.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { assert } from "@here/harp-utils";
 import { IntersectParams } from "./IntersectParams";
 import { PickResult } from "./PickHandler";
 
@@ -26,6 +27,7 @@ function defaultSort(lhs: PickResult, rhs: PickResult) {
 export class PickListener {
     private m_results: PickResult[] = [];
     private m_sorted: boolean = true;
+    private m_finished: boolean = true;
 
     /**
      * Constructs a new `PickListener`.
@@ -55,6 +57,7 @@ export class PickListener {
 
         if (foundFeatureIdx < 0) {
             this.m_sorted = false;
+            this.m_finished = false;
             this.m_results.push(result);
             return;
         }
@@ -64,6 +67,7 @@ export class PickListener {
         if (defaultSort(result, oldResult) < 0) {
             this.m_results[foundFeatureIdx] = result;
             this.m_sorted = false;
+            this.m_finished = false;
         }
     }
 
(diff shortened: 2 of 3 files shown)


atomicsulfate:
> Why do you make it now mandatory to call results explicity instead of having it as part of the call to `results` ?

It was precisely a review comment from Nino in a previous PR requesting this. I agree that side effects in a getter are unexpected.

nzjony:
> > Why do you make it now mandatory to call results explicity instead of having it as part of the call to `results` ?
> 
> It was precisely a review comment from Nino in a previous PR requesting this. I agree that side effects in a getter are unexpected.

True, but such side effects aren't visible to the external users through the existing interface?

atomicsulfate:
> > > Why do you make it now mandatory to call results explicity instead of having it as part of the call to `results` ?
> > 
> > 
> > It was precisely a review comment from Nino in a previous PR requesting this. I agree that side effects in a getter are unexpected.
> 
> True, but such side effects aren't visible to the external users through the existing interface?

The call to the getter potentially changes the state of PickListener  (set m_sorted to true). The caller cannot know this state change happened. Of course the getter documentation may give that hint, but it's still something unexpected a priori.

nzjony:
`m_sorted` is private. Even if `results` sorts the internal results, any user of this class can't observe what the state was before, and therefore can't prove that any side effect happened. The only observable side affect is that the `sortResults` method is a bit slower.

atomicsulfate:
> `m_sorted` is private. Even if `results` sorts the internal results, any user of this class can't observe what the state was before, and therefore can't prove that any side effect happened. The only observable side affect is that the `sortResults` method is a bit slower.

So you say that it's ok that the getter changes the private state of the object as long as there are no observable side effects.
Mind you, those side effects are actually observable: you add 2 results, then call the getter and you get just 1 (because maxResultCount was 1). 
But I see your point, as well as Nino's. I think any further discussion about this is not gonna be productive, it's hard to determine objectively what's the best option.
I'm good with how the code is right now. Unless @ninok cares to follow-up the discussion, I'll abandon this change since there's no consensus.

ninok:
> I'm good with how the code is right now. Unless @ninok cares to follow-up the discussion, I'll abandon this change since there's no consensus.

@nzjony I don't get your point. Yes, the side effect is not observable from the outside, but from the inside. I mean in the future this class might change and some methods could access `m_results` directly and some use `results`.
Why would you want to hide the fact, that there is some finalization step after all picking results are collected?

nzjony:
> Mind you, those side effects are actually observable: you add 2 results, then call the getter and you get just 1 (because maxResultCount was 1).

This I missed, good point. I would argue however that the user of this class can't know if the `addResult` or `results` method causes this behaviour, at least I can't think of a unit test where the side affect is observable.

nzjony:
> @nzjony I don't get your point. Yes, the side effect is not observable from the outside, but from the inside. I mean in the future this class might change and some methods could access `m_results` directly and some use `results`.
> Why would you want to hide the fact, that there is some finalization step after all picking results are collected?

@ninok , I see your point now, thanks. I would hope that anyone changing this would make a unit test to check their assumptions to avoid such regressions.

I would hide it for the fact that adding an extra method makes it more difficult / error prone to use, I would suggest then to either:

- change `results` to `sortedResults`, so it is clear that `this.m_results` != `results`. OR
- apply the sort after each `addResult`, sorting an already sorted list is (probably) quite quick, and we are talking less than 50 items right, so this is probably good enough (especially considering this happens when the map is not moving, so performance is unlikely to be a huge concern)

But this is more my personal opinion than a strong recommendation, so I won't veto this going through.

PR #1693, opened 2020-07-16
HARP-11091: Add DataProvider.onDidInvalidate

robertoraggi:
This change introduces an "invalidated" event to DataProvider
and adds an automatic way to clear caches when the
geojson data provider input is updated.

### DIFF
--- a/@here/harp-omv-datasource/test/OmvDataSourceTest.ts
+++ b/@here/harp-omv-datasource/test/OmvDataSourceTest.ts
@@ -7,12 +7,17 @@
 // tslint:disable:only-arrow-functions
 //    Mocha discourages using arrow functions, see https://mochajs.org/#arrow-functions
 
+import { FeatureCollection } from "@here/harp-datasource-protocol";
 import "@here/harp-fetch";
 import { TileKey } from "@here/harp-geoutils";
 import { DataProvider } from "@here/harp-mapview-decoder";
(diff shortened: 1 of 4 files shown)


@here/harp-omv-datasource/test/OmvDataSourceTest.ts
```diff
@@ -143,4 +148,59 @@ describe("DataProviders", function() {
...
         });
     });
+
+    it("DataProvider clears the cache", function() {
+        const geojson: FeatureCollection = {
+            type: "FeatureCollection",
+            features: []
+        };
+
+        const markTilesDirty = sinon.fake();
```
nzjony, line 158: why do you need this? It doesn't seem to be used.

robertoraggi: it is needed to get the decoder registration working, without it the test fails

```
(node:37761) UnhandledPromiseRejectionWarning: TypeError: this.mapView.markTilesDirty is not a function
    at OmvDataSource.configureDecoder (harp.gl/@here/harp-omv-datasource/lib/OmvDataSource.ts:419:22)
```

nzjony: Ok, in that case you could make it inline, but minor..

PR #1699, opened 2020-07-20
Add environment variable HARP_NO_HARD_SOURCE_CACHE.

robertoraggi:
The `HARP_NO_HARD_SOURCE_CACHE` env variable can be used to
control the usage of `hard-source-webpack-plugin`. For example:

    HARP_NO_HARD_SOURCE_CACHE=true yarn start

Also, this change disables the usage of the hard-source cache
in the CI.

Note that the value assigned to `HARP_NO_HARD_SOURCE_CACHE` is
not important, that is, as long as the environment variable
`HARP_NO_HARD_SOURCE_CACHE` is defined the hard-source cache will
be disabled.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,6 +49,8 @@ jobs:
         node-version: ${{ matrix.node_version }}
     - name: Install dependencies
       run: yarn
+    - name: Disable hard-source-webpack-plugin
+      run: echo "::set-env name=HARP_NO_HARD_SOURCE_CACHE::true"
     - name: Pretest
       run: yarn run pre-test
       shell: bash
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ branches:
 
 env:
   - DETECT_CHROMEDRIVER_VERSION=true
+  - HARP_NO_HARD_SOURCE_CACHE=true
 
 before_install:
   - yarn global add codecov
--- a/@here/harp-examples/webpack.config.js
+++ b/@here/harp-examples/webpack.config.js
@@ -102,12 +102,15 @@ const commonConfig = {
     // @ts-ignore
     mode: process.env.NODE_ENV || "development",
     plugins: [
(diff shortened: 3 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -49,6 +49,8 @@ jobs:
         node-version: ${{ matrix.node_version }}
     - name: Install dependencies
       run: yarn
+    - name: Disable hard-source-webpack-plugin
+      run: echo "::set-env name=HARP_NO_HARD_SOURCE_CACHE::true"
```
ninok, line 53: Wow, that's a really weird syntax. I saw this action in the marketplace but might be a bit overkill:
https://github.com/marketplace/actions/set-env

zbigg: The proper syntax for _global_ env variables - like this one is to use `env` section which is above in this file:

```
env:
  HARP_NO_HARD_SOURCE_CACHE: true
```

robertoraggi: @zbigg  but that is if you always want to have it. set-env as step can be configured, so we could re-enable in certain cases.

robertoraggi: @zbigg  also `env` is not what we want in our case because as far as I know it doesn't propagate to the next steps

from https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#about-environment-variables

You can also use the set-env workflow command to set an environment variable that the following steps in a workflow can use.

zbigg: Well, in this case we run few `webpack` throughout flow and i think that _global_ and _constant_ `HARP_NO_HARD_SOURCE_CACHE` would do the trick with no harm to steps that ignore it.

The `::set-env` is useful if you need to derive env variable value from output of command in very specific point of pipeline. In this case, i don't see any reason why `yarn` or cache setup would suffer from having our variable set.

YMMV though :)

PR #1708, opened 2020-07-21
HARP-10772: Add pickable flag to object's userData.

atomicsulfate:

The flag may be set to false to disable picking for particular objects. Picking is disabled for depth prepass, background, edges and center line of solid lines with outline (knowing whether the outline is picked is enough).

### DIFF
--- a/@here/harp-mapview/lib/MapObjectAdapter.ts
+++ b/@here/harp-mapview/lib/MapObjectAdapter.ts
@@ -6,7 +6,7 @@
 
 import * as THREE from "three";
 
-import { GeometryKind, Technique } from "@here/harp-datasource-protocol";
+import { GeometryKind, getPropertyValue, MapEnv, Technique } from "@here/harp-datasource-protocol";
 import { MapAdapterUpdateEnv, MapMaterialAdapter } from "./MapMaterialAdapter";
 
 /**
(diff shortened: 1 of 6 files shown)


atomicsulfate:
> how is this flag different than transient? 
pickable is a property used here to skip picking for some objects like deph prepasses and background that don't have their own associated style. Also, they are not transient, like the geometry created for selection.

> are they kept in sync somehow? are objects created with `transient = true` automatically get pickable == false?

Good idea, I'll set pickable to false for transient objects as well.

robertoraggi:
> > how is this flag different than transient?
> > pickable is a property used here to skip picking for some objects like deph prepasses and background that don't have their own associated style. Also, they are not transient, like the geometry created for selection.
> 
> > are they kept in sync somehow? are objects created with `transient = true` automatically get pickable == false?
> 
> Good idea, I'll set pickable to false for transient objects as well.

transient is a technique property so it can be a dynamic expressions. We should make sure that pickable doesn't conflict with transient.

atomicsulfate:
> > > how is this flag different than transient?
> > > pickable is a property used here to skip picking for some objects like deph prepasses and background that don't have their own associated style. Also, they are not transient, like the geometry created for selection.
> > 
> > 
> > > are they kept in sync somehow? are objects created with `transient = true` automatically get pickable == false?
> > 
> > 
> > Good idea, I'll set pickable to false for transient objects as well.
> 
> transient is a technique property so it can be a dynamic expression. We should make sure that pickable doesn't conflict with transient.

Ok, I've handled it as we've discussed, as a method in MapObjectAdapter.
I'll get rid of that code from decoders that removes ids for transient geometry as part of the ticket you created: https://saeljira.it.here.com/browse/HARP-11147

PR #1709, opened 2020-07-22
Change error message to warning to avoid confusion

ninok:


ninok:
> I guess we need to also change the logger type for this isnt it? because this is also showing as error.
> `[browser] error: THREE.WebGLRenderer: Error creating WebGL context.`

This message is comming from three.js. They catch the exception, log it to the console and re-throw it:
https://github.com/mrdoob/three.js/blob/master/src/renderers/WebGLRenderer.js#L256

So there is nothing we can do about it.

ninok:

ninok:

gairik:
> > I guess we need to also change the logger type for this isnt it? because this is also showing as error.
> > `[browser] error: THREE.WebGLRenderer: Error creating WebGL context.`
> 
> This message is comming from three.js. They catch the exception, log it to the console and re-throw it:
> https://github.com/mrdoob/three.js/blob/master/src/renderers/WebGLRenderer.js#L256
> 
> So there is nothing we can do about it.

`getWebGlInfo()` this itself is doing the work as well as logging. so we cannot do anything, like you mentioned. got it now.

PR #1716, opened 2020-07-23
Add action to generate the docs.

robertoraggi:
Add manually triggered action to generate the docs using
api-extractor/documenter.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,27 @@
+name: Generate Docs
+
+env:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,22 @@
+name: Docs
+
+env:
+  PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
```
ninok, line 4: Don't we need puppeteer to generate the UML diagrams?

PR #1721, opened 2020-07-23
HARP-10881: Support partial disabling of building edges

ninok:


test/rendering/StylingTest.ts
```diff
@@ -177,7 +178,7 @@ function mapViewFeaturesRenderingTest(
 
                 const grid = new THREE.GridHelper(gridSize, gridDivisions, 0xff0000, 0x00ff00);
                 grid.geometry.rotateX(Math.PI / 2);
-                (grid as MapAnchor).geoPosition = position;
+                (grid as MapAnchor).anchor = position;
```
ninok, line 181: This just removes an annoying deprecation warning

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -931,6 +931,7 @@ export class TileGeometryCreator {
 
                 if (renderDepthPrePass) {
                     const depthPassMesh = createDepthPrePassMesh(object as THREE.Mesh);
+                    this.addUserData(tile, srcGeometry, technique, depthPassMesh);
```
atomicsulfate, line 934: why do we need user data for depth prepass meshes?

ninok: b/c we also need to generate the appropriate draw calls for the enabled buildings in the depth-pre-pass.
Visually it makes no difference but the depth buffer would be polluted with all buildings (instead of only the enabled ones) and that might break usecases that rely on a proper depth buffer (e.g. car position marker).

Nevertheless, you have a point that we could reuse the userData from the "main" mesh so we would also avoid to re-compute the draw ranges again. Same for the outlines.
I will create a dedicated ticket for that.

ninok: Thinking about it again, it should even make a difference visually in a highly tilted view. Unfortunately that's not covered by our testcases

PR #1726, opened 2020-07-28
HARP-11250: Add support for style extensions.

robertoraggi:
Add support for extending and overriding existing style rules.

An existing style can be extended by referencing its `id` in
an `extends` property, for example:

base.json
```json
{
   "styles": [
     {
        "styleSet": "tilezen",
        "id": "extruded-building-rule",
        "layer": "buildings",
        "when": ["==", ["geometry-type"], "Polygon"],
        "technique": "extruded-building",
        "color": "red"
     }
  ]
}
```

can be extended by `derived.json` using

```json
{
   "extends": [
       "./base.json",
   ],
   "styles": [
     {
        "styleSet": "tilezen",
        "id": "extruded-building-rule",
        "color": "yellow",
        "minZoomLevel": 16
     }
  ]
}
```

Alternatively, the rule `extruded-building-rule` can be replaced by simply reusing its id in a derived theme, e.g.
```jsonc
{
   "extends": [
       "./base.json",
   ],
   "styles": [
     {
        "styleSet": "tilezen",
        "id": "extruded-building-rule",
        "layer": "buildings",
        "when": ["==", ["geometry-type"], "Polygon"],
        "technique": "fill", // use `fill` instead of `extruded-polygon`
        "color": "yellow"
     }
  ]
}
```

Also, this PR removes references to style and inline literal definitions.

### DIFF
--- a/@here/harp-datasource-protocol/lib/Theme.ts
+++ b/@here/harp-datasource-protocol/lib/Theme.ts
@@ -361,19 +361,6 @@ export interface StyleAttributes<Technique, Params> {
      */
     debug?: boolean;
 
-    // TODO: Make pixel units default.
-    /**
-     * Units in which different size properties are specified. Either `Meter` (default) or `Pixel`.
-     *
-     * @deprecated use "string encoded numerals" as documented in TODO, wher eis the doc ?
-     */
-    metricUnit?: "Meter" | "Pixel";
-
-    /**
-     * XYZ defines the property to display as text label of a feature in the styles.
-     */
-    labelProperty?: string;
-
     attr?: Partial<Params>;
 }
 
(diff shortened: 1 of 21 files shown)


@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -652,63 +505,54 @@ export interface BaseLight {
...
  */
 export interface DirectionalLight extends BaseLight {
+    /**
+     * The type of the light.
+     */
     type: "directional";
+
     /**
-     * @format color-hex
+     * The color of this ambient light.
```
ninok, line 537: fix the comments. It is not an ambient light but a directional light.

@here/harp-examples/src/styling_extend-a-theme.ts
```diff
@@ -87,13 +87,6 @@ export namespace HelloCustomThemeExample {
                     parkColor: {
                         type: "color",
                         value: "#00aa33"
-                    },
```
ninok, line 90: Can you rewrite this example to showcase the new extend functionality

@here/harp-datasource-protocol/lib/Theme.ts
robertoraggi: fixed

@here/harp-examples/src/styling_extend-a-theme.ts
robertoraggi: done, I extended the example to show how to override the extruded building style with a rule that renders the buildings with fill.

PR #1729, opened 2020-07-28
HARP-11228: Properly dispose MapView

StefanDachwitz:

# CI Check

Happy contributing!


@here/harp-mapview/lib/MapView.ts
```diff
@@ -141,7 +142,9 @@ export enum MapViewEventNames {
     /** Called when the WebGL context is restored. */
     ContextRestored = "webglcontext-restored",
     /** Called when camera position has been changed. */
-    CameraPositionChanged = "camera-changed"
+    CameraPositionChanged = "camera-changed",
+    /** Called when dispose has been called. */
+    Dispose = "dispose"
```
ninok, line 147: Shouldn't it be dispose**d** since the event is emitted after disposing happened

@here/harp-mapview/lib/EventDispatcher.ts
```diff
@@ -0,0 +1,130 @@
...
+ */
+
+import * as THREE from "three";
+
+/**
+ * JavaScript events for custom objects. Stores all listeners to allow removing all listeners for
+ * housekeeping.
+ *
+ */
+export class EventDispatcher {
```
ninok, line 14: Please add a note that we will switch back to `THREE.EventDispatcher` once https://github.com/mrdoob/three.js/pull/19844 is released

@here/harp-mapview/lib/MapView.ts
ninok: Just checked the code and it is called before the actual disposing happens. So please adjust the comment to make it more clear

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1261,6 +1269,12 @@ export class MapView extends THREE.EventDispatcher {
      * cleanup, you must ensure that all references to this `MapView` are removed.
      */
     dispose() {
+        this.m_disposed = true;
+
+        // Enforce listeners that we are about to dispose.
+        DISPOSE_EVENT.time = Date.now();
+        this.dispatchEvent(DISPOSE_EVENT);
```
ninok, line 1276: Since the event is fired before the actual disposal happens I would also move `this.m_disposed = true;` below the dispatch event call

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1281,8 +1295,19 @@ export class MapView extends THREE.EventDispatcher {
         this.m_textElementsRenderer.clearRenderStates();
         this.m_renderer.dispose();
         this.m_imageCache.clear();
+        this.m_tileGeometryManager.clear();
 
         this.m_movementDetector.dispose();
+
+        // Remove all event handlers.
+        super.dispose();
```
ninok, line 1303: We are still missing the termination of the workers.
I'm actually not sure how to do it b/c `ConcurrentDecoderFacade` and `ConcurrentTilerFacade` are using some singleton-ish pattern.
If we have multiple `MapView`s, we need to do some reference counting to only terminate the workers if no `MapView` needs them anymore.

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: Will do.

StefanDachwitz: Will do.

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: OK...

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: The WorkerBasedDecoder is disposed of, and the decoders are stopped, no events from them should arrive at MapView after that. I guess for all normal use cases this is good enough.

@here/harp-mapview/lib/EventDispatcher.ts
StefanDachwitz: will do.

@here/harp-mapview/lib/MapView.ts
ninok: I see, there is some reference counting in `WorkerBasedDecoder` and `WorkerBasedTiler`.

Nevertheless I cannot see that we call dispose on the `ConcurrentTilerFacade` anywhere.

StefanDachwitz: I will add that.

PR #1736, opened 2020-08-03
Rename the docs workflow file.

robertoraggi:
Apparently the docs workflow is marked as `disabled` and
as far as I know there is no way to change the state of a workflow.

This change renames docs.yaml to docs.yml with the home that
this is enough to get a new enabled workflow-id.


robertoraggi:
abandon, not needed.

PR #1743, opened 2020-08-05
(HARP-11092) Adds Bounds Generation

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-examples/src/bounds-generation.ts
+++ b/@here/harp-examples/src/bounds-generation.ts
@@ -1,0 +1,186 @@
+/*
+ * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {
+    FeaturesDataSource,
+    MapViewFeature,
+    MapViewMultiPointFeature,
+    MapViewPolygonFeature
+} from "@here/harp-features-datasource";
+import { GeoCoordinates, mercatorProjection } from "@here/harp-geoutils";
+import { MapControls, MapControlsUI } from "@here/harp-map-controls";
+import { BoundsGenerator, CopyrightElementHandler, MapView } from "@here/harp-mapview";
+import { OmvDataSource } from "@here/harp-omv-datasource";
+
+import { apikey } from "../config";
+
+export namespace BoundsExample {
+    // Create a new MapView for the HTMLCanvasElement of the given id.
+    function initializeMapView(id: string): MapView {
+        const canvas = document.getElementById(id) as HTMLCanvasElement;
+
+        // Look at BERLIN
+        const BERLIN = new GeoCoordinates(52.5186234, 13.373993);
+        const map = new MapView({
+            canvas,
+            projection: mercatorProjection,
+            tileWrappingEnabled: false,
+            theme: {
+                extends: "resources/berlin_tilezen_night_reduced.json",
+                styles: {
+                    geojson: [
+                        {
+                            when: ["all", ["==", ["geometry-type"], "Polygon"]],
+                            technique: "fill",
+                            renderOrder: 10003,
+                            attr: {
+                                color: ["get", "color"], // "#77ccff",
+                                opacity: 0.5,
+                                lineWidth: 1,
+                                lineColor: "#ff0000",
+                                enabled: true
+                            }
+                        },
+                        {
+                            when: "$geometryType == 'point'",
+                            technique: "circles",
+                            renderOrder: 10004,
+                            attr: {
+                                color: "#f0f",
+                                size: 10
+                            }
+                        }
+                    ]
+                }
+            },
+            target: BERLIN,
+            zoomLevel: 8,
+            tilt: 45,
+            heading: -80
+        });
+        map.renderLabels = false;
+
+        CopyrightElementHandler.install("copyrightNotice", map);
+
+        const mapControls = new MapControls(map);
+        mapControls.maxTiltAngle = 180;
+
+        const ui = new MapControlsUI(mapControls, { zoomLevel: "input" });
+        canvas.parentElement!.appendChild(ui.domElement);
+
+        map.resize(window.innerWidth, window.innerHeight);
+
+        window.addEventListener("resize", () => {
+            map.resize(window.innerWidth, window.innerHeight);
+        });
+
+        addOmvDataSource(map);
+        const featureList: MapViewFeature[] = []; //createFeatureList();
+        let featuresDataSource: FeaturesDataSource | undefined;
+        //@ts-ignore
+        featuresDataSource = addFeaturesDataSource(map, featureList);
+        let viewpoly: MapViewPolygonFeature;
+        let cornerpoints: MapViewMultiPointFeature;
+        const boundsGenerator = new BoundsGenerator(
+            map.camera,
+            map.projection,
+            map.tileWrappingEnabled
+        );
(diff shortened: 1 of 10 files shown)


@here/harp-mapview/lib/BoundsGenerator.ts
```diff
@@ -0,0 +1,296 @@
...
+} from "@here/harp-geoutils";
+import { Frustum, Line3, Matrix4, Plane, Vector2, Vector3 } from "three";
+
+import { TileGeometryCreator } from "./geometry/TileGeometryCreator";
+import { MapView } from "./MapView";
+import { MapViewUtils } from "./Utils";
+
+export class BoundsGenerator {
+    //TODO: check what we really need here
+    constructor(private m_mapView: MapView) {}
```
ninok, line 21: Please don't pass whole mapview but only the things that are needed (i.e. projection and camera)

@here/harp-mapview/lib/BoundsGenerator.ts
```diff
@@ -0,0 +1,296 @@
...
+        //TODO: support spherical projection
+
+        //!!!!!!!ALTITUDE IS NOT TAKEN INTO ACCOUNT!!!!!!!!!
+        const geoPolygon: GeoCoordinates[] = [];
+
+        //CASE A: FLAT PROJECTION
+        if (this.m_mapView.projection.type === ProjectionType.Planar) {
+            // 1.) Raycast into all for corners of the canvas
+            //     => if an intersection is found, add it to the polygon
+            this.raycastCanvasCorners(geoPolygon);
```
ninok, line 33: Not sure what we want to return in case not all rays hit the surface but wouldn't it work if we fallback to raycasting the far-plane in that case? That would reduce all the special case handling below.

@here/harp-mapview/lib/BoundsGenerator.ts
```diff
@@ -0,0 +1,296 @@
...
+            if (geoPolygon.length === 4) {
+                return geoPolygon;
+            }
+
+            //2.) Raycast into the two corners of the horizon cutting the canvas sides
+            //    => if intersection is found, add it to the polygon
+            this.rayCastOnHorizonCorners(geoPolygon);
+
+            //Setup the frustum for further checks
+            const frustum = new Frustum().setFromProjectionMatrix(
```
ninok, line 46: Can we not reuse the frustum that is also used for the visible tile intersection?

@here/harp-mapview/lib/BoundsGenerator.ts
```diff
@@ -0,0 +1,296 @@
...
+                new Matrix4().multiplyMatrices(
+                    this.m_mapView.camera.projectionMatrix,
+                    this.m_mapView.camera.matrixWorldInverse
+                )
+            );
+
+            // Setup the world corners for further checks
+            const worldCorners = TileGeometryCreator.instance.generateTilePlaneCorners(
+                mercatorTilingScheme.getGeoBox(new TileKey(0, 0, 0)),
```
ninok, line 55: Is mercatorTilingScheme correct if projection != webMercator?
I guess you should use projection.worldExtent instead

@here/harp-mapview/lib/BoundsGenerator.ts
FraukeF: I actually tried to to that, as it only needs the camera and the projection, but it has a dependency on MapViewUtils.rayCastGeoCoordinates, wich needs the mapView (although its also only using the camera and projection)

@here/harp-mapview/lib/BoundsGenerator.ts
FraukeF: Not sure I understand your point, in case no surface is hit, there would be no return value, as its not part of the searched polygon. 
What do you mean with raycasting the far-plane, raycast the CanvasCorners against the farplane will result in the back corners of the frustum, the intersection of the groundPlane with the farplane is done with the next check looking for the horizon,

@here/harp-mapview/lib/BoundsGenerator.ts
FraukeF: probably, it should be the same, I ll check how I can expose it for reuse

@here/harp-mapview/lib/BoundsGenerator.ts
FraukeF: good catch, yes, probably there couldbe differences in the bounding box, it should use the mapView.projection

@here/harp-mapview/lib/BoundsGenerator.ts
FraukeF: I added the option to only pass camera and projection to the MapViewUtils.rayCastWorldCoordinates, therefore mapview is not needed anymore

@here/harp-mapview/lib/BoundsGenerator.ts
FraukeF: The FrustumIntersection is nested inside the VisibleTileSet, and creates the frustum as private property. The only place this and BoundsGenerate have in common right now is the MapView. But i think adding the frustum to MapView would just overload MapView even more. Would be nice to have it directly on the camera... but we are using THREE.Camera, .. ideas are appreciated, but for now I would keep it here, it should not be to expensive to have this instance here.

ninok: It would just be a private property in `MapView` that is passed to the ctor of `VisibleTileSet` and to the ctor of `BoundGenerator`.
I agree the computation is not that expensive but it's at least fishy to have 2 frustums. If some magic is added to the computation of one of them we end up using two different frustums.
By just writing that sentence I realized it is actually the case. Check:
https://github.com/heremaps/harp.gl/blob/master/@here/harp-mapview/lib/FrustumIntersection.ts#L117

FraukeF: I had a deeper look at the Frustum being used in FrustumIntersection.ts, there is two reasons, why I think we should/can not share the two:
1. In some cases (terrain) the FrustumIntersection is using a frustum that does not match the visible frustum of the actual camera, to calculate the needed tiles. In the case of the BoundsGeneration, this frustum would not be correct, as the bounds calculation  needs the actually visible frustum.
2. The frustum changes with the camera, we already pass the camera to the BoundsGeneration which is enough to retrieve the Frustum, the camera is controlled and updated by the MapView in our case, but could basically be any camera that does not have to be related to MapView. If we create another dependency on a Frustum, the updating of this frustum depending on the given camera would need to be taken care of, which is redundant and error prone

@here/harp-examples/src/bounds-generation.ts
```diff
@@ -0,0 +1,186 @@
+/*
```
ninok, line 1: Can you add globe and wrap around options to the example please. (Globe is available in MapControlsUI already)

@here/harp-examples/src/bounds-generation.ts
```diff
@@ -0,0 +1,186 @@
...
+        featuresDataSource = addFeaturesDataSource(map, featureList);
+        let viewpoly: MapViewPolygonFeature;
+        let cornerpoints: MapViewMultiPointFeature;
+        const boundsGenerator = new BoundsGenerator(
+            map.camera,
+            map.projection,
+            map.tileWrappingEnabled
+        );
+
+        window.addEventListener("keyup", event => {
```
nzjony, line 92: please add some instructions how to use this example, otherwise without looking at the examples code, I don't know how to use it

PR #1744, opened 2020-08-05
HARP-8067: Rename OmvDataSource to VectorTileDataSource

robertoraggi:
Also, removed unneeded symbols like OmvTile.


@here/harp-omv-datasource/lib/OmvDataSource.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @example
+ * ```typescript
+ *    const dataSource = new OmvDataSource({
+ *        baseUrl: "https://vector.hereapi.com/v2/vectortiles/base/mc",
+ *        authenticationCode: apikey
+ *    });
+ *    mapView.addDataSource(dataSource);
+ *   ```
+ */
+export class OmvDataSource extends VectorTileDataSource {}
```
ninok, line 22: Can you mark it as deprecated or do we keep it as a utility class for easier HERE Vector Tile initialization?

robertoraggi: Good idea, I will mark all the symbols as deprecate in a follow up change.

PR #1745, opened 2020-08-06
HARP-8067: Move the data adapters to sub directories.

robertoraggi:
This change moves the data adapters to their own sub directory.


ninok:
Will this PR break user code if somebody is using data providers directly? We should at least announce that in the next release notes or ideally maintain alias for some time.

PR #1753, opened 2020-08-11
MINOR: Add safe version for screen to geo conversion

ninok:
In case the ray-casting of the screen coordinate is not hitting the earth surface a fallback to the far plane will be computed

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2646,6 +2646,38 @@ export class MapView extends EventDispatcher {
      * Returns the world space position from the given screen position.
      *
      * @remarks
+     * This method will always return a point. if the given `(x, y)` value is not intersecting
+     * the ground plane a fallback point on the far plane will be computed.
+     * Therfore the returned point might not be on the earth surface.
+     *
+     * @param x - The X position in css/client coordinates (without applied display ratio).
+     * @param y - The Y position in css/client coordinates (without applied display ratio).
+     */
+    getWorldPositionAtSafe(x: number, y: number): THREE.Vector3 {
+        this.m_raycaster.setFromCamera(this.getNormalizedScreenCoordinates(x, y), this.m_camera);
+        const worldPos =
+            this.projection.type === ProjectionType.Spherical
+                ? this.m_raycaster.ray.intersectSphere(this.m_sphere, cache.vector3[0])
+                : this.m_raycaster.ray.intersectPlane(this.m_plane, cache.vector3[0]);
+
+        if (worldPos === null) {
+            // Fall back to the far plane
+            const cosAlpha = this.m_camera
+                .getWorldDirection(cache.vector3[0])
+                .dot(this.m_raycaster.ray.direction);
+
+            return cache.vector3[0]
+                .copy(this.m_raycaster.ray.direction)
+                .multiplyScalar(this.m_camera.far / cosAlpha)
+                .add(this.m_camera.position);
+        }
+        return worldPos;
+    }
+
+    /**
+     * Returns the world space position from the given screen position.
+     *
+     * @remarks
      * The return value can be `null`, in case the camera is facing the horizon
      * and the given `(x, y)` value is not intersecting the ground plane.
      *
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -2642,6 +2642,38 @@ export class MapView extends EventDispatcher {
...
+     * @remarks
+     * This method will always return a point. if the given `(x, y)` value is not intersecting
+     * the ground plane a fallback point on the far plane will be computed.
+     * Therfore the returned point might not be on the earth surface.
+     *
+     * @param x - The X position in css/client coordinates (without applied display ratio).
+     * @param y - The Y position in css/client coordinates (without applied display ratio).
+     */
+    getWorldPositionAtSafe(x: number, y: number): THREE.Vector3 {
+        this.m_raycaster.setFromCamera(this.getNormalizedScreenCoordinates(x, y), this.m_camera);
```
nzjony, line 2657: 2657 to 2662 seem to be a repeat of the `getWorldPositionAt` method. Why not just call that method and set its return value to `worldPos`?

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2659,6 +2691,29 @@ export class MapView extends EventDispatcher {
...
+     * given screen position.
+     *
+     * @remarks
+     * This method will always return a point. if the given `(x, y)` value is not intersecting
+     * the ground plane a fallback point on the far plane will be computed.
+     * Therfore the returned point might not be on the earth surface (i.e. altitude > 0)
+     *
+     * @param x - The X position in css/client coordinates (without applied display ratio).
+     * @param y - The Y position in css/client coordinates (without applied display ratio).
+     * @param fallback - Fallback to
```
nzjony, line 2705: `fallback` not part of the parameter

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2659,6 +2691,29 @@ export class MapView extends EventDispatcher {
...
+     */
+    getGeoCoordinatesAtSafe(x: number, y: number): GeoCoordinates {
+        const worldPosition = this.getWorldPositionAtSafe(x, y);
+        const geoPos = this.projection.unprojectPoint(worldPosition);
+        if (!this.tileWrappingEnabled && this.projection.type === ProjectionType.Planar) {
+            // When the map is not wrapped we clamp the longitude
+            geoPos.longitude = THREE.MathUtils.clamp(geoPos.longitude, -180, 180);
```
nzjony, line 2712: won't this be handled by the normalized function?

ninok: no, normalize is for example converting a latitude of 181 to -179 what only makes sense if the world is repeated.

@here/harp-mapview/lib/MapView.ts
ninok: Oh, missed that one. Initially I was thinking about adding an additional parameter instead of creating a new function but I did not manage to convince TypeScript to have a different return type depending on the value of that parameter. Let me try a bit more.

@here/harp-mapview/lib/MapView.ts
nzjony: Ok, I get it now, it should clamp, which isn't handled as you mentioned.

@here/harp-mapview/test/MapViewTest.ts
```diff
@@ -839,6 +842,258 @@ describe("MapView", function() {
...
+            tileWrappingEnabled: false
+        });
+        const center = mapView.getGeoCoordinatesAt(canvas.clientWidth / 2, canvas.clientHeight / 2);
+        assert.isNotNull(center);
+        assert.closeTo(center!.latitude, target.latitude, eps);
+        assert.closeTo(center!.longitude, target.longitude, eps);
+        assert.isDefined(center!.altitude);
+        assert.closeTo(center!.altitude!, 0, eps);
+
+        // Clicking "behind" the anti-meridian should wrap around
```
nzjony, line 1087:
```suggestion
        // Clicking "behind" the anti-meridian should clamp
```

@here/harp-mapview/test/MapViewTest.ts
```diff
@@ -839,6 +842,258 @@ describe("MapView", function() {
...
```
nzjony, line 1058:
```suggestion
        // Clicking "behind" the anti-meridian should clamp
```

PR #1755, opened 2020-08-11
MINOR: Adds projection setter to BoundsGenerator

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/test/BoundsGeneratorTest.ts
+++ b/@here/harp-mapview/test/BoundsGeneratorTest.ts
@@ -106,6 +106,18 @@ describe("BoundsGenerator", function() {
         assert.isEmpty(geoPolygon);
     });
 
+    it("generates empty polygon for spherical projection,  after projection changes", function() {
+        mapView?.lookAt(lookAtParams);
+        mapView?.renderSync(); // render once to update camera parameter
+        let geoPolygon = (boundsGenerator as BoundsGenerator).generate();
+        assert.isNotEmpty(geoPolygon);
+        assert.equal(geoPolygon.length, 4);
+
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/test/BoundsGeneratorTest.ts
```diff
@@ -106,6 +106,18 @@ describe("BoundsGenerator", function() {
         assert.isEmpty(geoPolygon);
     });
 
+    it("generates empty polygon for spherical projection,  after projection changes", function() {
+        mapView?.lookAt(lookAtParams);
+        mapView?.renderSync(); // render once to update camera parameter
```
ninok, line 111: this should not be necessary. LookAt is already updating everything. If not it is a bug and should be fixed

FraukeF: It is actually not updating, but this is also not a bug in MapView, but an issue of how the stubbed MapView in the test is created. 
As we are running also in node context, basically the MapView is not complete and the Renderer and AnimationFrameRequest etc are just stubbed.
The MapView in this Test is not really needed but just used as a helper to set the camera parameters, that are used in the BoundsGenerator.

ninok: I think we should fix the near/far plane computation to be updated when look at is called and not only in the next animation frame. At least we could compute the near and far plane based on the camera and min/max elevation settings.
When the actual rendering happens we can then refine it based on the actual min/max elevation of the visible tiles.

FraukeF: I ll create a ticket for it

PR #1756, opened 2020-08-11
HARP-10238: Preparation to support WebGL2.

atomicsulfate:
New WebGL1RawShaderMaterial must be extended by all raw materials
using webgl1 compatible glsl code. The base class coverts the shaders
to webgl2 compatible code when a webgl2 context is used.

Rename 'sample' variable in text shaders, since it's a keyword in webgl2.

To support WebGL2, glsl version must be prepended to the shader code. This
can only be done in THREE.js, since onBeforeCompile() callback is called before
adding shader defines, and #version directive must be on the first line.
The following PR in THREE.js address this issue:
https://github.com/mrdoob/three.js/pull/20005

Once merged, new glslversion property must be set at WebGL1RawShaderMaterial, so
that THREE.js prepends the appropiate glsl version.

# CI Check

Happy contributing!


@here/harp-text-canvas/package.json
```diff
@@ -24,6 +24,7 @@
     },
     "devDependencies": {
         "@here/harp-geoutils": "^0.18.0",
+        "@here/harp-materials": "^0.18.0",
```
ninok, line 27: This should be a real dependency, otherwise the typings would not work for the users of this package.
Ideally I would prefer to not have this dependency b/c the plan was to factor out text canvas as a standalone module. Maybe we should move it to harp utils.
Do you know why we have geoutils as a dev dependency?

atomicsulfate: @ninok  thanks for the review.

> This should be a real dependency, otherwise the typings would not work for the users of this package.
sure will fix it.

> Ideally I would prefer to not have this dependency b/c the plan was to factor out text canvas as a standalone module. Maybe we should move it to harp utils.
If I move WebGL1RawShaderMaterial to harp-utils, harp-utils will have a dependency to three.js, which is even nastier.
The only alternative I can think of is to move only the glsl conversion functions to harp-utils, then duplicate the onBeforeCompile implementation directly in the text materials. What do you think?

> Do you know why we have geoutils as a dev dependency?
I'll remove it and check if it breaks anything.

@here/harp-materials/lib/WebGL1RawShaderMaterial.ts
```diff
@@ -0,0 +1,61 @@
...
+        "\n" +
+        fragmentShader
+    );
+}
+
+/**
+ * Material that converts webgl1 shaders to webgl2 compatible code when renderer uses a WebGL2
+ * context.
+ */
+export class WebGL1RawShaderMaterial extends THREE.RawShaderMaterial {
```
nzjony, line 42: The name is a bit confusing IMO, it seems to imply that the shader only works with WebGL1, however based on the comment, it seems to do the opposite.

atomicsulfate: This is the base class of all the materials using webgl1 compatible shaders, hence the name.
It does convert the shaders to webgl2 if necessary though. Please suggest an alternative name.

nzjony: I think just "RawShaderMaterial" is fine, the conversion of the shader is just some technical detail that isn't really relevant to the user.

atomicsulfate: ok.

@here/harp-text-canvas/package.json
atomicsulfate: done

PR #1759, opened 2020-08-12
MAPSJS-2709 Can toggle tile wrapping enabled

nzjony:


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1261,6 +1261,13 @@ export class MapView extends EventDispatcher {
         return this.m_tileWrappingEnabled;
     }
 
+    set tileWrappingEnabled(enabled: boolean) {
+        if (enabled !== this.m_tileWrappingEnabled) {
+            this.m_tileWrappingEnabled = enabled;
+            this.m_visibleTiles = this.createVisibleTileSet();
```
FraukeF, line 1267: You should check if the Projection is Planar before creating an new tile set, for the spherical projection the tileWrapping property should not change anything

nzjony: Opps, missed your comment, thanks, I will fix!

PR #1766, opened 2020-08-13
HARP-11458 Detached tiles don't dispose via VisibleTileSet

nzjony:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
+++ b/@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
@@ -105,6 +105,9 @@ describe("TileGeometryLoader", function() {
     });
 
     beforeEach(function() {
+        if (dataSource.isDetached()) {
+            dataSource.attach(mapView);
+        }
         tile = dataSource.getTile(tileKey)!;
         geometryLoader = new TileGeometryLoader(tile, mapView.taskQueue);
         tile.tileGeometryLoader = geometryLoader;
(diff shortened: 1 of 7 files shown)


@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
```diff
@@ -343,24 +340,11 @@ describe("TileGeometryLoader", function() {
             expect(mapView.taskQueue.numItemsLeft(TileTaskGroups.CREATE)).equal(0);
 
             expect(mapView.taskQueue.processNext(TileTaskGroups.CREATE)).equal(false);
-
-            await willEventually(() => {
```
nzjony, line 347: I had to remove this, because this condition is never reached, i.e. the `processTechniques` function is never called because the task is expired

FraukeF: The test should still have passed with `expect(spyProcessTechniques.callCount).equal(0); `

But anyway its probably not necessary to test this as its supposedly coverede with the empty TaskQueue test.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -278,7 +278,9 @@ export class TileGeometryLoader {
             execute: this.prepare.bind(this, enabledKinds, disabledKinds),
             group: TileTaskGroups.CREATE,
             getPriority: this.getPriority.bind(this),
-            isExpired: this.discardNeedlessTile.bind(this, this.tile),
+            isExpired: () => {
+                return !this.tile.isVisible || this.tile.dataSource.isDetached();
```
FraukeF, line 282: Have you checked if this maybe throws an error, because if I remember correctly a while ago there was an issue with Tile.isVisible accessing tile.datasource.mapView, so I think you might have to check for this.tile.dataSource.isDetached(); before this.tile.isVisible()

@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
nzjony: @FraukeF , good point, yes I thought that because they weren't finishing and that it was tested elsewhere, I could safely ignore it.

Looking at it, the only condition that isn't met is the `isFinished` condition is never true, I will check why.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
nzjony: Good spotting! I will fix.

I'm surprised my test where I detach the mapview doesn't pick this up however, because the `update` is called afterwards, and I would expect that is when the `isExpiring` method is called... will invistigate.

@here/harp-utils/lib/TaskQueue.ts
```diff
@@ -215,12 +215,13 @@ export class TaskQueue {
     }
 
     private updateTaskList(taskList: Task[]) {
-        taskList.forEach(task => {
+        for (let i = 0; i < taskList.length; i++) {
+            const task = taskList[i];
             if (task?.isExpired?.()) {
                 taskList.splice(taskList.indexOf(task), 1);
+                i--;
```
atomicsulfate, line 222: substitute taskList.indexOf(task) by i.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -187,13 +182,13 @@ export class TileGeometryLoader {
      * Start with or continue with loading geometry. Called repeatedly until `isFinished` is `true`.
      */
     update(
-        enabledKinds: GeometryKindSet | undefined,
-        disabledKinds: GeometryKindSet | undefined
+        enabledKinds?: GeometryKindSet | undefined,
+        disabledKinds?: GeometryKindSet | undefined
```
atomicsulfate, line 186: i guess now the union type with undefined is not needed anymore?

@here/harp-utils/lib/TaskQueue.ts
FraukeF: done

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
FraukeF: done

atomicsulfate (approved):

PR #1767, opened 2020-08-13
MINOR: tileWrappingEnabled fixes

nzjony:
- Added test
- Log warning when set and projection is spherical

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -1262,6 +1262,10 @@ export class MapView extends EventDispatcher {
     }
 
     set tileWrappingEnabled(enabled: boolean) {
+        if (this.projection.type === ProjectionType.Spherical) {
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -1262,6 +1262,10 @@ export class MapView extends EventDispatcher {
     }
 
     set tileWrappingEnabled(enabled: boolean) {
+        if (this.projection.type === ProjectionType.Spherical) {
+            logger.warn("Setting this with spherical projection has no affect. Was this intended?");
```
FraukeF, line 1266: The more I stumble over the tileWrapping, the more I think it should actually be part of the projection and not of MapView.
This probably goes beyond this fix, but we should at least consider moving it to the projection and see if this would not resolve a lot of this additional and somehow weird and repetitive checks

PR #1770, opened 2020-08-14
MINOR: Move example to public

nzjony:


@here/harp-examples/src/labelling_user-markers.ts
```diff
@@ -0,0 +1,226 @@
...
+import { apikey } from "../config";
+
+const USER_MARKER_DATASOURCE = "user-markers-datasource";
+
+/**
+ * Basic example showing how to create a custom data source providing user text elements
+ * that change on arbitrary events.
+ */
+export namespace LabellingUserMarkers {
+    class UserMarkerDataSource extends DataSource {
```
ninok, line 34: I think we should have a marker data source that can be reused and that maybe also adds some tiling for performance reasons.
@germanz also implemented a `MarkerDataSource` for JSAPI integration and it would be cool if they could be unified and made publicly available to harp users.

nzjony:
@ninok  added an example showing how to add user markers, so closing this PR

PR #1793, opened 2020-08-21
Add visit methods to classify interpolations in when conditions.

robertoraggi:
This change removes the exceptions raised by `StyleConditionClassifier`
when `step` and `interpolate` expressions are found during the
traversal. These exceptions are not needed and In fact,
it is perfectly acceptable to use `step` or `interpolate` expressions
with a feature attributes as their input in `when` conditions.

### DIFF
--- a/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
+++ b/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
@@ -625,4 +625,76 @@ describe("StyleSetEvaluator", function() {
         assert.deepStrictEqual(decodedTechnique.offset2, ["make-vector", 10, 20, 30]);
         assert.deepStrictEqual(decodedTechnique.offset3, ["make-vector", 10, 20, 30, 40]);
     });
+
+    describe("allow steps in filter conditions", () => {
+        const sse = new StyleSetEvaluator([
+            {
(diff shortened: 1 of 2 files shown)


@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
```diff
@@ -625,4 +625,76 @@ describe("StyleSetEvaluator", function() {
...
+                ],
+                technique: "solid-line"
+            },
+            {
+                id: "legacy-style-rule",
+                when: [
+                    "all",
+                    ["==", ["get", "$geometryType"], "line"],
+                    ["==", ["get", "$layer"], "buildings"],
+                    ["step", ["zoom"], false, 15, true]
```
ninok, line 647: This only works for integer zoomlevels right?!
Maybe we can point that out in the documentation.

ninok (approved):

@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
robertoraggi: yes, currently it works only with discrete zoom levels. We can make it work also work in general if we re-evaluate the expression in MapView before adding the objects to the scene.

PR #1802, opened 2020-08-27
HARP-11591: Add example showcasing geoToScreen and fixes tileWrapping…

FraukeF:
…Enabled setter

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-examples/src/markers_screen-anchor.ts
+++ b/@here/harp-examples/src/markers_screen-anchor.ts
@@ -1,0 +1,299 @@
+/*
+ * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { GeoCoordinates, mercatorProjection, sphereProjection } from "@here/harp-geoutils";
+import { MapControls, MapControlsUI } from "@here/harp-map-controls";
+import { CopyrightElementHandler, MapAnchor, MapView, MapViewEventNames } from "@here/harp-mapview";
+import { VectorTileDataSource } from "@here/harp-vectortile-datasource";
+import { GUI } from "dat.gui";
+import THREE = require("three");
+
+import { apikey } from "../config";
+
+export namespace GeoToScreenExample {
+    const BERLIN = new GeoCoordinates(52.5186234, 13.373993);
+    const geoPosition = {
+        lat: BERLIN.lat,
+        lng: BERLIN.lng,
+        alt: BERLIN.altitude ?? 0
+    };
+
+    // Create a new MapView for the HTMLCanvasElement of the given id.
+    function initializeMapView(id: string): MapView {
+        const canvas = document.getElementById(id) as HTMLCanvasElement;
+
+        // Look at BERLIN
+        const map = new MapView({
+            canvas,
+            projection: mercatorProjection,
+            tileWrappingEnabled: false,
+            theme: "resources/berlin_tilezen_night_reduced.json",
+            target: BERLIN,
+            zoomLevel: 8,
+            tilt: 45,
+            heading: -80
+        });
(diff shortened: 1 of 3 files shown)


@here/harp-examples/src/geo-to-screen.ts
```diff
@@ -0,0 +1,286 @@
+/*
```
ninok, line 1: I think we need to discuss what examples we want to have and how we want to structure them.
Also the examples should tell a story à la scrum: `As a user I want to add a marker in screen space at this geo location` 

At least we should have one section in the examples about the whole markers/map-objects topic.

@here/harp-examples/src/geo-to-screen.ts
```diff
@@ -0,0 +1,286 @@
...
+                "x: " + Math.round(screenpos.x) + " y: " + Math.round(screenpos.y);
+        } else {
+            screenAnchor.style.opacity = "0.1";
+            screenAnchorValues.innerHTML = "x: undefined , y: undefined";
+        }
+        if (mapAnchor !== undefined && mapAnchor.geoPosition !== undefined) {
+            const scaleFactor = mapView.targetDistance / 10000;
+            if (mapAnchor.scale.x !== scaleFactor) {
+                mapAnchor.scale.set(scaleFactor, scaleFactor, scaleFactor);
+                mapView.requestUpdateIfNeeded();
```
ninok, line 117: `requestUpdateIfNeeded` is deprecated. Just use `update`
It seems we lost the deprecation warnings when we switching from tslint to eslint. Can you create a ticket for that please.

@here/harp-examples/src/geo-to-screen.ts
```diff
@@ -0,0 +1,286 @@
...
+            }
+            if (
+                mapAnchor.geoPosition.latitude !== geoPosition.lat ||
```
ninok, line 120: I would skip this check to make the code easier to read.

@here/harp-examples/src/geo-to-screen.ts
```diff
@@ -0,0 +1,286 @@
...
+            case "ArrowRight":
+            case "ArrowUp":
+            case "ArrowDown":
+                stopAnchorUpdate();
+                break;
+            case "j":
+                mapView.lookAt({
+                    target: new GeoCoordinates(
+                        mapView.target.latitude,
+                        (mapView.target.longitude -= 360)
```
ninok, line 242: This is somehow not reflected in the GUI. I always see normalized longitude there

@here/harp-examples/src/geo-to-screen.ts
FraukeF: yes, definitely, we need some structure, I like writing these kind of little test apps to visually verify the changes, but ofter there is not yet a good place to put them, and I think, in the internal example repository things bitrot

@here/harp-examples/src/geo-to-screen.ts
FraukeF: done

@here/harp-examples/src/geo-to-screen.ts
FraukeF: The thing is, what you see in the GUI is the geoPosition that is used for the getScreenPosition call, ... this moves the mapView target

@here/harp-examples/src/geo-to-screen.ts
FraukeF: It is needed, as otherwise the update will run into a recursion with mapView.update => updateAnchors => mapView.update =>.....

@here/harp-examples/src/geo-to-screen.ts
ninok: ah, missed that. We don't move the object into the other world but we move all the worlds instead :smile:

ninok (approved):

PR #1803, opened 2020-08-27
HARP-11699: Adding minimum geometry height to tiles

StefanDachwitz:
* adds StaticElevationRangeSource for simple use cases
* demonstrates the use case in example datasource_custom

# CI Check

Happy contributing!


@here/harp-mapview/lib/ElevationRangeSource.ts
```diff
@@ -58,3 +59,81 @@ export interface ElevationRangeSource {
...
 }
+
+/**
+ * An [[ElevationRangeSource]] that delivers static values for the [[ElevationRange]], which are
+ * either setup from a [[DataSource]] or set in the constructor. Useful in situations where the
+ * geometry of the data is not (only) at ground level, where it may be clipped in certain
+ * situations.
+ */
+export class StaticElevationRangeSource implements ElevationRangeSource {
+    readonly m_tilingScheme: TilingScheme;
```
ninok, line 70: `private` with getters and setter or remove the `m_` if they really should be public.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -667,6 +667,12 @@ export interface MapViewOptions extends TextElementsRendererOptions, Partial<Loo
      * If set, the view will constrained within the given bounds in geo coordinates.
      */
     maxBounds?: GeoBox;
+
+    /**
+     * Optional [[ElevationProvider]] where [[DataSource]]s create geometry above and/or below
```
ninok, line 672: Optional `ElevationRangeSource`.
Initially the `ElevationRangeSource` was intended to only be used in combination with terrain (i.e. `ElevationProvider`). If the datasource has data that is not flat it would provide it's range in min/maxGeometry height (constant) or per tile.
When terrain is enabled and a ElevationRangeSource is set, the final tile bounding box would be the accumulation of terrain min/max height + datasource min/ max height.

As far as I understood the issue, it is  only about the min/max elevation range in the datasource.

@here/harp-vectortile-datasource/lib/VectorTileDataSource.ts
```diff
@@ -146,15 +146,26 @@ export interface VectorTileDataSourceParameters extends DataSourceOptions {
...
+    /**
+     * Minimum geometry height below ground level this `OmvDataSource` can produce. Negative values
+     * describe height below ground.
+     *
+     * @remarks
+     * Used in first stage of frustum culling before [[Tile.minGeometryHeight]] data is available.
+     *
+     * @default `0`.
+     */
+    minGeometryHeight?: number;
```
ninok, line 167: This should not be here but in `DataSourceOptions`. Same for `maxGeometryHeight`

ninok (changes requested):
I'm missing the actual change to the clipping plane computation (i.e. the issue that you want to fix).
See:
https://github.com/heremaps/harp.gl/blob/148feb2f0d7cd21d4693d1d9fa27b0a8a5c39a62/@here/harp-mapview/lib/MapView.ts#L3318

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: I disagree: The ElevationRangeSource is required for visibility testing during loading. Without it, probable tiles will not be loaded. The min/max elevation range is used for bounding box calculation during rendering. The ElevationRangeSource is required for this use case.

@here/harp-vectortile-datasource/lib/VectorTileDataSource.ts
StefanDachwitz: Done.

@here/harp-mapview/lib/ElevationRangeSource.ts
StefanDachwitz: Made them private.

@here/harp-mapview/lib/MapView.ts
ninok: Both, min/max height and elevation range source, should be used during loading.
The difference is that min/max height is global and the elevation range source is on tile level.

Additionally as mentioned was ElevationRangeSource intended for terrain.
Imagine putting vector data on top of terrain. To get propper bounding boxes we would need two elevation range sources, one for terrain and one for the vector data.
To avoid this complexity of accumulating multiple ElevationRangeSources the idea was that there is only one ElevationRangeSource for terrain and the min/max height for the vector data. The final bounding box of a tile would be the elevation range + min/max height.

StefanDachwitz:
The thing that caused most of the trouble with the IBCT tests was caused by setDepthPrePassStencil(), which failed to enable stencil tests, which led to z-fighting. Since I could not get the MSAA in MapRenderingManager to work, I enabled native AA in these tests, which causes some invisible changes at borders and also in most areas, due to rounding differences.

@here/harp-mapview/lib/DepthPrePass.ts
```diff
@@ -193,29 +193,45 @@ export function createDepthPrePassMesh(mesh: THREE.Mesh): THREE.Mesh {
...
+        colorMaterial.stencilFail = THREE.KeepStencilOp;
+        colorMaterial.stencilZFail = THREE.KeepStencilOp;
+        colorMaterial.stencilZPass = THREE.ZeroStencilOp;
+        colorMaterial.stencilFunc = THREE.EqualStencilFunc;
+        colorMaterial.stencilRef = 0xff;
+        (colorMaterial as any).stencilFuncMask = DEPTH_PRE_PASS_STENCIL_MASK;
+    }
+
+    if (depthMesh.material instanceof Array) {
+        depthMesh.material.map(setupDepthMaterialStencil);
```
ninok, line 227: foreach instead of map?

ninok (dismissed):

ninok (approved):

PR #1807, opened 2020-08-31
HARP-11780: Make events non-global to allow garbage collection of MapView and DataSource

### LINKED ISSUE #1798: MapView is not fully garbage collected after dispose()
**Describe the bug**

When attempting to clean up all the memory of the MapView, the MapView is still retained by several global data structures inside the Harp.gl library.

The "Events" data structures are global constants that retain the MapView because Three.js applies the MapView object to  the `target` property of the event when it is dispatched. This target property is not unset. This causes the MapView to linger in memory most likely until the next MapView is constructed that replaces the lingering references.

I would recommend these Events be placed inside the MapView/DataSource as private members to solve this problem.

**To Reproduce**

Build a MapView, attach it to the DOM, then detach and `dispose()`

**Expected behavior**

A look at Chrome memory management tool should show the MapView is no longer retained in memory.

ninok:
This fixes #1798

### DIFF
--- a/@here/harp-mapview/lib/DataSource.ts
+++ b/@here/harp-mapview/lib/DataSource.ts
@@ -13,7 +13,6 @@ import { MapView } from "./MapView";
 import { Tile } from "./Tile";
 
 const logger = LoggerManager.instance.create("DataSource");
-const UPDATE_EVENT = { type: "update" };
 
 /**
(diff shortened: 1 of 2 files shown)


FraukeF (approved):

J-Rojas:
Perfect! Thanks

PR #1811, opened 2020-09-01
HARP-11811: Added offset x/y for tilt and heading gestures

ninok:


ninok:

ninok:

@here/harp-mapview/lib/Utils.ts
```diff
@@ -223,40 +225,62 @@ export namespace MapViewUtils {
...
+            MathUtils.clamp(deltaTilt + currentTilt, 0, maxTiltAngle) - currentTilt;
+        if (Math.abs(clampedDeltaTilt) <= Number.EPSILON) {
+            return;
+        }
+        const tiltAxis = cache.vector3[0].set(1, 0, 0).applyQuaternion(mapView.camera.quaternion);
+        const tiltQuat = cache.quaternions[0].setFromAxisAngle(tiltAxis, clampedDeltaTilt);
+        mapView.camera.quaternion.premultiply(tiltQuat);
+        mapView.camera.position.sub(rotationTargetWorld);
+        mapView.camera.position.applyQuaternion(tiltQuat);
+        mapView.camera.position.add(rotationTargetWorld);
```
atomicsulfate, line 283: This can put the target out of the world. After that the orbit controls won't work anymore since we early exit above. It can be reproduced in your build both in sphere and mercator by increasing the maximum tilt angle and then tilting to max around and arbitrary screen point.

ninok: The problem is, that the `MapControls.maxTiltAngle` is handled inconsistently.
Previously the maximum tilt was clamped to 90 degrees but internally it was never possible to go beyond 89 degrees. I will check why the clamping is lost and will clamp consistently to 89 degrees.

ninok: found the culprit: `this.m_maxTiltAngle = Math.max(0, Math.min(90, THREE.MathUtils.degToRad(angle)));` So the angle was clamped to 90 degrees after converting to radians.
It works now, but still for globe we can get into a situation where the target is not existing anymore. I will investigate why this is happening.

atomicsulfate (approved):

PR #1818, opened 2020-09-02
HARP-11468 Empty payloads are set with an empty decodedTile so that the background plane is still set

nzjony:
Empty payload from the data provider now sets the decodedTile to be empty, this ensures that the
ground plane is properly created which is needed because otherwise, if the DataSource
isFullyCovering, it will remove the BackgroundDataSource's tiles and there will be holes in the
map.

### DIFF
--- a/@here/harp-mapview-decoder/lib/TileLoader.ts
+++ b/@here/harp-mapview-decoder/lib/TileLoader.ts
@@ -217,7 +217,10 @@ export class TileLoader implements ITileLoader {
             (payload.constructor === Object && Object.keys(payload).length === 0)
         ) {
             // Object is empty
-            this.onDone(TileLoaderState.Ready);
+            this.onDecoded({
+                geometries: [],
+                techniques: []
+            });
             return;
         }
 
--- a/@here/harp-mapview-decoder/test/TileLoaderTest.ts
+++ b/@here/harp-mapview-decoder/test/TileLoaderTest.ts
@@ -163,6 +163,8 @@ describe("TileLoader", function() {
                 getTileStub.resolves({});
                 loadPromise = tileLoader.loadAndDecode();
                 expect(loadPromise).to.not.be.undefined;
+                expect(tileLoader.decodedTile!.geometries.length).eq(0);
+                expect(tileLoader.decodedTile?.techniques.length).eq(0);
 
                 return expect(loadPromise).to.eventually.be.fulfilled.then(() => {
                     expect(decodeTileSpy.notCalled).to.be.true;


@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -323,18 +324,15 @@ export class TileGeometryCreator {
...
-        }
+        // The ground plane is required for when we change the zoom back and we fall back to the
+        // parent, in that case we reduce the renderOrder of the parent tile and this ground
+        // place ensures that parent doesn't come through. This value must be above the
+        // renderOrder of all objects in the fallback tile, otherwise there won't be a proper
+        // covering of the parent tile by the children, hence dividing by 2. To put a bit more
+        // concretely, we assume all objects are rendered with a renderOrder between 0 and
+        // FALLBACK_RENDER_ORDER_OFFSET / 2, i.e. 10000. The ground plane is put at -10000, and
+        // the fallback tiles have their renderOrder set between -20000 and -10000
+        this.addGroundPlane(tile, TileGeometryCreator.instance.fallBackRenderOffset, true);
```
ninok, line 335: Will this work for globe? I guess we would need to subdivide the ground plane the same way we subdivide the actual geometry.

ninok: Never mind, just saw that you moved the check inside the function

nzjony: I moved the check inside the `addGroundPlane` method, because otherwise I would need the check duplicated.

nzjony: Because I need to call the `addGroundPlane` twice, once when it is empty, and then once in the normal condition.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1356,8 +1354,15 @@ export class TileGeometryCreator {
      * Creates and add a background plane for the tile.
      * @param tile - Tile
      * @param renderOrder - Render order of the tile
+     * @param checkIfValid - Whether to add a check to see if this is necessary to add.
```
ninok, line 1357: Why would one add an invalid background plane?

nzjony: Good point, I didn't want to have the check for the BackgroundDataSource, but it is anyway correctly configured because it sets `addGroundPlane`, so it will pass this check. Will remove

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -210,13 +204,17 @@ export class TileGeometryLoader {
         if (
             (this.m_state === TileGeometryLoaderState.Initialized ||
                 this.m_state === TileGeometryLoaderState.Canceled) &&
-            tile.decodedTile !== undefined
+            tile.tileLoader?.isFinished === true
         ) {
-            if (this.m_state === TileGeometryLoaderState.Initialized) {
+            if (
+                this.m_state === TileGeometryLoaderState.Initialized &&
+                tile.decodedTile !== undefined
```
ninok, line 211: I wonder if it wouldn't make things easier if we instead of not producing a tile could just create a decoded tile that has no geometry.

ninok (dismissed):

@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
```diff
@@ -411,5 +396,26 @@ describe("TileGeometryLoader", function() {
...
+            const spyAddGroundPlane = sandbox.spy(geometryCreator, "addGroundPlane") as any;
+            expect(spyAddGroundPlane.callCount).equal(0);
+
+            geometryLoader!.update();
+            expect(mapView.taskQueue.numItemsLeft(TileTaskGroups.CREATE)).equal(1);
+            expect(mapView.taskQueue.processNext(TileTaskGroups.CREATE)).equal(true);
+
+            await willEventually(() => {
+                expect(spyAddGroundPlane.callCount).equal(1);
+                // The geometry loader doesn't finish, because we expect the task to expire
```
FraukeF, line 416: the comment is not matching the check in the next line .. "doesn#t finish.." ..isFinished).to.be.true

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
FraukeF: I think, as long as we wouldnt return a decoded Tile containing the background tile geometry, it does not make much of a difference

FraukeF (approved):

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -1358,6 +1356,9 @@ export class TileGeometryCreator {
      * @param renderOrder - Render order of the tile
      */
     addGroundPlane(tile: Tile, renderOrder: number) {
+        if (!tile.dataSource.isFullyCovering || tile.projection.type !== ProjectionType.Planar) {
```
FraukeF, line 1359: Actually looking at this a second time, I think, moving this check inside the method is error prone, it covers up information, that will lead to misusage and a lot of debugging :) .. it should be self contained and the code around decide to call it or not

FraukeF (changes requested):

@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
nzjony: Hahah, thanks.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: I guess it is a compromise, by moving the logic outside of this function, you may also introduce misusage, because the function may be called with different checks... and someone who wants to read the code would need to understand why there are different checks. For example, if this was fixed to work with sphere projection, then the person would need to find all calls to this function and update the external checks to remove the check for the Planar type (which they may miss) Easier if the ground plane logic is in one location and not spread in different places IMO.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
nzjony: I will take another look, the decodedTile that comes through is indeed not undefined, and has no geometry, not sure how I missed this, anyway, I think I don't need to change this condition.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: @FraukeF , I changed the logic, so this change is no longer necessary, I hope to push the new changes soon.

FraukeF (approved):

nzjony:

nzjony:

PR #1821, opened 2020-09-07
MINOR: Adds renderOrder support to WebTileDataSource

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-webtile-datasource/lib/WebTileDataSource.ts
+++ b/@here/harp-webtile-datasource/lib/WebTileDataSource.ts
@@ -6,6 +6,7 @@
 import { TileKey, TilingScheme, webMercatorTilingScheme } from "@here/harp-geoutils";
 import { CopyrightInfo, DataSource, DataSourceOptions, Tile } from "@here/harp-mapview";
 import { TileGeometryCreator } from "@here/harp-mapview/lib/geometry/TileGeometryCreator";
+import { enableBlending } from "@here/harp-materials";
 import { getOptionValue, LoggerManager } from "@here/harp-utils";
 import THREE = require("three");
 
@@ -17,15 +18,21 @@ const logger = LoggerManager.instance.create("MapView");
 export interface WebTileRenderingOptions {
     /**
      * Opacity of the rendered images.
-     * @default 1.0
+     * @defaultValue 1.0
      */
     opacity?: number;
 
     /**
      * Force Material to use transparency from texture if available
-     * @default false
+     * @defaultValue false
      */
     transparent?: boolean;
+
+    /**
+     * RenderOrder for order in which to render WebTileDataSouurces
+     * @defaultValue 0
+     */
+    renderOrder?: number;
 }
 
 export interface WebTileDataProvider {
@@ -84,8 +91,6 @@ export class WebTileDataSource extends DataSource {
 
         this.dataProvider = this.m_options.dataProvider;
         this.cacheable = true;
-        this.enablePicking = false;
-        this.storageLevelOffset = -1;
         this.m_resolution = getOptionValue(
             m_options.resolution,
             WebTileDataSource.resolutionValue.resolution512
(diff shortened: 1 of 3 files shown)


@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -121,23 +127,34 @@ export class WebTileDataSource extends DataSource {
...
-                    this.m_options.renderingOptions.opacity !== undefined
-                        ? this.m_options.renderingOptions.opacity
-                        : 1;
+
+                let transparent = false;
+                let opacity = 1;
+                let renderOrder = 0;
+                if (this.m_options.renderingOptions !== undefined) {
+                    opacity =
+                        this.m_options.renderingOptions.opacity !== undefined
```
nzjony, line 136: use the ?? operator

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -121,23 +127,34 @@ export class WebTileDataSource extends DataSource {
...
+                            ? this.m_options.renderingOptions.opacity
+                            : 1;
+                    transparent =
+                        this.m_options.renderingOptions.transparent === true ||
+                        (opacity !== undefined && opacity < 1);
+                    renderOrder =
+                        this.m_options.renderingOptions.renderOrder !== undefined
```
nzjony, line 143: use the ?? operator

FraukeF:

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
FraukeF: done

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
FraukeF: done

nzjony (dismissed):

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -121,23 +127,28 @@ export class WebTileDataSource extends DataSource {
...
+                }
                 const material = new THREE.MeshBasicMaterial({
                     map: texture,
-                    depthTest: false,
-                    depthWrite: false,
                     opacity,
-                    transparent: transparent || (opacity !== undefined && opacity < 1.0)
+                    transparent,
+                    polygonOffset: renderOrder > 0 ? true : false,
+                    polygonOffsetFactor: -1,
```
ninok, line 146: Why not keeping the `depthTest: false`. Especially when combining satellite + omv this could cause issues.

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -26,6 +26,12 @@ export interface WebTileRenderingOptions {
      * @default false
      */
     transparent?: boolean;
+
+    /**
+     * RenderOrder for order in which to render WebTileDataSouurces
+     * @default 0
+     */
+    renderOrder?: number;
```
ninok, line 34: Just wondering if we go in the right direction here. We add more and more styling options to the API but maybe we should just make this datasource styleable (we also have a ticket to make landmarks styleable).

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -26,6 +26,12 @@ export interface WebTileRenderingOptions {
      * @default false
      */
     transparent?: boolean;
+
+    /**
+     * RenderOrder for order in which to render WebTileDataSouurces
+     * @default 0
```
ninok, line 32: We are using api-extractor(i.e. tsdoc) tags, so tit should be `@defaultValue`
https://api-extractor.com/pages/tsdoc/tag_defaultvalue/

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
FraukeF: ok, done

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
FraukeF: done

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
FraukeF: Very good point, yes, I think it would make sense to actually make it stylable. As this will need some more evaluation and gathering of requirements etc, I ll create a ticket for it, to not block the current work.

@here/harp-webtile-datasource/lib/WebTileDataSource.ts
ninok: I actually meant to use `depthTest: false` instead of `polgonOffset` to not pollute the depth buffer with "wrong" values

FraukeF: Yes, agreed, but the thing is, that with pure renderOrder we cannot interleave correctly transparent and non-transparent layers, as they are in different render groups. It´s probably a not so common case to put a opaque layer on top of a transparent one, but still it would not handle it as expected. 
I could somehow make it optional to use either or, but this would again blow up the API..

ninok: We should keep all objects in the opaque queue as we also do for the OMV layers.
We have a utility method to enable blending even if `transparent=false`:
https://github.com/heremaps/harp.gl/blob/master/@here/harp-materials/lib/Utils.ts#L81

FraukeF: ok, nice. I ll adjust to that then

ninok (approved):

PR #1831, opened 2020-09-09
MINOR: Adds opacity accessors to WebTileDataSource

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-webtile-datasource/lib/WebTileDataSource.ts
+++ b/@here/harp-webtile-datasource/lib/WebTileDataSource.ts
@@ -86,6 +86,10 @@ export class WebTileDataSource extends DataSource {
     protected readonly m_resolution: WebTileDataSource.resolutionValue;
     protected dataProvider: WebTileDataProvider;
 
+    private m_opacity: number = 1;
+    private readonly m_renderOrder: number = 0;
+    private m_transparent: boolean = false;
+
     /**
      * Constructs a new `WebTileDataSource`.
      *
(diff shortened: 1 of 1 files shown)


@here/harp-webtile-datasource/lib/WebTileDataSource.ts
```diff
@@ -96,12 +100,38 @@ export class WebTileDataSource extends DataSource {
...
+     * Sets the opacity for the WebTileDataSource, will only affect not yet loaded or not cached
+     * tiles.
+     *
+     * Use WebTileDataSource:clearCache and MapView:markTilesDirty to reload all tiles with the
+     * new opacity setting.
+     */
+    set opacity(value: number) {
+        this.m_opacity = value;
+        if (this.m_opacity < 1) {
+            this.m_transparent = true;
```
ninok, line 124: In the construtor m_transparent is set to false if `opacity === 1` I guess you want to do the same here.

FraukeF: No, I dont want to override it, in case it was set in the renderOptions, i.e. to force the material to handle transparencies in the texture

ninok: but you override it. If transparent is false in the options and you set opacity to 0.9 and later back to 1.0 it will stay true even though it is false in the options.

FraukeF: yes, thats true, but it will visually not really make a difference. Thats why I kept it on the same variable. 
But still, you are right, I ll add a condition that solves both cases

ninok (approved):

FraukeF:

PR #1835, opened 2020-09-09
HARP-11531: Sphere horizon computation.

atomicsulfate:
SphereHorizon computes the horizon for sphere projection, which is
a circle. It provides functions to get any point on it or subdivide
an arc into any number of divisions.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/SphereHorizon.ts
+++ b/@here/harp-mapview/lib/SphereHorizon.ts
@@ -1,0 +1,382 @@
+/*
+ * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { EarthConstants, MathUtils, sphereProjection } from "@here/harp-geoutils";
+import { Math2D } from "@here/harp-utils";
+import * as THREE from "three";
+
+import { MapViewUtils } from "./Utils";
+
+const twoPi = Math.PI * 2;
+
+// Keep counter clockwise order.
+export enum CanvasSide {
+    Bottom,
+    Right,
+    Top,
+    Left
+}
+
+/**
+ * Class computing horizon tangent points and intersections with canvas for spherical projection.
+ *
+ * @remarks
+ *
+ * The horizon for a sphere is a circle formed by all intersections of tangent lines passing through
+ * the camera with said sphere. It lies on a plane perpendicular to the sphere normal at the camera
+ * and it's center is at the line segment joining the sphere center and the camera.
+ *
+ * The further the camera is, the nearer the horizon center gets to the sphere center, only reaching
+ * the sphere center when the camera is at infinity. In other words, the horizon observed from a
+ * finite distance is always smaller than a great circle (a circle with the sphere radius, dividing
+ * the sphere in two hemispheres, and therefore it's radius is smaller than the sphere's.
+ *
+ * @internal
+ */
+export class SphereHorizon {
+    private readonly m_matrix: THREE.Matrix4;
+    private readonly m_radius: number;
+    private readonly m_normalToTangentAngle: number;
+    private readonly m_distanceToHorizonCenter: number;
+    private readonly m_intersections: number[][] = [];
+    private m_cameraPitch?: number;
+    private m_hFovVertical?: number;
+    private m_hFovHorizontal?: number;
+
+    /**
+     * Constructs the SphereHorizon for the given camera.
+     *
+     * @param m_camera - The camera used as a reference to compute the horizon.
+     */
+    constructor(private readonly m_camera: THREE.PerspectiveCamera) {
+        //
+        //          TL :,,,,,,,,,,,,,,,,,,,,,,,,,,,,! TR canvas corner (proj. on horizon plane)
+        //             >                            +
+        //             >                            +
+        //             ::                           +
+        //              +                           >
+        //              >         `::::::'         !"
+        //              >     `:::"      ':::,     +
+        //              +   :!,              .!!`  +
+        //              ,,:;`                   :+`>
+        //              >T                        T# <-- Sphere radius to tangent angle (90 deg).
+        //             !!>                       `+!(`
+        //            =' +                      `.+  {:
+        //           /.  +                     `` >   |:`
+        //          /`   ::                   .`  >    |:.
+        //         /`     >         camZ     '`  `+     /`'`
+        //        ;,      +           *,`   '    !,      ( `'
+        //       ,^       +              `.~     +       ./  .`
+        //       )        +               ' ```  >        !~  `.
+        //      *`        '~             '    ``.>         (    .` tangentDistance
+        //     `\          +           `'        ?``       '?    `'
+        //     /`          +          `.        :.  ```     )      `.
+        //    `\           >         `.         >     ```   .=       .`
+        //    |`           +        ``          +        ``` (        `.
+        //    /            "!      .`           +           `?!         '`
+        //   ,+             !,,,,,!:,,,,,,,,,,,,,             {.`        `.
+        //   /`            BL    '               BR           )  ```       .`
+        //   /                  '                             >`    ``      `'
+        //   /                 '                              `*      ```     .`
+        //  ,:                ' Sphere radius                  )     camZ```   `.
+        //  ?`              `.                                 )           ```   `.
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/SphereHorizon.ts
```diff
@@ -0,0 +1,333 @@
...
+ * Class computing horizon tangent points and intersections with canvas for spherical projection.
+ *
+ * @remarks
+ *
+ * The horizon for a sphere is a circle on a plane with a normal matching the sphere normal at the
+ * camera position, and its center is at the interesection of the camera position vector with that
+ * plane. This circle is formed by all intersection points on the sphere with all tangents passing
+ * through the current camera position.
+ *
+//                                            ::::::::::::::::::*`
```
ninok, line 33: Awesome ASCII art, but to be honest I cannot see anything. Could also be the Death Star :smile: 
Alternatively you could just add a svg file to the repository (e.g. in docs/@docs) and add the link to the raw file (that will only be available after merging)

@here/harp-mapview/lib/SphereHorizon.ts
```diff
@@ -0,0 +1,333 @@
...
+
+    /**
+     * Constructs the SphereHorizon for the given camera and projection.
+     *
+     * @param m_camera - The camera used as a reference to compute the horizon.
+     * @param m_projection - A spherical projection.
+     */
+    constructor(
+        private readonly m_camera: THREE.PerspectiveCamera,
+        private readonly m_projection: Projection
```
FraukeF, line 100: Will this work for any other projection, then the sphereProjection? If not, there is no need, to pass the projection, but just use sphereProjection

@here/harp-mapview/lib/SphereHorizon.ts
```diff
@@ -0,0 +1,333 @@
...
+    Top,
+    Left
+}
+
+/**
+ * Class computing horizon tangent points and intersections with canvas for spherical projection.
+ *
+ * @remarks
+ *
+ * The horizon for a sphere is a circle on a plane with a normal matching the sphere normal at the
```
FraukeF, line 28: It might be me, but I have read this now a couple of times, and it still confuses me.

So we have a circle on a plane as the horizon, I get that..
Now with the normal and center point I am not sure I understand: 

The normal of the plane is then a vector from the sphere center to the camera position??
The Plane always contains the sphereCenter, right? 
So the Circle Center Point would be the same as the Sphere Center Point?
The circle then would be the intersection of the plane with the sphere?

Maybe be a bit more verbose, and I have to agree with Nino, maybe it´s due to the formatting in the github diff, but the ascii does not help.

@here/harp-mapview/lib/SphereHorizon.ts
ninok: I already created some svg when working on the rotation around arbitrary screen point:
https://raw.githubusercontent.com/heremaps/harp.gl/f1e7f4c55e7c472048baaf97639ab8fe185be661/docs/%40docs/globe-math/globe-tilt.svg
I was using Inkscape to create it. If you want I can also share the source file.

atomicsulfate: thanks for the review, but you jumped in slightly too soon and I didn't have the documentation completed. I was toying with an app that converts diagrams to ascii art, hence that uninformative ascii blob. I'm tweaking it right now and adding text so that's useful.

@here/harp-mapview/lib/SphereHorizon.ts
atomicsulfate: In theory it works for any Sphere projection (that is any instance of Sphere projection with any radius, not necessarily the earth radius). In practice, it's true that we only use the sphere projection "singleton" with the earth radius.

@here/harp-mapview/lib/SphereHorizon.ts
atomicsulfate: sorry, I was still working on the documentation, I'll upload more informative diagrams today. 

But answering your question: the horizon circle does not have generally the center on the sphere center. The horizon center is always at the line joining the earth center and the camera. The further away the camera is, the closer  the horizon center to the earth center becomes. When the camera is at infinity, the horizon center is at the earth center and therefore it becomes a great circle.

@here/harp-mapview/lib/SphereHorizon.ts
atomicsulfate: actually i lied, since I'm using the earth radius constant everywhere.

atomicsulfate: I've removed the projection.

@here/harp-mapview/lib/SphereHorizon.ts
atomicsulfate: I've rewritten the remarks and added several diagrams.

FraukeF: thanks, for the clarification. Make sense to me now.

@here/harp-mapview/lib/SphereHorizon.ts
```diff
@@ -0,0 +1,382 @@
...
+
+    /**
+     * Subdivides and arc of the horizon circle, providing the world coordinates of the divisions.
+     *
+     * @param callback - Function called for every division point, getting the the point world
+     * coordinates as parameter.
+     * @param tStart - Angular parameter of the arc's start point [0,1].
+     * @param tEnd - Angular parameter of the arc's end point [0,1].
+     */
+    getDivisionPoints(callback: (point: THREE.Vector3) => void, tStart = 0, tEnd = 1) {
```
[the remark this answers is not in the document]
atomicsulfate: The callback avoids the unnecessary memory allocations of the array. BoundsGenerator is just getting every point, unprojecting it to geo space and pushing it into another array, so the array created here would be useless.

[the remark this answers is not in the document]
FraukeF: ah ok, thanks for the clarification

@here/harp-mapview/lib/SphereHorizon.ts
```diff
@@ -0,0 +1,382 @@
...
+        // 10 divisions for a whole horizon circle.
```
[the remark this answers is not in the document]
atomicsulfate: done

@here/harp-mapview/lib/SphereHorizon.ts
```diff
@@ -0,0 +1,382 @@
...
+        for (let d = 0; d < divisionCount; d++) {
+            callback(this.getPoint(d / divisionCount, tStart, tEnd));
+        }
+    }
+
+    /**
+     * Indicates whether the horizon circle is fully visible.
+     * @returns 'True' if horizon is fully visible, false otherwise.
+     */
+    isFullyVisible(): boolean {
```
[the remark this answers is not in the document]
atomicsulfate: done. Reimplemented as a boolean property set on computeIntersections()

atomicsulfate:

@here/harp-mapview/lib/SphereHorizon.ts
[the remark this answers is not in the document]
FraukeF: nice, although I think you never set it to "true" it I saw right.

FraukeF (approved):

@here/harp-mapview/test/SphereHorizonTest.ts
```diff
@@ -0,0 +1,211 @@
...
+        checkPointInHorizon(point);
+        const pointVec = point.clone().normalize();
+        const x = pointVec.dot(camXAxis);
+        const y = pointVec.dot(camYAxis);
+        const angle = Math.atan2(y, x);
+        return angle >= 0 ? angle / (2 * Math.PI) : 1 + angle / (2 * Math.PI);
+    }
+
+    before(function() {
+        focalLength = MapViewUtils.calculateFocalLengthByVerticalFov(
```
nzjony, line 67: MINOR: vFov and canvasHeight are both const, why not just calculate this once?

atomicsulfate:

@here/harp-mapview/test/SphereHorizonTest.ts
atomicsulfate: It is calculated once for the whole test suite. The "before" function is run once before any test in the suite is run.

nzjony: Sorry, I mistook for beforeEach :(

PR #1838, opened 2020-09-11
HARP-5550: Adds support for GeoBox to wrap around Antimeridian

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-geoutils/lib/coordinates/GeoBox.ts
+++ b/@here/harp-geoutils/lib/coordinates/GeoBox.ts
@@ -6,7 +6,7 @@
 import * as THREE from "three";
 
 import { GeoBoxExtentLike } from "./GeoBoxExtentLike";
-import { GeoCoordinates } from "./GeoCoordinates";
+import { GeoCoordinates, MAX_LONGITUDE } from "./GeoCoordinates";
 import { GeoCoordinatesLike } from "./GeoCoordinatesLike";
 
 /**
@@ -48,7 +48,11 @@ export class GeoBox implements GeoBoxExtentLike {
      * @param southWest - The south west position in geo coordinates.
      * @param northEast - The north east position in geo coordinates.
      */
-    constructor(readonly southWest: GeoCoordinates, readonly northEast: GeoCoordinates) {}
+    constructor(readonly southWest: GeoCoordinates, readonly northEast: GeoCoordinates) {
+        if (this.west > this.east) {
+            this.northEast.longitude += 360;
+        }
+    }
 
     /**
      * Returns the minimum altitude or `undefined`.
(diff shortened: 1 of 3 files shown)


@here/harp-geoutils/lib/coordinates/GeoCoordinatesLike.ts
```diff
@@ -29,3 +29,33 @@ export function isGeoCoordinatesLike(object: any): object is GeoCoordinatesLike
...
+        const currentLon = coords[index].longitude;
+        coords[index].longitude += 360;
+        const nextLon = coords[(index + 1) % coords.length].longitude;
+
+        if (currentLon < -90 && nextLon > 90) {
+            // new crossing in opposite direction, stop.
+            break;
+        }
+    }
+}
```
atomicsulfate, line 61: This function is limited to polygons with sides spanning less than 180 degrees longitude.
It'd be simpler and more robust to just compute the GeoBox for the polygon to see where the southEast and northWest points are, then wrap around points with longitudes smaller than eastmost.
I can do that change in my PR, since here as I said above you don't need this function for GeoBox.

@here/harp-geoutils/lib/coordinates/GeoBox.ts
```diff
@@ -47,8 +58,17 @@ export class GeoBox implements GeoBoxExtentLike {
      *
      * @param southWest - The south west position in geo coordinates.
      * @param northEast - The north east position in geo coordinates.
+     * @param wrapAround - If `true` wraps the coordinates around the antimeridian.
      */
-    constructor(readonly southWest: GeoCoordinates, readonly northEast: GeoCoordinates) {}
+    constructor(
+        readonly southWest: GeoCoordinates,
+        readonly northEast: GeoCoordinates,
+        wrapAround: boolean = false
```
nzjony, line 66: Do you need the user to specify this? Why not just infer it based on the coordinates, i.e. if the northEast's longitude is less than the southWest's longitude, then wrap, as Andres mentioned below.

@here/harp-geoutils/lib/coordinates/GeoBox.ts
```diff
@@ -47,8 +58,17 @@ export class GeoBox implements GeoBoxExtentLike {
...
+    ) {
+        if (wrapAround) {
+            wrapCoordinatesAround([this.southWest, this.northEast]);
+        }
+    }
```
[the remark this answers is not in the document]
FraukeF: ok, done

@here/harp-geoutils/lib/coordinates/GeoCoordinatesLike.ts
FraukeF: ok

@here/harp-geoutils/lib/coordinates/GeoBox.ts
FraukeF: its default behaviour now

@here/harp-geoutils/lib/coordinates/GeoBox.ts
```diff
@@ -254,10 +258,19 @@ export class GeoBox implements GeoBoxExtentLike {
 
         const { west, east } = this;
 
-        if (east > west) {
-            return point.longitude >= west && point.longitude < east;
+        let longitude = point.longitude;
```
nzjony, line 261: is this tested somewhere?

FraukeF: yes, I added a test for GeoBox.contains which implicitly tests this... see following test in GeoBoxTest: 
it("contains works for non normalized coordinates", function() {

nzjony (approved):

@here/harp-geoutils/lib/coordinates/GeoBox.ts
nzjony: Sorry, I missed that.

PR #1859, opened 2020-09-22
HARP-12111: Fix handling of negative tilt angle

ninok:


@here/harp-mapview/lib/Utils.ts
```diff
@@ -266,12 +266,16 @@ export namespace MapViewUtils {
         }
         const mapTargetNormal = projection.surfaceNormal(mapTargetWorld, new THREE.Vector3());
 
-        const dirMapTarget = camera.position
-            .clone()
-            .sub(mapTargetWorld)
-            .normalize();
+        camera.updateMatrixWorld(true);
+        const cm = camera.matrixWorld.elements;
+        const dir = cache.vector3[1].set(cm[8], cm[9], cm[10]);
```
nzjony, line 271: Consider `camera.getWorldDirection`, or `extractBasis` from the `matrixWorld` ?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -266,12 +266,16 @@ export namespace MapViewUtils {
...
+        const up = cache.vector3[2].set(cm[4], cm[5], cm[6]);
+
+        // Due to inaccuracies it can happen that the tilt angle gets less than 0.
+        // Using the dot product of the direction vector and the surface normal alone does not give us the sign.
+        const tiltSign = Math.sign(up.dot(mapTargetNormal));
 
-        const currentTilt = dirMapTarget.angleTo(mapTargetNormal);
+        const currentTilt = (tiltSign === 0 ? 1 : tiltSign) * dir.angleTo(mapTargetNormal);
```
nzjony, line 278: If tiltSign is -ve, then currentTilt can be -ve, is that desired? It sounds weird to have a -ve tilt.

@here/harp-mapview/lib/Utils.ts
ninok: I actually copied the code from camera.getWorldDirection to avoid an extra updateWorldMatrix call.
Maybe `camera.matrixWorld.setFromMatrixColumn` is more readable

@here/harp-mapview/lib/Utils.ts
ninok: Due to precision issue with sphere projection it could happen that the tilt get's negative but the clamping in line 282 will take care of it.

@here/harp-mapview/lib/Utils.ts
nzjony: Sounds good.

@here/harp-mapview/lib/Utils.ts
nzjony: Ok, cool. I think I was getting confused with the name, it isn't so much the current tilt, but the "requested" tilt, but whatever, it is fine for now.

nzjony (approved):

PR #1874, opened 2020-09-25
Support for WebGL2 and GLSL 3.0 shaders.

### LINKED ISSUE #1575: Supports for WebGL2?
Hi,is there any plan to support webgl2? I tried to pass WebGL2RenderingContext to MapViewOptions,but it's not works for all functionalities,especially for mesh with RawMaterial.
Thanks so much.

J-Rojas:
Commit to close #1575 

Determine WebGL2 support via canvas capabilities via MapView
Use new glslVersion setting (as of Three@0.120) in ShaderMaterials to support GLSL 3.0 features.

Signed-off-by: Jose Rojas <jrojas@redlinesolutions.co>

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -2146,6 +2146,13 @@ export class MapView extends EventDispatcher {
     }
 
     /**
+     * Returns the glsl version compatible with the MapView's renderer
+     */
+    get glslVersion(): number {
+        return this.m_renderer.capabilities.isWebGL2 ? 3.0 : 1.0;
+    }
+
+    /**
      * Set's the way in which the fov is calculated on the map view.
      *
      * @remarks
--- a/@here/harp-materials/lib/EdgeMaterial.ts
+++ b/@here/harp-materials/lib/EdgeMaterial.ts
@@ -131,6 +131,11 @@ export interface EdgeMaterialParameters
      * @defaultValue false
      */
     vertexColors?: boolean;
+
+    /**
+     * GLSL version
+     */
+    glslVersion?: number;
 }
 
 /**
@@ -164,7 +169,8 @@ export class EdgeMaterial extends RawShaderMaterial
         if (params?.vertexColors === true) {
             setShaderDefine(defines, "USE_COLOR", true);
         }
-        const shaderParams = {
+
+        const shaderParams: THREE.ShaderMaterialParameters = {
             name: "EdgeMaterial",
             vertexShader: vertexSource,
             fragmentShader: fragmentSource,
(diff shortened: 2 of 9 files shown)


J-Rojas:
Unfortunately I don't have much time to fix these tests. The code works well for me. If anyone would like to address the intergration tests, I'd appreciate it.

nzjony:

@here/harp-materials/lib/EdgeMaterial.ts
```diff
@@ -182,6 +188,9 @@ export class EdgeMaterial extends RawShaderMaterial
             depthWrite: false,
             defines
         };
+        if (params?.glslVersion === 3.0) {
+            shaderParams.glslVersion = THREE.GLSL3;
+        }
```
atomicsulfate, line 193: If we set THREE.ShaderMaterialParameters' glslVersion directly in function "createMaterial" instead, then we don't have to repeat the same logic for every material derived from THREE.RawShaderMaterial. It's also easy to forget adding doing this whenever we add a new material.

"createMaterial" can get as parameter the renderer capabilities and set the glsl version for instances of THREE.RawShaderMaterial when the renderer is WebGL2.

@here/harp-text-canvas/lib/TextCanvas.ts
```diff
@@ -255,7 +255,10 @@ export class TextCanvas {
 
         if (params.material === undefined) {
             this.m_ownsMaterial = true;
-            this.m_material = createSdfTextMaterial({ fontCatalog: params.fontCatalog });
+            this.m_material = createSdfTextMaterial({
+                fontCatalog: params.fontCatalog,
+                glslVersion: this.m_renderer.capabilities.isWebGL2 ? 3.0 : undefined
+            });
```
atomicsulfate, line 261: Same here, just pass the renderer capabilities to createSdfTextMaterial and set there directly the THREE.ShaderMaterialParameters' glslVersion. That way you don't need to repeat the same logic below.

Unfortunately, the same code will have to be duplicated in createMaterial and createSdfTextMaterial, cause harp-text-canvas is meant to be a standalone module.

atomicsulfate (changes requested):
I'd rather move the glsl version handling to the create material functions, please have a look at my comments.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2145,6 +2145,13 @@ export class MapView extends EventDispatcher {
         return this.m_tileDataSources;
     }
 
+    /**
+     * Returns the glsl version compatible with the MapView's renderer
+     */
+    get glslVersion(): number {
+        return this.m_renderer.capabilities.isWebGL2 ? 3.0 : 1.0;
+    }
+
```
atomicsulfate, line 2154: This is a too low-level detail for MapView's interface. We try to move away from using MapView's as a facade for lower level interfaces.
You can access the renderer capabilities doing "mapView.renderer.capabilities", see my other comments.

nzjony:
Closing, because this was superseded by: https://github.com/heremaps/harp.gl/pull/1887

PR #1875, opened 2020-09-25
HARP-11672: Add TileGeometryLoader's finished promise

atomicsulfate:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
+++ b/@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
@@ -412,5 +412,104 @@ describe("TileGeometryLoader", function() {
                 expect(geometryLoader.isFinished).to.be.true;
             });
         });
+
+        it("should reload geometry for already loaded tile that was reset", async function() {
+            tile.decodedTile = createFakeDecodedTile();
+
+            const geometryCreator = TileGeometryCreator.instance;
+            const spyProcessTechniques = sandbox.spy(geometryCreator, "processTechniques") as any;
+            const spyCreateGeometries = sandbox.spy(geometryCreator, "createAllGeometries") as any;
+            expect(spyCreateGeometries.callCount).equal(0);
+            expect(spyProcessTechniques.callCount).equal(0);
+
+            geometryLoader!.update();
+
+            expect(mapView.taskQueue.numItemsLeft(TileTaskGroups.CREATE)).equal(1);
+            expect(mapView.taskQueue.processNext(TileTaskGroups.CREATE)).equal(true);
+
+            geometryLoader.waitFinished().should.be.fulfilled.then(() => {
+                expect(spyProcessTechniques.callCount).equal(1);
+                expect(spyCreateGeometries.callCount).equal(1);
+                expect(geometryLoader.isFinished).to.be.true;
+
+                geometryLoader.reset();
+                expect(geometryLoader.isFinished).to.be.false;
+
+                expect(mapView.taskQueue.numItemsLeft(TileTaskGroups.CREATE)).equal(1);
+                expect(mapView.taskQueue.processNext(TileTaskGroups.CREATE)).equal(true);
+
+                geometryLoader.waitFinished().should.be.fulfilled.then(() => {
+                    expect(spyProcessTechniques.callCount).equal(2);
+                    expect(spyCreateGeometries.callCount).equal(2);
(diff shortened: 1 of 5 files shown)


@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
```diff
@@ -25,6 +24,8 @@ import * as sinon from "sinon";
 
 chai.use(chaiAsPromised);
 const { expect } = chai;
+chai.should();
```
FraukeF, line 27: This is probably not needed?

FraukeF (dismissed):

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -225,32 +242,46 @@ export class TileGeometryLoader {
...
         this.m_state = TileGeometryLoaderState.Disposed;
+        this.m_rejectFinishedPromise?.();
     }
 
     /**
      * Reset the loader to its initial state and cancels any asynchronous work.
      */
     reset(): void {
+        this.clear();
+
```
nzjony, line 255: Should the `this.m_rejectFinishedPromise`  be called first? Otherwise anyone waiting on the current promise will never be informed of the new promise and it will never be resolved / finished.

If you intend it, then at least document it.

atomicsulfate: reset() does not change the state of the loader to any final state (cancelled, finished or disposed). It's done to reload the tile again (e.g. when tile is dirty), so waiters can still wait until the new tile geometry is loaded. I'll add a note to clarify it.

@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
atomicsulfate: It's part of the setup to be able to use "should" with "chai-as-promised". Otherwise the chai-as-promised extensions don't work with should.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
nzjony: I am referring to the condition below, where the ` this.m_rejectFinishedPromise` is set to a new function in the case where the state is Finished/Canceled/Disposed.

But you anyway call the promise in those cases, so I now see that it isn't an issue.

nzjony: Actually, I notice this method is never tested in the `TileGeometryLoaderTest.ts`, so please add a test for it.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -225,32 +242,46 @@ export class TileGeometryLoader {
...
-        this.m_tile.removeDecodedTile();
-
         this.m_decodedTile = undefined;
         this.m_state = TileGeometryLoaderState.Finished;
+        this.m_resolveFinishedPromise?.();
     }
 
     private cancel() {
         this.m_state = TileGeometryLoaderState.Canceled;
+        this.m_rejectFinishedPromise?.();
```
nzjony, line 284: Also, can you create a test for the state canceled and check that this promise is called?

@here/harp-mapview-decoder/test/TileGeometryLoaderTest.ts
FraukeF: ah ok, thanks for clarifiing

FraukeF (dismissed):

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
atomicsulfate: This is already covered in two tests: 
"should not start geometry creation for invisible tile"
"should not start geometry loading for invisible tile"

Note that cancel is private. It's called indirectly whenever the update() method is called for an invisible tile.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
atomicsulfate: I've added tests to cover the reset method.

atomicsulfate:

nzjony (approved):

PR #1876, opened 2020-09-28
HARP-11672: Make TileGeometryLoader protected.

atomicsulfate:
A Tile subclass may still need access (e.g. to attach callbacks to the finished promise).

Also, fix uncaught promised exception when a tile is disposed
before it waits on the TileGeometryDecoder promise.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -1025,11 +1025,12 @@ export class Tile implements CachedResource {
             this.m_tileLoader.cancel();
             this.m_tileLoader = undefined;
         }
-        this.m_tileGeometryLoader?.dispose();
         this.clear();
-        this.m_disposed = true;
         // Ensure that tile is removable from tile cache.
         this.frameNumLastRequested = 0;
+        this.m_disposed = true;
+        this.m_tileGeometryLoader?.dispose();
+
         if (this.m_disposeCallback) {
             this.m_disposeCallback(this);
         }
@@ -1048,6 +1049,7 @@ export class Tile implements CachedResource {
     /**
      * Update tile for current map view zoom level
      * @param zoomLevel - Zoom level of the map view
+     * @internal
      */
     update(zoomLevel: number): void {
         for (const object of this.objects) {
@@ -1080,7 +1082,7 @@ export class Tile implements CachedResource {
     ): boolean {
         if (this.m_tileGeometryLoader) {
             if (priority !== undefined) {
-                this.m_tileGeometryLoader.priority;
+                this.m_tileGeometryLoader.priority = priority;
             }
             this.m_tileGeometryLoader.update(enabledKinds, disabledKinds);
             return true;
@@ -1115,7 +1117,13 @@ export class Tile implements CachedResource {
                 this.removeDecodedTile();
             })
             .catch(() => {
-                /* ignore */
+                if (this.disposed) {
+                    return;
+                }
+                // Loader was canceled, dispose tile.
+                if (!this.dataSource.isDetached()) {
+                    this.mapView.visibleTileSet.disposeTile(this);
+                }
             });
     }
 
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -391,8 +391,16 @@ export class Tile implements CachedResource {
         }
     }
 
+    /**
+     * Sets the tile visibility status.
+     * @param visible - `True` to mark the tile as visible, `False` otherwise.
+     */
     set isVisible(visible: boolean) {
         this.frameNumLastRequested = visible ? this.dataSource.mapView.frameNumber : -1;
+
+        if (!visible && this.m_tileGeometryLoader && !this.m_tileGeometryLoader.isSettled) {
+            this.m_tileGeometryLoader.cancel();
+        }
     }
 
     /**
@@ -1088,6 +1096,11 @@ export class Tile implements CachedResource {
             return true;
         }
 
+        if (this.dataSource.isDetached()) {
+            this.m_tileGeometryLoader.cancel();
+            return true;
+        }
+
         if (this.tileLoader) {
             if (!this.tileLoader.isFinished) {
                 return true;
@@ -1098,10 +1111,6 @@ export class Tile implements CachedResource {
             }
         }
 
-        if (!this.isVisible) {
-            this.m_tileGeometryLoader.cancel();
-            return true;
-        }
         if (priority !== undefined) {
             this.m_tileGeometryLoader.priority = priority;
         }
--- a/@here/harp-mapview/lib/Tile.ts
+++ b/@here/harp-mapview/lib/Tile.ts
@@ -3,7 +3,12 @@
  * Licensed under Apache 2.0, see full license in LICENSE
  * SPDX-License-Identifier: Apache-2.0
  */
-import { DecodedTile, GeometryType, TextPathGeometry } from "@here/harp-datasource-protocol";
+import {
+    DecodedTile,
+    GeometryKindSet,
+    GeometryType,
+    TextPathGeometry
+} from "@here/harp-datasource-protocol";
 import { GeoBox, OrientedBox3, Projection, TileKey } from "@here/harp-geoutils";
 import { assert, CachedResource, chainCallbacks, LoggerManager } from "@here/harp-utils";
 import * as THREE from "three";
@@ -290,6 +295,7 @@ export class Tile implements CachedResource {
      * Prepared text geometries optimized for display.
      */
     protected preparedTextPaths: TextPathGeometry[] | undefined;
+    protected readonly m_tileGeometryLoader?: TileGeometryLoader;
 
     /**
      * The bounding box of this `Tile` in world coordinates.
@@ -304,7 +310,6 @@ export class Tile implements CachedResource {
 
     private m_tileLoader?: ITileLoader;
     private m_decodedTile?: DecodedTile;
-    private m_tileGeometryLoader?: TileGeometryLoader;
 
     // Used for {@link TextElement}s that are stored in the data, and that are placed explicitly,
     // fading in and out.
@@ -360,6 +365,10 @@ export class Tile implements CachedResource {
         this.m_textStyleCache = new TileTextStyleCache(this);
         this.m_offset = offset;
         this.m_uniqueKey = TileOffsetUtils.getKeyForTileKeyAndOffset(this.tileKey, this.offset);
+        if (dataSource.useGeometryLoader) {
+            this.m_tileGeometryLoader = new TileGeometryLoader(this, this.mapView.taskQueue);
+            this.attachGeometryLoadedCallback();
+        }
     }
 
     /**
@@ -747,16 +756,6 @@ export class Tile implements CachedResource {
             this.forceHasGeometry(true);
         }
 
-        this.m_tileGeometryLoader
-            ?.waitFinished()
-            .then(() => {
-                this.loadingFinished();
-                this.removeDecodedTile();
-            })
-            .catch(() => {
-                /* ignore */
-            });
-
         // If the decoder provides a more accurate bounding box than the one we computed from
         // the flat geo box we take it instead. Otherwise, if an elevation range was set, elevate
         // bounding box to match the elevated geometry.
@@ -823,31 +822,10 @@ export class Tile implements CachedResource {
     }
 
     /**
-     * Gets the [[TileGeometryLoader]] that manages this tile.
-     * @internal
-     */
-    get tileGeometryLoader(): TileGeometryLoader | undefined {
-        return this.m_tileGeometryLoader;
-    }
-
-    /**
-     * Sets the [[TileGeometryLoader]] to manage this tile.
-     *
-     * @param tileGeometryLoader - A [[TileGeometryLoader]] instance to manage the geometry creation
-     *      for this tile.
-     * @internal
-     */
-    set tileGeometryLoader(tileGeometryLoader: TileGeometryLoader | undefined) {
-        this.m_tileGeometryLoader = tileGeometryLoader;
-    }
-
-    /**
      * `True` if all geometry of the `Tile` has been loaded.
      */
     get allGeometryLoaded(): boolean {
-        return this.m_tileGeometryLoader === undefined
-            ? this.hasGeometry
-            : this.m_tileGeometryLoader.isFinished;
+        return this.m_tileGeometryLoader?.isFinished ?? this.hasGeometry;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


@here/harp-mapview/lib/Utils.ts
```diff
@@ -1631,7 +1631,7 @@ export namespace MapViewUtils {
                 if (tile.tileLoader !== undefined && !tile.tileLoader.isFinished) {
```
FraukeF, line 1631: Put both conditions of tile.tileLoader and tileGeometryLoader into soemthing like tile.isLoading, no need for Utils then to know the internals

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -1241,30 +1240,25 @@ export class VisibleTileSet {
         const dataSourceCache = this.m_dataSourceCache;
         const retainedTiles: Set<TileCacheId> = new Set();
 
-        function markTileDirty(tile: Tile, tileGeometryManager: TileGeometryManager) {
+        function markTileDirty(tile: Tile) {
             const tileKey = DataSourceCache.getKeyForTile(tile);
             if (!retainedTiles.has(tileKey)) {
                 retainedTiles.add(tileKey);
-                if (tile.tileGeometryLoader !== undefined) {
```
FraukeF, line 1248: doesnt it still need some kind of tile.dispose() call or something similar?

@here/harp-mapview/lib/Utils.ts
FraukeF: there is also 2 references to tile.tileLoader in the VisibleTileSet which look like they could be using this too, something like if(tile.isLoading) {tile.canccel())

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -513,6 +517,13 @@ export class TileGeometryLoader {
         return true;
     }
 
+    /**
+     * `True` if TileGeometryLoader was canceled
+     */
+    private get isCanceled(): boolean {
+        return this.m_state === TileGeometryLoaderState.Canceled;
+    }
+
```
FraukeF, line 526: Can you have a look at TileGeometryLoader.discardNeedlessTile, it has not much to do with the tileGEometryLoader, and has some back reference to the visibleTileSet for disposal, keeping this in tile, might make things easier, maybe it could become something like  if(!tile.needed){tile.dispose()}

@here/harp-mapview/lib/Utils.ts
FraukeF: this might lead to more loading tiles being found and canceled, if the check includes the tileGeometryLoading

@here/harp-mapview/lib/Tile.ts
```diff
@@ -909,6 +887,14 @@ export class Tile implements CachedResource {
             return await Promise.resolve();
         }
 
+        if (this.m_tileGeometryLoader) {
```
nzjony, line 890: truthy

@here/harp-mapview/lib/Utils.ts
atomicsulfate: This code is wrong. If the tileLoader is not finished then tile.allGeometryLoaded is false as well. Just checking allGeometryLoaded should suffice.

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: I've moved the reset of the tileGeometryLoader to tile.load(). This reset is done to mark a tile as dirty, so we want to keep the tile contents till the new contents are loaded.

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
```diff
@@ -80,36 +79,22 @@ export class TileGeometryManager {
...
-                }
+            //this assumes the tiles are ordered by priority, this is currently done in
+            // the visible tile set with 0 as the highest priority
+            const tilePriority = prio++;
+            const updateDone = tile.updateGeometry(
+                tilePriority,
+                this.enableFilterByKind ? this.enabledGeometryKinds : undefined,
+                this.enableFilterByKind ? this.disabledGeometryKinds : undefined
+            );
+            if (updateDone && this.m_tileUpdateCallback) {
```
nzjony, line 96: truthy

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1081,6 +1064,38 @@ export class Tile implements CachedResource {
...
+     * @param disabledKinds - {@link GeometryKind}s that will not be created.
+     * @return `true` if geometry update was done, `false` otherwise.
+     * @internal
+     */
+    updateGeometry(
+        priority?: number,
+        enabledKinds?: GeometryKindSet,
+        disabledKinds?: GeometryKindSet
+    ): boolean {
+        if (this.m_tileGeometryLoader) {
```
nzjony, line 1081: truthy

@here/harp-mapview/lib/Tile.ts
```diff
@@ -1081,6 +1064,38 @@ export class Tile implements CachedResource {
...
+            if (priority !== undefined) {
+                this.m_tileGeometryLoader.priority;
```
nzjony, line 1083: what is this line doing?

@here/harp-mapview/lib/Tile.ts
nzjony, line 890: Minor: why not move this into the then below, that means, if the tile is disposed before it is loaded, that this extra callback isn't queued.

nzjony (commented):
In general, this whole logic is like chasing a rabbit hole, we should prioritize to make some sequence diagram to make it more easier to understand.

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
```diff
@@ -80,36 +79,22 @@ export class TileGeometryManager {
...
-                );
-                if (this.m_tileUpdateCallback) {
-                    this.m_tileUpdateCallback(tile);
-                }
+            //this assumes the tiles are ordered by priority, this is currently done in
+            // the visible tile set with 0 as the highest priority
+            const tilePriority = prio++;
+            const updateDone = tile.updateGeometry(
+                tilePriority,
+                this.enableFilterByKind ? this.enabledGeometryKinds : undefined,
```
nzjony, line 93: Also, would it make sense to remove this logic to enable / disable geometry kinds at run time, and instead, when the kinds change, just reload the tiles.

That would simplify the logic quite a bit, instead of constantly calling `update`, it could just be one single call to await on the TileLoader to finish loading and then load the geometry.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
atomicsulfate: I did something in the middle, let me know what you think.

@here/harp-mapview/lib/Tile.ts
atomicsulfate: done

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
atomicsulfate: Makes sense. Could you bring it up in the chat and create a ticket if nobody sees any problem?

@here/harp-mapview/lib/Tile.ts
atomicsulfate: The tile might be disposed before the decoding finishes. Then, the TileGeometryLoader rejects its promise and that'll cause and exception cause there's no reject callback attached.
That's actually the issue we have right now in master.

nzjony: I got ya.

You could alternatively in that use `finally`, so that even if it is disposed, that the reset method is called. But not super important, makes it a bit easier to read, but also ok.

atomicsulfate: as discussed, here it's ok.

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
atomicsulfate: as discussed, here it's ok.

@here/harp-mapview/lib/Tile.ts
atomicsulfate: as discussed, here it's ok.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -207,8 +230,7 @@ export class TileGeometryLoader {
         // If the tile is not ready for display, or if it has become invisible while being loaded,
         // for example by moving the camera, the tile is not finished and its geometry is not
         // created. This is an optimization for fast camera movements and zooms.
-        if (this.discardNeedlessTile(tile)) {
-            this.cancel();
+        if (this.cancelNeedlessTile(tile)) {
```
FraukeF, line 233: its better then before, but I still think that it would be simpler to have just one direction of command and call on tileGeometryLoader just to do what it should do. And as hopefully we can get rid of the update function, .. I think the Tile should check if its still needed before even calling on the tileGeometryLoader.

So before even calling on tileGEometryLoader.update the tile could check for:

if (!tile.isVisible || tile.disposed) {....

and cancel the tileGEometryLoading

@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -659,7 +659,7 @@ export class VisibleTileSet {
             // Remove all tiles that are still being loaded, but are no longer visible. They have to
             // be reloaded when they become visible again. Hopefully, they are still in the browser
             // cache by then.
-            if (!tile.isVisible && tile.tileLoader !== undefined && !tile.tileLoader.isFinished) {
+            if (!tile.isVisible && !tile.allGeometryLoaded) {
```
FraukeF, line 662: is there more references to tile.tileLoader? otherwise we could now make it private too

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
FraukeF: I left a comment there

@here/harp-mapview/lib/geometry/TileGeometryManager.ts
FraukeF: would be great if we can get rid of it and remove the update function

atomicsulfate: Yes, but let's do it in a separate ticket, otherwise this change is gonna blow up outside of the scope of this ticket.
I could do it after my current ticket.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
atomicsulfate: Have you seen that createGeomety() (executed in the async task) is also calling cancelNeedlessTile? 
I don't see it makes much sense to remove it from here but leave it there.

When we remove the whole tile.update() workflow all this code will disappear anyway.

FraukeF: yes, I saw, but I think, this should not happen, as the isExpired check for the task should cancel the task before we even get there.
But, I am ok, to wait this out, until tile.update() hopefully disappears

@here/harp-mapview/lib/VisibleTileSet.ts
FraukeF: If TileDataSource would add getters on dataProvider and decoder, it could be constructed in the Tile without needing a signature change

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
FraukeF: The whole idea is to separate the resposibilities to avoid the "rabbithole chasing" as @nzjony  mentioned, wh should let tileGeometryLoader do what its supposed to do, that is load the geometry, .. disposing the tile it belongs to, shouldnt be part of that.

atomicsulfate: done. Moved the check to the isExpired function for the task.

@here/harp-mapview/lib/VisibleTileSet.ts
atomicsulfate: That's precisely what I meant by a workaround, those getters shouldn't be needed. Anyway, let's see what can be done, but in a separate PR.

atomicsulfate:

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -25,6 +25,21 @@ import { PerformanceStatistics } from "../Statistics";
 import { Tile } from "../Tile";
 import { TileGeometryCreator } from "./TileGeometryCreator";
 
+function addDiscardedTileToStats(tile: Tile) {
```
FraukeF, line 28: this should move to Tile.ts

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -301,7 +307,14 @@ export class TileGeometryLoader {
             group: TileTaskGroups.CREATE,
             getPriority: this.getPriority.bind(this),
             isExpired: () => {
-                return this.m_state !== TileGeometryLoaderState.CreationQueued;
+                if (this.m_state !== TileGeometryLoaderState.CreationQueued) {
```
FraukeF, line 310: This change is not needed, as the tile.visible setter cancels the tileGeometryLoader, the state should be changed at this point.
Maybe its missing in the tile.dispose setter, but basically if the tile becomes "needless" it changes the state of TileGEometryLoader and isExpired can stay the same checking for the state only

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
atomicsulfate: This must be done on every call to TileGeometryLoader.cancel(). The simplest way to do it is to do it inside the cancel function. Otherwise, we need to create a new function in tile to ensure that we call TileGeometryLoader.cancel() and log the stats.
Furthermore, the rest of the stats are also logged from TileGeometryLoader.

FraukeF: thanks for clarifiing, I did not look at the bigger picture there, that the other stats are also there.

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
atomicsulfate: done

FraukeF (approved):

atomicsulfate:

PR #1885, opened 2020-10-02
HARP-12026 Apply tilt limit also to rotation center

nzjony:

### DIFF
--- a/@here/harp-mapview/lib/Utils.ts
+++ b/@here/harp-mapview/lib/Utils.ts
@@ -248,6 +248,14 @@ export namespace MapViewUtils {
             return;
         }
 
+        const mapTargetWorld =
+            offsetX === 0 && offsetY === 0
+                ? rotationTargetWorld
+                : MapViewUtils.rayCastWorldCoordinates(mapView, 0, 0);
+        if (mapTargetWorld === null) {
+            return;
+        }
+
         applyAzimuthAroundTarget(mapView, rotationTargetWorld, -deltaAzimuth);
 
         const clampedDeltaTilt = computeClampedDeltaTilt(
@@ -256,6 +264,7 @@ export namespace MapViewUtils {
             offsetY,
             deltaTilt,
             maxTiltAngle,
+            mapTargetWorld,
             rotationTargetWorld
         );
 
@@ -289,7 +298,9 @@ export namespace MapViewUtils {
      * @internal
      *
      * Clamps the supplied `deltaTilt` to the `maxTiltAngle` supplied. Note, when a non-zero offset
-     * is applied, we apply
+     * is applied, we apply another max angle of 89 degrees to the rotation center to prevent some
+     * corner cases where the angle at the rotation center is 90 degrees and therefore intersects
+     * the geometry with the near plane.
      */
     function computeClampedDeltaTilt(
         mapView: MapView,
@@ -297,38 +308,25 @@ export namespace MapViewUtils {
         offsetY: number,
         deltaTilt: number,
         maxTiltAngle: number,
+        mapTargetWorld: THREE.Vector3,
         rotationTargetWorld: THREE.Vector3
     ): number {
         const camera = mapView.camera;
         const projection = mapView.projection;
 
-        const mapTargetWorld =
-            offsetX === 0 && offsetY === 0
-                ? rotationTargetWorld
-                : MapViewUtils.rayCastWorldCoordinates(mapView, 0, 0);
-        if (mapTargetWorld === null) {
-            return 0;
+        const tilt = extractTiltAngleFromLocation(projection, camera, mapTargetWorld);
+        if (tilt + deltaTilt < 0) {
+            // Clamp the final tilt to 0
+            return -tilt;
+        } else if (deltaTilt <= 0) {
+            // Reducing the tilt isn't clamped (apart from above).
+            return deltaTilt;
         }
 
-        let currentTilt = 0;
-        {
-            const mapTargetNormal = projection.surfaceNormal(mapTargetWorld, cache.vector3[0]);
-            camera.updateMatrixWorld(true);
-
-            const dir = cache.vector3[1].setFromMatrixColumn(camera.matrixWorld, 2);
-            const up = cache.vector3[2].setFromMatrixColumn(camera.matrixWorld, 1);
-
-            // Due to inaccuracies it can happen that the tilt angle gets less than 0.
-            // Using the dot product of the direction vector and the surface normal alone does not give us the sign.
-            const tiltSign = Math.sign(up.dot(mapTargetNormal));
-            currentTilt = (tiltSign === 0 ? 1 : tiltSign) * dir.angleTo(mapTargetNormal);
-        }
-
-        const currentTiltRotationCenter = MapViewUtils.extractTiltAngleFromLocation(
-            mapView.projection,
-            mapView.camera,
-            rotationTargetWorld!
-        );
+        const rotationCenterTilt =
+            offsetX === 0 && offsetX === offsetY
+                ? tilt
+                : extractTiltAngleFromLocation(projection, camera, rotationTargetWorld!);
 
         // This is used to find the max tilt angle, because the difference in normals is needed
         // to correct the triangle used to find the max tilt angle at the rotation center, please
@@ -344,65 +342,59 @@ export namespace MapViewUtils {
         }
 
         const ninetyRad = THREE.MathUtils.degToRad(90);
-        // This calculates the angle between the direction vector to the rotation center from the
-        // camera and the direction vector to the map view center (offsetX and offsetY == 0). This
-        // value stays constant when tilting but using the same X & Y offsets. We use the triangle
-        // with these two vectors from the camera position and the intersection of the flat plane.
-        // We know two angles, and the third is easy to compute (just 180 minus the sum of the
-        // rest). the first angle is 90 degrees + the current tilt, i.e. the tilt at the map
-        // center. The second angle is 90 - the tilt at the rotation center. Note, this doesn't work
-        // for sphere, it would have to be fixed for that.
+        // The following terminology will be used:
+        // R = rotation target, i.e. the point about which we are rotating: `rotationTargetWorld`
+        // C = camera position
+        // M = map target, i.e. the point which the camera is looking at: `mapTargetWorld`
+        // The following variable `angleBetweenRotationTargetAndMapTarget` is the angle between the
+        // two rays C->R and C->M, i.e. the angle RCM. This angle remains constant when tilting with
+        // a fixed `offsetX` and `offsetY`. It is calculated by using the intersection of the two
+        // ways with the earth. Where:
+        // MRC is 90 + `angleBetweenNormals` - tilt at R.
+        // CMR is 90 + `tilt`
+        // RCM is 180 - CRM - RMC, aka `angleBetweenRotationTargetAndMapTarget`
+        // Note the use of `angleBetweenNormals` to ensure this works for spherical projections.
         const angleBetweenRotationTargetAndMapTarget =
             ninetyRad * 2 -
-            (ninetyRad + angleBetweenNormals - currentTiltRotationCenter + ninetyRad + currentTilt);
+            (ninetyRad + angleBetweenNormals - rotationCenterTilt + ninetyRad + tilt);
 
-        // This uses the same formula as above, just we change `currentTilt` to the maxTiltAngle,
+        // This uses the same formula as above, just we change `tilt` to the maxTiltAngle,
         // i.e we want to find the greatest angle at the rotation target that gives us the max
         // angle at the map center target.
+        // i.e. CMR is 90 + `maxTiltAngle`
         const maxRotationTargetAngle =
             ninetyRad * 2 - angleBetweenRotationTargetAndMapTarget - ninetyRad - maxTiltAngle;
 
-        // We compute the max tilt based on the angle of the inner triangle, note, for sphere
-        // projection this is just an approximation, in order to compute this exactly, we would
-        // need to shift the camera to the new position, compute the new normal at the map center
-        // and use this to then compute the angleBetweenNormals, it would then actually need to be
-        // done iteratively, because when shifting the camera, it is possible that we move it to a
-        // position which is outside the max tilt given the new normal. Because the `deltaTilt` is
-        // generally small, the difference in normals should be for most cases very small and the
-        // extra computation required to compute such a small value isn't worth it. The value we
-        // compute is anyway conservative, so we don't risk computing something that will put the
-        // camera outside the valid range.
+        // Converting the `maxRotationTargetAngle` back to a tilt is as easy as subtracting it from
+        // 90 and the `angleBetweenNormals`, i.e. this gives us the maximum allowed tilt at R
+        // that satisfies the `maxTiltAngle` constraint.
         const maxTilt = ninetyRad + angleBetweenNormals - maxRotationTargetAngle;
 
         const clampedDeltaTiltMapTarget =
-            MathUtils.clamp(deltaTilt + currentTiltRotationCenter, 0, maxTilt) -
-            currentTiltRotationCenter;
+            MathUtils.clamp(deltaTilt + rotationCenterTilt, 0, maxTilt) - rotationCenterTilt;
 
         const maxRotationTiltAngle = THREE.MathUtils.degToRad(89);
-        // Here we clamp to 89 degrees, to prevent it intersecting with the world at 90, note, we
-        // don't use the `maxTiltAngle`, because then special handling needs to be applied when for
-        // example the tilt at the rotation center is more than that at the map target, then the
-        // tilt feels different at the center / offset positions. Using 89 gives the most natual
-        // feeling when applying the tilt angle.
+        // Here we clamp to 89 degrees, to prevent it intersecting with the world at 90. This is
+        // possible for example when the R position is near the horizon. If the angle RCM is say 5
+        // degrees, then an angle of say 89 degrees at R, plus 5 degrees means the tilt at M would
+        // be 84 degrees. If the `maxTiltAngle` is however 89, then we still have what seems like 5
+        // degrees at M, so, without this check, we would go to 90 degrees.
         const clampedDeltaTiltRotationTarget =
-            MathUtils.clamp(deltaTilt + currentTiltRotationCenter, 0, maxRotationTiltAngle) -
-            currentTiltRotationCenter;
+            MathUtils.clamp(deltaTilt + rotationCenterTilt, 0, maxRotationTiltAngle) -
+            rotationCenterTilt;
 
         // Math.min doesn't work, because we need to keep the sign of the delta as well.
-        const clampedDeltaTilt =
-            Math.abs(clampedDeltaTiltMapTarget) < Math.abs(clampedDeltaTiltRotationTarget)
-                ? clampedDeltaTiltMapTarget
-                : clampedDeltaTiltRotationTarget;
+        const clampedDeltaTilt = Math.min(
+            clampedDeltaTiltMapTarget,
+            clampedDeltaTiltRotationTarget
+        );
 
         // It can happen in some cases that the maxRotationTiltAngle is exceeded because the
         // `clampedDeltaTilt` is taken from the smaller value clampedDeltaTiltMapTarget. We check
         // that we don't exceed 0, because otherwise it will snap to 89, so we let the user ease
         // back to 89, but not any further.
-        if (
-            currentTiltRotationCenter + clampedDeltaTilt >= maxRotationTiltAngle &&
-            clampedDeltaTilt > 0
-        ) {
-            return maxRotationTiltAngle - currentTiltRotationCenter;
+        if (rotationCenterTilt + clampedDeltaTilt >= maxRotationTiltAngle) {
+            return maxRotationTiltAngle - rotationCenterTilt;
         }
 
         return clampedDeltaTilt;
@@ -420,10 +412,10 @@ export namespace MapViewUtils {
         deltaTilt: number
     ) {
         const camera = mapView.camera;
-        // We use index 2, because these values are required if the method: rayCastWorldCoordinates
-        // fails, which itself uses indexes 0 and 1.
-        const posBackup = cache.vector3[2].copy(camera.position);
-        const quatBackup = cache.quaternions[1].copy(camera.quaternion);
+        // Consider to use the cache if necessary, but beware, because the `rayCastWorldCoordinates`
+        // also uses this cache.
+        const posBackup = camera.position.clone();
+        const quatBackup = camera.quaternion.clone();
 
         const tiltAxis = cache.vector3[0].set(1, 0, 0).applyQuaternion(camera.quaternion);
         const tiltQuat = cache.quaternions[0].setFromAxisAngle(tiltAxis, deltaTilt);
--- a/@here/harp-mapview/lib/Utils.ts
+++ b/@here/harp-mapview/lib/Utils.ts
@@ -11,6 +11,7 @@ import {
     OrientedBox3,
     Projection,
     ProjectionType,
+    sphereProjection,
     TileKey,
     Vector3Like
 } from "@here/harp-geoutils";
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/Utils.ts
```diff
@@ -278,8 +278,31 @@ export namespace MapViewUtils {
...
+        // mapTargetWorld then has 3 angles, 5 degrees, 90 + 84.5, that then puts the angle between
+        // the camera and the world plane at 0.5 degrees (or as a tilt, 89.5), which is outside the
+        // tilt limit of 89.
+        const dirToRotationTarget = cache.vector3[0];
+        dirToRotationTarget.copy(camera.position).sub(rotationTargetWorld);
+        dirToRotationTarget.normalize();
+        const currentTiltRotationCenter = dirToRotationTarget.angleTo(mapTargetNormal);
+        // Set the delta in such a way that the final angle will be maxTiltAngle
+        if (currentTiltRotationCenter + deltaTilt > maxTiltAngle) {
+            clampedDeltaTilt = maxTiltAngle - currentTiltRotationCenter;
```
atomicsulfate, line 303: here you're overwriting the clamped delta tilt with respect to the map target. what if clampedDeltaTiltAtMapTarget < clampedDeltaTiltAtRotTarget ?
Then you will apply clampedDeltaTiltAtRotTarget and the camera will be at a tilt higher than maxTilt at the map target.

I think here you should take the most restrictive of the two clamped delta tilts (the minimum).

@here/harp-mapview/lib/Utils.ts
```diff
@@ -278,8 +278,31 @@ export namespace MapViewUtils {
...
+        // maxTiltAngle. As an example, imagine our max tilt angle is 89 degrees and that the angle
+        // between the center / rotation target is 5 degrees and that the current tilt to the
+        // mapTargetWorld is 84.5 degrees, the triangle from the rotaionTargetWorld -> camera ->
+        // mapTargetWorld then has 3 angles, 5 degrees, 90 + 84.5, that then puts the angle between
+        // the camera and the world plane at 0.5 degrees (or as a tilt, 89.5), which is outside the
+        // tilt limit of 89.
+        const dirToRotationTarget = cache.vector3[0];
+        dirToRotationTarget.copy(camera.position).sub(rotationTargetWorld);
+        dirToRotationTarget.normalize();
+        const currentTiltRotationCenter = dirToRotationTarget.angleTo(mapTargetNormal);
```
atomicsulfate, line 300: I think you should use here the normal at the rotation center instead of the normal at the map target. For planar is the same. For sphere, I'm not sure what's the meaning of the angle between the rotation target direction and the normal at the map target.

nzjony: I forgot about sphere.... will take a re-look, thanks for noticing.

@here/harp-mapview/lib/Utils.ts
nzjony: Good spotting, I will fix.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
+     */
+    function applyTiltAroundTarget(
+        mapView: MapView,
+        rotationTargetWorld: THREE.Vector3,
+        deltaTilt: number
+    ) {
+        const camera = mapView.camera;
+        // We use index 2, because these values are required if the method: rayCastWorldCoordinates
+        // fails, which itself uses indexes 0 and 1.
+        const posBackup = cache.vector3[2].copy(camera.position);
```
atomicsulfate, line 425: These problems with functions in Utils.ts trashing each other's temp variables are getting scary.
We are abusing temp caches in this file, having a cache object of temp variable arrays is just too much.

If you had to add a comment to clarify why you have to use that tmp variable and no other, it's a good sign we have to stop. I suggest you just clone camera position. If later we see a problem with that, we'll see what we can do.

I would vote to dismantle that cache object we use here.

nzjony: Fully, agree, it is a nightmare to program with, I had numerous times when bugs showed just because I forgot I was overwriting memory with a function call that I needed later.

Will remove them in this case that you mention. In the other cases, the cache usage is ok, because something is calculated immediately and not required anymore.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
+        const projection = mapView.projection;
 
         const mapTargetWorld =
             offsetX === 0 && offsetY === 0
                 ? rotationTargetWorld
                 : MapViewUtils.rayCastWorldCoordinates(mapView, 0, 0);
         if (mapTargetWorld === null) {
-            return;
+            return 0;
         }
```
atomicsulfate, line 311: Before we were returning from "orbitAroundScreenPoint" if the map target is not on the ground. Now we return 0 tilt and go on an apply it. We could return immediately as before.
Maybe this piece of code belongs to "orbitAroundScreenPoint" instead, that way you can return from the function directly without forcing this one to return some undefined tilt value.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
+            currentTilt = (tiltSign === 0 ? 1 : tiltSign) * dir.angleTo(mapTargetNormal);
+        }
 
-        //FIXME(HARP-11926): For globe tilt in the map center is different from the tilt in the rotation center,
-        //hence the clamped tilt is too conservative.
+        const currentTiltRotationCenter = MapViewUtils.extractTiltAngleFromLocation(
+            mapView.projection,
+            mapView.camera,
+            rotationTargetWorld!
+        );
```
atomicsulfate, line 331: we should skip doing this if offset is 0 (when mapTarget === rotationTarget). In this case currentTiltRotationCenter === currentTilt

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
+
+        const maxRotationTiltAngle = THREE.MathUtils.degToRad(89);
+        // Here we clamp to 89 degrees, to prevent it intersecting with the world at 90, note, we
+        // don't use the `maxTiltAngle`, because then special handling needs to be applied when for
+        // example the tilt at the rotation center is more than that at the map target, then the
+        // tilt feels different at the center / offset positions. Using 89 gives the most natual
+        // feeling when applying the tilt angle.
+        const clampedDeltaTiltRotationTarget =
+            MathUtils.clamp(deltaTilt + currentTiltRotationCenter, 0, maxRotationTiltAngle) -
+            currentTiltRotationCenter;
```
atomicsulfate, line 389: after this: 
clampedDeltaTiltRotationTarget <= (89 - currentTiltRotationCenter)

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
+
+        // Math.min doesn't work, because we need to keep the sign of the delta as well.
         const clampedDeltaTilt =
-            MathUtils.clamp(deltaTilt + currentTilt, 0, maxTiltAngle) - currentTilt;
-        if (Math.abs(clampedDeltaTilt) <= Number.EPSILON) {
-            return;
+            Math.abs(clampedDeltaTiltMapTarget) < Math.abs(clampedDeltaTiltRotationTarget)
+                ? clampedDeltaTiltMapTarget
+                : clampedDeltaTiltRotationTarget;
```
atomicsulfate, line 395: after this:
clampedDeltaTilt <= clampedDeltaTiltRotationTarget <=   (89 - currentTiltRotationCenter)

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
+
+        // It can happen in some cases that the maxRotationTiltAngle is exceeded because the
+        // `clampedDeltaTilt` is taken from the smaller value clampedDeltaTiltMapTarget. We check
+        // that we don't exceed 0, because otherwise it will snap to 89, so we let the user ease
+        // back to 89, but not any further.
+        if (
+            currentTiltRotationCenter + clampedDeltaTilt >= maxRotationTiltAngle &&
+            clampedDeltaTilt > 0
```
atomicsulfate, line 403: as I mention above clampedDeltaTilt <= (89 - currentTiltRotationCenter), that means:

currentTiltRotationCenter + clampedDeltaTilt  <= 89

And therefore that condition in the if statement should always be false right?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
         if (mapTargetWorld === null) {
-            return;
+            return 0;
         }
-        const mapTargetNormal = projection.surfaceNormal(mapTargetWorld, new THREE.Vector3());
 
-        camera.updateMatrixWorld(true);
-        const dir = cache.vector3[1].setFromMatrixColumn(camera.matrixWorld, 2);
-        const up = cache.vector3[2].setFromMatrixColumn(camera.matrixWorld, 1);
+        let currentTilt = 0;
```
atomicsulfate, line 313: before doing anything we could check if deltaTilt is negative. If we are reducing the tilt, we don't really need to clamp it. Only if it's positive there's a possibility that we exceed the maximum values.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
+
+        // It can happen in some cases that the maxRotationTiltAngle is exceeded because the
+        // `clampedDeltaTilt` is taken from the smaller value clampedDeltaTiltMapTarget. We check
+        // that we don't exceed 0, because otherwise it will snap to 89, so we let the user ease
+        // back to 89, but not any further.
+        if (
+            currentTiltRotationCenter + clampedDeltaTilt >= maxRotationTiltAngle &&
+            clampedDeltaTilt > 0
+        ) {
+            return maxRotationTiltAngle - currentTiltRotationCenter;
```
atomicsulfate, line 405: If we return early for negative delta tilts as I mentioned at the top of the function, this part of the function could e simpler. You could use Math.min above and the last if shouldn't be needed.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
+        // projection this is just an approximation, in order to compute this exactly, we would
+        // need to shift the camera to the new position, compute the new normal at the map center
+        // and use this to then compute the angleBetweenNormals, it would then actually need to be
+        // done iteratively, because when shifting the camera, it is possible that we move it to a
+        // position which is outside the max tilt given the new normal. Because the `deltaTilt` is
+        // generally small, the difference in normals should be for most cases very small and the
+        // extra computation required to compute such a small value isn't worth it. The value we
+        // compute is anyway conservative, so we don't risk computing something that will put the
+        // camera outside the valid range.
+        const maxTilt = ninetyRad + angleBetweenNormals - maxRotationTargetAngle;
```
atomicsulfate, line 375: I'm having trouble understanding what you're doing here.
The last time we discussed we agreed on just limiting the tilt at the rotation center to 89 and at the map target to maxTilt. Did you a find a problem with that solution?

Anyway, I'm not sure which triangle you mention here. I think you mean the triangle CRT, being C = camera, R = rotation target, T = map target. If that's the case, the you make the following angle calculations:

angle at T (CTR) = 90 + currentTilt
angle at R (CRT) = 90 - currentTiltRotationCenter

I don't understand those calculations.

atomicsulfate (commented):
I'm having issues understanding parts of the code, see my comments.

@here/harp-mapview/lib/Utils.ts
nzjony: I will move.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,190 @@ export namespace MapViewUtils {
...
-        // Using the dot product of the direction vector and the surface normal alone does not give us the sign.
-        const tiltSign = Math.sign(up.dot(mapTargetNormal));
+            const dir = cache.vector3[1].setFromMatrixColumn(camera.matrixWorld, 2);
+            const up = cache.vector3[2].setFromMatrixColumn(camera.matrixWorld, 1);
 
-        const currentTilt = (tiltSign === 0 ? 1 : tiltSign) * dir.angleTo(mapTargetNormal);
+            // Due to inaccuracies it can happen that the tilt angle gets less than 0.
+            // Using the dot product of the direction vector and the surface normal alone does not give us the sign.
+            const tiltSign = Math.sign(up.dot(mapTargetNormal));
+            currentTilt = (tiltSign === 0 ? 1 : tiltSign) * dir.angleTo(mapTargetNormal);
```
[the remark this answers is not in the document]
nzjony: Good idea, FYI I see the check for tilt angle less than zero is already part of the function:

```
if (cosTheta > 1 - epsilon) {
            // Top down view.
            return 0;
        }
```

@here/harp-mapview/lib/Utils.ts
nzjony: Good point, will do. I only needed to clamp to ensure that if applying the deltaTilt would make the tilt -ve, to stop at 0 tilt.

@here/harp-mapview/lib/Utils.ts
nzjony: Yes, the other solution works in Planar, but I was getting issues with the sphere case, and I spent a few days trying to find an adequate solution, but never could, so I decided to write the algorithm again  from scratch.

nzjony: > and at the map target to maxTilt

It isn't actually limiting to the maxTilt, it just looks like it, we were applying the deltaTilt around the rotation target, and not the map target. As such it is easy to get the map target past the horizon, for example, if your rotation center is below the map target, then you have less tilt, say the map target has tilt 45, and the tilt at rotation center is 30, which gives us 15 more degrees of freedom until we reach the max tilt, hence it can easily hit the horizon (with both values < 89).

nzjony: I will reword using the terminology you suggested above.

nzjony:
> I'm having issues understanding parts of the code, see my comments.

Thanks for the review @atomicsulfate , I have addressed the comments, see my commit: https://github.com/heremaps/harp.gl/pull/1885/commits/84ee84e521f2ea441a3ae840f7f7fc5629995cfd

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,182 @@ export namespace MapViewUtils {
...
+     * the geometry with the near plane.
+     */
+    function computeClampedDeltaTilt(
+        mapView: MapView,
+        offsetX: number,
+        offsetY: number,
+        deltaTilt: number,
+        maxTiltAngle: number,
+        mapTargetWorld: THREE.Vector3,
+        rotationTargetWorld: THREE.Vector3
```
atomicsulfate, line 312: This function doesn't need the offsets anymore. It only needs to know whether the rotation target is the same as the map target. You can make rotationTargetWorld optional. If undefined, the rotation has mapTargetWorld as center.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,182 @@ export namespace MapViewUtils {
...
+        // Here we clamp to 89 degrees, to prevent it intersecting with the world at 90. This is
+        // possible for example when the R position is near the horizon. If the angle RCM is say 5
+        // degrees, then an angle of say 89 degrees at R, plus 5 degrees means the tilt at M would
+        // be 84 degrees. If the `maxTiltAngle` is however 89, then we still have what seems like 5
+        // degrees at M, so, without this check, we would go to 90 degrees.
+        const clampedDeltaTiltRotationTarget =
+            MathUtils.clamp(deltaTilt + rotationCenterTilt, 0, maxRotationTiltAngle) -
+            rotationCenterTilt;
+
+        // Math.min doesn't work, because we need to keep the sign of the delta as well.
```
atomicsulfate, line 386: Outdated comment.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,182 @@ export namespace MapViewUtils {
...
+            return -tilt;
+        } else if (deltaTilt <= 0) {
+            // Reducing the tilt isn't clamped (apart from above).
+            return deltaTilt;
+        }
+
+        const rotationCenterTilt =
+            offsetX === 0 && offsetX === offsetY
+                ? tilt
+                : extractTiltAngleFromLocation(projection, camera, rotationTargetWorld!);
```
atomicsulfate, line 329: if the rotation and map targets are the same point, most of the code after this is not needed anymore, i.e., there's no triangle anymore. We only need to check the tilt on the map target.

@here/harp-mapview/lib/Utils.ts
nzjony: I need them to check whether the points are the same or not, I thought it would be better to check the offset rather than do an exact check, i.e. `mapTargetWorld.equals(rotationTargetWorld)`, because of potential rounding issues? Though I guess they would be the same if the offsets are the same, so maybe I should just go with the `equals`

@here/harp-mapview/lib/Utils.ts
nzjony: Good point, thanks.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,182 @@ export namespace MapViewUtils {
...
+        mapTargetWorld: THREE.Vector3,
+        rotationTargetWorld: THREE.Vector3
+    ): number {
+        const camera = mapView.camera;
+        const projection = mapView.projection;
+
+        const tilt = extractTiltAngleFromLocation(projection, camera, mapTargetWorld);
+        if (tilt + deltaTilt < 0) {
+            // Clamp the final tilt to 0
+            return -tilt;
```
atomicsulfate, line 320: Shouldn't we return 0 here then? 
We should also check later than if we return 0, we don't call applyTiltAroundTarget later, or applyTiltAroundTarget returns early.

nzjony: This function returns a delta, so if we have current tilt as `tilt`, then returning `-tilt` will result in a final tilt value of 0.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,182 @@ export namespace MapViewUtils {
...
+        // degrees at M, so, without this check, we would go to 90 degrees.
+        const clampedDeltaTiltRotationTarget =
+            MathUtils.clamp(deltaTilt + rotationCenterTilt, 0, maxRotationTiltAngle) -
+            rotationCenterTilt;
+
+        // Math.min doesn't work, because we need to keep the sign of the delta as well.
+        const clampedDeltaTilt = Math.min(
+            clampedDeltaTiltMapTarget,
+            clampedDeltaTiltRotationTarget
+        );
```
atomicsulfate, line 391: Here you're doing:

clampedDeltaTilt = min(clampedDeltaTiltM,clampedDeltaTiltR) = 
min( clamp(deltaTilt + rotCenterTilt, 0, maxTilt) - rotCenterTilt, clamp(deltaTilt + rotCenterTilt, 0, maxRotTiltAngle) - rotationCenterTilt ) =
min( clamp(deltaTilt + rotCenterTilt, 0, maxTilt) , clamp(deltaTilt + rotCenterTilt, 0, maxRotTiltAngle) ) - rotCenterTilt

So this can be simplified to:

min( clamp(deltaTilt + rotCenterTilt, 0, min(maxTilt,maxRotTiltAngle)) - rotCenterTilt

Is this this intended? Does it make sense to compute clampedDeltaTiltMapTarget using rotationCenterTilt ?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,182 @@ export namespace MapViewUtils {
...
+
+        // It can happen in some cases that the maxRotationTiltAngle is exceeded because the
+        // `clampedDeltaTilt` is taken from the smaller value clampedDeltaTiltMapTarget. We check
+        // that we don't exceed 0, because otherwise it will snap to 89, so we let the user ease
+        // back to 89, but not any further.
+        if (rotationCenterTilt + clampedDeltaTilt >= maxRotationTiltAngle) {
```
atomicsulfate, line 397: Continuing from my last comment, this if condition would never be true:

rotationCenterTilt + clampedDeltaTilt =
rotationCenterTilt + clamp(deltaTilt + rotCenterTilt, 0, min(maxTilt,maxRotTiltAngle)) - rotCenterTilt =
clamp(deltaTilt + rotCenterTilt, 0, min(maxTilt,maxRotTiltAngle)) <= maxRotTiltAngle

@here/harp-mapview/lib/Utils.ts
nzjony: Nice, I hadn't considered to simplify this, but it is a good idea.

> Does it make sense to compute clampedDeltaTiltMapTarget using rotationCenterTilt ?

Yes, the `clampedDeltaTiltMapTarget` is named a bit incorrectly, it is the delta tilt clamped that, when applied to the rotation center, would give a correctly clamped value at `mapTargetWorld`

@here/harp-mapview/lib/Utils.ts
nzjony: Yes, good point, I need to however still detect this condition when `rotationCenterTilt > maxRotationTiltAngle`, because then I want to return 0, otherwise the `clampedDeltaTilt` is negative and it will snap back to the 89 degrees at the rotation center world, which is a bit unnatural. Here, we should let the user not increase the `rotationCenterTilt`, but not snap it back either, I will add a condition above to prevent this case.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,182 @@ export namespace MapViewUtils {
...
+        // R = rotation target, i.e. the point about which we are rotating: `rotationTargetWorld`
+        // C = camera position
+        // M = map target, i.e. the point which the camera is looking at: `mapTargetWorld`
+        // The following variable `angleBetweenRotationTargetAndMapTarget` is the angle between the
+        // two rays C->R and C->M, i.e. the angle RCM. This angle remains constant when tilting with
+        // a fixed `offsetX` and `offsetY`. It is calculated by using the intersection of the two
+        // ways with the earth. Where:
+        // MRC is 90 + `angleBetweenNormals` - tilt at R.
+        // CMR is 90 + `tilt`
+        // RCM is 180 - CRM - RMC, aka `angleBetweenRotationTargetAndMapTarget`
```
atomicsulfate, line 365: These formulas  don't seem to work. See a counterexample:
https://www.geogebra.org/classic/grpzmz9v

atomicsulfate (changes requested):
See my comments, I think the triangle calculation you're doing is wrong.

@here/harp-mapview/lib/Utils.ts
atomicsulfate: Ok, but if rotationCenterTilt > maxRotationTiltAngle, it means you can return 0 deltaTilt much sooner in the function right?

@here/harp-mapview/lib/Utils.ts
nzjony: Very cool demo, I will have to remember this tool for next time.

I see the problem, I solved the problem only for R above M, i.e. `offsetY` > 0.

I will fix, thanks a lot for spotting.

nzjony: Thinking about this a bit more, if `offsetY` is < 0, then I can just return maxTiltAngle, because then the tilt is guaranteed to be less at M.

@here/harp-mapview/lib/Utils.ts
nzjony: Yeah, check my latest commit, it should be in there.

@here/harp-mapview/lib/Utils.ts
nzjony: I added a comment to help those that may arrive at the same conclusion you did.

nzjony:

nzjony:

@here/harp-mapview/lib/Utils.ts
```diff
@@ -317,8 +319,9 @@ export namespace MapViewUtils {
         } else if (deltaTilt <= 0) {
             // Reducing the tilt isn't clamped (apart from above).
             return deltaTilt;
-        } else if (mapTargetWorld.equals(rotationTargetWorld)) {
-            // When the rotation target is the center, then we have a simple formula
+        } else if (mapTargetWorld.equals(rotationTargetWorld) || offsetY < 0) {
+            // When the rotation target is the center, or the offsetY is < 0, i.e. the angle at the
+            // `mapTargetWorld` is always bigger, then we have a simple formula
```
atomicsulfate, line 324: >or the offsetY is < 0, i.e. the angle at the
// `mapTargetWorld` is always bigger, then we have a simple formula

That's not necessarily true for globe. Imaging you see the whole globe with tilt 0 at the target, and you pick a rotation center below the target near the horizon.

@here/harp-mapview/lib/Utils.ts
atomicsulfate: > Thinking about this a bit more, if `offsetY` is < 0, then I can just return maxTiltAngle, because then the tilt is guaranteed to be less at M.

Besides the issue that this doesn't work for globe (see my new comment), I don't understand why this change fixes my review comment. Isn't the triangle calculation still wrong? I don't see why it would hold for offsetY > 0.

@here/harp-mapview/lib/Utils.ts
nzjony: In that case it has -ve tilt (relative to the rotation target)

@here/harp-mapview/lib/Utils.ts
```diff
@@ -1319,21 +1459,27 @@ export namespace MapViewUtils {
...
             zAxis: tangentSpace.z,
             position: cache.vector3[0]
         });
-
         // Get point to object vector (dirVec) and compute the `tilt` as the angle with tangent Z.
-        const dirVec = cache.vector3[1].copy(object.position).sub(cache.vector3[0]);
+        const dirVec = cache.vector3[2].copy(object.position).sub(cache.vector3[0]);
+        if (tiltAxis) {
+            dirVec.projectOnPlane(tiltAxis);
+        }
```
atomicsulfate, line 1482: I see two problems with this:
- In planar, since you're projecting the orbiting center to the "tilt plane", the computed tilt will be lower than the tilt of the camera wrt to the real orbiting center. SInce you underestimate the real tilt, you'll let the camera reach tilts higher than the limit of 89 degrees.
- In sphere, I don't understand the meaning of the tilt value you compute, since you mix the normal at the real rotation center with the direction vector to the projected center on the tilt plane.

atomicsulfate (commented):
I don't have yet a clear idea of what should be the behaviour of tilting wrt an arbitrary point. If you do, we can go on with the code review. Otherwise, maybe we could go back to the drawing board, think it through, and later arrange a call together and discuss. What do you say?

@here/harp-mapview/lib/Utils.ts
nzjony: @atomicsulfate ,

The tilt of the camera wrt to the real orbiting center is anyway clamped to 89 degrees in the second step.

```
const clampedDeltaTilt =
            MathUtils.clamp(
                deltaTilt + rotationCenterTilt,
                0,
                Math.min(maxTilt, maxRotationTiltAngle)
            ) - rotationCenterTilt;
```

For the spheres, yes, I need to also project the normal. Good spotting.

nzjony:
> I don't have yet a clear idea of what should be the behaviour of tilting wrt an arbitrary point. If you do, we can go on with the code review. Otherwise, maybe we could go back to the drawing board, think it through, and later arrange a call together and discuss. What do you say?

Functionally, the most important things are that the map target never exceeds the tilt limit and that the camera never reaches 90 and intersects the world.

The actual behaviour wrt to the arbitrary point can be chosen many ways and it will have different impact on the user experience. By projecting and computing the max tilt limit from the plane perpendicular to the world, it is IMO the most intuitive, because it makes the computation simpler and it also gives the camera the most range of freedom. But yes, if you have more concerns, we can chat for sure.

@here/harp-mapview/lib/Utils.ts
atomicsulfate: > @atomicsulfate ,
> 
> The tilt of the camera wrt to the real orbiting center is anyway clamped to 89 degrees in the second step.
> 
> ```
> const clampedDeltaTilt =
>             MathUtils.clamp(
>                 deltaTilt + rotationCenterTilt,
>                 0,
>                 Math.min(maxTilt, maxRotationTiltAngle)
>             ) - rotationCenterTilt;
> ```

I've added some logs and I've confirmed that the tilt at the projected rotation center is smaller than that at the original rotation center. In planar the difference is small, hence the original issue cannot be reproduced.
In Sphere the difference can be bigger and therefore you can put the rotation center beyond the horizon.
It's a minor thing, I guess we can leave it like that.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -1319,21 +1461,27 @@ export namespace MapViewUtils {
...
             yAxis: tangentSpace.y,
             zAxis: tangentSpace.z,
             position: cache.vector3[0]
         });
-
         // Get point to object vector (dirVec) and compute the `tilt` as the angle with tangent Z.
-        const dirVec = cache.vector3[1].copy(object.position).sub(cache.vector3[0]);
+        const dirVec = cache.vector3[2].copy(object.position).sub(cache.vector3[0]);
+        if (tiltAxis) {
+            dirVec.projectOnPlane(tiltAxis);
```
atomicsulfate, line 1483: You're still using the normal at the rotation center to compute the tilt, that is, tangentSpace.z. You should project that on the plane as well, as you do here:
https://github.com/heremaps/harp.gl/pull/1885/files#diff-305bd59a411397d3c92eafbdc5c800c3bb6f0bb0040b12d5bfc6c411cdda288fR353-R354

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,194 @@ export namespace MapViewUtils {
...
 
-        const posBackup = cache.vector3[1].copy(camera.position);
-        const quatBackup = cache.quaternions[1].copy(camera.quaternion);
+        // This is used to find the max tilt angle, because the difference in normals is needed
+        // to correct the triangle used to find the max tilt angle at the rotation center.
+        let angleBetweenNormals = 0;
+        if (projection === sphereProjection) {
+            const projectedRotationTargetNormal = projection
+                .surfaceNormal(rotationTargetWorld, cache.vector3[0])
+                .projectOnPlane(tiltAxis);
```
atomicsulfate, line 354: Here you're taking the surface normal at the rotation target and projecting it to the tilt plane. Instead, you should project the rotation target on the tilt plane, and the get the surface normal at that location.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,194 @@ export namespace MapViewUtils {
...
+        // two rays C->Rp and C->M, i.e. the angle RpCM. This angle remains constant when tilting with
+        // a fixed `offsetX` and `offsetY`. It is calculated by using the intersection of the two
+        // ways with the earth. Where:
+        // MRpC is 90 + `angleBetweenNormals` - tilt at Rp.
+        // CMRp is 90 + `tilt`
+        // RpCM is 180 - CRpM - RpMC, aka `angleBetweenRotationTargetAndMapTarget`
+        // Note the use of `angleBetweenNormals` to ensure this works for spherical projections.
+        // Note, also, the above calculation only works when the tilt at the M is less than the tilt
+        // at Rp, otherwise the above formula won't work. We however don't need to worry about this
+        // case because this happens only when offsetY is less than zero, and this is handled above.
```
atomicsulfate, line 378: Could you add some simple ascii art of the triangle. That would be make it clearer and probably you could then reduce the textual explanation.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,194 @@ export namespace MapViewUtils {
...
+        const angleBetweenRotationTargetAndMapTarget =
+            ninetyRad * 2 -
+            (ninetyRad + angleBetweenNormals - rotationCenterTilt + ninetyRad + tilt);
```
atomicsulfate, line 381: Instead of explaining in comments above how to compute the angles, just write constants with the angle names. I rather see self-explanatory code than lengthy comments. For example:

```suggestion
        const MRC = 90 + angleBetweenNormals - tilt;
        const CMR = 90 + tilt;
        const RCM = 180 - MRC - CMR;
```

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,194 @@ export namespace MapViewUtils {
...
+
+        // This uses the same formula as above, just we change `tilt` to the maxTiltAngle,
+        // i.e we want to find the greatest angle at the rotation target that gives us the max
+        // angle at the map center target.
+        // i.e. CMR is 90 + `maxTiltAngle`
+        const maxRotationTargetAngle =
+            ninetyRad * 2 - angleBetweenRotationTargetAndMapTarget - ninetyRad - maxTiltAngle;
```
atomicsulfate, line 388:
```suggestion
       const maxCMR = 90 + maxTiltAngle;
       const maxMRC = 180 - RCM - maxCMR;
```

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,194 @@ export namespace MapViewUtils {
...
+        // difference in normals is reduced to some epsilon. I don't apply this because it is
+        // computationally expensive and the user would never notice this in practice.
+        const maxTilt = ninetyRad + angleBetweenNormals - maxRotationTargetAngle;
+
+        // Here we clamp to the min of `maxTilt` and 89 degrees. The check for 89 is to prevent it
+        // intersecting with the world at 90. This is possible for example when the R position is
+        // near the horizon. If the angle RCM is say 5 degrees, then an angle of say 89 degrees at
+        // R, plus 5 degrees means the tilt at M would be 84 degrees. If the `maxTiltAngle` is
+        // however 89, then we still have 5 degrees at M, so, without this check, we would go to 90
+        // degrees.
```
atomicsulfate, line 404: Could you try making these comments a little bit more concise?

atomicsulfate (commented):
I'm finished with the review.

@here/harp-mapview/lib/Utils.ts
```diff
@@ -242,53 +243,194 @@ export namespace MapViewUtils {
...
+    ) {
+        const camera = mapView.camera;
+        // Consider to use the cache if necessary, but beware, because the `rayCastWorldCoordinates`
+        // also uses this cache.
+        const posBackup = camera.position.clone();
+        const quatBackup = camera.quaternion.clone();
 
-        const tiltAxis = cache.vector3[0].set(1, 0, 0).applyQuaternion(camera.quaternion);
-        const tiltQuat = cache.quaternions[0].setFromAxisAngle(tiltAxis, clampedDeltaTilt);
+        const tiltQuat = cache.quaternions[0].setFromAxisAngle(tiltAxis, deltaTilt);
```
atomicsulfate, line 433: Just a comment in case we come back to this change in the future. I thought it makes more sense to use as tilt axis the cross product of the up camera vector with the direction vector from the camera to the rotation center.
I've tried it and it works (it keeps the rotation center in place), the only problem is that it rolls the camera. Maybe we could investigate how to undo or avoid this roll.

nzjony: Please make a ticket.

@here/harp-mapview/lib/Utils.ts
nzjony: Why? The surface normal on the plane and the projected surface normal to the plane have the same direction.

@here/harp-mapview/lib/Utils.ts
nzjony: I'm not a fan of ascii art, as it is difficult to maintain / create, I prefer to describe the problem in text. If you want, I will create a diagram and put in the readme.md.

@here/harp-mapview/lib/Utils.ts
nzjony: Agreed, will do.

@here/harp-mapview/lib/Utils.ts
nzjony: Please highlight which part of the text is superfluous.

@here/harp-mapview/lib/Utils.ts
nzjony: The projected rotation center isn't enough, because it will be under the earths surface, I will need to also adjust the length.

I will just project the normal to the tilt axis as well.

@here/harp-mapview/lib/Utils.ts
nzjony: Yes, what matters is that the camera doesn't intersect the ground and that the camera behaviours in both modes behaviour similiarly. We can restrict the camera yes, but then the user will get some different tilting behaviour based on the rotation point. I think it is most intuitive to the user to let the tilt range not depend on the rotation center, unless the projected rotation center will cause the camera to intersect the ground.

@here/harp-mapview/lib/Utils.ts
nzjony: But... forgot to normalize()

@here/harp-mapview/lib/Utils.ts
atomicsulfate: nevermind, it's ok.

@here/harp-mapview/lib/Utils.ts
atomicsulfate: so far we don't need to do anything, as long as this works right. Not sure we need a ticket, it was just a comment for reference.

@here/harp-mapview/lib/Utils.ts
atomicsulfate: I don't like ascii art either, but when it's something simple it makes things much clearer than text. Anyway, it's fine.

@here/harp-mapview/lib/Utils.ts
nzjony: I made a ticket: HARP-12600

@here/harp-mapview/lib/Utils.ts
nzjony: I made some changes anyway with the changes you mentioned, so it should be a bit better now.

atomicsulfate (approved):

PR #1888, opened 2020-10-07
HARP-11899: Cull only labels that are completely off screen

StefanDachwitz:
* cull using near/far plane
* render partial text and icons near border

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/ScreenProjector.ts
+++ b/@here/harp-mapview/lib/ScreenProjector.ts
@@ -16,6 +16,15 @@ function isOnScreen(ndc: THREE.Vector3) {
 }
 
 /**
+ * Determines whether a position in NDC (Normalized Device Coordinates) is between the near
+ * and far plane.
+ * @param ndc - The position to check.
+ */
+function isInRange(ndc: THREE.Vector3) {
+    return ndc.z > -1 && ndc.z < 1;
+}
+
+/**
  * @hidden
  * Handles the projection of world coordinates to screen coordinates.
  */
@@ -70,14 +79,14 @@ export class ScreenProjector {
      * @param {(Vector3Like)} source The source vector to project.
      * @param {THREE.Vector2} target The target vector.
      * @returns {THREE.Vector2} The projected vector (the parameter 'target') or undefined if
-     * outside the screen.
+     * outside of the near/far plane. The point may be outside the screen.
      */
-    projectOnScreen(
+    projectToScreen(
         source: Vector3Like,
         target: THREE.Vector2 = new THREE.Vector2()
     ): THREE.Vector2 | undefined {
         const p = this.projectVector(source, ScreenProjector.tempV3);
-        if (isOnScreen(p)) {
+        if (isInRange(p)) {
             return this.ndcToScreen(p, target);
         }
         return undefined;
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -348,6 +348,7 @@ const tmpCollisionBox = new CollisionBox();
 const tmpScreenPosition = new THREE.Vector2();
 const tmpTextOffset = new THREE.Vector2();
 const tmp2DBox = new Math2D.Box();
+const tmp2DBox2 = new Math2D.Box();
 const tmpCenter = new THREE.Vector2();
 const tmpSize = new THREE.Vector2();
 
@@ -398,10 +399,7 @@ export function placeIcon(
     const iconSpaceAvailable =
         poiInfo.mayOverlap === true || !screenCollisions.isAllocated(tmp2DBox);
 
-    if (!iconSpaceAvailable) {
-        return iconRenderState.isVisible() ? PlacementResult.Rejected : PlacementResult.Invisible;
-    }
-    return PlacementResult.Ok;
+    return !iconSpaceAvailable ? PlacementResult.Rejected : PlacementResult.Ok;
 }
 
 /**
@@ -421,10 +419,8 @@ export function placeIcon(
  * @param multiAnchor - The parameter decides if multi-anchor placement algorithm should be
  * used, be default [[false]] meaning try to place label using current alignment settings only.
  * @returns `PlacementResult.Ok` if point __label can be placed__ at the base or any optional
- * anchor point. `PlacementResult.Rejected` if there's a collision for all placements or it's
- * __persistent label with icon rejected and text visible__. Finally `PlacementResult.Invisible`
- * if it's text is not visible at any placement position or it's __new label with text or icon__
- * __rejected__.
+ * anchor point. `PlacementResult.Rejected` if there's a collision for all placements. Finally
+ * `PlacementResult.Invisible` if it's text is not visible at any placement position.
  */
 export function placePointLabel(
     labelState: TextElementState,
@@ -442,10 +438,9 @@ export function placePointLabel(
     const layoutStyle = labelState.element.layoutStyle!;
 
     // For the new labels with rejected icons we don't need to go further.
-    // Make them invisible.
     const newLabel = !labelState.visible;
     if (isRejected && newLabel) {
-        return PlacementResult.Invisible;
+        return PlacementResult.Rejected;
     }
     // Check if alternative placements have been provided.
     multiAnchor =
@@ -697,6 +692,15 @@ function placePointLabelAtAnchor(
     // Compute text offset from the anchor point
     const textOffset = computePointTextOffset(label, placement, scale, env, tmpTextOffset);
     textOffset.add(screenPosition);
+
+    // Update output screen position.
+    outScreenPosition.set(textOffset.x, textOffset.y, labelState.renderDistance);
+
+    // Check if icon's label was already rejected.
+    if (isRejected) {
+        return PlacementResult.Rejected;
+    }
+
     tmpBox.copy(label.bounds!);
     // Apply additional persistent margin, keep in mind that text bounds just calculated
     // are not (0, 0, w, h) based, so their coords usually are also non-zero.
@@ -710,28 +714,13 @@ function placePointLabelAtAnchor(
     tmpSize.multiplyScalar(scale);
     tmp2DBox.set(tmpCenter.x - tmpSize.x / 2, tmpCenter.y - tmpSize.y / 2, tmpSize.x, tmpSize.y);
 
-    // Update output screen position.
-    outScreenPosition.set(textOffset.x, textOffset.y, labelState.renderDistance);
-
-    // Check the text visibility if invisible finish immediately
-    // regardless of the persistence state - no fading required.
-    if (!screenCollisions.isVisible(tmp2DBox)) {
-        return PlacementResult.Invisible;
-    }
-
-    const persistent = labelState.visible;
-    // Check if icon's label was already rejected.
-    if (isRejected) {
-        // Allows to fade out persistent label and simply ignore new one.
-        // NOTE:
-        // It might be changed if we would like to render text without icon (at border, etc.).
-        return persistent ? PlacementResult.Rejected : PlacementResult.Invisible;
-    }
+    // Save original bounds for visibility check
+    tmp2DBox2.copy(tmp2DBox);
 
     if (measureText) {
-        // Up-scaled label bounds are used only for new labels and after visibility check, this is
-        // intentional to avoid processing labels out of the screen due to increased bounds, such
-        // labels would be again invisible in the next frame.
+        // Up-scaled label bounds are used only for new labels and only for collision check, this
+        // is intentional to avoid processing labels out of the screen due to increased bounds,
+        // such labels would be again invisible in the next frame.
         tmpBox.getSize(tmpSize);
         tmpSize.multiplyScalar(scale * (1 + newPointLabelTextMarginPercent));
         tmp2DBox.set(
@@ -742,33 +731,39 @@ function placePointLabelAtAnchor(
         );
     }
 
-    // Check label's text collision.
+    // Check label's text collision. Collision is more important than visibility (now), because for
+    // icon/text combinations the icon should be rendered if the text is out of bounds, but it may
+    // _not_ be rendered if the text is colliding with another label.
     if (!label.textMayOverlap && screenCollisions.isAllocated(tmp2DBox)) {
-        // Allows to fade persistent and ignore new label.
-        return persistent ? PlacementResult.Rejected : PlacementResult.Invisible;
+        return PlacementResult.Rejected;
     }
 
-    // Don't allocate space for rejected text. When zooming, this allows placement of a
-    // lower priority text element that was displaced by a higher priority one (not
-    // present in the new zoom level) before an even lower priority one takes the space.
-    // Otherwise the lowest priority text will fade in and back out.
-    // TODO: Add a unit test for this scenario.
-    if (label.textReservesSpace) {
-        screenCollisions.allocate(tmp2DBox);
+    // Visibility check with non-enlarged bounds
+    if (screenCollisions.isVisible(tmp2DBox2)) {
+        // Don't allocate space for rejected text. When zooming, this allows placement of a
+        // lower priority text element that was displaced by a higher priority one (not
+        // present in the new zoom level) before an even lower priority one takes the space.
+        // Otherwise the lowest priority text will fade in and back out.
+        // TODO: Add a unit test for this scenario.
+        if (label.textReservesSpace) {
+            screenCollisions.allocate(tmp2DBox);
+        }
+
+        // Glyphs arrangement have been changed remove text buffer object which needs to be
+        // re-created.
+        if (measureText) {
+            label.textBufferObject = undefined;
+        }
+
+        // Save current placement in label state.
+        // TextElementState creates layout snapshot solely for alternative placements which saves
+        // memory that could be wasted on unnecessary objects construction.
+        labelState.textPlacement = placement;
+
+        return PlacementResult.Ok;
     }
 
-    // Glyphs arrangement have been changed remove text buffer object which needs to be
-    // re-created.
-    if (measureText) {
-        label.textBufferObject = undefined;
-    }
-
-    // Save current placement in label state.
-    // TextElementState creates layout snapshot solely for alternative placements which saves
-    // memory that could be wasted on unnecessary objects construction.
-    labelState.textPlacement = placement;
-
-    return PlacementResult.Ok;
+    return PlacementResult.Invisible;
 }
 
 /**
@@ -828,8 +823,6 @@ export function placePathLabel(
         checkGlyphCollision = candidateBoxes.length > 0;
     }
 
-    const checkGlyphVisible = !screenCollisions.isFullyVisible(tmpCollisionBox);
-
     // Perform per-character collision checks.
     tmpCollisionBoxes.length = tmpBoxes.length;
     for (let i = 0; i < tmpBoxes.length; ++i) {
@@ -840,9 +833,6 @@ export function placePathLabel(
             tmpCollisionBoxes[i] = collisionBox;
         } else {
             collisionBox.copy(glyphBox);
-        }
-        if (checkGlyphVisible && !screenCollisions.isVisible(collisionBox)) {
-            return PlacementResult.Invisible;
         }
 
         if (
@@ -884,7 +874,7 @@ export function isPathLabelTooSmall(
         // Skip invisible points at the beginning of the path.
         const screenPoint = anyPointVisible
             ? screenProjector.project(pt, tmpScreenPosition)
-            : screenProjector.projectOnScreen(pt, tmpScreenPosition);
+            : screenProjector.projectToScreen(pt, tmpScreenPosition);
         if (screenPoint === undefined) {
             continue;
         }
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -454,7 +454,6 @@ export function placePointLabel(
             textCanvas,
             env,
             screenCollisions,
-            isRejected,
             outScreenPosition
         );
     }
@@ -536,7 +535,6 @@ function placePointLabelChoosingAnchor(
             textCanvas,
             env,
             screenCollisions,
-            false,
             !isLastPlacement,
             outScreenPosition
         );
@@ -609,7 +607,6 @@ function placePointLabelAtCurrentAnchor(
     textCanvas: TextCanvas,
     env: Env,
     screenCollisions: ScreenCollisions,
-    isRejected: boolean,
     outScreenPosition: THREE.Vector3
 ): PlacementResult {
     assert(labelState.element.layoutStyle !== undefined);
@@ -624,7 +621,6 @@ function placePointLabelAtCurrentAnchor(
         textCanvas,
         env,
         screenCollisions,
-        isRejected,
         !labelState.visible,
         outScreenPosition
     );
@@ -663,7 +659,6 @@ function placePointLabelAtAnchor(
     textCanvas: TextCanvas,
     env: Env,
     screenCollisions: ScreenCollisions,
-    isRejected: boolean,
     forceInvalidation: boolean,
     outScreenPosition: THREE.Vector3
 ): PlacementResult {
@@ -696,11 +691,6 @@ function placePointLabelAtAnchor(
     // Update output screen position.
     outScreenPosition.set(textOffset.x, textOffset.y, labelState.renderDistance);
 
-    // Check if icon's label was already rejected.
-    if (isRejected) {
-        return PlacementResult.Rejected;
-    }
-
     tmpBox.copy(label.bounds!);
     // Apply additional persistent margin, keep in mind that text bounds just calculated
     // are not (0, 0, w, h) based, so their coords usually are also non-zero.
@@ -713,6 +703,12 @@ function placePointLabelAtAnchor(
 
     tmpSize.multiplyScalar(scale);
     tmp2DBox.set(tmpCenter.x - tmpSize.x / 2, tmpCenter.y - tmpSize.y / 2, tmpSize.x, tmpSize.y);
+
+    // Check the text visibility if invisible finish immediately
+    // regardless of the persistence state - no fading required.
+    if (!screenCollisions.isVisible(tmp2DBox)) {
+        return PlacementResult.Invisible;
+    }
 
     // Save original bounds for visibility check
     tmp2DBox2.copy(tmp2DBox);
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -428,7 +428,6 @@ export function placePointLabel(
     textCanvas: TextCanvas,
     env: Env,
     screenCollisions: ScreenCollisions,
-    isRejected: boolean,
     outScreenPosition: THREE.Vector3,
     multiAnchor: boolean = false
 ): PlacementResult {
@@ -436,16 +435,11 @@ export function placePointLabel(
 
     const layoutStyle = labelState.element.layoutStyle!;
 
-    // For the new labels with rejected icons we don't need to go further.
-    const newLabel = !labelState.visible;
-    if (isRejected && newLabel) {
-        return PlacementResult.Rejected;
-    }
     // Check if alternative placements have been provided.
     multiAnchor =
         multiAnchor && layoutStyle.placements !== undefined && layoutStyle.placements.length > 1;
     // For single placement labels or labels with icon rejected, do only current anchor testing.
-    if (!multiAnchor || isRejected) {
+    if (!multiAnchor) {
         return placePointLabelAtCurrentAnchor(
             labelState,
             screenPosition,
--- a/@here/harp-mapview/lib/text/Placement.ts
+++ b/@here/harp-mapview/lib/text/Placement.ts
@@ -348,7 +348,6 @@ const tmpCollisionBox = new CollisionBox();
 const tmpScreenPosition = new THREE.Vector2();
 const tmpTextOffset = new THREE.Vector2();
 const tmp2DBox = new Math2D.Box();
-const tmp2DBox2 = new Math2D.Box();
 const tmpCenter = new THREE.Vector2();
 const tmpSize = new THREE.Vector2();
 
@@ -710,9 +709,6 @@ function placePointLabelAtAnchor(
         return PlacementResult.Invisible;
     }
 
-    // Save original bounds for visibility check
-    tmp2DBox2.copy(tmp2DBox);
-
     if (measureText) {
         // Up-scaled label bounds are used only for new labels and only for collision check, this
         // is intentional to avoid processing labels out of the screen due to increased bounds,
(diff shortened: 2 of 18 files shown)


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -2018,4 +2038,20 @@ export class TextElementsRenderer {
...
+     * Project point to screen and check if it is on screen or within a fixed distance to the border.
+     *
+     * @param point center point of label.
+     * @param outPoint projected screen point of label.
+     */
+    private labelPotentiallyVisible(point: Vector3Like, outPoint: THREE.Vector2): boolean {
+        const projectionResult = this.m_screenProjector.projectAreaToScreen(
+            point,
+            MAX_HORIZONTAL_DISTANCE_TO_BORDER * 2 * this.m_pixelToNDC,
+            MAX_VERTICAL_DISTANCE_TO_BORDER * 2 * this.m_pixelToNDC,
```
atomicnitrate, line 2052: This conversion to NDC can be done in ScreenProjector, it already has the screen size in pixels.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2777,6 +2777,30 @@ export class MapView extends EventDispatcher {
...
+            const pixelRatio = this.m_renderer.getPixelRatio();
+            return {
+                width: Math.round(this.canvas.width / pixelRatio),
+                height: Math.round(this.canvas.height / pixelRatio)
+            };
+        } else {
+            return { width: clientWidth, height: clientHeight };
+        }
+    }
+
```
atomicnitrate, line 2803: No need to make it public if the pixel to NDC conversion is done in ScreenProjector.

@here/harp-mapview/lib/ScreenProjector.ts
```diff
@@ -70,14 +79,43 @@ export class ScreenProjector {
...
         target: THREE.Vector2 = new THREE.Vector2()
     ): THREE.Vector2 | undefined {
         const p = this.projectVector(source, ScreenProjector.tempV3);
-        if (isOnScreen(p)) {
+        if (
+            isInRange(p) &&
+            p.x + width * 0.5 >= -1 &&
+            p.x - width * 0.5 <= 1 &&
+            p.y + height * 0.5 >= -1 &&
+            p.y - height * 0.5 <= 1
```
atomicnitrate, line 117:
```suggestion
            p.x + width * 0.5 <= 1 &&
            p.x - width * 0.5 >= -1 &&
            p.y + height * 0.5 <= 1 &&
            p.y - height * 0.5 >= -1
```
Better check why the unit tests didn't catch this.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -122,6 +122,13 @@ const OVERLOAD_UPDATE_TIME_LIMIT = 5;
  */
 const OVERLOAD_PLACE_TIME_LIMIT = 10;
 
+// Allowed number of pixels outside the screen to test for placement.
+// A value of 0 will lead to labels popping in at the border of the screen. A large value will
+// lead to many labels being placed outside the screen, with all the required actions for measuring
+// and loading glyphs.
+const MAX_HORIZONTAL_DISTANCE_TO_BORDER = 100;
+const MAX_VERTICAL_DISTANCE_TO_BORDER = 40;
```
atomicnitrate, line 130: It's very difficult these absolute thresholds are gonna fit all cases: different fonts, font sizes, label lengths, clients...
As in harp native, I would put this as a styling property, so that it can be adjusted on a case by case basis.
I would also use thresholds that are relative to the total label size. Harp native has a "minVisibility" factor [0,1], representing how much area of the label must be inside the screen before it's made visible.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1756,10 +1780,9 @@ export class TextElementsRenderer {
...
             this.m_tmpVector3
         );
-        // Only process labels frustum-clipped labels
-        if (
-            this.m_screenProjector.projectOnScreen(worldPosition, tempScreenPosition) === undefined
-        ) {
+
+        // Only process labels that are potentially within the frustum.
+        if (!this.labelPotentiallyVisible(worldPosition, tempScreenPosition)) {
             return false;
```
atomicnitrate, line 1786: This is too soon to do check label bound visibility. The label has not been yet placed, therefore you don't even know the alignment wrt the anchor point. You're assuming the label is centered but it could be left-aligned, and then it would be skipped even being fully inside the screen.

Here we can only check if it's within the far distance range. The other checks have to be done after placement.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1665,22 +1678,27 @@ export class TextElementsRenderer {
                 iconIndex === undefined
             );
             if (placeResult === PlacementResult.Invisible) {
-                if (placementStats) {
+                if (placementStats !== undefined) {
```
atomicnitrate, line 1681: It's fine as it was. Our coding style guidelines favor implicit conversion to boolean for object types.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1812,10 +1835,9 @@ export class TextElementsRenderer {
         if (minDistanceSqr > 0 && shieldGroup !== undefined) {
             for (let pointIndex = 0; pointIndex < path.length; ++pointIndex) {
                 const point = path[pointIndex];
-                // Only process labels frustum-clipped labels
-                if (
-                    this.m_screenProjector.projectOnScreen(point, tempScreenPosition) !== undefined
-                ) {
+
+                // Only process potentially visible labels
+                if (this.labelPotentiallyVisible(point, tempScreenPosition)) {
```
atomicnitrate, line 1840: same here, label is not yet placed, we don't know its alignment.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1855,10 +1877,8 @@ export class TextElementsRenderer {
         else {
             for (let pointIndex = 0; pointIndex < path.length; ++pointIndex) {
                 const point = path[pointIndex];
-                // Only process labels frustum-clipped labels
-                if (
-                    this.m_screenProjector.projectOnScreen(point, tempScreenPosition) !== undefined
-                ) {
+                // Only process potentially visible labels
+                if (this.labelPotentiallyVisible(point, tempScreenPosition)) {
```
atomicnitrate, line 1881: same here, label is not yet placed, we don't know its alignment.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1627,17 +1638,19 @@ export class TextElementsRenderer {
                 this.m_screenCollisions
             );
             if (result === PlacementResult.Invisible) {
+                // Reset the icon and keep it from popping up before fading in.
+                iconInvisible = true;
                 iconRenderState.reset();
-
-                if (placementStats) {
+                iconRenderState.opacity = 0;
```
atomicnitrate, line 1644: this line is only reached if placementStats is defined, i guess that was unintended.
Anyway, iconRenderState.reset() should take care of setting the opacity to 0. So far it sets it to 1, i don't understand why.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1627,17 +1638,19 @@ export class TextElementsRenderer {
...
+                if (placementStats !== undefined) {
```
atomicnitrate, line 1645: ditto, no need for explicit conversion.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1627,17 +1638,19 @@ export class TextElementsRenderer {
...
                     ++placementStats.numNotVisible;
                 }
-                return false;
+                // Proceed with the label text, which may still be rendered.
```
atomicnitrate, line 1648: Please add a unit test in TextElementsRendererTest to cover this change.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -878,29 +868,16 @@ export function isPathLabelTooSmall(
...
 
-    // TODO: (HARP-3515)
-    //      The rendering of a path label that contains just a single point that is not
-    //      visible is impossible, which is problematic with long paths.
-    //      Fix: Skip/clip the invisible points at beginning and end of the path to get
-    //      the visible part of the path.
-
-    // If not a single point is visible, skip the path
-    if (!anyPointVisible) {
+    // If not a single point is visible, skip the path.
```
atomicnitrate, line 879:
```suggestion
    // If there aren't at least two points visible, skip the path.
```

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -878,29 +868,16 @@ export function isPathLabelTooSmall(
...
-    let anyPointVisible = false;
 
     for (const pt of textElement.points as THREE.Vector3[]) {
-        // Skip invisible points at the beginning of the path.
-        const screenPoint = anyPointVisible
-            ? screenProjector.project(pt, tmpScreenPosition)
-            : screenProjector.projectOnScreen(pt, tmpScreenPosition);
-        if (screenPoint === undefined) {
+        if (screenProjector.projectToScreen(pt, tmpScreenPosition) === undefined) {
             continue;
```
atomicnitrate, line 874: By skipping invisible points in the middle of the path, you're altering the shape of the path. The text might not follow the path anymore.
Before we were only skipping invisible points at the beginning of the path, which is (more or less) ok. 
Probably we should still add the last point that was out of the screen, like in this case:


out&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;in
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|
0--1--|---2
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|&nbsp;&nbsp;&nbsp;&nbsp;|    
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;4--|---3
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|&nbsp;&nbsp;&nbsp;&nbsp;|
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;5--|---6

Here, we were removing 0 and 1. We should probably keep 1 though. But surely we cannot skip 4 and 5 as you're doing with your change, otherwise you'll write on a straight line from 3 to 6.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -828,8 +823,6 @@ export function placePathLabel(
         checkGlyphCollision = candidateBoxes.length > 0;
     }
 
-    const checkGlyphVisible = !screenCollisions.isFullyVisible(tmpCollisionBox);
-
```
atomicnitrate, line 832: I guess you removed this because you're only passing visible path points from isPathLabelTooSmall. That's wrong, check my comment there.
As long as we pass here invisible path points, we need these checks. Another option is to clamp the invisible points to the screen borders and the I guess we can get rid of this.

But frankly i wouldn't do any further changes as part of this PR, is too difficult to follow all changes already.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -841,9 +834,6 @@ export function placePathLabel(
         } else {
             collisionBox.copy(glyphBox);
         }
-        if (checkGlyphVisible && !screenCollisions.isVisible(collisionBox)) {
-            return PlacementResult.Invisible;
-        }
```
atomicnitrate, line 846: same as above.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -697,6 +692,15 @@ function placePointLabelAtAnchor(
...
     const textOffset = computePointTextOffset(label, placement, scale, env, tmpTextOffset);
     textOffset.add(screenPosition);
+
+    // Update output screen position.
+    outScreenPosition.set(textOffset.x, textOffset.y, labelState.renderDistance);
+
+    // Check if icon's label was already rejected.
+    if (isRejected) {
+        return PlacementResult.Rejected;
+    }
```
atomicnitrate, line 702: With this change, it doens't make sense any more to pass the isRejected parameter to this function or even to placePointLabel.
If the icon is rejected, we can't just skip the call to placePointLabel and reject the text as well.

@here/harp-mapview/lib/text/Placement.ts
```diff
@@ -710,28 +714,13 @@ function placePointLabelAtAnchor(
...
     tmp2DBox.set(tmpCenter.x - tmpSize.x / 2, tmpCenter.y - tmpSize.y / 2, tmpSize.x, tmpSize.y);
 
-    // Update output screen position.
-    outScreenPosition.set(textOffset.x, textOffset.y, labelState.renderDistance);
-
-    // Check the text visibility if invisible finish immediately
-    // regardless of the persistence state - no fading required.
-    if (!screenCollisions.isVisible(tmp2DBox)) {
-        return PlacementResult.Invisible;
-    }
```
atomicnitrate, line 720: I don't understand why you postpone this check. Why do you wan't to check collisions if the text is invisible?

@here/harp-mapview/lib/text/RenderState.ts
```diff
@@ -119,6 +119,18 @@ export class RenderState {
...
+     * than 0.
+     */
+    isReallyVisible(): boolean {
+        return (
+            this.m_state !== FadingState.FadedOut &&
+            this.m_state !== FadingState.Undefined &&
+            this.opacity > 0
+        );
+    }
+
```
atomicnitrate, line 133: couldn't we instead check the opacity in isVisible() ?

@here/harp-mapview/lib/text/TextElementState.ts
```diff
@@ -80,6 +80,32 @@ export class TextElementState {
...
+            return false;
+        }
+
+        for (const state of iconRenderStates) {
+            if (state.isReallyVisible()) {
+                return true;
+            }
+        }
+        return false;
+    }
```
atomicnitrate, line 107: I'd rather avoid having the almost exact copy of isVisible(), better change isVisible behaviour so that it checks opacity.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1665,22 +1678,27 @@ export class TextElementsRenderer {
                 iconIndex === undefined
             );
```
atomicnitrate, line 1679: See my comments in Placement.ts. Probably we could skip placing the text if the icon is rejected. Makes the code simpler with 'placePointLabel'. We were placing the text just to mark it invisible instead of rejected to avoid the fade out, but I'm not sure it's worth the confusion it add to placement code.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1709,8 +1727,14 @@ export class TextElementsRenderer {
...
         if (iconReady) {
-            if (iconRejected) {
-                iconRenderState!.startFadeOut(renderParams.time);
+            if (iconRejected || iconInvisible) {
+                if (iconRenderState.isReallyVisible()) {
+                    iconRenderState.startFadeOut(renderParams.time);
+                } else if (!iconRenderState.isFadingOut()) {
+                    // Reset the icon and keep it from popping up before fading in.
+                    iconRenderState.reset();
+                    iconRenderState.opacity = 0;
```
atomicnitrate, line 1736: ditto, this should be done in reset()

atomicnitrate, line 1736: ditto, the opacity should be set to 0 in reset()

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1665,22 +1678,27 @@ export class TextElementsRenderer {
...
                     pointLabel.poiInfo !== undefined && pointLabel.poiInfo.textIsOptional === true;
                 iconRejected = textRejected && !textIsOptional;
             }
 
             if (textRejected) {
-                textRenderState!.startFadeOut(renderParams.time);
+                if (textRenderState!.isReallyVisible()) {
+                    textRenderState!.startFadeOut(renderParams.time);
+                }
+                iconRejected = true;
```
atomicnitrate, line 1701: we already reject the icon above based on whether the text is optional or not.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1665,22 +1678,27 @@ export class TextElementsRenderer {
...
-            if (!iconRejected) {
+            if (!iconRejected && !iconInvisible) {
                 const textIsOptional: boolean =
                     pointLabel.poiInfo !== undefined && pointLabel.poiInfo.textIsOptional === true;
                 iconRejected = textRejected && !textIsOptional;
             }
 
             if (textRejected) {
-                textRenderState!.startFadeOut(renderParams.time);
+                if (textRenderState!.isReallyVisible()) {
```
atomicnitrate, line 1698: This check shouldn't be needed.

If the label is FadingOut or FadedOut, startFadeOut() does nothing. We should add a check there to also return if the state is undefined.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1709,8 +1727,14 @@ export class TextElementsRenderer {
         }
         // ... and render the icon (if any).
         if (iconReady) {
```
atomicnitrate, line 1729:
```suggestion
        if (iconReady && !iconInvisible) {
```
that way you don't need to handle invisible icons below

@here/harp-mapview/lib/text/RenderState.ts
atomicnitrate: actually, shouldn't it always be the case that opacity > 0 for any states other than faded out or undefined?

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1709,8 +1727,14 @@ export class TextElementsRenderer {
...
-            if (iconRejected) {
-                iconRenderState!.startFadeOut(renderParams.time);
+            if (iconRejected || iconInvisible) {
+                if (iconRenderState.isReallyVisible()) {
+                    iconRenderState.startFadeOut(renderParams.time);
```
atomicnitrate, line 1732: I already commented above that startFadeOut should internally have all checks needed avoid doing anything when it's not needed (already fading out, faded out or undefined labels).

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicnitrate, line 1736: I don't understand the need of this.
If the label is not visible and not fading out then it must be in undefined or faded out state, either way you don't need to reset it.

atomicnitrate (changes requested):
Please have a look at how harp native checks the label bounds visibility:
https://gerrit.it.here.com/gitweb?p=harp/core.git;a=blob;f=label/src/placement/Positioning.cpp;h=32481bd86c16f78ceac048b50895e29a9c181799;hb=HEAD

Also, please add unit tests in TextElementsRendererTest for any changes/fixes you do in label fading behaviour. It's very difficult to avoid regressions otherwise.

Please add follow-up changes as commits on top, otherwise it'll be really difficult for me keep up with what you changed.

@here/harp-mapview/lib/MapView.ts
StefanDachwitz: Missed that, fixing it.

@here/harp-mapview/lib/ScreenProjector.ts
StefanDachwitz: The unit tests seem not to catch this. BUT, the code that you propose does not work, the icons near the border do disappear while they should be visible. I keep the modified version.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: > The alternative is to do layout for **all** labels on a tile, no? 
> Ass described below, this performance optimization has been removed against my judgement.

It's not the only alternative, I'm not sure why you think I asked you to do that. Quoting my comment, this was my suggestion:

> As in harp native, I would put this as a styling property, so that it can be adjusted on a case by case basis.
> I would also use thresholds that are relative to the total label size. Harp native has a "minVisibility" factor [0,1], representing how much area of the label must be inside the screen before it's made visible.

So I'm just suggesting to change label bbox culling to use **relative** thresholds instead of the absolute ones you were using.
I also mentioned in my last comment where you can find HARP native's code to cull labels:

> Please have a look at how harp native checks the label bounds visibility:
> https://gerrit.it.here.com/gitweb?p=harp/core.git;a=blob;f=label/src/placement/Positioning.cpp;h=32481bd86c16f78ceac048b50895e29a9c181799;hb=HEAD

StefanDachwitz: I was thinking that, to do a proper computation of the label area the layout has to be computed before the visible part can be computed. The measuring of text is pretty expensive due to the complex machinery required, with fonts and glyphs required to be loaded in addition to placement checks to be done for optional layout positions. If you have an approach that does not require that, I am open for that.
I would rather go for something that is fast and works fine for the standard settings, and live with minor visual issues in border cases with unusual large fonts or such things.

@here/harp-mapview/lib/ScreenProjector.ts
atomicsulfate: I still don't get why is like that, but check my suggestion below.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: Again, I misunderstood. This should be fine.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: I think this check was wrong. Instead of returning invisible if a glyph is not visible, what we should do is skip the collision test for  that glyph that we do just after this.

@here/harp-mapview/lib/ScreenProjector.ts
StefanDachwitz: This code has been removed. I suggest that we measure the impact of this PR, and optimize it in another PR if necessary.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
StefanDachwitz: This code has been removed. I suggest that we measure the impact of this PR, and optimize it in another PR if necessary.

@here/harp-mapview/lib/text/Placement.ts
StefanDachwitz: This code has been reverted. A new commit is ready soon.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: Yet here you're rejecting the text if the icon was rejected without checking whether it's optional or not. Then why not reject it sooner?

There's also other places that wouldn't work for an optional icon:

- placePointLabel is rejecting new labels with rejected icon
https://github.com/heremaps/harp.gl/commit/9584f668f38a121b8881dc9f64052207313cf5cf#diff-cbaa0c98ceb02c4586a463f1982b813f5555d33eab176cd4f180be8eaad25e6cR442-R443

- placePointLabel is checking only the current anchor for permanent labels with rejected icon
https://github.com/heremaps/harp.gl/commit/9584f668f38a121b8881dc9f64052207313cf5cf#diff-cbaa0c98ceb02c4586a463f1982b813f5555d33eab176cd4f180be8eaad25e6cR449-R458

Afaik, MarkerTechnique's iconIsOptional only applies when the icon image is not available for some reason.See:

https://github.com/heremaps/harp.gl/blob/69af4d17600970a9653951e234716da6d1f0ac4b/%40here/harp-datasource-protocol/lib/TechniqueParams.ts#L656
`
/**
     * If true, the text will appear even if the icon cannot be rendered because of missing icon
     * graphics.
     ...
`
That's why currently we only check that property in shouldRenderPointText() if the poiInfo is not valid.

I'm fine changing the behaviour of 'iconIsOptional', actually I think the current behaviour is weird. But for that we can change shouldRenderPointText() to return false if the icon is required but it was rejected. Then I don't see why placePointText needs to know whether the icon was rejected.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1668,12 +1661,14 @@ export class TextElementsRenderer {
                 if (placementStats) {
                     placementStats.numPoiTextsInvisible++;
                 }
-                labelState.reset();
-                return false;
+                if (!renderIcon || iconInvisible) {
+                    labelState.reset();
+                    return false;
+                }
```
atomicsulfate, line 1667: with this change, an invisible text without icon will be faded in later, since textNeedsDraw is evaluated true and textRejected is false.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: > If we declare all text parts of the poi labels "invisible", in the current flow of the code, the icons are not rendered as well. 

With your changes that's not true anymore:
https://github.com/heremaps/harp.gl/pull/1888/files#diff-6b6f266bec84f376c6f75070ad562145d6e14b8e4031e0e34c85be854c939f39R1664-R1667 

> it makes sense to reserve space outside the current visible space

We aren't reserving any space for invisible text later in the function:
https://github.com/heremaps/harp.gl/pull/1888/files#diff-cbaa0c98ceb02c4586a463f1982b813f5555d33eab176cd4f180be8eaad25e6cR742

So the only difference I see is that we would mark colliding labels outside of the screen as rejected instead of invisible.
It'd be good to know exactly why this behaviour is better, cause we do it at the expense of checking more collisions (a perf bottleneck).

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: You forgot to add the check for Undefined state in startFadeOut()

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: I think that is because we're not check for undefined in startFadeOut(). If the label is in undefined state, startFadeOut() should return early like for FadedOut and FadingOut states.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: I think that is because we're not check for undefined in startFadeOut(). If the label is in undefined state, startFadeOut() should return early like for FadedOut and FadingOut states.

atomicsulfate (commented):
I've finished the 2nd review round.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: > I was thinking that, to do a proper computation of the label area the layout has to be computed before the visible part can be computed. The measuring of text is pretty expensive due to the complex machinery required, with fonts and glyphs required to be loaded in addition to placement checks to be done for optional layout positions. If you have an approach that does not require that, I am open for that.
> I would rather go for something that is fast and works fine for the standard settings, and live with minor visual issues in border cases with unusual large fonts or such things.

That's clear, as I said, disregard my first comment, it's really unrelated to what you were trying to achieve here. I agree we should early reject labels that are far away from the screen.

> This code has been removed. I suggest that we measure the impact of this PR, and optimize it in another PR if necessary.

Ok, makes sense to reduce the scope of this PR. But we can only do it if FPS is not impacted. Please, run manual perf tests on macbook to verify there's no noticeable drop in performance:
https://render-verity.jenkins.release.in.here.com/view/Tests/job/verity-perf-device-test-macbook-manual

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: A label in Undefined state is not visible, it doesn't make sense for it to fade out, only to fade in.

atomicsulfate (changes requested):
See my latest comments.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
StefanDachwitz: Modified startFadeOut() and the unit tests accordingly. I hope the rest of the code plays along.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
StefanDachwitz: outdated

@here/harp-mapview/lib/text/Placement.ts
StefanDachwitz: outdated

@here/harp-mapview/lib/text/Placement.ts
StefanDachwitz: removed the line.

@here/harp-mapview/lib/text/Placement.ts
StefanDachwitz: re-added the test

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1626,18 +1626,11 @@ export class TextElementsRenderer {
...
             );
-            if (result === PlacementResult.Invisible) {
-                iconRenderState.reset();
-
-                if (placementStats) {
-                    ++placementStats.numNotVisible;
-                }
-                return false;
-            }
+            iconInvisible = result === PlacementResult.Invisible;
```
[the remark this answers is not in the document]
StefanDachwitz: added the reset if invisible

@here/harp-mapview/lib/text/TextElementsRenderer.ts
StefanDachwitz: fixed it by adding the reset for the textRenderState and by handling the invisible text label in the code below (not rendering it).

@here/harp-mapview/lib/text/TextElementsRenderer.ts
StefanDachwitz: Simplified back kto

            if (iconRejected) {
                iconRenderState.startFadeOut(renderParams.time);
            } else {
                iconRenderState!.startFadeIn(renderParams.time, this.m_options.disableFading);
            }

StefanDachwitz: ... and defaulting the opacity value in RenderState to 0.0 instead of 1.0, which leads to issues.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: This condition is not needed anymore then:
https://github.com/heremaps/harp.gl/pull/1888/files#diff-cbaa0c98ceb02c4586a463f1982b813f5555d33eab176cd4f180be8eaad25e6cR738

It's the same test you added back.

@here/harp-mapview/lib/text/Placement.ts
atomicsulfate: Ok, but we still reject new text with rejected icon:
https://github.com/heremaps/harp.gl/pull/1888/files#diff-cbaa0c98ceb02c4586a463f1982b813f5555d33eab176cd4f180be8eaad25e6cR442-R444

Both new and permanent labels should do the same if the icon is rejected. I suggest to check if the icon is rejected and not optional in shouldRenderPointText() and skip completely the call to placePointLabel. Then placePointLabel() won't need to know if the icon is rejected anymore.

atomicsulfate (changes requested):
I've finished the review.
Please have a look at the unresolved comments and add some tests in TextElementsRendererTest to cover scenarios with invisible/rejected icon and visible text.

@here/harp-mapview/lib/text/Placement.ts
StefanDachwitz: Moved test to shouldRenderPointText()

StefanDachwitz: ... had to move the test back into the calling method, since the case where a rejected icon has to be rendered because it is being faded out would keep the text from being rendered (in which case shouldRenderPointText() would return false).
I think that this should be OK now.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
StefanDachwitz: I ran a test, and it showed some impact. I would strongly suggest to put the optimization into another PR.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1672,7 +1677,13 @@ export class TextElementsRenderer {
                 textRenderState!.reset();
             }
 
-            const textRejected = placeResult === PlacementResult.Rejected;
+            const iconIsRequired = !(poiInfo?.iconIsOptional === false);
+            const iconIsReady = poiInfo?.isValid === true;
```
atomicsulfate, line 1681: could you use the iconReady constant from line 1618? 
https://github.com/heremaps/harp.gl/pull/1888/commits/572d9eb15efd00d3463c2f94e6588ce43c810a63#diff-6b6f266bec84f376c6f75070ad562145d6e14b8e4031e0e34c85be854c939f39R1618

atomicsulfate (changes requested):
I finished my review.

atomicsulfate (approved):

PR #1891, opened 2020-10-08
HARP-11672: Support web tile request cancellation

atomicsulfate:


@here/harp-mapview-decoder/lib/TileLoader.ts
```diff
@@ -79,159 +80,79 @@ export class TileLoader implements ITileLoader {
...
             this.requestController.priority = priority;
         }
     }
 
     /**
-     * Start loading. Only call if loading did not start yet.
+     * @override
      */
-    protected startLoading() {
-        const myLoadCancellationToken = this.loadAbortController.signal;
```
nzjony, line 176: I notice you haven't removed the `this.loadAbortController`, is there a reason for it?

@here/harp-mapview-decoder/lib/TileLoader.ts
```diff
@@ -79,159 +80,79 @@ export class TileLoader implements ITileLoader {
...
+    protected loadImpl(
+        abortSignal: AbortSignal,
```
nzjony, line 114: this value is protected in the base class, so it can be exposed here, do you need to pass it in?

@here/harp-mapview-decoder/lib/TileLoader.ts
```diff
@@ -79,159 +80,79 @@ export class TileLoader implements ITileLoader {
...
+        onDone: (doneState: TileLoaderState) => void,
```
nzjony, line 115: I see that there are 4 methods that have these values passed in, if onDone and onError were protected, you wouldn't need to pass them through like this, the onLoaded method could access them directly, but as you prefer.

@here/harp-mapview-decoder/lib/TileLoader.ts
```diff
@@ -255,77 +176,60 @@ export class TileLoader implements ITileLoader {
                     return;
                 }
 
-                this.onDecoded(decodedTile);
+                this.decodedTile = decodedTile;
```
nzjony, line 179: you have a private method: `onDecoded` that does the same thing as these two lines

@here/harp-mapview-decoder/lib/TileLoader.ts
atomicsulfate: No, i just missed it. Done.

@here/harp-mapview-decoder/lib/TileLoader.ts
atomicsulfate: Good point. I forgot to make the abort controller and all members related to the done promise private. I'll do that. That way the base class is as closed for modification as it's possible in Typescript.

@here/harp-mapview-decoder/lib/TileLoader.ts
atomicsulfate: I want to follow SOLID as much as possible with this class hierarchy. If I make them protected, base classes could override them, thus breaking the open-closed principle.

@here/harp-mapview-decoder/lib/TileLoader.ts
atomicsulfate: done

@here/harp-mapview-decoder/lib/TileLoader.ts
nzjony: Good point.

nzjony (approved):

PR #1908, opened 2020-10-20
HARP-12549 Fix bug with max retries + add test

nzjony:


@here/harp-transfer-manager/src/TransferManager.ts
```diff
@@ -85,21 +85,25 @@ export class TransferManager implements ITransferManager {
...
                 err.hasOwnProperty("isCancelled") ||
                 (err.hasOwnProperty("name") && err.name === "AbortError") ||
-                retryCount > maxRetries
+                retryCount >= maxRetries
             ) {
                 throw err;
             }
         }
         return await TransferManager.waitFor(TransferManager.retryTimeout * retryCount).then(() =>
-            TransferManager.fetchRepeatedly(fetchFunction, maxRetries, retryCount + 1, url, init)
```
nzjony, line 102: This is the bug, the `maxRetries` was swapped with the `retryCount + 1`.

@here/harp-transfer-manager/src/TransferManager.ts
```diff
@@ -85,21 +85,25 @@ export class TransferManager implements ITransferManager {
         init?: RequestInit
     ): Promise<Response> {
         try {
-            const response = await fetchFunction(url, init);
-            if (response.status !== 503 || retryCount > maxRetries) {
```
nzjony, line 89: The `retryCount > maxRetries` is also wrong, because `retryCount` is 0 based, and we wait to get past `maxRetries`, so we actually get 2 more than maxRetries, i.e. if maxRetries is 3, then retryCount is 0,1,2,3,4, i.e. 5 attempts, hence this is also fixed.

nzjony:
retest this please

FraukeF (approved):

PR #1909, opened 2020-10-20
HARP-12514: Add DataSource.dataSourceOrder.

robertoraggi:
When `dataSourceOrder` is defined, all the objects created by the
`DataSource` will be rendered on top of the objects created by other
`DataSource`s with lower `dataSourceOrder` values.


@here/harp-mapview/lib/DataSource.ts
```diff
@@ -169,6 +169,18 @@ export abstract class DataSource extends THREE.EventDispatcher {
...
+     * Overrides the default rendering order of this `DataSource`.
+     *
+     * @remarks
+     * When `dataSourceOrder` is defined, all the objects created by this `DataSource`
+     * will be rendered on top of the objects created by other `DataSource`s with
+     * lower `dataSourceOrder` values.
+     *
+     * @defaultValue undefined
+     */
+    dataSourceOrder?: number;
```
FraukeF, line 182: please mark as @beta

FraukeF: Align/Remove the WebTileDataSource renderOrder Parameter with this change

germanz: ``WebTileDataSourceOptions.renderingOptions.renderOrder`` should be deprecated/removed in favour of common ``dataSourceOrder`` option.

germanz:
The following code has to be added in **WebTileLoader.ts**
``
TileGeometryCreator.instance.registerTileObject(this.tile, mesh, undefined);
``

after the following lines

``
const mesh = TileGeometryCreator.instance.createGroundPlane(
    this.tile,
    material,
    true
);
``

in order to fix the rendering order of web tile layer.

nzjony:
retest this please

nzjony:
retest this please

nzjony:
retest this please

nzjony:
retest this please

@here/harp-mapview/lib/DataSource.ts
nzjony: @germanz , I have done it in a follow up commit

nzjony:
after discussion with @germanz , @atomicsulfate , we need to rethink this to work with labels, so will put this back to draft. Don't review for now.

@here/harp-mapview/lib/MapObjectAdapter.ts
```diff
@@ -136,6 +141,10 @@ export class MapObjectAdapter {
         );
     }
 
+    get level(): number {
+        return this.object.userData.tileKey?.level;
```
atomicsulfate, line 145: Either assert the existence of tileKey:
this.object.userData.tileKey!.level;
or the return type should be:
number|undefined

@here/harp-mapview/lib/TileObjectsRenderer.ts
```diff
@@ -66,6 +72,86 @@ export class TileObjectRenderer {
...
+         */
+        const stableSort = (a: RenderItem, b: RenderItem): number => {
+            if (a.groupOrder !== b.groupOrder) {
+                return a.groupOrder - b.groupOrder;
+            } else if (a.renderOrder !== b.renderOrder) {
+                return a.renderOrder - b.renderOrder;
+            } else if (
+                a.object.userData.tileKey !== undefined &&
+                b.object.userData.tileKey !== undefined &&
+                a.object.userData.tileKey?.mortonCode() !== b.object.userData.tileKey?.mortonCode()
```
atomicsulfate, line 96:
```suggestion
                a.object.userData.tileKey.mortonCode() !== b.object.userData.tileKey.mortonCode()
```

@here/harp-mapview/lib/TileObjectsRenderer.ts
```diff
@@ -66,6 +72,86 @@ export class TileObjectRenderer {
...
+            ) {
+                return (
+                    a.object.userData.tileKey?.mortonCode() -
+                    b.object.userData.tileKey?.mortonCode()
```
atomicsulfate, line 100:
```suggestion
                    a.object.userData.tileKey.mortonCode() -
                    b.object.userData.tileKey.mortonCode()
```

atomicsulfate (changes requested):
It'd be better to cover this with some IBCTs.

@here/harp-mapview/lib/MapObjectAdapter.ts
nzjony: @atomicsulfate , good spotting.

Actually, I looked a bit further, and the way I am doing this is not so great, because the `userData.tileKey` is set in the `TileGeometryCreator.ts` and this `level` getter relies on that.

I think this is for legacy reasons, so I won't touch it now, but I have added `level` as a separate parameter so I won't make it worse.

@here/harp-mapview/lib/TileObjectsRenderer.ts
nzjony: I removed the `undefined` checks too

nzjony:
@atomicsulfate , this is already covered by IBCTs, see: test/rendering/WebTileDataRendering.ts, and the changed ref's internally.

atomicsulfate (approved):

PR #1910, opened 2020-10-20
HARP-11672: Pass client parameter to ITileLoader's loadAndDecode() an…

atomicsulfate:
…d cancel().

Used to support TileLoader sharing for terrain.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/ITileLoader.ts
+++ b/@here/harp-mapview/lib/ITileLoader.ts
@@ -46,9 +46,10 @@ export interface ITileLoader {
     /**
      * Start loading and/or proceed through the various states of loading of this tile.
      *
+     * @param client - Optional client requesting the load.
      * @returns A promise which resolves the [[TileLoaderState]].
      */
-    loadAndDecode(): Promise<TileLoaderState>;
+    loadAndDecode(client?: any): Promise<TileLoaderState>;
 
     /**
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/ITileLoader.ts
```diff
@@ -46,9 +46,10 @@ export interface ITileLoader {
     /**
      * Start loading and/or proceed through the various states of loading of this tile.
      *
+     * @param client - Optional client requesting the load.
      * @returns A promise which resolves the [[TileLoaderState]].
      */
-    loadAndDecode(): Promise<TileLoaderState>;
+    loadAndDecode(client?: any): Promise<TileLoaderState>;
```
nzjony, line 52: It seems that client can actually be type `ITileLoader` ?

atomicsulfate: For TerrainTileLoader implementation, it's true that all clients implement ITileLoader, but at the ITileLoader interface level, that doesn't need to be the case. So far the client parameter is only needed to track how many clients are requesting the tile, that can be implemented just by storing the values in a set. So the client type can be anything, any object type, a number, a string... 
Any future implementation of ITileLoader can narrow the type down to whatever it uses.

nzjony (approved):

PR #1915, opened 2020-10-23
HARP-12587: Moves TileObject rendering to TileObjectRenderer class

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -5,12 +5,8 @@
  */
 import {
     Env,
-    Expr,
-    getFeatureId,
-    getPropertyValue,
     GradientSky,
     ImageTexture,
-    IndexedTechnique,
     Light,
     MapEnv,
     PostEffects,
@@ -61,12 +57,10 @@ import { ElevationRangeSource } from "./ElevationRangeSource";
 import { EventDispatcher } from "./EventDispatcher";
 import { FrustumIntersection } from "./FrustumIntersection";
 import { overlayOnElevation } from "./geometry/overlayOnElevation";
-import { SolidLineMesh } from "./geometry/SolidLineMesh";
 import { TileGeometryManager } from "./geometry/TileGeometryManager";
 import { MapViewImageCache } from "./image/MapViewImageCache";
 import { IntersectParams } from "./IntersectParams";
 import { MapAnchors } from "./MapAnchors";
-import { MapObjectAdapter } from "./MapObjectAdapter";
 import { MapViewFog } from "./MapViewFog";
 import { MapViewTaskScheduler } from "./MapViewTaskScheduler";
 import { PickHandler, PickResult } from "./PickHandler";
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/MapView.ts
```diff
@@ -3417,13 +3379,13 @@ export class MapView extends EventDispatcher {
...
-        this.m_env.entries.$zoom = this.m_zoomLevel;
+        this.m_tileObjectRenderer.env.entries.$zoom = this.m_zoomLevel;
 
         // This one introduces unnecessary calculation of pixelToWorld, even if it's barely
         // used in our styles.
-        this.m_env.entries.$pixelToMeters = this.pixelToWorld;
+        this.m_tileObjectRenderer.env.entries.$pixelToMeters = this.pixelToWorld;
 
-        this.m_env.entries.$frameNumber = this.m_frameNumber;
+        this.m_tileObjectRenderer.env.entries.$frameNumber = this.m_frameNumber;
```
nzjony, line 3388: There seem to be a few cases where MapView interacts with MapEnv so I wonder if it doesn't make more sense to keep it in MapView and just pass it to the `TileObjectRenderer` ?

Otherwise move `updateEnv` to the `TileObjectRenderer`, so the classes try to be as encapsulated as possible.

@here/harp-mapview/lib/MapView.ts
```diff
@@ -3701,7 +3662,12 @@ export class MapView extends EventDispatcher {
         // no need to check everything if we're not going to create text renderer.
         renderList.forEach(({ zoomLevel, renderedTiles }) => {
             renderedTiles.forEach(tile => {
-                this.renderTileObjects(tile, zoomLevel);
+                this.m_tileObjectRenderer.render(
+                    tile,
+                    zoomLevel,
+                    this.m_camera.position,
+                    this.m_sceneRoot
```
nzjony, line 3669: I wonder if the scene root node can be owned by the `TileObjectRenderer` ?

It would require moving the `MapAnchors` update method as well to the `TileObjectRenderer`, but that may be actually a good idea?

At least the `update` method has similar logic to the `renderTileObjects` method, i.e. position and add to the root

FraukeF: Hmm, I see TileObjectsRenderer more in parallel to the MapAnchors, although the naming does not obviously say so. I could call the new class only TileObjects without the "Renderer" suffix, that might clear things up a bit. what do you think?

@here/harp-mapview/lib/MapView.ts
FraukeF: yeah, good point, I also saw that it broke some tests, I ll review that

FraukeF: I moved the MapEnv back to MapView

FraukeF:
@harpgl-bot  retest this please

nzjony (approved):

@here/harp-mapview/lib/MapView.ts
nzjony: I valid name is hard to come up, the change is anyway good.

PR #1919, opened 2020-10-27
HARP-12192: Request Map view update if camera settings are not final.

atomicsulfate:
- Add final flag to getTargetAndDistance return type. The flag will be false if the elevation is not yet ready.

- Also replace getHeightAboveTerrain() by one line of code.

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/Utils.ts
+++ b/@here/harp-mapview/lib/Utils.ts
@@ -498,33 +498,6 @@ export namespace MapViewUtils {
     }
 
     /**
-     * Returns the height of the camera above the earths surface.
-     *
-     * @remarks
-     * If there is an ElevationProvider, this is used. Otherwise the projection is used to determine
-     * how high the camera is above the surface.
-     *
-     * @param level - Which level to request the surface height from.
-     * @return Height in world units.
-     */
-    function getHeightAboveTerrain(
-        location: THREE.Vector3,
-        projection: Projection,
-        elevationProvider?: ElevationProvider,
-        level?: number
-    ): number {
-        if (elevationProvider !== undefined) {
-            const geoLocation = projection.unprojectPoint(location);
-            const heightAboveTerrain = elevationProvider.getHeight(geoLocation, level);
-            if (heightAboveTerrain !== undefined) {
-                const height = projection.unprojectAltitude(location) - heightAboveTerrain;
-                return Math.max(height, 1);
-            }
-        }
-        return Math.abs(projection.groundDistance(location));
-    }
-
-    /**
      * Constrains given camera target and distance to {@link MapView.maxBounds}.
      *
      * @remarks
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/Utils.ts
```diff
@@ -665,13 +638,18 @@ export namespace MapViewUtils {
     }
     /**
      * @internal
-     * @deprecated This method will be moved to MapView.
```
nzjony, line 668: Who put this here, and do you know why?

Does it still make sense to keep this deprecated so that it will eventually be moved to MapView.. or not?

@here/harp-mapview/lib/Utils.ts
```diff
@@ -685,6 +663,7 @@ export namespace MapViewUtils {
                   TERRAIN_ZOOM_LEVEL
               )
             : undefined;
+        const final = !elevationProvider || elevation !== undefined;
```
nzjony, line 666: Minor, elevation is undefined if !evelationProvider is undefined.

@here/harp-mapview/lib/Utils.ts
atomicsulfate: Nino deprecated several functions in MapView's Utils because there were not usages outside MapView at that time, here:
https://github.com/heremaps/harp.gl/pull/1365

But now getTargetAndDistance() is used within Utils.

@here/harp-mapview/lib/Utils.ts
atomicsulfate: If elevationProvider is undefined, then !elevationProvider is true and the expression is shortcircuited, resulting in final === true.
I don't see the problem.

nzjony: Sorry, I wasn't clear, line 660 already does the check for `elevationProvider`, and if it is undefined, then `elevation` is undefined, hence you can use just `const final = elevation !== undefined;`

nzjony: @atomicsulfate , will you change this or keep it? It is a minor, if you prefer not to change, then I will just approve.

atomicsulfate: > Sorry, I wasn't clear, line 660 already does the check for `elevationProvider`, and if it is undefined, then `elevation` is undefined, hence you can use just `const final = elevation !== undefined;`

If elevationProvider is undefined, final should be true, cause that means we are not using elevation and therefore the computed target is final.
final = elevation !== undefined  <- That's only applicable if there's an elevation provider.

nzjony: Ahh, I missed that it should be true if it is undefined, because no better result will come, ok will approve.

nzjony (approved):

PR #1924, opened 2020-10-28
HARP-12403: Add support for clipping lines against the tile bounds.

robertoraggi:
This change introduces the function clipLineString that can be used
to clip lines against the tile bounding box.

### DIFF
--- a/@here/harp-geometry/lib/ClipLineString.ts
+++ b/@here/harp-geometry/lib/ClipLineString.ts
@@ -1,0 +1,155 @@
+/*
+ * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Math2D } from "@here/harp-utils";
+import { Vector2 } from "three";
+
+/**
+ * A clipping edge.
+ *
+ * @remarks
+ * Clip lines using the Sutherland-Hodgman algorithm.
+ *
+ * @internal
+ */
+class ClipEdge {
+    readonly p0: Vector2;
+    readonly p1: Vector2;
+
+    /**
+     * Creates a clipping edge.
+     *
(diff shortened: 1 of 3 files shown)


@here/harp-geometry/lib/ClipLineString.ts
```diff
@@ -0,0 +1,155 @@
...
+            const currentPoint = inputList[i];
+            const prevPoint = i > 0 ? inputList[i - 1] : undefined;
+
+            if (this.inside(currentPoint)) {
+                if (prevPoint !== undefined && !this.inside(prevPoint)) {
+                    if (lineString.length > 0) {
+                        lineString = [];
+                        result.push(lineString);
+                    }
+                    pushPoint(this.computeIntersection(prevPoint, currentPoint));
```
nzjony, line 99: `computeIntersection` can returned undefined, should this be checked?

@here/harp-vectortile-datasource/lib/adapters/geojson/GeoJsonDataAdapter.ts
```diff
@@ -199,14 +200,36 @@ export class GeoJsonDataAdapter implements DataAdapter {
...
-                        $level
-                    );
+                    let geometry = convertLineGeometry(feature.geometry, decodeInfo);
+
+                    const clippedGeometries: ILineGeometry[] = [];
+
+                    const DEFAULT_BORDER = 100;
+
+                    geometry.forEach(g => {
+                        const cliipped = clipLineString(
```
nzjony, line 210:
```suggestion
                        const clipped = clipLineString(
```

nzjony (approved):

@here/harp-geometry/lib/ClipLineString.ts
robertoraggi: it should not be needed because `computeIntersection` is called only after the end points are tested against the clipping edge and an intersection is known to exist.

PR #1945, opened 2020-11-03
FlatTheme support

FraukeF:

# CI Check

Happy contributing!


@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -252,6 +252,33 @@ export function isJsonExprReference(value: any): value is JsonExprReference {
...
 
+/**
+ * Extracts a StyleSet for a styleSetName from a Theme
+ *
+ * @param theme - The theme
+ * @param styleSetName - The name of the styleSet to be extracted.
+ *
+ * @return The Extracted StyleSet
+ */
+export function extractStyleSet(theme: Theme | FlatTheme, styleSetName: string): StyleSet {
```
robertoraggi, line 263: please mark this function with `@internal`

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -198,15 +205,15 @@ export class TileDataSource<TileType extends Tile = Tile> extends DataSource {
...
         }
 
-        const styleSet =
-            this.styleSetName !== undefined && theme.styles
-                ? theme.styles[this.styleSetName]
-                : undefined;
+        let styleSet: StyleSet | undefined;
+        if (this.styleSetName !== undefined) {
+            styleSet = extractStyleSet(theme, this.styleSetName);
+        }
```
robertoraggi, line 216: filtering the style rules here should be deprecated and considered only as a temporary solution until we are with the deprecation of the old theme format.

Our goal should be to deprecate (and eventually remove) the concept that there is only one single "styleSet" associated to a datasource. That is, currently the datasources must set the property `"styleSetName"` at construction type and then later this property is used to filter the style rules. This is not how the unified theme is supposed to work.

In the unified theme format the `styleSet` is just a way to group together style rules, for example a styleSet called "roads" could be used to group layers for the roads, another one called city-center could collect the layers for the city centers and so on.. These groups of rules could later be references in `theme.priorities` and `theme.labelPriorities` to source items. Eventually we could also have APIs to enable/disable these style sets.

@here/harp-mapview-decoder/lib/TileDataSource.ts
```diff
@@ -198,15 +205,15 @@ export class TileDataSource<TileType extends Tile = Tile> extends DataSource {
      * `styleSetName` property) is found in `theme`.
      * @override
      */
-    setTheme(theme: Theme, languages?: string[], styleSetName?: string): void {
+    setTheme(theme: Theme | FlatTheme, languages?: string[], styleSetName?: string): void {
```
robertoraggi, line 208: i'm not sure that the signature of this function is correct. This code above is assuming that the `theme` set to this data source is already fully resolved (e.g. all the extends are evaluated), which is not the case.
I would expect setTheme to be an `async` function, accept a url or a theme and resolve the theme using the ThemeLoader.

robertoraggi: also i would remove `styleSetName` and languages from the signature of this function.

FraukeF: Will be added that in #1949

@here/harp-datasource-protocol/lib/Theme.ts
FraukeF: removed it

@here/harp-mapview-decoder/lib/TileDataSource.ts
FraukeF: I added your comment to the ticket I created for the full FlatTheme support: HARP-12835

@here/harp-mapview-decoder/lib/TileDataSource.ts
FraukeF: I squashed the two PRs, so the resolving of the themes is done... as well as the removal of the styleSetName I had added...

robertoraggi (approved):

PR #1952, opened 2020-11-05
HARP-12773: Roadshield icon disappear, while text remains visible

StefanDachwitz:
* also fixes infinite render loop due to only part of road shields rendered

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -1574,6 +1574,7 @@ export class TextElementsRenderer {
     ): boolean {
         const pointLabel: TextElement = labelState.element;
         const textRenderState: RenderState | undefined = labelState.textRenderState;
+        const isLineMarker = iconIndex !== undefined;
 
         assert(iconIndex === undefined || labelState.iconRenderStates !== undefined);
         const iconRenderState: RenderState =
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1574,6 +1574,7 @@ export class TextElementsRenderer {
     ): boolean {
         const pointLabel: TextElement = labelState.element;
         const textRenderState: RenderState | undefined = labelState.textRenderState;
+        const isRoadShield = iconIndex !== undefined;
```
FraukeF, line 1577: Just a minor thing: This name brings more meaning to this as the condition gives, shouldn´t it be just something like "hasIcon"

FraukeF (dismissed):

@here/harp-mapview/lib/text/TextElementsRenderer.ts
atomicsulfate: This constant is to differentiate between a point label or a line marker. Better name it "isLineMarker" to use the same naming as in the rest of the code.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1831,6 +1840,7 @@ export class TextElementsRenderer {
 
         // Process markers (with shield groups).
         if (minDistanceSqr > 0 && shieldGroup !== undefined) {
+            let numShieldsVisible = 0;
```
atomicsulfate, line 1843: Use numMarkersVisible instead.

atomicsulfate (changes requested):
Please add a unit test in TextElementsRenderer to verify the fix.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
StefanDachwitz: renamed to "isLineMarker"

atomicsulfate (approved):

PR #1954, opened 2020-11-05
HARP-12352: subsequent DataProvider.register() return early

StefanDachwitz:
* subsequent register operations should resolve after first connect

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview-decoder/test/DataProviderTest.ts
+++ b/@here/harp-mapview-decoder/test/DataProviderTest.ts
@@ -29,6 +29,17 @@ class FakeDataProvider extends DataProvider {
     }
 }
 
+class TimedFakeDataProvider extends FakeDataProvider {
+    /** @override */ async connect(): Promise<void> {
+        const promise = new Promise<void>(resolve => {
+            setTimeout(() => {
+                resolve();
+            }, 100);
+        });
+        return promise;
+    }
+}
+
 describe("DataProvider", function() {
     const clients = [{}, {}];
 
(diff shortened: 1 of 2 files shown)


@here/harp-mapview-decoder/test/DataProviderTest.ts
```diff
@@ -60,4 +71,44 @@ describe("DataProvider", function() {
...
+
+        await dataProvider.register(clients[0]);
+        expect(connectSpy.calledOnce).is.true;
+
+        await dataProvider.register(clients[0]);
+        await dataProvider.register(clients[1]);
+        expect(connectSpy.calledOnce).is.true;
+    });
+
+    it("calls connect only on first registered client", async function() {
```
FraukeF, line 87: this test is duplicating the first part of the one above, also has the same name...

@here/harp-mapview-decoder/test/DataProviderTest.ts
```diff
@@ -60,4 +71,44 @@ describe("DataProvider", function() {
...
+        const dataProvider = new TimedFakeDataProvider();
+        const connectSpy = sinon.spy(dataProvider, "connect");
+
+        dataProvider.register(clients[0]);
+        dataProvider.register(clients[1]);
+        expect(connectSpy.calledOnce).is.true;
+    });
+
+    it("subsequent register operations wait for connect", async function() {
```
FraukeF, line 96: I am not sure, but would it work if while awaiting the Promise, the client gets unregistered and another or same one registered, or would this need disposal of the Promise: 

const promise0 = dataProvider.register(clients[0]);
dataProvider.unregister(clients[0]);
const promise0 = dataProvider.register(clients[1]);

@here/harp-mapview-decoder/test/DataProviderTest.ts
StefanDachwitz: Yes, it covers the same functionality as the first one, removed.

@here/harp-mapview-decoder/test/DataProviderTest.ts
StefanDachwitz: The current implementation does not reject the promise if the client is un-registered, for which I added tests. Currently, all clients get the same resolve (since they are using the same promise), even those that have been un-registered. 
There is also the issue of DataProviders that have their only client un-registered, then their dispose() method is called.
I am not sure if these cases need an improved specification.

@here/harp-mapview-decoder/test/DataProviderTest.ts
```diff
@@ -60,4 +71,75 @@ describe("DataProvider", function() {
...
+
+    it("calls connect after one provider call unregister", async function() {
+        const dataProvider = new TimedFakeDataProvider();
+
+        const promise0 = dataProvider.register(clients[0]);
+        const promise1 = dataProvider.register(clients[1]);
+        dataProvider.unregister(clients[0]);
+
+        let promise0Resolved = false;
+        promise0.then(() => {
```
nzjony, line 109: you should await on the `then` also

@here/harp-mapview-decoder/test/DataProviderTest.ts
```diff
@@ -60,4 +71,75 @@ describe("DataProvider", function() {
...
+            promise0Resolved = true;
+        });
+
+        await promise0;
+
+        let promise1Resolved = false;
+        promise1.then(() => {
```
nzjony, line 116: ditto

@here/harp-mapview-decoder/test/DataProviderTest.ts
nzjony: or if you don't want to await on it now, store a reference to it then await on it when you need to

@here/harp-mapview-decoder/test/DataProviderTest.ts
nzjony: @FraukeF  that example you mentioned would dispose the dataProvider.

I guess also the promise should be rejected, @StefanDachwitz is that possible? Maybe the `connect()` method can return the Promise + the resolve / reject functions and you can reject the promise when the `dispose()` method is called?

StefanDachwitz: @nzjony The question is, if this is in the scope of the ticket.

@here/harp-mapview-decoder/test/DataProviderTest.ts
StefanDachwitz: done

@here/harp-mapview-decoder/test/DataProviderTest.ts
StefanDachwitz: done

nzjony (approved):

nzjony (approved):

PR #1956, opened 2020-11-06
MINOR: Decoder.decodeTile does not throw

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
+++ b/@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
@@ -12,6 +12,9 @@ import {
 } from "@here/harp-datasource-protocol";
 import { StyleSetEvaluator, StyleSetOptions } from "@here/harp-datasource-protocol/index-decoder";
 import { Projection, TileKey } from "@here/harp-geoutils";
+import { LoggerManager } from "@here/harp-utils";
+
+const logger = LoggerManager.instance.create("ThemedTileDecoder");
 
 /**
  * `ThemedTileDecoder` implements an [[ITileDecoder]] which uses a [[Theme]] to apply styles to the
@@ -35,9 +38,10 @@ export abstract class ThemedTileDecoder implements ITileDecoder {
         data: ArrayBufferLike,
         tileKey: TileKey,
         projection: Projection
-    ): Promise<DecodedTile> {
+    ): Promise<DecodedTile | undefined> {
         if (this.m_styleSetEvaluator === undefined) {
-            return Promise.reject(new Error("No style is defined"));
+            logger.info("cannot decode tile, as there is not style available");
+            return Promise.resolve(undefined);
         }
 
         return this.decodeThemedTile(data, tileKey, this.m_styleSetEvaluator, projection);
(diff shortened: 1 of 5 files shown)


@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
```diff
@@ -35,9 +38,10 @@ export abstract class ThemedTileDecoder implements ITileDecoder {
         data: ArrayBufferLike,
         tileKey: TileKey,
         projection: Projection
-    ): Promise<DecodedTile> {
+    ): Promise<DecodedTile | undefined> {
         if (this.m_styleSetEvaluator === undefined) {
-            return Promise.reject(new Error("No style is defined"));
+            logger.info("cannot decode tile, as there is not style available");
```
nzjony, line 43: would logger.warn or logger.error be better?

@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
```diff
@@ -35,9 +38,10 @@ export abstract class ThemedTileDecoder implements ITileDecoder {
         data: ArrayBufferLike,
         tileKey: TileKey,
         projection: Projection
-    ): Promise<DecodedTile> {
+    ): Promise<DecodedTile | undefined> {
         if (this.m_styleSetEvaluator === undefined) {
```
nzjony, line 42: Another question, if there is no style available, why do we even request tiles ?

@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
FraukeF: I think its actually not an error, e.g. if we set a Theme to a Datasource that has not Styles for the Type of Data, we will run into this, which is not wrong.
Or if no theme has yet been supplied

@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
FraukeF: Good question...I think it´s based on the one-way communication between DataSource and DataDecoder in a Worker and that a DataSource can be created and added to MapView without a Theme assigned to neither of both, that would contain style information for the Data.
Neither MapView not DataSource does know the state of the decoder and just sends their commands and see what comes back. So if the Decoder did not yet get a theme but gets a request to decode a tile the result of the decoding is undefined.

nzjony: Yes, I see the problem now, there would need to be some two way communication which we don't support. Ok, sounds good.

nzjony (approved):

@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
robertoraggi: I also wonder if we can avoid sending the payload to the decoder in this case. I mean, we know that the resulting DecodedTile will be empty so why bothering creating the request to decode the tile.

@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
robertoraggi: i'm not sure about using logger.info. I'd say it's a `warning` if we want the user to change its code because otherwise they are creating useless requests that are slow and don't produce anything.

Actually, I'm not sure what logger.info() is useful for, do we have any guidelines for our logger?

PR #1973, opened 2020-11-16
HARP-12959 Support GeoJSON point features with altitude for pre-tiled data.

atomicsulfate:
- GeoJsonDataProvider forwards now altitude for point features to decoder.
- Data tiled with geojson-vt does not yet support altitude.
- Add public example and IBCT tests.

### DIFF
--- a/@here/harp-geoutils/lib/math/Vector3Like.ts
+++ b/@here/harp-geoutils/lib/math/Vector3Like.ts
@@ -4,20 +4,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { isVector2Like, Vector2Like } from "./Vector2Like";
+
 /**
  * Interface representing a Vector3.
  */
-export interface Vector3Like {
-    /**
-     * The X position.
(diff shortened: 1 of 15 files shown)


@here/harp-geoutils/lib/math/Vector3Like.ts
```diff
@@ -4,26 +4,18 @@
...
-     * The X position.
-     */
-    x: number;
-
-    /**
-     * The Y position.
-     */
-    y: number;
-
+export interface Vector3Like extends Vector2Like {
```
robertoraggi, line 12: there shouldn't be any reason for this. Any Vector3Like is already a Vector2Like

robertoraggi (dismissed):

@here/harp-geoutils/lib/math/Vector3Like.ts
atomicsulfate: done

robertoraggi (approved):

PR #1974, opened 2020-11-17
HARP-12989 Adding null channel for CI

nzjony:
Adding NullChannel for CI


nzjony:
Abandoning, it makes more sense to explicitly flag those messages which are showing, incase new messages come in the future they are ignored, and it shows an actual bug.

PR #1980, opened 2020-11-19
HARP-12581: Fix render order for labels

atomicsulfate:
- Set Render order from the technique in TextElement or PoiInfo.
- Use DataSource.dataSourceRenderOrder to compose with Technique's renderOrder into a single bigint renderOrder.
- Add unit and IBCT tests cases.

### DIFF
--- a/@here/harp-mapview/lib/poi/BoxBuffer.ts
+++ b/@here/harp-mapview/lib/poi/BoxBuffer.ts
@@ -145,7 +145,7 @@ export class BoxBuffer {
      */
     constructor(
         private readonly m_material: THREE.Material | THREE.Material[],
-        private readonly m_renderOrder: number = 0,
+        private readonly m_renderOrder: bigint,
         startElementCount = START_BOX_BUFFER_SIZE,
         private readonly m_maxElementCount = MAX_BOX_BUFFER_SIZE
     ) {
@@ -480,7 +480,6 @@ export class BoxBuffer {
 
         if (this.m_mesh === undefined) {
             this.m_mesh = new BoxBufferMesh(this.m_geometry, this.m_material);
-            this.m_mesh.renderOrder = this.m_renderOrder;
         } else {
             this.m_mesh.geometry = this.m_geometry;
         }
--- a/@here/harp-mapview/lib/poi/BoxBuffer.ts
+++ b/@here/harp-mapview/lib/poi/BoxBuffer.ts
@@ -145,7 +145,7 @@ export class BoxBuffer {
      */
     constructor(
         private readonly m_material: THREE.Material | THREE.Material[],
-        private readonly m_renderOrder: bigint,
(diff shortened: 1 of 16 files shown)


@here/harp-mapview/lib/text/TextElementBuilder.ts
```diff
@@ -83,8 +99,11 @@ export class TextElementBuilder {
      */
     constructor(
         private readonly m_env: MapEnv | Env,
-        private readonly m_styleCache: TileTextStyleCache
-    ) {}
+        private readonly m_styleCache: TileTextStyleCache,
+        private readonly m_baseRenderOrder: number
```
FraukeF, line 103: I saw that later in the tests, you pass a lot of zeros for the baseRenderOrder, maybe you can default to 0 instead

@here/harp-mapview/lib/poi/BoxBuffer.ts
```diff
@@ -145,7 +145,7 @@ export class BoxBuffer {
      */
     constructor(
         private readonly m_material: THREE.Material | THREE.Material[],
-        private readonly m_renderOrder: number = 0,
+        private readonly m_renderOrder: bigint,
```
FraukeF, line 148: should we keep the default?

@here/harp-mapview/lib/text/TextElement.ts
```diff
@@ -287,7 +287,7 @@ export class TextElement {
      * A `TextElement` with a higher `renderOrder` will be rendered after a `TextElement` with a
      * lower `renderOrder`.
      */
-    renderOrder?: number = 0;
+    renderOrder = BigInt(0);
```
FraukeF, line 290: this removes the explicit type information, consider:

renderOrder: bigint = 0n;

instead

@here/harp-vectortile-datasource/test/resources/geoJsonData.ts
```diff
@@ -104,7 +104,7 @@ export const THEME: Theme = {
                     size: 8,
                     priority: 100
                 },
-                renderOrder: 10.3
+                renderOrder: 10
             }
```
FraukeF, line 108: maybe I overlooked it, but I think there is no test for the introduced asserts if renderOrder is no integer, might be good to add  a test for it

robertoraggi: I believe we have many styles were the render order is not an integer. Is there a way to avoid to use bigint in this or?

@here/harp-mapview/lib/text/TextElementBuilder.ts
atomicsulfate: I rather pass the extra zeros explicitely in the test. Having a default value has the risk that someone forgets to pass the datasourceOrder.

FraukeF (approved):

PR #1981, opened 2020-11-19
HARP-13055: Individual RenderStates for text in line markers

StefanDachwitz:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementState.ts
+++ b/@here/harp-mapview/lib/text/TextElementState.ts
@@ -41,7 +41,7 @@ export class TextElementState {
      * @hidden
      * Used during rendering.
      */
-    private m_textRenderState?: RenderState;
+    private m_textRenderStates?: RenderState | RenderState[];
     /**
      * @hidden
      * Used to store recently used text layout.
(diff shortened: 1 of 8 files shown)


@here/harp-mapview/lib/text/TextElementState.ts
```diff
@@ -41,7 +41,7 @@ export class TextElementState {
      * @hidden
      * Used during rendering.
      */
-    private m_textRenderState?: RenderState;
+    private m_textRenderStates?: RenderState | RenderState[];
```
atomicsulfate, line 44: I remember we discussed that we should create a separate TextElementState for each marker on a line, thus getting rid of the icon stata array above. What made you change your mind?

This implementation is still wrong because viewDistance and textLayoutState should be different for each marker on a line. Also I think splitting markers in separate TextElementState objects will simplify code.

atomicsulfate (changes requested):
I stopped reviewing cause a I think we should discuss my comment above before continuing.

@here/harp-mapview/lib/text/TextElementState.ts
StefanDachwitz: I probably mixed up TextRenderState and TextElementState, in my head I always meant individual TextRenderState as implemented in this PR.
If I were to defend the this PR, I would say that
# The visual difference of having separate **TextElementState** is only visible at extreme viewing angles and unusually long line markers (if used for road shields)
# It works _now_, i.e. it is finished
# I would question having separate textLayoutState's for line marker is making sense, at least not for road shields.
# The advantage is less memory usage, I guess.

I would suggest to implement this PR as is, but if the better shall be the enemy of the good, I can offer another way to implement the ticket.

atomicsulfate: Even if it works now, it's wrong conceptually, because of the limitations you mention in 1 and 3. It also makes the code more complicated, cause we to handle the case when the TextElementState has one RenderState vs an array of RenderStates.
The memory usage difference is 3 extra numbers per line marker, it doesn't sound relevant.

StefanDachwitz:
Resolved in another PR

PR #1983, opened 2020-11-19
Remove verbose debug messages.

robertoraggi:
This change removes diagnostic messages that show up often when
all the logging channels are enabled.

### DIFF
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -494,10 +494,6 @@ export class TextElementsRenderer {
             tileTextElementsChanged ||
             this.m_viewState.renderedTilesChanged;
 
-        logger.debug(
-            `FRAME: ${this.m_viewState.frameNumber}, ZOOM LEVEL: ${this.m_viewState.zoomLevel}`
-        );
-
         if (updateTextElements && this.m_addNewLabels) {
             this.m_textElementStateCache.clearVisited();
             this.updateTextElements(dataSourceTileList);
@@ -1111,8 +1107,6 @@ export class TextElementsRenderer {
      * @param dataSourceTileList - List of tiles to be rendered for each data source.
      */
     private updateTextElements(dataSourceTileList: DataSourceTileList[]) {
-        logger.debug("updateTextElements");
-
         if (updateStats) {
             updateStats.clear();
         }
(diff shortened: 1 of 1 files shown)


@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1350,9 +1344,6 @@ export class TextElementsRenderer {
 
         const placeNew = this.m_forceNewLabelsPass || placeNewTextElements;
         if (this.m_forceNewLabelsPass) {
-            if (!placeNewTextElements) {
-                logger.debug("Force new label pass");
-            }
```
atomicsulfate, line 1355: This log indicates when a pass for new labels is done because there were glyphs still loading, or labels are elevated on terrain and terrain data was not yet loaded. It's useful to debug issues related to that and to know why the text renderer keeps doing text placement even if there's no new labels added.

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -1111,8 +1107,6 @@ export class TextElementsRenderer {
      * @param dataSourceTileList - List of tiles to be rendered for each data source.
      */
     private updateTextElements(dataSourceTileList: DataSourceTileList[]) {
-        logger.debug("updateTextElements");
-
```
atomicsulfate, line 1115: This log is useful to know when the text elements are updated due to cache or tile invalidation.

atomicsulfate (changes requested):
It's ok if you want to remove the log with the frame number.

However, the other 2 are much less frequent and more informative. They appear many times in the test because the tests are design specifically to trigger those scenarios. 
They won't appear anymore in the tests once I merge this LoggerManagerTest fix:
https://github.com/heremaps/harp.gl/pull/1982

@here/harp-mapview/lib/text/TextElementsRenderer.ts
robertoraggi: yeah, I get that but if all the code starts to add these type of log messages then there's no way to enable the tracing and get something decent out of it. Can you add an additional check to enable these messages when you need them?

@here/harp-mapview/lib/text/TextElementsRenderer.ts
robertoraggi: I don't think this is a good reason.

atomicsulfate (approved):

PR #1984, opened 2020-11-20
MapView loadPromise method

nzjony:
Adding `loadPromise()` method to MapView

### DIFF
--- a/@here/harp-mapview/lib/MapView.ts
+++ b/@here/harp-mapview/lib/MapView.ts
@@ -881,6 +881,7 @@ export class MapView extends EventDispatcher {
     private readonly m_failedDataSources = new Set<string>();
     private readonly m_polarDataSource?: PolarTileDataSource;
     private readonly m_enablePolarDataSource: boolean = true;
+    private readonly m_promises: Array<Promise<any>> = [];
 
     // gestures
     private readonly m_raycaster: PickingRaycaster;
@@ -1154,7 +1155,9 @@ export class MapView extends EventDispatcher {
         // will initialize with an empty theme and updated when theme is loaded and set
         this.m_textElementsRenderer = this.createTextRenderer();
 
-        this.setTheme(getOptionValue(this.m_options.theme, MapViewDefaults.theme));
+        this.m_promises.push(
+            this.setTheme(getOptionValue(this.m_options.theme, MapViewDefaults.theme))
+        );
 
         this.update();
     }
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/test/BoundsGeneratorTest.ts
```diff
@@ -91,7 +91,7 @@ describe("BoundsGenerator", function() {
 
     afterEach(async function() {
         if (mapView !== undefined) {
-            await mapView.getTheme(); // Needed otherwise the dispose will cause log messages
+            await mapView.loadPromise(); // Needed otherwise the dispose will cause log messages
```
FraukeF, line 94: This looks like we should better fix the dispose method to dispose the promises correctly then to add a function for waiting for them. 

And expect a user to keep this order of API calls

@here/harp-mapview/lib/MapView.ts
```diff
@@ -1154,7 +1155,9 @@ export class MapView extends EventDispatcher {
         // will initialize with an empty theme and updated when theme is loaded and set
         this.m_textElementsRenderer = this.createTextRenderer();
 
-        this.setTheme(getOptionValue(this.m_options.theme, MapViewDefaults.theme));
+        this.m_promises.push(
```
FraukeF, line 1158: I am a bit confused about the approach, where I see that having a promise to wait for is a nice alternative to attaching to the THEME_LOADED Event, I dont see why we need to add the addDatasources/promises array as addDataSource is already async and returns a Promise that can be kept

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2133,54 +2136,74 @@ export class MapView extends EventDispatcher {
      * @param dataSource - The data source.
      */
     async addDataSource(dataSource: DataSource): Promise<void> {
-        const twinDataSource = this.getDataSourceByName(dataSource.name);
-        if (twinDataSource !== undefined) {
-            throw new Error(
-                `A DataSource with the name "${dataSource.name}" already exists in this MapView.`
-            );
-        }
+        const dsPromise = new Promise<void>(async resolve => {
```
FraukeF, line 2139: which addDataSource promises do you want to catch here?
Are we not just interessted in DataSources added internally without knowledge of the user? Then you could add the promise to the array where the actual addDataSource call happens, not internally here.

@here/harp-mapview/test/MapViewTest.ts
```diff
@@ -105,17 +105,13 @@ describe("MapView", function() {
...
-            addBackgroundDatasource: false,
-            enablePolarDataSource: false
+            canvas
         };
     });
 
     afterEach(async function() {
         if (mapView !== undefined) {
-            await mapView.getTheme();
+            await mapView.loadPromise();
```
FraukeF, line 114: same here, I think dispose should work without any API call to be waited for, or dispose should do so internally if needed

@here/harp-mapview/lib/MapView.ts
```diff
@@ -2133,54 +2136,74 @@ export class MapView extends EventDispatcher {
...
+        return dsPromise;
+    }
+
+    /**
+     * Promise that resolves when the theme and all DataSources added via `addDataSource` are added.
+     *
+     * Only necessary if you need to `await` for the `MapView` to be fully loaded before doing some
+     * next step.
+     */
+    async loadPromise() {
```
FraukeF, line 2202: Not sure we should add this API, whats the use case apart from the dispose method not working correctly?

@here/harp-mapview/lib/MapView.ts
nzjony: The problem is that it happens in the MapViewEnvironment, so I would somehow need to expose it externally.

@here/harp-mapview/lib/MapView.ts
nzjony: Yes, I originally exposed the promise from `addDataSource`  in `MapViewEnvironment`, but @atomicsulfate didn't like that approach.

@here/harp-mapview/test/BoundsGeneratorTest.ts
nzjony: Yes, that is another good point. At least with the `loadPromise` approach, you can be sure that the `MapView` is in some valid usage state. 

By `await` ing the dispose, you are making the test easier, but users can't know when the `MapView` is fully loaded.

I see pro/contra for both sides.

@here/harp-mapview/lib/MapView.ts
FraukeF: Hmm, then maybe we should recheck how we create the BackgroundDataSource, is it actually async, I think its neither connecting to anything nor loading not needing a theme... so it should be sync.. maybe we can solve how we handle the BackgroundDataSource...

@here/harp-mapview/lib/MapView.ts
nzjony: That is the only use case I know of so far.

@here/harp-mapview/lib/MapView.ts
nzjony: Yes, if we can sync its creation, and avoid the `loadPromise`, that would be best.

@here/harp-mapview/lib/MapView.ts
FraukeF: Then I think it might not be worth a new API to maintain, at least make it internal, but as discussed above, I think the whole thing boils down to the BackgroundDAtaSource being artificially async which should not be the case and solved there

nzjony: I don't know how you will solve this, because a constructor can't be async. Do you have any ideas?

FraukeF: Good point, it would need to basically call most of the content of MapView.addDatasource without the async unneeded stuff, which does not scale... 

If I understand right, we want to fix the dispose method with this... I still tend to thing that adding another function to call before might not solve the issue, as the dispose method can still be called without awaiting... 

So, if I get this right doing something like:

const mapView = new MapView({theme: sometheme});
mapView.dispose();

somehow crashes, caused by the theme loading and and the background datasource being added...

The theme loading can actually be canceled, the MapViewThemeManager is doing so, if a new theme is added... 
if we expose this to be used internally in MapView.dispose() could this solve the issue?

For the BackgroundDataSource, which part of it actually fails on disposal?

nzjony:
After discussion with @robertoraggi , @FraukeF , it was decided that it is best to wait until there is a clear use case to fix this rather than just being able to `dispose` quickly.

PR #1988, opened 2020-11-23
HARP-12581: Move TextElement construction into builders.

atomicsulfate:
Cleanup text element construction code to remove duplication. Preparation to set renderOrder for text elements.


@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -8,25 +8,20 @@ import {
...
-    LineMarkerTechnique,
-    PoiGeometry,
-    PoiTechnique
+    MapEnv,
+    PoiGeometry
 } from "@here/harp-datasource-protocol";
-import { ContextualArabicConverter } from "@here/harp-text-canvas";
 import { assert, assertExists, LoggerManager } from "@here/harp-utils";
 import * as THREE from "three";
+import { Vector3 } from "three";
```
FraukeF, line 20: THREE is already imported use THREE.Vector3

@here/harp-mapview/lib/poi/PoiManager.ts
```diff
@@ -44,6 +39,38 @@ interface ImageTextureDef {
     pixelRatio?: number;
 }
 
+function getPoiImageTexture(poiGeometry: PoiGeometry, index: number = 0): string | undefined {
```
FraukeF, line 42: just another minor, maybe you can drop the *Poi* in the name as its inside the PoiManager, getImageTexture might be enough... same for the funtions below

FraukeF (dismissed):
just some very minors... all in all I think this change is great and clears up a lot... 
I wish we would not have the name "Poi" all over the code but something  like "Marker" or "PointMarker" instead, but there is not much we can do about it at this point

@here/harp-mapview/lib/poi/PoiManager.ts
atomicsulfate: done

@here/harp-mapview/lib/poi/PoiManager.ts
atomicsulfate: done

FraukeF (approved):

PR #1991, opened 2020-11-27
HARP-13251: Add support for wrapping linestrings crossing the antimeridian.

robertoraggi:
This change adds the function `wrapLineString` that can be used
to wrap line features crossing the antimeridian.

### DIFF
--- a/@here/harp-geometry/lib/ClipLineString.ts
+++ b/@here/harp-geometry/lib/ClipLineString.ts
@@ -4,8 +4,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { EarthConstants, GeoCoordinates, webMercatorProjection } from "@here/harp-geoutils";
 import { Math2D } from "@here/harp-utils";
-import { Vector2 } from "three";
+import { Vector2, Vector3 } from "three";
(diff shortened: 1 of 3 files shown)


@here/harp-geometry/lib/ClipLineString.ts
```diff
@@ -153,3 +154,101 @@ export function clipLineString(
...
+
+    return {
+        left: wrapMultiLineStringHelper(multiLineString, WRAP_LEFT_CLIP_EDGES, 360),
+        middle: wrapMultiLineStringHelper(multiLineString, WRAP_MIDDLE_CLIP_EDGES, 0),
+        right: wrapMultiLineStringHelper(multiLineString, WRAP_RIGHT_CLIP_EDGES, -360)
+    };
+}
+
+const ec = EarthConstants.EQUATORIAL_CIRCUMFERENCE;
+const border = 0;
```
robertoraggi, line 239: the default border is 0 for now, I need to understand what to do with borders for these multi line strings, but if it turns out that we need to add borders we can do that in a follow up PR.

FraukeF (approved):

PR #1992, opened 2020-11-27
Enforce closing the rings of the GeoJson polygons.

robertoraggi:
Close the rings of the polygons when decoding GeoJson Polygon and
MultiPolygon features.


FraukeF (approved):
hmm, I am not a fan of doing special fixes that break specification, the data should be corrected on user side,  but I guess this one wont hurt

robertoraggi:
I also prefer to not close the polygons.

PR #2006, opened 2020-12-08
Remove duplicate readdirp from the yarn lockfile.

robertoraggi:
yarn-deduplicated didn't remove this entry.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8915,13 +8915,6 @@ readdirp@~3.2.0:
   dependencies:
     picomatch "^2.0.4"
 
-readdirp@~3.4.0:
-  version "3.4.0"
-  resolved "https://registry.yarnpkg.com/readdirp/-/readdirp-3.4.0.tgz#9fdccdf9e9155805449221ac645e8303ab5b9ada"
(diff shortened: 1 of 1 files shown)


FraukeF (commented):
It should not be removed as it is asking for a ~ version, not a ^  version,
so basically readdirp@~3.4.0 and readdirp@~3.5.0 have no overlapping version they can use, as the first has the scope of readdirp@~3.4.x and the second readdirp@~3.5.x

FraukeF:
I just checked, its even more restringent https://github.com/npm/node-semver#tilde-ranges-123-12-1 ... so definitly no overlapping versions there :)

FraukeF (dismissed):
As this is the removal of a unreferenced package and not as mentioned a duplicate one, approved

robertoraggi:
> I just checked, its even more restringent https://github.com/npm/node-semver#tilde-ranges-123-12-1 ... so definitly no overlapping versions there :)

In this particular case the issue is that yarn-deduplicate removed all the reachable references to `readdirp@~3.4.0` but it left both `readdirp@~3.4.0` and `readdirp@~3.5.0` in the lockfile. So running `yarn` on a fresh build resulted in a modified yarn.lock file. So assuming that we want to keep using yarn deduplicate the workflow will be to run again `yarn` after `yarn deduplicate` to prune the lock file.

FraukeF (approved):

PR #2010, opened 2020-12-10
HARP-13367: Update Three JS (0.123.0)

germanz:

# CI Check

Happy contributing!


@here/harp-lines/lib/HighPrecisionLines.ts
```diff
@@ -162,7 +162,7 @@ export class HighPrecisionLine extends THREE.Mesh implements HighPrecisionObject
         super.updateMatrixWorld(force);
 
         if (doUpdateMatrixWorldInverse) {
-            this.matrixWorldInverse.getInverse(this.matrixWorld);
+            this.matrixWorldInverse.copy(this.matrixWorld).invert();
```
robertoraggi, line 165: I believe `invert()` is new API introduced in 0.123. So if we merge this change harp.gl will not work anymore with 0.122. If we want to keep support 122 a bit longer it may be better to keep using the deprecate `Matrix.getInverse()` a bit longer and add a cast to fix the typing e.g.
`this.matrixWorldInverse.getInverse(this.matrixWorld) as Matrix4`  or something like that.

robertoraggi (approved):

@here/harp-lines/lib/HighPrecisionLines.ts
germanz: As discussed we decided to stick to the latest v0.123.

PR #2013, opened 2020-12-14
HARP-13368: update typescript to 4.1.2

AndriiHeonia:
Update typescript to 4.1.2

### DIFF
--- a/@here/harp-mapview/lib/poi/PoiRenderer.ts
+++ b/@here/harp-mapview/lib/poi/PoiRenderer.ts
@@ -680,7 +680,7 @@ export class PoiRenderer {
             .loadImage()
             .then(loadedImageItem => {
                 // Skip setup if image was not loaded (cancelled).
-                if (loadedImageItem.image) {
+                if (loadedImageItem && loadedImageItem.image) {
                     this.setupPoiInfo(poiInfo, loadedImageItem, env, imageTexture);
                 }
             })
--- a/@here/harp-mapview/lib/text/TextElementsRenderer.ts
+++ b/@here/harp-mapview/lib/text/TextElementsRenderer.ts
@@ -530,7 +530,7 @@ export class TextElementsRenderer {
         // TODO: this is an intermeditate solution, in the end this
         // should not create a new cache, but update the former one
         this.m_textStyleCache = new TextStyleCache(textStyles, defaultTextStyle);
-        this.m_textStyleCache.initializeDefaultTextElementStyle(DEFAULT_FONT_CATALOG_NAME);
+        this.m_textStyleCache.initializeDefaultTextElementStyle();
         await this.waitLoaded();
         this.m_textStyleCache.initializeTextElementStyles(this.m_textCanvases);
     }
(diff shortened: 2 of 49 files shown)


AndriiHeonia:

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -680,7 +680,7 @@ export class PoiRenderer {
             .loadImage()
             .then(loadedImageItem => {
                 // Skip setup if image was not loaded (cancelled).
-                if (loadedImageItem.image) {
+                if (loadedImageItem && loadedImageItem.image) {
```
atomicsulfate, line 683:
```suggestion
                if (loadedImageItem?.image) {
```

@here/harp-mapview/lib/text/TextElementsRenderer.ts
```diff
@@ -530,7 +530,7 @@ export class TextElementsRenderer {
         // TODO: this is an intermeditate solution, in the end this
         // should not create a new cache, but update the former one
         this.m_textStyleCache = new TextStyleCache(textStyles, defaultTextStyle);
-        this.m_textStyleCache.initializeDefaultTextElementStyle(DEFAULT_FONT_CATALOG_NAME);
+        this.m_textStyleCache.initializeDefaultTextElementStyle();
```
atomicsulfate, line 533: This change comes from Frauke's commit. Please rebase this PR on top of Frauke's merged change.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -589,38 +588,16 @@ export class SolidLineMaterial extends RawShaderMaterial
...
-     */
-    set fog(enable: boolean) {
-        this.m_fog = enable;
-        // Function may be called from THREE.js cause we override setter,
-        // in this case defines are not yet initialized and require late invalidation in
-        // SolidLineMaterial c-tor.
-        if (this.defines !== undefined) {
-            setShaderMaterialDefine(this, "USE_FOG", enable);
-        }
-    }
```
atomicsulfate, line 615: The problem now is that the material define is not updated when fog is enabled/disabled, so the shader won't change even if it's recompiled.
Since the fog settings are controlled only from MapViewFog, we can add a new method setFog in RawShaderMaterial that sets this.fog and then calls invalidateFog().
Then we just need to change MapViewFog.setFogInRawShaderMaterials() so that it calls setFog in raw materials.

@here/harp-materials/lib/RawShaderMaterial.ts
```diff
@@ -50,5 +52,12 @@ export class RawShaderMaterial extends THREE.RawShaderMaterial {
             delete (shaderParams as any).rendererCapabilities;
         }
         super(shaderParams);
+        this.invalidateFog();
+    }
+
+    protected invalidateFog() {
+        if (this.defines !== undefined && this.fog !== getShaderMaterialDefine(this, "USE_FOG")) {
+            setShaderMaterialDefine(this, "USE_FOG", this.fog);
+        }
```
atomicsulfate, line 61: The problem now is that without the setter in SolidLineMaterial, the material defines are not updated when fog is enabled/disabled, so the shader won't change even if it's recompiled.
Since the fog settings are controlled only from MapViewFog, we can make this method invalidateFog() public and call it in MapViewFog.setFogInRawShaderMaterials().

PR #2014, opened 2020-12-14
HARP-13369: update package dependencies

dbacarel:

# CI Check

Happy contributing!


dbacarel:

dbacarel:

dbacarel:

nzjony:

nzjony:
@dbacarel , when you use @, you can find the bot, just search "@harp", that might help because "@harpgl_bot" doesn't exist.

dbacarel:
interesting, thank you @nzjony. Maybe it was a coincidence but also via "harpgl_bot" it triggered the retest 🤷🏻‍♂️

nzjony:
> interesting, thank you @nzjony. Maybe it was a coincidence but also via "harpgl_bot" it triggered the retest 🤷🏻‍♂️

Hmm.. that reminds me, I remember that the @harpgl-bot is actually redundant, and that just "retest this please" is needed, see: https://plugins.jenkins.io/ghprb/ I forgot :(

dbacarel:
retest this please

.eslintrc.json
```diff
@@ -5,7 +5,9 @@
         "project": "./tsconfig.json"
     },
     "rules": {
-        "prettier/prettier": "error",
+        "prettier/prettier": ["error", {
```
FraukeF, line 8: why did you add this condition, did the default of eslint change?

.prettierrc.json
```diff
@@ -1,4 +1,5 @@
 {
     "printWidth": 100,
-    "tabWidth": 4
+    "tabWidth": 4,
+    "trailingComma": "none"
```
FraukeF, line 4: same here, did the default change?

@here/harp-atlas-tools/src/AtlasGenerator.ts
```diff
@@ -320,7 +320,7 @@ async function readAndProcessImage(
     logger.log(LogLevel.DEBUG, `Processing file: '${inFile}`);
     const imageDecoder = await imageFactory.loadImage(inFile);
     let imageOutput: ImageEncoder = imageDecoder.copy();
-    operations.forEach(imageProcess => {
+    operations.forEach((imageProcess) => {
```
FraukeF, line 323: why is this needed now, it changes almost all files, is there a way to configure the new packages to work as before?

FraukeF (commented):
Please double check to configure the packages, or maybe its only your local copy to not do all these reformattings

.eslintrc.json
dbacarel: I changed the rule because Prettier v2 default for trailing commas changed and the eslint fails without this extra rule.
Although I find it strange since I explicitly set "trailingComma": "none" in .prettierrc.json

.prettierrc.json
dbacarel: Yes, Prettier v2 has now trailing commas set to "es5" by default, before it was "none"
https://prettier.io/docs/en/options.html#trailing-commas

@here/harp-atlas-tools/src/AtlasGenerator.ts
dbacarel: Prettier v2 has now arrow-function-parentheses set to "always" by default, before it was "avoid".
https://prettier.io/docs/en/options.html#arrow-function-parentheses

I can add an additional rule in .prettierrc.json and remove those parentheses in case we want to stick with the previous style

FraukeF: yes, please, lets stick to the previous formatting rules as far as possible.

gairik (commented):
`
"prettier": "prettier -l "**/*.ts" "**/*.tsx" "**/*.json"",
"prettier:fix": "prettier --write "**/*.ts" "**/*.tsx" "**/*.json"",
`

@here/harp-atlas-tools/src/AtlasGenerator.ts
dbacarel: I added the rule to disable the parenthesis.
However, I couldn't do much about the space after the keyword 'function' on function expressions, e.g. `const foo = function () {..}`.
They don't seem to provide a way to suppress the rule - https://prettier.io/blog/2020/03/21/2.0.0.html#other-changes

dbacarel:
retest this please

FraukeF (commented):
This change is still enormous due to the prettier stuff, could you split it into two seperate PRs one without the prettier v2 update and one for the prettier v2 update

dbacarel:
Good point, alright, I excluded Prettier's update from this PR. I'll create a separate one for it

.eslintrc.json
```diff
@@ -41,7 +41,7 @@
         "@typescript-eslint/no-non-null-assertion": "off",
         "@typescript-eslint/explicit-function-return-type": "off",
 
-        "simple-import-sort/sort": "error",
+        "simple-import-sort/imports": "error",
```
FraukeF, line 44: why did this change?

@here/create-harp.gl-app/package.json
```diff
@@ -43,7 +43,7 @@
         "log-symbols": "^4.0.0"
     },
     "devDependencies": {
-        "mocha": "^7.2.0"
+        "mocha": "8.2.1"
```
FraukeF, line 46: please use ^8.2.1 (higher then version for all the updated packages)

FraukeF: also for all the following

@here/harp-examples/decoder/custom_decoder.ts
```diff
@@ -30,9 +30,8 @@ import { CUSTOM_DECODER_SERVICE_TYPE } from "./custom_decoder_defs";
 const MIN_GEOMETRY_HEIGHT = -100;
 
 export // snippet:custom_datasource_example_custom_decoder.ts
-class CustomDecoder extends ThemedTileDecoder
-// end:custom_datasource_example_custom_decoder.ts
-{
+class CustomDecoder extends ThemedTileDecoder {
```
FraukeF, line 33: will this throw an error if not changed?

@here/harp-examples/src/datasource_custom.ts
```diff
@@ -108,9 +108,8 @@ export namespace CustomDatasourceExample {
     const MIN_GEOMETRY_HEIGHT = -100;
 
     // snippet:custom_datasource_example_custom_data_provider.ts
-    class CustomDataProvider extends DataProvider
-    // end:custom_datasource_example_custom_data_provider.ts
-    {
+    class CustomDataProvider extends DataProvider {
```
FraukeF, line 111: same here, is  this a necessary reformatting?

@here/harp-map-controls/lib/MapAnimations.ts
```diff
@@ -452,7 +452,9 @@ export class CameraPanAnimation extends CameraAnimation {
             })
             .onUpdate(({ latitude, longitude, altitude }) => {
                 this.mapView.geoCenter = new GeoCoordinates(latitude, longitude, altitude);
-                this.mapView.camera.position.z = altitude;
+                if (altitude) {
```
FraukeF, line 455: this will not set the altitude if its 0, this changes the former behaviour... why is this change needed?

@here/harp-mapview/test/UtilsTest.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable simple-import-sort/sort */
+/* eslint-disable simple-import-sort/imports */
```
FraukeF, line 1: why do we disable the sort here at all, maybe we can remove theline?

@here/harp-materials/lib/MapMeshMaterials.ts
```diff
@@ -251,7 +251,7 @@ export function hasExtrusionFeature(material: any): material is ExtrusionFeature
     return "extrusionRatio" in material;
 }
 
-namespace DisplacementFeature {
+namespace DisplacementFeatureNS {
```
FraukeF, line 254: why change the name?

FraukeF (commented):
Thanks for removing the prettier update, its already a bit better, but I still think this change is way to big, if it causes issues later on, it will be difficult to break it down to where it actually came from... e.g. the yarn.lock file changes are unreviewable :( 

Please update always to ^<version> not to fixed ones ...

I am not 100% sure how a good split would look, one package one PR, seems to create a lot of PRs... 

I could imagine something like: 
maybe one PR with the testing devDependencies like mocha, sinoon, chai....
another PR with the packaging utilities the webpack, webpack-merge, ts-loader ... stuff

one for the tween update
one for the olp-sdk update
one for the eslint related stuff

let me know what you think

dbacarel:
>> Please update always to ^ not to fixed ones ...

Sure, I used `./scripts/reset-dependency-version` to update the packages and it set to a fixed version (so, without ^). I thought it was the intended approach. Since it's not, I'll add the ^ manually.

>> [...] it will be difficult to break it down to where it actually came from... e.g. the yarn.lock file changes are unreviewable :(

Agree.

I'll create individual PRs for the following:

- eslint related packages
- packing utils
- tween
- olp-sdk

After that, I'll revert the changes related to these updates from this PR and see if it becomes of a reasonable size.

FraukeF:
> > > Please update always to ^ not to fixed ones ...
> 
> Sure, I used `./scripts/reset-dependency-version` to update the packages and it set to a fixed version (so, without ^). I thought it was the intended approach. Since it's not, I'll add the ^ manually.

you can use the script also with the ^version, so you dont have to do this manually

dbacarel:
>> you can use the script also with the ^version, so you don't have to do this manually

Thank you, I thought that it accepted only the numeric version.


So, I split this into the following PRs:

### Packaging tools - #2039
### ESLint and deps - #2033
### OLP SDKs - #2036 
### Tween - #2035 
### Testing tools - #2037 


I realized that the rest of the updates in this PR are mostly patches, so nothing that's worth dedicating a PR.
My suggestion is to close this one and move our attention to the ones above.

PR #2016, opened 2020-12-16
HARP-13493: Make maxZoomLevel value in technique non inclusive

germanz:
Signed-off-by: German Zargaryan <german.zargaryan@here.com>

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
+++ b/@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
@@ -693,7 +693,7 @@ export class StyleSetEvaluator {
                 }
             }
 
-            if (typeof maxZoomLevel === "number" && zoomLevel > maxZoomLevel) {
+            if (typeof maxZoomLevel === "number" && zoomLevel >= maxZoomLevel) {
                 return false;
             }
         }
--- a/@here/harp-datasource-protocol/lib/TechniqueParams.ts
+++ b/@here/harp-datasource-protocol/lib/TechniqueParams.ts
@@ -264,7 +264,7 @@ export interface BaseTechniqueParams {
     minZoomLevel?: DynamicProperty<number>;
 
     /**
-     * Maximum zoom level. If the current zoom level is larger, the technique will not be used.
+     * Maximum zoom level. If the current zoom level is equal to or greater, the technique will not be used.
      */
     maxZoomLevel?: DynamicProperty<number>;
 }
--- a/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
+++ b/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
@@ -588,8 +588,8 @@ describe("StyleSetEvaluator", function() {
             ])
         );
 
-        assert.isNotEmpty(
-            getMatchingTechniques({ ...defaultProperties, $zoom: 15 }, [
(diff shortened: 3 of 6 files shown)


@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
```diff
@@ -693,7 +693,7 @@ export class StyleSetEvaluator {
                 }
             }
 
-            if (typeof maxZoomLevel === "number" && zoomLevel > maxZoomLevel) {
```
atomicsulfate, line 696: We need to document this change in BaseTechniqueParams in @here/harp-datasource-protocol/lib/TechniqueParams.ts.

atomicsulfate (commented):
Please document the behaviour change as part of the next release notes:
https://github.com/heremaps/harp.gl/releases/

There's not draft of next release (0.22) so you can create a new one. Just follow the same format as the previous releases. Once done click on "Save draft", not on "Publish release".

@here/harp-datasource-protocol/lib/StyleSetEvaluator.ts
germanz: done

atomicsulfate (dismissed):

nzjony:
I agree with the idea of making it non-inclusive, however the naming is unfortunate, the `max` in `maxZoomLevel` is short for "maximum", which (according to google), means: "the greatest amount, extent, or intensity possible, permitted, or recorded.", i.e. the maximum always refers to an attainable value, however this changes breaks this concept.

I think a better name could be start `beginZoomLevel` and `limitZoomLevel`, where "limit" is (at least in mathematics) clear that it is a value that we get close to, but don't reach.

germanz:
retest this please

germanz:
@nzjony I agree with you about the definition of the term. However, I think it is better to keep it simple and document the range `[minZoomLevel, maxZoomLevel)` so that the upper edge is not reachable. 

It will be unfortunate to break the compatibility by renaming the property or introducing another one and cause confusion. Other Map rendering APIs also use the same terms.

@here/harp-datasource-protocol/lib/TechniqueParams.ts
```diff
@@ -264,7 +264,7 @@ export interface BaseTechniqueParams {
     minZoomLevel?: DynamicProperty<number>;
 
     /**
-     * Maximum zoom level. If the current zoom level is larger, the technique will not be used.
+     * Maximum zoom level. If the current zoom level is equal to or greater, the technique will not be used.
```
robertoraggi, line 267: You may want to check with our Design team before going forward with this change of semantics.

robertoraggi: Also, shouldn't you update the semantics of all the zoom/data level constraints? e.g. `DataSource.maxDisplayLevel`, `DataSource.maxDataLevel`, `MapView.maxZoomLevel`, `MapControls.maxZoomLevel`?

robertoraggi (approved):

@here/harp-datasource-protocol/lib/TechniqueParams.ts
germanz: As agreed for the time being we leave the semantic of those properties as-is, both edges inclusive.

PR #2026, opened 2021-01-05
Update typescript to 4.1.2

AndriiHeonia:
Update typescript to 4.1.2

### DIFF
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -738,7 +738,7 @@ export function applyBaseColorToMaterial(
     }
 
     opacity = THREE.MathUtils.clamp(opacity, 0, 1);
-    if (material instanceof CirclePointsMaterial || material instanceof SolidLineMaterial) {
+    if (material instanceof RawShaderMaterial) {
         material.setOpacity(opacity);
     } else {
         material.opacity = opacity;
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -738,7 +738,12 @@ export function applyBaseColorToMaterial(
     }
 
     opacity = THREE.MathUtils.clamp(opacity, 0, 1);
-    material.opacity = opacity;
+    if (material instanceof CirclePointsMaterial || material instanceof SolidLineMaterial) {
+        material.setOpacity(opacity);
+    } else {
+        material.opacity = opacity;
+    }
+
     materialColor.setRGB(r, g, b);
 
     const opaque = opacity >= 1.0;
--- a/@here/harp-mapview/lib/MapMaterialAdapter.ts
+++ b/@here/harp-mapview/lib/MapMaterialAdapter.ts
@@ -5,7 +5,12 @@
  */
 
 import { ColorUtils, Expr, getPropertyValue, Value } from "@here/harp-datasource-protocol";
-import { disableBlending, enableBlending } from "@here/harp-materials";
+import {
+    CirclePointsMaterial,
+    disableBlending,
+    enableBlending,
+    SolidLineMaterial
+} from "@here/harp-materials";
 import * as THREE from "three";
 
 import { evaluateColorProperty } from "./DecodedTileHelpers";
@@ -237,8 +242,17 @@ export class MapMaterialAdapter {
             color = parsed;
         }
         const { r, g, b, a } = ColorUtils.getRgbaFromHex(color ?? 0xff0000);
+
         const actualOpacity = a * THREE.MathUtils.clamp(opacity ?? 1, 0, 1);
-        this.material.opacity = actualOpacity;
+        if (
+            this.material instanceof CirclePointsMaterial ||
+            this.material instanceof SolidLineMaterial
+        ) {
+            this.material.setOpacity(actualOpacity);
+        } else {
+            this.material.opacity = actualOpacity;
+        }
+
         (this.material as any).color.setRGB(r, g, b);
 
         const opaque = actualOpacity >= 1.0;
--- a/@here/harp-mapview/lib/MapMaterialAdapter.ts
+++ b/@here/harp-mapview/lib/MapMaterialAdapter.ts
@@ -5,12 +5,7 @@
  */
 
 import { ColorUtils, Expr, getPropertyValue, Value } from "@here/harp-datasource-protocol";
-import {
-    CirclePointsMaterial,
-    disableBlending,
-    enableBlending,
-    SolidLineMaterial
-} from "@here/harp-materials";
+import { disableBlending, enableBlending, RawShaderMaterial } from "@here/harp-materials";
 import * as THREE from "three";
 
 import { evaluateColorProperty } from "./DecodedTileHelpers";
@@ -244,10 +239,7 @@ export class MapMaterialAdapter {
         const { r, g, b, a } = ColorUtils.getRgbaFromHex(color ?? 0xff0000);
 
         const actualOpacity = a * THREE.MathUtils.clamp(opacity ?? 1, 0, 1);
-        if (
-            this.material instanceof CirclePointsMaterial ||
-            this.material instanceof SolidLineMaterial
-        ) {
+        if (this.material instanceof RawShaderMaterial) {
             this.material.setOpacity(actualOpacity);
         } else {
             this.material.opacity = actualOpacity;
--- a/@here/harp-mapview/lib/MapViewFog.ts
+++ b/@here/harp-mapview/lib/MapViewFog.ts
@@ -5,6 +5,7 @@
  */
 import { Fog } from "@here/harp-datasource-protocol";
 import { HighPrecisionLineMaterial } from "@here/harp-materials";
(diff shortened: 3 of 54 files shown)


AndriiHeonia:

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -738,7 +738,12 @@ export function applyBaseColorToMaterial(
     }
 
     opacity = THREE.MathUtils.clamp(opacity, 0, 1);
-    material.opacity = opacity;
+    if (material instanceof CirclePointsMaterial || material instanceof SolidLineMaterial) {
```
atomicsulfate, line 741: Instead of having to enumerate here all our custom materials, we can do this for any material than inherits from RawShaderMaterial, like SolidLineMaterial does. CirclePointsMaterial can be changed to derive from RawShaderMaterial too. The only difference is that we need to pass any defines explicitly, but the shader code for that material doesn't use any, so I guess it's only a matter of changing the base class.
That way we can move setOpacity to RawShaderMaterial.

@here/harp-materials/lib/CirclePointsMaterial.ts
```diff
@@ -117,20 +116,9 @@ export class CirclePointsMaterial extends THREE.ShaderMaterial {
...
-    /**
-     * Set circle opacity.
-     */
-    set opacity(opacity: number) {
-        this.m_opacity = opacity;
-
-        // Base constructor may set opacity before uniform being created.
+    setOpacity(opacity: number) {
+        this.opacity = opacity;
+        // Base constructor may set opacity before uniform being created:
```
atomicsulfate, line 121: Derive this class from RawShaderMaterial and move this method there.

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -652,21 +623,17 @@ export class SolidLineMaterial extends RawShaderMaterial
...
+    setOpacity(value: number) {
+        this.opacity = value;
         // Setting opacity before uniform being created requires late invalidation,
         // call to invalidateOpacity() is done at the end of c-tor.
         if (this.uniforms?.opacity) {
             this.uniforms.opacity.value = value;
         }
 
-        this.stencilWrite = this.m_opacity < 0.98;
+        this.stencilWrite = value < 0.98;
```
atomicsulfate, line 636: As I mentioned, setOpacity could be moved to RawShaderMaterial. If this stencilWrite flag is specific of SoliLineMaterial,override setOpacity here.

@here/harp-mapview/lib/MapMaterialAdapter.ts
```diff
@@ -237,8 +242,17 @@ export class MapMaterialAdapter {
             color = parsed;
         }
         const { r, g, b, a } = ColorUtils.getRgbaFromHex(color ?? 0xff0000);
+
         const actualOpacity = a * THREE.MathUtils.clamp(opacity ?? 1, 0, 1);
-        this.material.opacity = actualOpacity;
+        if (
+            this.material instanceof CirclePointsMaterial ||
+            this.material instanceof SolidLineMaterial
```
atomicsulfate, line 249:
```suggestion
        if (
            this.material instanceof RawShaderMaterial
```

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -31,12 +31,18 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture):
 
     if (Array.isArray(material)) {
         material.forEach(mat => {
-            if (hasDisplacementFeature(mat)) {
+            if (hasDisplacementFeature(mat) && mat.displacementMap !== displacementMap) {
                 mat.displacementMap = displacementMap;
+                mat.needsUpdate = true;
```
atomicsulfate, line 36:
```suggestion
            if (hasDisplacementFeature(mat) && mat.displacementMap !== displacementMap) {
                mat.displacementMap = displacementMap;
                mat.needsUpdate = true;
                mat.displacementMap.needsUpdate = true;
```

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
```diff
@@ -31,12 +31,18 @@ function overlayObject(object: TileObject, displacementMap: THREE.DataTexture):
...
             }
         });
-    } else if (material && hasDisplacementFeature(material)) {
+    } else if (
+        material &&
+        hasDisplacementFeature(material) &&
+        material.displacementMap !== displacementMap
+    ) {
         material.displacementMap = displacementMap;
+        material.needsUpdate = true;
```
atomicsulfate, line 45: Same here, Missing setting the needsUpdate flag for the displacement texture. 
You missed setting the flags in setupTerrainMaterial() in TileGeometryCreator, next to the line doing:
```
stdMaterial.displacementMap = null;
```
Maybe it makes sense to move out to a function.

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
atomicsulfate: Missing setting the needsUpdate flag for the displacement texture

atomicsulfate (changes requested):

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
AndriiHeonia: I moved this code to DisplacementFeature.ts#setDisplacementMapToMaterial and set needsUpdate flag on the texture as well

@here/harp-mapview/lib/geometry/overlayOnElevation.ts
AndriiHeonia: I fixed this one as well.

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -680,7 +680,7 @@ export class PoiRenderer {
             .loadImage()
             .then(loadedImageItem => {
                 // Skip setup if image was not loaded (cancelled).
-                if (loadedImageItem.image) {
+                if (loadedImageItem && loadedImageItem.image) {
```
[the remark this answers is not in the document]
AndriiHeonia: Fixed by @ThFischer

@here/harp-mapview/lib/MapViewFog.ts
```diff
@@ -144,25 +145,25 @@ export class MapViewFog {
...
+                    // We may skip redundant updates:
+                    material.fog !== enableFog
+                ) {
+                    material.fog = enableFog;
+                    if (material instanceof RawShaderMaterial) {
+                        // Fog properties can't be easily changed at runtime (once the material
+                        // is rendered at least once) and thus requires building of new shader
+                        // program - force material update.
+                        material.invalidateFog();
+                        material.needsUpdate = true;
```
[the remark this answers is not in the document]
AndriiHeonia: material.needsUpdate = true; call was removed by @ThFischer

@here/harp-materials/lib/SolidLineMaterial.ts
```diff
@@ -866,20 +833,14 @@ export class SolidLineMaterial extends RawShaderMaterial
...
-        if (this.m_fog !== getShaderMaterialDefine(this, "USE_FOG")) {
-            setShaderMaterialDefine(this, "USE_FOG", this.m_fog);
-        }
-    }
-
     private invalidateOpacity() {
-        if (this.m_opacity !== this.uniforms.opacity.value) {
-            this.uniforms.opacity.value = this.m_opacity;
+        if (this.opacity !== this.uniforms.opacity.value) {
+            this.uniforms.opacity.value = this.opacity;
```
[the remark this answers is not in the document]
AndriiHeonia: invalidateOpacity method was removed by @ThFischer

@here/harp-mapview/lib/DecodedTileHelpers.ts
AndriiHeonia: - CirclePointsMaterial now extends RawShaderMaterial
- setOpacity was moved to RawShaderMaterial
- those two instanceof checks are gone and replaced by one "instanceof RawShaderMaterial". 
- It is still possible to call `material = new CirclePointsMaterial()` without parameters, but `material.size` will not be accessible right after initialization. As discussed, in real use cases, this constructor is always called with parameters, and constructor without parameters is there just because three.js `Material.clone()` requires it, but `clone` doesn't need to access our custom properties.  I adjusted unit-tests accordingly.

@here/harp-materials/lib/CirclePointsMaterial.ts
AndriiHeonia: done

@here/harp-materials/lib/SolidLineMaterial.ts
AndriiHeonia: done

@here/harp-mapview/lib/MapMaterialAdapter.ts
AndriiHeonia: done

atomicsulfate (approved):

PR #2032, opened 2021-01-07
HARP-13721: update Prettier to v2

dbacarel:
Signed-off-by: Daniele Bacarella <daniele.bacarella@here.com>

# CI Check

Happy contributing!


dbacarel:
This is a separate PR from #2014 as it touches several files and introduces minor modifications to the code style.

### Prettier v2
- ` arrowParens` is set to `"always"` by default, before it was` "avoid"` -https://prettier.io/docs/en/options.html#arrow-function-parentheses
This rule was overridden in `.prettierrc.json` to stick with the previous style in use.

- `trailingComma` is set to `"es5"` by default, before it was `"none"` - https://prettier.io/docs/en/options.html#trailing-commas
This rule was overridden in `.prettierrc.json` to stick with the previous style in use.

- Prettier now insert a space before the parenthesis and the keyword ```function```, for example:
`const foo = function() {...}` -> `const foo = function () {...}`
It doesn't provide a rule to suppress it and I couldn't find a workaround either

robertoraggi (dismissed):

nzjony (approved):

PR #2040, opened 2021-01-12
HARP-11894 Fixing partially missing tiles

nzjony:
The problem was that the perspective division looses information about
whether the tile was behind the camera or behind the far plane, which
means we can't sort the tiles correctly. Tiles behind the near plane
are therefore given a distance of -1 and those past the far plane
are given a distance of +1. Note, we know that the tiles are visible
because they intersect the frustum.


nzjony:
retest this please

FraukeF (approved):

nzjony:
retest this please

Issue #2019, opened 2021-01-01
Satellite/vector data misalignment


brent-williams:
Overlaying HERE satellite & vector data shows a clear systematic misalignment at some zoom levels.

To reproduce, overlay Vector Tile API data over e.g. the 'datasource_satellite-tile' example by adding the code below.  There is a clear bias at many zoom levels whereby the vectors are drawn several pixels above and to the left of the satellite data.  

```
  const omvDataSourceStreets = new OmvDataSource({
    baseUrl: "https://vector.hereapi.com/v2/vectortiles/base/mc",
    apiFormat: APIFormat.HereV1,
    authenticationCode: apikey,
  });
  mapView.addDataSource(omvDataSourceStreets).then(() => {
    const styles: any = [
      {
        when: `$geometryType ^= 'poly'`,
        technique: "solid-line",
        renderOrder: 2010,
        attr: {
          color: "#ffffff",
          lineWidth: "1px"
        }
      }
    ];
    omvDataSourceStreets.setStyleSet(styles);
    mapView.update();
  });
```

Expected reasonable correspondence between satellite & vector data or at least no systematic bias towards the upper-left.

Attached harp.gl example + Google Earth Pro for reference.  Bias is clear everywhere but especially obvious on islands, a few circled in red for clarity.  Taken on macOS Catalina, latest Chrome Version 87.0.4280.88 (Official Build) (x86_64), harp.gl 0.21.1.

![a](https://user-images.githubusercontent.com/1694687/103435145-8825a900-4c56-11eb-94f2-2983204dd7fd.png)
<img width="1391" alt="b" src="https://user-images.githubusercontent.com/1694687/103435148-9247a780-4c56-11eb-87d8-986976581139.png">

FraukeF:
Hi @brent-williams ,

thanks for reporting this, I can actually reproduce it also with the other raster tile formats, I ll create an internal issue to investigate this issue, and let you know when there is a solution.

nzjony:
@brent-williams , I did some investigation, and it seems there is an offset in our satellite images, but, as far as I can tell, it isn't related to harp.gl. I compared also in another internal renderer and got the same issue.

I also compared with a google satellite image, and there seems to be a 1px shift south east.

https://user-images.githubusercontent.com/2533428/104513544-1a7f6280-55f0-11eb-9fb8-0496d94c2034.mp4

brent-williams:
[...]

PR #2047, opened 2021-01-14
MINOR: Relates to #2019, fix broken storageLevelOffset

nzjony:
storageLevelOffset was not passed through to DataSource correctly.

### DIFF
--- a/@here/harp-webtile-datasource/lib/HereWebTileDataSource.ts
+++ b/@here/harp-webtile-datasource/lib/HereWebTileDataSource.ts
@@ -399,10 +399,10 @@ export class HereWebTileDataSource extends WebTileDataSource {
             minDataLevel: 1,
             maxDataLevel: 20,
             resolution: m_options.resolution,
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/lib/DataSource.ts
```diff
@@ -63,6 +63,7 @@ export interface DataSourceOptions {
     maxDisplayLevel?: number;
     /**
      * Storage level offset applied to this `DataSource`.
+     * @default 0
```
robertoraggi, line 66: Isn't the default "datasource dependent"? e.g. in HereWebTileDataSource.ts the default seem to be -1.

nzjony: Yes, I refer to the DataSource class itself, which defaults to 0. But I can remove if you prefer?

nzjony: I have removed it.

robertoraggi (approved):

PR #2050, opened 2021-01-14
Generate a JSON schema for both the legacy and the new flat theme format

robertoraggi:
Simplify the typings describing legacy and flat themes.

This change makes it possible to generate a JSON schema that can validate both theme format.

### DIFF
--- a/@here/harp-datasource-protocol/lib/Theme.ts
+++ b/@here/harp-datasource-protocol/lib/Theme.ts
@@ -29,7 +29,7 @@ import {
  * Map theme is used to define what features are shown and how the map is styled, for example
  * which lightning is used or whether fog should be displayed.
  */
-export interface Theme {
+export interface BaseTheme {
     /**
      * The URI of the JSON schema describing themes.
      */
@@ -43,7 +43,7 @@ export interface Theme {
      * overwrite these defined in base theme.
      */
 
-    extends?: string | Theme | Array<string | Theme>;
+    extends?: string | Theme | Array<string | Theme | FlatTheme>;
 
     //
     // TODO: We support also [[FlatTheme]], but it's not exposed here since ts-json-schema-generator
@@ -94,11 +94,6 @@ export interface Theme {
      * The definitions exported by these theme.
      */
     definitions?: Definitions;
-
-    /**
-     * Map styles available for datasources used to render the map.
-     */
-    styles?: Styles;
 
     /**
      * Define the style to render different types of text used on the map.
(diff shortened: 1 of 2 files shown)


@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -196,17 +191,34 @@ export interface StylePriority {
...
 /**
- * A type representing HARP themes with all the styleset declarations
- * grouped in one `Array`.
- *
- * @internal This type will merge with {@link Theme}.
+ * Map theme is used to define what features are shown and how the map is styled, for example
+ * which lightning is used or whether fog should be displayed.
  */
-export type FlatTheme = Omit<Theme, "styles"> & {
+export interface Theme extends BaseTheme {
```
robertoraggi, line 197: this should probably be renamed to LegacyTheme and marked as deprecate.

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -196,17 +191,34 @@ export interface StylePriority {
...
+     * Map styles available for datasources used to render the map.
+     */
+    styles?: Styles;
+}
+
+/**
+ * Map theme is used to define what features are shown and how the map is styled, for example
+ * which lightning is used or whether fog should be displayed.
+ */
+export interface FlatTheme extends BaseTheme {
```
robertoraggi, line 208: This should be renamed to Theme.

@here/harp-datasource-protocol/lib/Theme.ts
```diff
@@ -196,17 +191,34 @@ export interface StylePriority {
...
-};
+}
+
+/**
+ * Map theme is used to define what features are shown and how the map is styled, for example
+ * which lightning is used or whether fog should be displayed.
+ *
+ * @internal
+ */
+export type ThemeOrFlatTheme = Theme | FlatTheme;
```
robertoraggi, line 221: This is the entry point used to generate a JSON schema supporting both the legacy and the new flat format. In theory we should remove this as soon as we stop supporting LegacyThemes.

@here/harp-datasource-protocol/lib/Theme.ts
FraukeF: I agree, I think we should update Theme to be the FlatTheme and deprecate the former as something like LegacyTheme... have you seen: HARP-12835

robertoraggi:
rebase please

nzjony:
rebase please

PR #2054, opened 2021-01-15
HARP-13770: Update the copyright header

ThFischer:
Signed-off-by: Fischer, Thomas <Thomas.Fischer@HERE.com>

### DIFF
--- a/@here/create-harp.gl-app/test/test.js
+++ b/@here/create-harp.gl-app/test/test.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Copyright (C) 2017-2021 HERE Europe B.V.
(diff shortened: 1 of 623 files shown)


germanz:
According to the convention, we should keep the copyright notice in the format:

``Copyright (C) 201<x>-202<y> HERE Global B.V. and its affiliate(s)``

where ``x`` is created and ``y`` is the modified date of the file. When those are equal format it is simply 

``Copyright (C) 201<x> HERE Global B.V. and its affiliate(s)``

nzjony (approved):

ThFischer:
A new solution uses `git log` to determine the first and last commit to define the expected created and lastModified year of the copyright notice. Therefore I have to close this PR/Draft because it unnecessarily modified copyright headers.

PR #2055, opened 2021-01-15
HARP-13235: Fix altitude in mercator

atomicsulfate:


@here/harp-geoutils/lib/projection/MercatorProjection.ts
```diff
@@ -92,7 +92,7 @@ class MercatorProjection extends Projection {
         result.y =
             (MercatorProjection.latitudeClampProject(geoPoint.latitudeInRadians) * 0.5 + 0.5) *
             this.unitScale;
-        result.z = geoPoint.altitude ?? 0;
+        result.z = (geoPoint.altitude ?? 0) * this.getScaleFactor(result);
```
robertoraggi, line 95: i don't think this is a good idea. This will enforce applying the scaling factor to Z at projection time, resulting always in different altitudes for objects with large latitude span. So many objects will have oblique planes and the their normal vector will also different.

atomicsulfate: As discussed, I now scale the height on decoding depending on the technique parameter constantHeight, which now is available in all techniques.

robertoraggi (dismissed):

robertoraggi (approved):

PR #2062, opened 2021-01-19
HARP-13852 Increase precision, otherwise tilt is too inaccurate

nzjony:
- Improve the epsilon used when comparing the dot product.
- When the tilt animation has finished, a final check is done to see if the targeted tilt is reached, and if not, we apply the delta needed to reach the desired tilt.

### DIFF
--- a/@here/harp-map-controls/lib/MapControls.ts
+++ b/@here/harp-map-controls/lib/MapControls.ts
@@ -618,6 +618,9 @@ export class MapControls extends EventDispatcher {
             this.m_tiltRequested = undefined;
         }
 
+        // Whether the tilt animation has reached full duration & a final frame is rendered. We need
+        // this to know when to stop the tilt (and hence deregister the methon )
+        let tiltAnimationFinished = false;
         if (this.inertiaEnabled) {
             if (!this.m_tiltIsAnimated) {
                 this.m_tiltIsAnimated = true;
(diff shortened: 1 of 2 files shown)


@here/harp-map-controls/lib/MapControls.ts
```diff
@@ -628,6 +628,22 @@ export class MapControls extends EventDispatcher {
...
+                        MapViewUtils.orbitAroundScreenPoint(
+                            this.mapView,
+                            0,
+                            0,
+                            0,
+                            deltaAngle,
+                            this.m_maxTiltAngle
+                        );
+                        this.updateMapView();
+                    }
```
atomicsulfate, line 646: I think the call to this.stopTilt() is done to soon. It sets:
 this.m_targetedTilt = this.m_currentTilt = undefined;

Then later we pass to easeOutCubic() as target value this.targetedTilt, which returns the current mapView.tilt if this.m_targetedTilt  is undefined.

If the stopTilt() call is done after easOutCubic, together with your epsilon changes, it works fine for me.

atomicsulfate, line 646: I think the call to this.stopTilt() is done to soon. It sets:
this.m_targetedTilt = this.m_currentTilt = undefined;

Then later we pass to easeOutCubic() as target value this.targetedTilt, which returns the current mapView.tilt if this.m_targetedTilt is undefined.

If the stopTilt() call is done after easOutCubic, together with your epsilon changes, it works fine for me.

nzjony: Thanks @atomicsulfate. I have pushed a fix.

atomicsulfate (dismissed):
Looks good. It'd be even better if we add a unit test to check we actually reach 0 tilt.

nzjony:
@atomicsulfate , I agree, I will put the test in a separate PR, so we can at least get this into the release.

nzjony:
retest this please

atomicsulfate (approved):

PR #2065, opened 2021-01-20
HARP-11233: Store the clipping states in the ring vertices

robertoraggi:
This change modifies `clipPolygon` to store the state of the clipping
in the vertices. Also, this change modifies the clipping planes
to ensure that vertices on the edge are not marked as "clipped".


nzjony:
@robertoraggi , reviewing, I just want to try something locally.

nzjony (approved):

PR #2092, opened 2021-02-02
MINOR: Remove travis CI, because github actions now covers same funct…

nzjony:
…ionality


FraukeF:
Lets keep it a couple of days, to see all goes smooth and until we have moved the release process over too

PR #2101, opened 2021-02-05
HARP-13229: Fix possible precision issues when converting coords from world to local space

robertoraggi:
There is a precision issue and the clipping fails to correctly
classify the inner rings of clipped polygons at level 2.


robertoraggi:
rebase please

atomicsulfate (approved):

PR #2104, opened 2021-02-09
HARP-11013: Support GeoJson feature ids

atomicsulfate:
So far feature ids were expected to be numbers, therefore original feature ids coming from geojson data were ignored.

### DIFF
--- a/@here/harp-datasource-protocol/lib/DecodedTile.ts
+++ b/@here/harp-datasource-protocol/lib/DecodedTile.ts
@@ -87,9 +87,10 @@ export interface PathGeometry {
 
 /**
  * Attributes corresponding to some decoded geometry. It may be either a map
- * of multiple attributes or just a number with the geometry's feature id.
+ * of multiple attributes or just a string with the geometry's feature id (id numbers are
+ * deprecated).
  */
-export type AttributeMap = {} | number;
+export type AttributeMap = {} | string | number;
 
 /**
  * This object keeps textual data together with metadata to place it on the map.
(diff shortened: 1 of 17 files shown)


atomicsulfate:

@here/harp-mapview-decoder/lib/GeoJsonTiler.ts
```diff
@@ -69,7 +69,7 @@ export class GeoJsonTiler implements ITiler {
             buffer: BUFFER, // tile buffer on each side
             lineMetrics: false, // whether to calculate line metrics
             promoteId: null, // name of a feature property to be promoted to feature.id
-            generateId: true, // whether to generate feature ids. Cannot be used with promoteId
```
atomicsulfate, line 72: this was overwritting featureIds coming from geojson data. It was used to generate sequential numbers that could be use later to match the output features with the original geojson features to add the original geometry, see below.

@here/harp-mapview-decoder/lib/GeoJsonTiler.ts
```diff
@@ -85,27 +85,7 @@ export class GeoJsonTiler implements ITiler {
         const tile = index.getTile(tileKey.level, tileKey.column, tileKey.row);
         if (tile !== null) {
             tile.layer = indexId;
-            for (const feature of tile.features) {
-                feature.originalGeometry = this.getOriginalGeometry(feature, index.geojson);
```
atomicsulfate, line 89: Here the original geometry was attached to the output features, but it's never used. It was previously used for lines, but Roberto changed that code half a year ago, already discussed with him.

@here/harp-datasource-protocol/lib/DecodedTile.ts
```diff
@@ -87,9 +87,10 @@ export interface PathGeometry {
 
 /**
  * Attributes corresponding to some decoded geometry. It may be either a map
- * of multiple attributes or just a number with the geometry's feature id.
+ * of multiple attributes or just a string with the geometry's feature id (id numbers are
+ * deprecated).
  */
-export type AttributeMap = {} | number;
+export type AttributeMap = {} | string | number;
```
germanz, line 93: Shall we introduce an alias type like ``FeatureId``? BTW we also have some type definitions for GeoJSON data here ``@here/harp-datasource-protocol/lib/GeoJsonDataType.ts`` and there it is still defined as ``id?: string;``

atomicsulfate: GeoJsonDataType doesn't need to be changed, it just reflects the types according to the geojson standard. We just retain the number temporally for backwards compatibility with anyone that uses OMV data, cause we are now decoding numerical ids.

I don't think it's worth adding a FeatureId type since it'll be a simple string once we remove the backwards compatibility.

germanz: OK

robertoraggi, line 93: the type of AttrinbuteMap is a bit strange. Maybe it would be better to use `JsonObject | string | number` or JsonValue ?

robertoraggi (approved):

PR #2112, opened 2021-02-12
HARP-13952: Handle error states, like 404 in OmvRestClient

germanz:
Thie change handles HTTP error messages which in the case of the OMV client can be due to various reasons:

- **401** Invalid credentials
- **404** No data, when for example the supported ``maxDataLevel`` is exceeded

Error messages obtained from the server are logged for easy troubleshooting. Furthermore, in the case of HTTP client error codes, we do not attempt to fetch the data again.


AndriiHeonia (approved):

germanz:

PR #2124, opened 2021-02-25
MINOR: Fix error message on Datasource connection failure.

atomicsulfate:
Caught error might be just a string from a rejected promise.


@here/harp-vectortile-datasource/test/OmvDataSourceTest.ts
```diff
@@ -13,7 +13,10 @@ import { TileKey } from "@here/harp-geoutils";
 import { DataProvider } from "@here/harp-mapview-decoder";
 import { GeoJsonTiler } from "@here/harp-mapview-decoder/index-worker";
 import { silenceLoggingAroundFunction } from "@here/harp-test-utils";
-import { assert } from "chai";
+import * as chai from "chai";
```
FraukeF, line 16: this looks interessting :) .. can you add a comment why its needed

atomicsulfate: Comment added. chai-as-promised is a chai plugin to support assertions on promises, to install it you need to import chai, then the plugin, and finally do chai.use(chai_as_promised). We already use it in other test suites.

FraukeF (approved):

PR #2125, opened 2021-02-26
Enable shadows on WebTiles

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-examples/src/real-time-shadows.ts
+++ b/@here/harp-examples/src/real-time-shadows.ts
@@ -3,7 +3,6 @@
  * Licensed under Apache 2.0, see full license in LICENSE
  * SPDX-License-Identifier: Apache-2.0
  */
-
 import "three/examples/js/controls/TrackballControls";
 
 import { Theme } from "@here/harp-datasource-protocol";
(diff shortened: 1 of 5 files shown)


@here/harp-examples/src/real-time-shadows.ts
```diff
@@ -61,6 +62,12 @@ function swapCamera() {
     shadowCameraHelper.visible = !shadowCameraHelper.visible;
 }
 
+const hereWebTileDataSource = new HereWebTileDataSource({
```
nzjony, line 65: Thanks for expanding the example!

@here/harp-webtile-datasource/lib/WebTileLoader.ts
```diff
@@ -56,23 +55,18 @@ export class WebTileLoader extends BaseTileLoader {
...
-                    enableBlending(material);
-                }
-                addGroundPlane(
+                const planeMesh = addGroundPlane(
                     this.tile,
-                    this.dataSource.renderOrder, // Remove, as `renderOrder` will be deprecated.
-                    material,
-                    true,
-                    false
+                    this.dataSource.dataSourceOrder,
```
nzjony, line 60: Hmm, I'm not sure if this is quite correct, because the dataSourceOrder is not the renderOrder, and this value is used internally as the renderOrder.

I think it would make more sense to remove the `renderOrder` from this function and just set the order to some small value.

FraukeF: right, the datasourceOrder is evaluated somewhere else. I ll remove it

nzjony (approved):

PR #2130, opened 2021-03-02
Fix for use with Figma environment

mviranyi:
When running HARP within Chromium app of Figma (also called Figma Desktop App), functioning is hindered by function getUrlHostAndProtocol from @here/harp-utils/lib/UrlUtils.ts:

image

Reason being Figma setting the iframe's src attribute in UI (is a HTML view) to an base64 encoded string, thus internally when calling `location.href` there is no correct URI to be expected. For this the above stated function, correctly, does not give positive result.
With this fix, the plugin renders correctly within Figma and vanilla Chrome:

image

## Post Scriptum:

For now I've fixed the actual base function to give any caller the benefit of not throwing an exception. This also requires parent caller to handle error themselves. But there might be an actual bug along the way with `@here\harp-mapview\lib\ThemeLoader.ts` in line `132` - `139`:

```
            const themeUrl = uriResolver !== undefined ? uriResolver.resolveUri(theme) : theme;
            const response = await fetch(themeUrl, { signal: options.signal });
            if (!response.ok) {
                throw new Error(`ThemeLoader#load: cannot load theme: ${response.statusText}`);
            }
            theme = (await response.json()) as Theme;
            theme.url = resolveReferenceUri(getAppBaseUrl(), themeUrl); // <-------------- !!!
            theme = this.resolveUrls(theme, options);
```
To me this looks like it should read like `theme.url = themeUrl`, as the themeUrl successfully was fetched already (proven to be correct)?! Or am I to naiv in my assumption here?


@here/harp-utils/lib/UrlUtils.ts
```diff
@@ -130,7 +130,10 @@ export function getUrlHostAndProtocol(
 
     const match = url.match(urlOriginRe);
     if (!match) {
-        throw new Error(`getUrlHostAndProtocol: unable to parse URL '${url}'`);
+        return {
```
nzjony, line 133: Instead of returning empty fields, I think best to try `window.atob(url)` and then re-run the regex and return the protocol / host as on the lines 139

nzjony (commented):
Thanks for the PR and contribution @mviranyi !

Just one comment and it can be submitted.

@here/harp-utils/lib/UrlUtils.ts
```diff
@@ -130,7 +130,10 @@ export function getUrlHostAndProtocol(
...
+            protocol: "",
```
nzjony, line 134: something like

```
if(!match) {
try {
    const base64Url = window.atob(str);
    base64Url.match(...)
    return {
    protocol:...
    host:...
    }
} catch(e) {
    throw new Error(`getUrlHostAndProtocol: unable to parse URL '${url}'`);
}

@here/harp-utils/lib/UrlUtils.ts
nzjony, line 133: I don’t like silent failures, because it makes debugging much harder, because program execution continues in some invalid state and trying to find back to the point the program enters this state is hard.

I would prefer that `getUrlHostAndProtocol` stays the way it is and you just check the `window.location.href` (which anyway is the root cause of the issue) for an invalid value.

PR #2146, opened 2021-03-10
HARP-14495: Updates three to version 0.126.0

germanz:
* Updates three.js to version 0.126.0


@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -835,10 +835,21 @@ function getBaseColorPropName(technique: Technique): string | undefined {
...
+    | Int32Array
+    | Uint32Array
+    | Float32Array
+    | Float64Array;
+
 function getTextureBuffer(
     buffer: ArrayBuffer,
     textureDataType: THREE.TextureDataType | undefined
-): THREE.TypedArray {
+): TypedArray {
```
nzjony, line 852: I think this should be `BufferSource`, at least the `DataTexture` takes a BufferSource, see: https://github.com/three-types/three-ts-types/commit/4ef80d3e9e76a44271e21d8f536b784213375dbe

@here/harp-mapview/lib/TileObjectsRenderer.ts
```diff
@@ -157,7 +156,9 @@ export class TileObjectRenderer {
             return stableSort(a, b);
         };
 
-        this.m_renderer.setOpaqueSort(painterSortStable);
+        // Temporary workaround due to incorrect comparator type definition:
```
nzjony, line 159: Please create a bug on https://github.com/three-types/three-ts-types and mention it here.

@here/harp-datasource-protocol/lib/ThreeBufferUtils.ts
```diff
@@ -19,6 +18,17 @@ import {
     InterleavedBufferAttribute
 } from "./DecodedTile";
 
+type TypedArray =
```
nzjony, line 21: This is necessary, yes, but we could consider to deprecate this type and instead use ArrayLike<number>, because that is the type in the three-ts-types, see: https://github.com/three-types/three-ts-types/blob/master/types/three/src/core/BufferAttribute.d.ts

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -835,10 +835,21 @@ function getBaseColorPropName(technique: Technique): string | undefined {
     return getTechniqueDescriptor(technique)?.attrTransparencyColor;
 }
 
+type TypedArray =
```
nzjony, line 838: I don't think this is necessary, see below.

@here/harp-mapview/lib/TileObjectsRenderer.ts
germanz: Done.
https://github.com/three-types/three-ts-types/issues/41

@here/harp-datasource-protocol/lib/ThreeBufferUtils.ts
germanz: What about ``ArrayBufferView``?

nzjony: Internally it uses ArrayLike<number> in the three-ts-types, but I guess that works also.

nzjony (approved):

PR #2165, opened 2021-04-08
HARP-14417: fix picking of polygon outlines

oleksmarkh:
This PR fixes picking at polygonal outlines for the `fill` technique.

Before the fix:

picking example

* Expected result: when an outline is picked, all resulting objects (belonging to a given datasource) contain `.featureId` and `.userData` derived from original GeoJSON features that were used to render both `Mesh` and `LineSegments` objects.
* Actual result: while for `Mesh` objects results are correct (underlying polygons), for `LineSegments` it's always the first polygon of a given tile that's being returned instead. E.g. on the screenshot above, "Piemonte" is returned instead of "Abruzzo".

Reason:

faceIndex value

`PickHandler` was comparing `.faceIndex` with `undefined` when deciding whether to take `.faceIndex * 3` or `.index` in order to recognise the proper feature. In case of `LineSegments` it's `null`, as that's the default value coming from THREE's [`.intersectObject()`](https://threejs.org/docs/#api/en/core/Raycaster.intersectObject).

### DIFF
--- a/@here/harp-mapview/lib/PickHandler.ts
+++ b/@here/harp-mapview/lib/PickHandler.ts
@@ -356,7 +356,7 @@ export class PickHandler {
         if (
             featureData.starts === undefined ||
             featureData.starts.length === 0 ||
-            (intersect.faceIndex === undefined && intersect.index === undefined)
+            (typeof intersect.faceIndex !== "number" && intersect.index === undefined)
         ) {
             if (featureData.objInfos.length === 1) {
                 pickResult.userData = featureData.objInfos[0];
@@ -370,7 +370,7 @@ export class PickHandler {
         }
 
         const intersectIndex =
-            intersect.faceIndex !== undefined ? intersect.faceIndex * 3 : intersect.index!;
+            typeof intersect.faceIndex === "number" ? intersect.faceIndex * 3 : intersect.index!;
 
         // TODO: Implement binary search.
         let objInfosIndex = 0;
--- a/@here/harp-mapview/test/PickHandlerTest.ts
+++ b/@here/harp-mapview/test/PickHandlerTest.ts
@@ -1,0 +1,295 @@
+/*
+ * Copyright (C) 2021 HERE Europe B.V.
+ * Licensed under Apache 2.0, see full license in LICENSE
+ * SPDX-License-Identifier: Apache-2.0
+ */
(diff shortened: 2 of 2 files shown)


atomicsulfate (changes requested):
I know there's no PickHandler unit test, part is my fault since I changed this class a lot. So far we've been testing picking at a functional test "MapViewPickingTest.ts". But I think this class needs unit testing.
Could you please create one? So far it's enough to verify what you fixed, we'll add further coverage progressively.
Looks like the setup won't be straightforward. We can have a fake mapView with a single tile whose bbox must be intersected by the rayCaster's ray.
Then we could fake the rayCaster's (which is created in raycasterFromScreenPoint) intersectObjects() so that it returns the intersections that we want to test.

oleksmarkh:
@atomicsulfate, I've added two test cases ensuring the `.faceIndex` and `.index` lookups. The test setup is relying on minimal fake `MapView` substitute object (with a single fake rendered tile). I had to use the following combination of stubs in order to provide the test data:

```js
beforeEach(function () {
    ... // instantiating "PickHandler"
});

describe("#intersectMapObjects", function () {
    let raycasterFromScreenPointStub: sinon.SinonStub<[x: number, y: number], THREE.Raycaster>;

    beforeEach(function () {
        raycasterFromScreenPointStub = sinon.stub(pickHandler, "raycasterFromScreenPoint");
    });

    it("collects results for objects based on '.faceIndex' of intersection", function () {
        // these two stubs are defined inside test cases because
        // picking logic is tested against a combination of input coords and indexing data
        raycasterFromScreenPointStub.callsFake((x: number, y: number) => {
            // stubbing a real "Raycaster" instance
            const raycaster = raycasterFromScreenPointStub.wrappedMethod.call(pickHandler, x, y);
            sinon
                .stub(raycaster, "intersectObjects")
                .callsFake((objects, recursive, target: THREE.Intersection[] = []) => {
                    // test data, containing the index
                    target.push({...});
                });
            return raycaster;
        });

        const results = pickHandler.intersectMapObjects(467, 279);

        // checking for ".featureId" and ".userData"
        expect(...);
    });
});
```

oleksmarkh:
rebase please

atomicsulfate (approved):

PR #2173, opened 2021-04-19
HARP 14470

nzjony:
- nyc only supports code coverage in node, however we have a number of browser based tests, and it is the primary use case, so we have decided to do the code coverage in karma.
- the karma.conf.js previously used the pre-bundled test bundle, but this doesn't give us the code coverage per line number, I could have tried to find a way to remap the bundle to lines, but because the karma-typescript project is well maintained, I went this road. This however means there is a lot of configuration to get the tests running correctly, especially those that need to load resources.
- note also, there are quite a few tests that time out when run in the browser, hence the need to change the timeout in some cases
- note also, karma outputs the code coverage format using lcov, this is different to nyc and means that the code coverage drops a lot, this is because the lcov format can tell if a line is fully covered, i.e. for an if statement, it records if both true and false variants are executed. As such, lines without full branch coverage are treated as not covered, and this reduces the code coverage in the codecov.io report available via github. The local text report which is shown when running `yarn cov-test` doesn't have this behaviour, i.e lines that have just one branch executed are counted as part of the code coverage.
- moved most of the `karma.conf.js` to `karma.options.js`, so it can be shared and extended with the sdk and added the ability to prefix the files / exclude so that this repo works next to the sdk correctly.


nzjony:
Closing / reopening to retrigger sdk pipeline.

@here/harp-test-utils/lib/ProfileHelper.ts
```diff
@@ -11,14 +11,15 @@ import * as fs from "fs";
...
 // from `perf_hooks` in node env.
-//
-if (typeof window === "undefined") {
-    const perfHooks = require("perf_hooks");
-
-    (global as any).performance = perfHooks.performance;
-    (global as any).PerformanceObserver = perfHooks.PerformanceObserver;
-    (global as any).PerformanceEntry = perfHooks.PerformanceEntry;
-}
+// Note, the following had to be commented out, because "karma-typescript" tries to resolve the
```
dbacarel, line 14: can it be removed at once?

Also, why it doesn't apply to the same code in
`test/performance/LinesPerformanceTest.ts` and
`test/performance/OmvDecoderPerformanceTest.ts`?

dbacarel: discussed internally - it will be taken care of in a separate ticket.

dbacarel (approved):

PR #2177, opened 2021-04-22
Change URL caching inside ThemeLoader.

mviranyi:
This is another way of patching the issue we're facing with Figma application's plugin sandbox.

### DIFF
--- a/@here/harp-mapview/lib/ThemeLoader.ts
+++ b/@here/harp-mapview/lib/ThemeLoader.ts
@@ -24,7 +24,6 @@ import {
     IContextLogger,
     ISimpleChannel,
     RelativeUriResolver,
-    resolveReferenceUri,
     UriResolver
 } from "@here/harp-utils";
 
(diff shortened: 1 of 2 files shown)


mviranyi:
I definitely propose this as fix for the issue with Figma.

To me this is the wrong spot to evaluate anything URI related. Especially this being after the response got back and we just cache the URL.

Can we enlighten this discussion again please?

nzjony:
@mviranyi , agreed.

I see this line was written by @zbigg , do you remember why we have the call to `resolveReferenceUri`, is it some left over code from some refactoring? As mentioned by @mviranyi , we already do the `fetch`, so this shouldn't be needed?

zbigg:
@nzjony Unfortunately i don't remember all details about this [patch](9383825cada068b6639556674d58ced1af92c865). Brief look confirms that indeed, this is wrong place to "resolve" relative url to absolute.

I also got gut feeling that we needed to resolve `theme.url` to absolute but i don't remember if that's still the case.

PR #2182, opened 2021-04-26
MAPSJS-2783: Faster creation of keys for techniques with dynamic array buffer properties.

atomicsulfate:
Instead of stringifying the array buffers, an id is assigned to them
and used as part of the technique key.

### DIFF
--- a/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
+++ b/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
@@ -308,6 +308,29 @@ describe("StyleSetEvaluator", function () {
             assert.deepEqual(techniquesTileA[0], techniquesTileB[1]);
             assert.deepEqual(techniquesTileA[1], techniquesTileB[0]);
         });
+
+        it("returns same technique when same array buffer is used twice", function () {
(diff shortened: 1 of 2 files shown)


@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
```diff
@@ -288,6 +288,25 @@ describe("StyleSetEvaluator", function () {
             })();
 
             assert.equal(techniquesTileA.length, 2);
+
+            const techniquesTileB = (() => {
+                const ev = new StyleSetEvaluator({ styleSet: texturedStyle });
+                ev.getMatchingTechniques(new MapEnv({ kind: "park", map: { buffer: buffer2 } }));
```
nzjony, line 294: Please also add a test where the buffer: both point to `buffer1` and the length is 1

nzjony (commented):
Looks good, just please extend the unit test a bit.

@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
atomicsulfate: done

nzjony (approved):

PR #2186, opened 2021-04-30
HARP-14678 Fix tile dependencies not working

nzjony:
HARP-14678 Fix issue with morton codes being stored as 32 bit
    
The tile dependency example wasn't working with two separate data sources because the morton code was stored in a Float32Array, which doesn't have enough precision and caused issues. This change stores the row, column and level separately and then in the decoder, the mortonCode is computed and stored as a `number`, which has more accuracy that a 32 bit float.

### DIFF
--- a/@here/harp-mapview/lib/VisibleTileSet.ts
+++ b/@here/harp-mapview/lib/VisibleTileSet.ts
@@ -962,6 +962,11 @@ export class VisibleTileSet {
                         visibleTileKeys.find(
                             tileKeyEntry =>
                                 tileKeyEntry.tileKey.mortonCode() === tileKey.mortonCode()
+                        ) === undefined &&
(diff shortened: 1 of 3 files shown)


@here/harp-mapview/lib/VisibleTileSet.ts
```diff
@@ -962,6 +962,11 @@ export class VisibleTileSet {
                         visibleTileKeys.find(
                             tileKeyEntry =>
                                 tileKeyEntry.tileKey.mortonCode() === tileKey.mortonCode()
+                        ) === undefined &&
+                        // Check that we haven't already added this TileKey
+                        dependentTiles.find(
```
nzjony, line 967: Note, this isn't related to the bug, it is just a minor optimization.

FraukeF (approved):

PR #2202, opened 2021-05-20
HARP-15180 Upgrading yeoman because it has a transitive dependency (axios) which is out of date

nzjony:

# CI Check

Happy contributing!

### DIFF
--- a/@here/generator-harp.gl/package.json
+++ b/@here/generator-harp.gl/package.json
@@ -23,13 +23,14 @@
     ],
     "dependencies": {
         "mkdirp": "^1.0.4",
-        "yeoman-generator": "^4.10.1"
+        "yeoman-generator": "^5.2.0"
     },
     "devDependencies": {
         "chai": "^4.2.0",
         "fs.extra": "^1.3.2",
         "mocha": "^8.2.1",
-        "yeoman-test": "^2.6.0"
+        "yeoman-environment": "^3.4.1",
+        "yeoman-test": "^6.1.0"
     },
     "publishConfig": {
         "access": "public"
--- a/@here/generator-harp.gl/test/test.js
+++ b/@here/generator-harp.gl/test/test.js
@@ -48,7 +48,7 @@ describe("harp.gl:app", function () {
 
     after(function () {
         if (process.env.KEEP_TEMP_DIR) {
-            console.log("Keeping", runContext);
+            console.log("Keeping", runContext.cwd);
         } else {
             fsExtra.rmrfSync(runContext.cwd);
         }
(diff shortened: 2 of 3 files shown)


@here/generator-harp.gl/test/test.js
```diff
@@ -17,34 +13,35 @@ const {
...
+            .run(path.join(__dirname, "../generators/app"))
             .inTmpDir(function (dir) {
                 if (process.env.USE_NPMRC) {
                     const targetPath = path.join(dir, ".npmrc");
                     fs.copyFileSync(process.env.USE_NPMRC, targetPath);
                 }
             })
             .withPrompts({
-                'access_token': 'test_token'
+                language: "typescript",
```
nzjony, line 30: these match what is queried in the Generator, i.e. language & apikey. The `packagename` already has a default, so this isn't required.

@here/generator-harp.gl/test/test.js
```diff
@@ -17,34 +13,35 @@ const {
...
+                apikey: "test_token"
             });
 
-        assert.isString(helperDirectory);
+        assert.isString(helperDirectory.cwd);
```
dbacarel, line 34: `cwd` is not public according to https://yeoman.github.io/yeoman-test/YeomanTest.html, we might need to use a separate variable.
Can we assume that the current cwd is `path.join(__dirname, "../generators/app")` ?

dbacarel (changes requested):

@here/generator-harp.gl/package.json
```diff
@@ -23,13 +23,14 @@
...
         "mkdirp": "^1.0.4",
-        "yeoman-generator": "^4.10.1"
+        "yeoman-generator": "^5.2.0"
     },
     "devDependencies": {
         "chai": "^4.2.0",
         "fs.extra": "^1.3.2",
         "mocha": "^8.2.1",
-        "yeoman-test": "^2.6.0"
+        "yeoman-environment": "^3.4.0",
```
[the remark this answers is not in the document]
nzjony: Good spotting, thank you.

@here/generator-harp.gl/test/test.js
nzjony: Good point, yes I guess that makes sense.

Actually checking the `isString` is anyway some what ambiguous, I will check that the run result is not undefined.

nzjony: I see the code: https://yeoman.github.io/yeoman-test/run-result.js.html

And there is a `cwd`, so whilst not part of the documentation, it is part of the public API, at least according to the code.

We call `inTmpDir`, which gives us some random tmp directory, so it is different that the path.join you mentioned and not so easy to get the `cwd`, at least I don't see a method to return it.

dbacarel (approved):

PR #2203, opened 2021-05-20
MAPSJS-2953: Support custom DataAdapters for VectorTileDecoder.

atomicsulfate:

VectorTileDecoder has now a method getDataAdapter() whose default
implementation returns one of the 3 default adapters (omv, geojson,
or geojson-vt), the one that supports the provided data.

The method can be overriden by a subclass to provide a custom adapter:

```javascript
class CustomDataAdapter implements DataAdapter{
    // This method can be used to select an appropiate adapter for the given data when a decoder is meant to support,
    // multiple formats. It may just return true if there's only one supported format.
    // @override
    canProcess(data: ArrayBufferLike | {}): boolean { return true;} 

     // @override
    process(data: ArrayBufferLike | {},  decodeInfo: DecodeInfo,  geometryProcessor: IGeometryProcessor): void {
         // TODO
   }
}

class CustomVectorTileDecoder extends VectorTileDecoder {
    private readonly m_customDataAdapter = new CustomDataAdapter();
/* @override */
getDataAdapter(data: ArrayBufferLike | {}): DataAdapter | undefined {
        return this.m_customDataAdapter;
    }
}
````

### CODE REVIEW

File: @here/harp-vectortile-datasource/lib/VectorTileDecoder.ts [hunk 1]
```diff
@@ -317,11 +300,16 @@ export class VectorTileDecoder extends ThemedTileDecoder {
         projection: Projection
     ): Promise<DecodedTile> {
         const startTime = PerformanceTimer.now();
+        const dataAdapter = this.getDataAdapter(data);
```
nzjony, line 303: Does this need to be called for every tile key? Could it be lazy initialized? Perhaps not if the class has to support different formats at run time.

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
```diff
@@ -56,28 +56,24 @@ export class VectorTileDataProcessor implements IGeometryProcessor {
...
-
-        // Register the default adapters.
-
-        const omvDataAdapter = new OmvDataAdapter(this, dataPreFilter, logger);
-        omvDataAdapter.roundUpCoordinatesIfNeeded = m_roundUpCoordinatesIfNeeded;
-        this.m_dataAdapters.push(omvDataAdapter);
-
-        this.m_dataAdapters.push(new GeoJsonVtDataAdapter(this, dataPreFilter, logger));
-        this.m_dataAdapters.push(new GeoJsonDataAdapter(this, dataPreFilter, logger));
+        if (this.m_dataAdapter instanceof OmvDataAdapter) {
```
nzjony, line 69: Can this be done outside this class?

I find it a bit ugly that VectorTileDataProcessor has to know about the OmvDataAdapter internals.

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
```diff
@@ -125,7 +102,7 @@ export class VectorTileDataProcessor implements IGeometryProcessor {
             this.m_languages
         );
 
-        dataAdapter.process(data, decodeInfo);
+        this.m_dataAdapter.process(data, decodeInfo, this);
```
nzjony, line 105: Not related to your code, but this is a bit weird to have a `this` pointer.

It shows that the `VectorTileDataProcessor` does two things. It is a geometry processor and it also does the wireup between DataAdapter and the DecodedTileEmitter.

I think the IGeometryProcessor should be separated into a separate class from the VectorTileDataProcessor, but this can be done later.

@here/harp-vectortile-datasource/lib/DataAdapter.ts
```diff
@@ -5,22 +5,19 @@
...
-     * `DataAdapter`'s id.
-     */
-    id: string;
-
     /**
      * Checks if the given data can be processed by this `DataAdapter`.
      *
      * @param data - The raw data to adapt.
      */
+
```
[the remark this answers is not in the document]
atomicsulfate: done

@here/harp-vectortile-datasource/lib/DataAdapter.ts
```diff
@@ -29,5 +26,9 @@ export interface DataAdapter {
      * @param data - The raw data to process.
      * @param decodeInfo - The `DecodeInfo` of the tile to process.
      */
-    process(data: ArrayBufferLike | {}, decodeInfo: DecodeInfo): void;
+    process(
+        data: ArrayBufferLike | {},
+        decodeInfo: DecodeInfo,
+        geometryProcessor: IGeometryProcessor
```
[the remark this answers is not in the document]
atomicsulfate: done

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
atomicsulfate: Changed to do lazy initialization.

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
```diff
@@ -427,6 +413,25 @@ export class VectorTileDecoder extends ThemedTileDecoder {
         }
     }
 
+    /**
```
germanz, line 416: MInor spelling fixes.

```
    /**
     * Returns the appropriate data adapter to convert the given data into the format expected by
     * VectorTileDecoder.
     * @note Default adapters are available for GeoJson and OMV formats.
     * Child classes may override this function to support additional formats.
     *
     * @param data - The input data to be converted.
     * @returns The DataAdapter to convert the data, or undefined if there's no adapter for that
     * data format.
     */
```

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
```diff
@@ -111,149 +113,145 @@ export class VectorTileDataProcessor implements IGeometryProcessor {
...
-            }
-        }
-
-        const techniques = this.applyKindFilter(
-            this.m_styleSetEvaluator.getMatchingTechniques(env, layer, "line"),
-            GeometryKind.Line
+        assert(this.m_decodedTileEmitter !== undefined);
+        // Pass feature modifier method to processFeature if there's any modifier. Get it from any
+        // modifier, processFeature will later apply it to all using Function.apply().
+        const featureModifierFunc = this.m_featureModifiers?.[0].doProcessLineFeature;
```
nzjony, line 147: Why do you use just the first feature modifier function? It is an array so there could be multiple ones?

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
atomicsulfate: I've moved all OMV-specific decode options to OmvDataAdapter.

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
atomicsulfate: It explains in the comment above, to get a method you need to get it from any instance of the class, it doesn't matter which. Later I'm binding that method to different objects using Function.apply(). 
Anyway, I've refactored the code and now that's not there anymore. I've moved the code to OmvDataAdapter.

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
atomicsulfate: Yes, makes sense. But I'd rather leave it for a separate PR.

@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
```diff
@@ -35,7 +35,7 @@ export abstract class ThemedTileDecoder implements ITileDecoder {
     }
 
     decodeTile(
-        data: ArrayBufferLike,
+        data: ArrayBufferLike | {},
```
nzjony, line 38: I already have a PR for this: https://github.com/heremaps/harp.gl/pull/2198/files and I invited you to review, but I guess you missed it.

@here/harp-vectortile-datasource/lib/IGeometryProcessor.ts
```diff
@@ -56,11 +56,6 @@ export interface ILineGeometry {
  * line and polygon geometries, see [[VectorTileDecoder]].
  */
 export interface IGeometryProcessor {
-    /**
-     * The offset to apply the storage level of a tile to compute the zoom level.
-     */
-    storageLevelOffset?: number;
```
nzjony, line 63: should this be first deprecated?

@here/harp-vectortile-datasource/lib/OmvDataFilter.ts
```diff
@@ -713,7 +713,7 @@ export class ComposedDataFilter implements OmvFeatureFilter {
      */
     get hasKindFilter() {
         return this.filters.reduce<boolean>(
-            (result, filter) => result && filter.hasKindFilter,
```
nzjony, line 716: should be a separate commit, it seems unrelated to the rest of the PR

@here/harp-vectortile-datasource/lib/VectorTileDataEmitter.ts
```diff
@@ -226,6 +226,17 @@ enum LineType {
 type TexCoordsFunction = (tilePos: THREE.Vector2, tileExtents: number) => THREE.Vector2;
 const tmpColor = new THREE.Color();
 
+/**
+ * Options for VectorTileDataEmitter, @see {@link DecoderOptions} and {@link OmvDecoderOptions}.
```
nzjony, line 230: This doesn't seem to be rendered correctly, please try this locally and fix.

@here/harp-vectortile-datasource/lib/VectorTileDataEmitter.ts
```diff
@@ -226,6 +226,17 @@ enum LineType {
...
+ * @internal
+ */
+export interface VectorTileDataEmitterOptions {
```
nzjony, line 233: Ideally this should be in a separate commit / PR, because it seems just a refactoring and not related to the actual ticket?

@here/harp-vectortile-datasource/lib/VectorTileDataEmitter.ts
```diff
@@ -242,14 +253,20 @@ export class VectorTileDataEmitter {
     private m_maxGeometryHeight: number = 0;
     private m_minGeometryHeight: number = 0;
 
+    private readonly m_gatherFeatureAttributes: boolean;
```
nzjony, line 256: Why don't you just store the options object here instead of repeating the instance methods? At least that is something that we seem to do regularly in other classes.

@here/harp-mapview-decoder/lib/ThemedTileDecoder.ts
atomicsulfate: Oh, sorry, I didn't see that. I just merged it.

nzjony: No worries.

@here/harp-vectortile-datasource/lib/IGeometryProcessor.ts
atomicsulfate: This wasn't part of the API, it's not exported in the package index.ts. Now it's only exposed indirectly because it's a parameter type of DataAdapter interface.

@here/harp-vectortile-datasource/lib/OmvDataFilter.ts
atomicsulfate: I had to add a lot of missing unit tests for VectorTileDecoder, VectorTileDataProcessor, OmvDataAdapter... I found this bug while unit testing VectorTileDataProcessor, that's why I fixed it here.
It could be moved to a separate PR, I'd have to move the failing test as well. But since it's a one liner, is it ok if I leave it here?

@here/harp-vectortile-datasource/lib/VectorTileDataEmitter.ts
atomicsulfate: DecoderOptions works. OmvDecoderOptions doesn't because it's marked as "@hidden". I've changed it to "@internal" instead, so that we can see it in our documentation but still can omit it from public documentation by not exporting internals.

@here/harp-vectortile-datasource/lib/VectorTileDataEmitter.ts
atomicsulfate: True. But since it hardly adds any complexity to the change, may we keep it here? Otherwise I have to rewrite all the usages back to the original signature and then introduce the interface again in a separate PR.

I see you try to reduce the size of this PR, but most of it comes from the unit tests (+600 lines) and then moving OMV specific options out of VectorTileDecoder and VectorTileDataProcessor.
The changes related to custom adapters are in the first two commits. The rest of the commits are unit tests and refactoring.

@here/harp-vectortile-datasource/lib/VectorTileDataEmitter.ts
atomicsulfate: makes sense. Wouldn't you rather do this in a separate PR?

@here/harp-vectortile-datasource/lib/IGeometryProcessor.ts
nzjony: Good point, I should also check the index.ts...

@here/harp-vectortile-datasource/lib/OmvDataFilter.ts
nzjony: All good, ok, sounds fine.

@here/harp-vectortile-datasource/lib/VectorTileDataEmitter.ts
nzjony: Ok, fair enough.

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
```diff
@@ -311,21 +264,22 @@ export class VectorTileDecoder extends ThemedTileDecoder {
         }
         const dataAdapter = this.m_dataAdapter!;
         assert(dataAdapter.canProcess(data));
+        if (this.m_optionsChanged) {
+            if (dataAdapter instanceof OmvDataAdapter) {
+                const omvOptions = this.m_options as OmvDecoderOptions;
+                dataAdapter.configure(omvOptions, styleSetEvaluator);
```
nzjony, line 270: why not put this method into the DataAdapter interface (Open-closed principle if I'm not mistaken)...?

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
```diff
@@ -337,122 +291,32 @@ export class VectorTileDecoder extends ThemedTileDecoder {
...
-                this.m_enableElevationOverlay = omvOptions.enableElevationOverlay;
-            }
-            if (omvOptions.roundUpCoordinatesIfNeeded !== undefined) {
-                this.m_roundUpCoordinatesIfNeeded = omvOptions.roundUpCoordinatesIfNeeded;
-            }
-        }
-        if (options?.languages !== undefined) {
-            this.languages = options.languages;
-        }
+        this.m_options = {
```
nzjony, line 294: Minor, you could merge the `m_options` and `m_optionsChanged` into a class and then add a method update which would encapsulate the `m_optionsChanged = true` change.

@here/harp-vectortile-datasource/lib/adapters/omv/OmvDataAdapter.ts
```diff
@@ -248,20 +345,25 @@ export class OmvDataAdapter implements DataAdapter, OmvVisitor {
      */
     visitPointFeature(feature: com.mapbox.pb.Tile.IFeature): void {
```
nzjony, line 346: this should be void | undefined

@here/harp-vectortile-datasource/lib/adapters/omv/OmvDataAdapter.ts
```diff
@@ -460,27 +512,58 @@ export class OmvDataAdapter implements DataAdapter, OmvVisitor {
...
+        }
+
+        const properties = readAttributes(this.m_layer, feature);
+        const env = new MapEnv(properties);
+        if (
+            this.m_featureModifiers?.find(fm => {
+                // TODO: The logic of feature ignore should be actually in the feature filtering
+                // mechanism - see OmvFeatureFilter.
+                assert(modifierFunc !== undefined);
+                return !modifierFunc!.apply(fm, [layerName, env, this.m_tileKey.level]);
```
nzjony, line 555: this is some genius level typescript, I tried for an hour to figure it out and I just understand...

nzjony: Well, I found it is possible to do it by setting:

modifierName: "doProcessPointFeature" | "doProcessLineFeature" | "doProcessPolygonFeature",

And then go `fm[modifierName]`

nzjony: However.... that isn't so nice to have to hard code the methods there. I tried to find a way to extract it from the OmvFeatureModifier but to no avail

nzjony: Nevermind,

atomicsulfate: ok

@here/harp-vectortile-datasource/lib/adapters/omv/OmvDataAdapter.ts
atomicsulfate: I removed the "return undefined".

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
atomicsulfate: Yes, it's kind of hacky. If we only had a configure method getting an option map, I'd move it to DataAdapter interface. But there's two more things specific of OMV:
1. You have to pass the StyleSetEvaluator to create a StyleSetDataFilter.
2. You have to retrieve back a OmvFeatureFilter created by OmvDataAdapter to use it to filter kinds in VectorTileDataProcessor (see a couple of lines below the call to configure).

These details are so far specific of OMV. We could generalize them for all formats, but the question is if we really want to do it:
- Kinds filtering is deprecated (see BaseTechniqueParams.kind), eventually we'll remove it.
- All kinds of filtering could be implemented directly in the style, using dynamic properties if necessary.
- FeatureModifiers where introduced for political views, but I think it should be doable via styling  as well.

I think this needs a revisit for further refactoring, maybe when we add OCM.

I can put a comment and create a ticket for this.

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
atomicsulfate: done.

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
nzjony: @atomicsulfate , you could theoretically pass the `This` pointer, so the data adapter could extract the options / stylesetevaluator internally and not here.

However I get your point, if this is mostly things that can be deprecated, then a comment + ticket should be enough to let future developers no that this shouldn't be permanent.

nzjony (approved):

@here/harp-vectortile-datasource/lib/VectorTileDecoder.ts
atomicsulfate: Ticket created

PR #2225, opened 2021-06-22
HARP-15210: Wait for texture GPU upload on TileGeometryCreator.

atomicsulfate:

TileGeometryCreator.createAllGeometries() returns now a promise that
is resolved when all textures in the tile are uploaded in GPU. This ensures that all textures are rendered on frame complete event.

### DIFF
--- a/@here/harp-mapview/lib/DecodedTileHelpers.ts
+++ b/@here/harp-mapview/lib/DecodedTileHelpers.ts
@@ -112,7 +112,7 @@ function createTextureFromRawImage(
     onError: (error: ErrorEvent | string) => void
 ) {
     const properties = textureBuffer.dataTextureProperties;
-    if (properties !== undefined) {
+    if (properties) {
         const textureDataType: THREE.TextureDataType | undefined = properties.type
             ? toTextureDataType(properties.type)
             : undefined;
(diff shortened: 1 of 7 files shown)


@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
```diff
@@ -347,13 +347,13 @@ export class TileGeometryLoader {
...
         // Set up techniques which should be processed.
         geometryCreator.initDecodedTile(decodedTile, enabledKinds, disabledKinds);
-        geometryCreator.createAllGeometries(tile, decodedTile);
-
-        if (stats.enabled) {
-            this.addStats(stats, now);
-        }
-        this.finish();
-        tile.dataSource.requestUpdate();
+        geometryCreator.createAllGeometries(tile, decodedTile).then(() => {
```
nzjony, line 350: I think it is nicer to mark the function `createGeometry` as async (from a documentation point of view). The added benefit is that you can await here.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
```diff
@@ -275,13 +276,35 @@ export class TileGeometryCreator {
      *
      * @param tile - The {@link Tile} to process.
      * @param decodedTile - The decodedTile containing the actual tile map data.
+     * @returns Promise resolved when all textures are ready to render.
      */
-    createAllGeometries(tile: Tile, decodedTile: DecodedTile) {
+    createAllGeometries(tile: Tile, decodedTile: DecodedTile): Promise<void> {
```
nzjony, line 281: You can (but don't strictly have to), mark this async, I think from a documentation POV, it is nice.

@here/harp-mapview/lib/DecodedTileHelpers.ts
```diff
@@ -131,102 +131,106 @@ function createTextureFromRawImage(
...
+    }
+    if (properties.repeatU !== undefined) {
+        texture.repeat.x = properties.repeatU;
+    }
+    if (properties.repeatV !== undefined) {
+        texture.repeat.y = properties.repeatV;
+    }
+}
+
 function createTexture(
```
nzjony, line 161: consider to mark as async

@here/harp-mapview/lib/geometry/TileGeometryLoader.ts
atomicsulfate: done

@here/harp-mapview/test/DecodedTileHelpersTest.ts
```diff
@@ -187,29 +194,43 @@ describe("DecodedTileHelpers", function () {
...
+                },
+                callbackSpy
+            ) as MapMeshStandardMaterial;
+            assert.isDefined(onImageLoad);
+            assert.isTrue(callbackSpy.called);
+
+            // Check that texture promised is not yet resolved.
+            let textureLoaded = false;
+            const texturePromise = callbackSpy.firstCall.args[0];
+            texturePromise.then(() => {
```
nzjony, line 220: I guess this works because of the `await wait` below, but to be safe, you should `await` here too...? Or put the rest of the test in the `.then()` ?

nzjony: Wait... not here, that would break the test, but before `await wait` ?

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
atomicsulfate: I'm not sure what you mean.
a) If you mean I should just add "async" in front of the method without any further changes, that would actually be misleading from the documentation POV. The method is fully synchronous, ESLint would complain about the missing "async" otherwise.
- If you mean rewrite it to make it asynchronous, we could do it by changing at the very end ```return texturesReady;``` to  ```await texturesReady;```. But I don't see any benefit in that. Since this is a public method, better leave the decision to the calling code. Async callers may await on the promise, sync callers can work with the promise.

Awaiting on every single texture promise is not an option either, cause the promises are generated sequentially, that would make the texture loading sequential.

@here/harp-mapview/lib/DecodedTileHelpers.ts
atomicsulfate: It's actually a synchronous function, marking it async is actually misleading.

As for making it asynchronous, we don't want to await on the texture here. We need to trigger the loading of all textures in the tile and later wait for all of them to be resolved concurrently.

@here/harp-mapview/test/DecodedTileHelpersTest.ts
atomicsulfate: done, fixed similar cases in other tests as well.

@here/harp-mapview/lib/geometry/TileGeometryCreator.ts
nzjony: @atomicsulfate , I mean option 1.

Technically the function is synchronous, yes, but it returns a Promise, so the result isn't fully known. You can argue that the function returns without any need to wait, but users of this are going to likely anyway `await/.then()` on the result of this function, so they won't know if it was synchronous or not, they will just use the result.

The [docs](https://stackoverflow.com/questions/45594596/async-function-without-await-in-javascript#comment87236078_45594870) on Mozilla.org also mention:

> Async functions can contain zero or more await expressions.

Another benefit to using async is that any error thrown will cause the promise to reject, so users of the API can use `.catch` and react accordingly. Otherwise they need to wrap this function in a try/catch block, see: https://stackoverflow.com/questions/45594596/async-function-without-await-in-javascript#comment87236078_45594870

nzjony (approved):

PR #2230, opened 2021-06-30
MINOR: Fix theme extends to respect multiple image textures

germanz:
The specified allows theme extends to work properly when multiple images (textures) are specified.

### DIFF
--- a/@here/harp-mapview/test/ThemeLoaderTest.ts
+++ b/@here/harp-mapview/test/ThemeLoaderTest.ts
@@ -306,6 +306,71 @@ describe("ThemeLoader", function () {
             assert.deepEqual(result.definitions!.roadColor, { type: "color", value: "#fff" });
             assert.deepEqual(result.definitions!.waterColor, { type: "color", value: "#0f0" });
         });
+        it("supports multiple inheritance with textures", async function () {
+            const inheritedTheme: Theme = {
+                extends: [
+                    {
+                        images: {
+                            foo: {
+                                url: "icons://maki_icons.png",
+                                preload: true
+                            },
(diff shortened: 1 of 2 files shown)


@here/harp-mapview/test/ThemeLoaderTest.ts
```diff
@@ -306,6 +306,48 @@ describe("ThemeLoader", function () {
...
+                            }
+                        ]
+                    }
+                ]
+            };
+
+            const result = await ThemeLoader.load(inheritedTheme);
+            assert.exists(result.images?.foo);
+            assert.exists(result.images?.bar);
+            assert.deepEqual(result.imageTextures?.map(imageTexture => imageTexture.name).sort(), [
```
nzjony, line 346: Please change one of the imageTextures to have a collision and check that the second one overrides the first one.

germanz: Good point. Done.

nzjony: Sorry, didn't mean to resolve.

It is better, but if I understand correctly, the test still doesn't test that the collision is resolved correctly because the "baz" points to the same image as the base theme. So you can't say for certain that you have the ImageTexture from the theme.

If you change the `image:"baz"` to something different (but keep the "name" the same), then you can be sure that the collision is correctly resolved.

nzjony: Nevermind, I read the code wrong. Resolving and will approve.

nzjony (approved):

PR #2234, opened 2021-07-06
HARP-14531: sort picking results by 'dataSourceOrder'

oleksmarkh:
Signed-off-by: Oleksii Markhovskyi <oleksii.markhovskyi@here.com>

This PR introduces a higher tie-breaker (comparator criterion) for sorting picking results - by `.dataSourceOrder`. It defaults those values to zeros to ensure consistent sorting.

It also orders zero-distance results (screen-space objects, e.g. labels) higher, because those are currently rendered on top.

Sorting by `.distance` and `.renderOrder` remains in place for results with same `.dataSourceOrder`.

### DIFF
--- a/@here/harp-mapview/lib/PickHandler.ts
+++ b/@here/harp-mapview/lib/PickHandler.ts
@@ -83,6 +83,14 @@ export interface PickResult {
     dataSourceName: string | undefined;
 
     /**
+     * Data source order, useful for sorting a collection of picking results.
+     * A number for objects/features coming from tiles (as those have data sources attached),
+     * an undefined when objects are added via "mapView.mapAnchors.add(object)" - those are treated as
+     * base layer objects during picking (same as "dataSourceOrder: 0").
+     */
+    dataSourceOrder: number | undefined;
+
+    /**
      * Render order of the intersected object.
      */
     renderOrder?: number;
@@ -90,7 +98,7 @@ export interface PickResult {
     /**
      * An optional feature ID of the picked object.
      * @remarks The ID may be assigned by the object's {@link DataSource}, for example in case of
-     * Optimized Map Vector (OMV) and GeoJSON datata sources.
+     * Optimized Map Vector (OMV) and GeoJSON data sources.
      */
     featureId?: number | string;
 
@@ -259,6 +267,7 @@ export class PickHandler {
             point: intersection.point,
             distance: intersection.distance,
             dataSourceName: intersection.object.userData?.dataSource,
+            dataSourceOrder: tile?.dataSource?.dataSourceOrder,
             intersection,
             tileKey: tile?.tileKey
         };
--- a/@here/harp-mapview/lib/PickListener.ts
+++ b/@here/harp-mapview/lib/PickListener.ts
@@ -9,8 +9,28 @@ import { assert } from "@here/harp-utils";
 import { IntersectParams } from "./IntersectParams";
 import { PickResult } from "./PickHandler";
 
-// Default sorting by distance first and then by reversed render order.
+// Default sorting by:
+// 1. reversed data source order (higher first)
+// 2. distance to the camera (closer first)
+// 3. reversed render order (higher first)
+//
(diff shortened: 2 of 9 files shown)


@here/harp-mapview/lib/PickListener.ts
```diff
@@ -133,9 +148,22 @@ export class PickListener {
...
-            this.m_results.sort(defaultSort);
-            this.m_sorted = true;
+        if (this.m_sorted) {
+            return;
         }
+
+        // HARP-14531: group and sort zero-distance results separately and merge them back
+        // to ensure that screen-space objects (e.g. labels) are returned first,
+        // as they are currently rendered on top.
+        const zeroDistanceGroup: PickResult[] = [];
```
nzjony, line 158: Why do you need to split & merge?

Why not just change the `defaultSort` function to sort those with zero distance to be higher?

@here/harp-mapview/lib/PickHandler.ts
```diff
@@ -82,6 +82,11 @@ export interface PickResult {
      */
     dataSourceName: string | undefined;
 
+    /**
+     * Data source order, useful for sorting a collection of picking results.
```
nzjony, line 86: Please add a comment that objects added via the `mapAnchor` API on `MapView` will have an undefined value here, because there is no such order.

@here/harp-mapview/lib/PickListener.ts
oleksmarkh: I wanted the `defaultSort()` to remain a relatively small comparator function, meaning it applies criteria one by one and compares same properties within a given criterion.

Basically, I didn't want to add `if (lhs.distance === 0 && rhs.distance !== 0)` kind of conditions to this function, thinking that splitting->sorting->merging is expressing the end result better, like "first there are zero-distant items sorted according to 3-step criteria, then the remaining ones, using the same criteria".

@here/harp-mapview/lib/PickHandler.ts
oleksmarkh: But is it expected to have such objects treated as belonging to the base layer during picking? Or they should be returned first instead, like text labels?

@here/harp-mapview/lib/PickListener.ts
nzjony: Ok, I agree, it makes things more complicated.

@here/harp-mapview/lib/PickListener.ts
```diff
@@ -133,9 +148,22 @@ export class PickListener {
...
+        const nonZeroDistanceGroup: PickResult[] = [];
+
+        this.m_results.forEach(result =>
+            (result.distance === 0 ? zeroDistanceGroup : nonZeroDistanceGroup).push(result)
+        );
+        this.m_results = zeroDistanceGroup
```
nzjony, line 164: Why not sort `this.m_results` first, then you don't need to sort the `zeroDistanceGroup` and the `nonZeroDistanceGroup`, because they are already sorted?

@here/harp-mapview/lib/PickHandler.ts
nzjony: I'm not so sure. 

You make the assumption anyway now that they belong to a datasource with order 0, so it doesn't hurt to add it, but yes, we can think about this.

oleksmarkh: I agree, added a comment. Will update it if we decide to sort map anchor objects first.

@here/harp-mapview/lib/PickListener.ts
oleksmarkh: That's true! Updated, thanks!

nzjony (dismissed):

nzjony (approved):

nzjony (dismissed):

nzjony:
rebase please

nzjony (approved):

PR #2243, opened 2021-07-22
HARP-16283 build-* now builds with the production flag set

nzjony:


nzjony:
No longer needed. The production builds, at least in GitHub are handled within the `ci.yaml` file.

PR #2250, opened 2021-07-27
MAPSJS-2983: Missing texture refs with Harp

dbacarel:
Change-Id: I6f343979e9e8fbf0d66de7f0e8e1ca0266d9c35c
Signed-off-by: Daniele Bacarella <daniele.bacarella@here.com>

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/BoundsGenerator.ts
+++ b/@here/harp-mapview/lib/BoundsGenerator.ts
@@ -366,7 +366,7 @@ export class BoundsGenerator {
     }
 
     private generateOnPlane(): GeoPolygon | undefined {
-        //!!!!!!!ALTITUDE IS NOT TAKEN INTO ACCOUNT!!!!!!!!!
+        // ALTITUDE IS NOT TAKEN INTO ACCOUNT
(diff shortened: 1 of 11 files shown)


nzjony:
I looked at this a bit and thanks @dbacarel for the draft PR.

I think another approach could be instead to isolate the issue to the PoiRenderer, to avoid having to make big changes elsewhere. Because it should be possible for the code to still work even if the POI still doesn't have an image.

My suggestion would be to make the MapViewImageCache trigger an event when the cache is updated and the PoiRenderer can listen for this event.
The findImageItem can return a promise which is then resolved when the cache finally contains the image that is requested, and then it can be rendered.

nzjony:
@dbacarel , I look into this further and with help from @atomicsulfate , we agreed to have a simple approach, i.e. we remove the line to invalidate the `PoiInfo` and just try again, until we reach some limit and then print an error / give up.

I found some other ways to tackle the issue, but this seems the simplest.

Thanks anyway for starting to look into this, it was a very tricky problem and I needed a long time to fully understand what was going on.

PR #2257, opened 2021-08-03
HARP-16327, Switch to the flat list index page for the examples

AndriiHeonia:
Adjust examples to avoid writing into document.body.innerHTML. This fix prevents destruction of event listeners attached by the main (template) page.

### DIFF
--- a/@here/harp-examples/src/datasource_features_lines-and-points.ts
+++ b/@here/harp-examples/src/datasource_features_lines-and-points.ts
@@ -205,7 +205,11 @@ export namespace LinesPointsFeaturesExample {
     }
 
     function createBaseMap(theme: Theme): MapView {
-        document.body.innerHTML += getExampleHTML();
+        const template = document.createElement("template");
+        template.innerHTML = getExampleHTML();
+        template.content.childNodes.forEach(node => {
+            document.body.appendChild(node);
+        });
 
         const canvas = document.getElementById("mapCanvas") as HTMLCanvasElement;
         const mapView = new MapView({
@@ -236,8 +240,7 @@ export namespace LinesPointsFeaturesExample {
 
     function getExampleHTML() {
         return (
-            `
-            <style>
+            `<style>
                 #mapCanvas {
                     top: 0;
                 }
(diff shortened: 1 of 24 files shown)


@here/harp-examples/src/datasource_features_lines-and-points.ts
```diff
@@ -247,7 +250,7 @@ export namespace LinesPointsFeaturesExample {
                     text-align: center;
                     font-family: monospace;
                     left: 50%;
```
germanz, line 252: Alternatively, we can use the following CSS, as the current approach with absolute positioning and then adjustment with percentage is IMHO a bit strange:

```
  margin: 0 auto;
  font-size: 15px;
  padding-top: 10px;
```

@here/harp-examples/src/getting-started_orbiting-view.ts
```diff
@@ -68,7 +68,29 @@ export namespace CameraOrbitExample {
...
-        document.body.innerHTML += getExampleHTML();
+        const template = document.createElement("template");
+        template.innerHTML = `<style>
+            #mapCanvas {
+                top:0
+            }
+            #info {
+                color: #fff;
+                width: 80%;
+                left: 50%;
```
germanz, line 79: Same as above.

@here/harp-examples/src/markers_dynamic.ts
```diff
@@ -189,8 +189,10 @@ export namespace DynamicMarkersExample {
     function addInstructions() {
         const message = document.createElement("div");
         message.innerHTML = `Tap map to add a marker, long press on a marker to remove it`;
-        message.style.position = "relative";
+        message.style.position = "absolute";
```
germanz, line 192: For the sake of consistency, I recommend using the same styling for the **info** element as in ``datasource_features_lines-and-points`` example.

@here/harp-examples/src/threejs_raycast.ts
```diff
@@ -112,18 +112,18 @@ export namespace ThreejsRaycast {
...
+    const template = document.createElement("template");
+    template.innerHTML =
         `<style>
-            #mapCanvas{
-                top:0;
-            }
             #info{
-                color: #000;
+                position: absolute;
+                left: 0;
```
germanz, line 120: Why not stretch it to 100% like in other examples instead of specifying left/right properties and margin?

PR #2259, opened 2021-08-03
Release 0.26.0 back to master.

nzjony:
Merge release branch back to master.


.github/workflows/deploy_to_npm.yaml
```diff
@@ -35,7 +35,6 @@ jobs:
                   yarn test-cov
                   yarn karma-headless
                   yarn karma-headless-firefox
-                  ./scripts/test-npm-packages.sh
```
atomicsulfate, line 38: Now we'll never test the npm packages right? Shouldn't we instead skip this step for release branch but keep doing it on master?

atomicsulfate, line 38: Now we'll never test the npm packages right? Shouldn't we instead skip this step for release branch but keep doing it on master?

nzjony: If you search for `test-npm-packages.sh` it is on line 103 in `ci.yaml`, this just removes it from the publishing step (see line 1).

nzjony: Ahhh ic, this also is to be interpreted as running `test-npm-packages.sh` on post-submit.

I would argue that this is not necessary, because it is anyway tested in pre-submit, but because we can rebase, there is the slight chance we could break something.

Is it ok to remove from the post-submit to master, or should I filter out for release?

atomicsulfate (approved):

gairik (approved):

PR #2274, opened 2021-08-11
MAPSJS-2983 Don't create POI if image is missing in cache

nzjony:
The user can modify the cache so we need to check if the image exists before preparing the POI

Change-Id: I9a68e523e9e3aa8a815a53037063a9d83cd6a436

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-mapview/lib/poi/PoiRenderer.ts
+++ b/@here/harp-mapview/lib/poi/PoiRenderer.ts
@@ -470,8 +470,9 @@ export class PoiRenderer {
      * @param env - TODO! The current zoomLevel level of {@link MapView}
      *
      * @returns `True` if the space is not already allocated by another object (text label or POI)
+     * or a Promise if the image has to be loaded first.
      */
-    prepareRender(pointLabel: TextElement, env: Env): boolean {
+    prepareRender(pointLabel: TextElement, env: Env): boolean | Promise<boolean> {
         const poiInfo = pointLabel.poiInfo;
         if (poiInfo === undefined) {
             return false;
@@ -490,9 +491,9 @@ export class PoiRenderer {
             return false;
         }
         if (poiInfo.buffer === undefined) {
-            this.preparePoi(pointLabel, env, imageItem, imageTexture);
+            return this.preparePoi(pointLabel, env, imageItem, imageTexture);
         }
-        return poiInfo.buffer !== undefined;
+        return false;
     }
 
     /**
@@ -650,56 +651,65 @@ export class PoiRenderer {
      * loading of the actual image.
      *
      * Assumes that the image requested exists already in the cache.
+     *
+     * If the ImageItem hasn't had a request to load, this is done here as well, this is returned so
+     * that the user can await on this to know when to use the image.
+     *
+     * @returns if the POI was setup, or not (as a Promise because of the reason above, or
+     * directly).
      */
     private preparePoi(
         pointLabel: TextElement,
         env: Env,
         imageItem: ImageItem,
         imageTexture?: ImageTexture
-    ): void {
+    ): boolean | Promise<boolean> {
         const poiInfo = pointLabel.poiInfo;
         if (!poiInfo || !pointLabel.visible) {
-            return;
+            return false;
         }
 
         if (poiInfo.buffer !== undefined || poiInfo.isValid === false) {
             // Already set up, nothing to be done here.
-            return;
+            return false;
         }
 
         if (poiInfo.poiTableName !== undefined) {
             if (this.m_poiManager.updatePoiFromPoiTable(pointLabel)) {
                 if (!pointLabel.visible) {
                     // PoiTable set this POI to not visible.
-                    return;
+                    return false;
                 }
             } else {
                 // PoiTable has not been loaded, but is required to determine visibility.
-                return;
+                return false;
             }
         }
 
         if (imageItem.loaded) {
             this.setupPoiInfo(poiInfo, imageItem, env, imageTexture);
-            return;
+            return poiInfo.isValid === true;
         }
 
         if (imageItem.loading) {
             // already being loaded, will be rendered once available
-            return;
+            return false;
         }
 
-        imageItem
+        return imageItem
             .loadImage()
             .then(loadedImageItem => {
                 // Skip setup if image was not loaded (cancelled).
                 if (loadedImageItem?.image) {
                     this.setupPoiInfo(poiInfo, loadedImageItem, env, imageTexture);
+                    return poiInfo.isValid === true;
                 }
+                return false;
             })
             .catch(error => {
                 logger.error(`preparePoi: Failed to load imageItem: '${imageItem.url}`, error);
                 poiInfo.isValid = false;
+                return false;
             });
     }
 
--- a/@here/harp-mapview/lib/poi/PoiRenderer.ts
+++ b/@here/harp-mapview/lib/poi/PoiRenderer.ts
@@ -494,7 +494,7 @@ export class PoiRenderer {
         if (poiInfo.buffer === undefined) {
             return this.preparePoi(pointLabel, env, imageItem, imageTexture);
         }
-        return false;
+        return poiInfo.buffer !== undefined;
     }
 
     /**
--- a/@here/harp-mapview/lib/poi/PoiRenderer.ts
+++ b/@here/harp-mapview/lib/poi/PoiRenderer.ts
@@ -469,32 +469,19 @@ export class PoiRenderer {
      * @param pointLabel - TextElement with PoiInfo for rendering the POI icon.
      * @param env - TODO! The current zoomLevel level of {@link MapView}
      *
-     * @returns `True` if the space is not already allocated by another object (text label or POI)
-     * or a Promise if the image has to be loaded first. This happens when the user explicitly
-     * doesn't load the image when it is put into the cache.
+     * @returns `True` if the label can be rendered or a Promise if the image has to be loaded
+     * first. This happens when the user explicitly doesn't load the image when it is put into the
+     * cache.
      */
     prepareRender(pointLabel: TextElement, env: Env): boolean | Promise<boolean> {
         const poiInfo = pointLabel.poiInfo;
         if (poiInfo === undefined) {
             return false;
         }
-        const { imageItem, imageTexture } = this.imageCached(poiInfo.imageTextureName);
-        if (!imageItem) {
-            // Set the POI to be invalid, so it will not be rendered, nor will it be re-checked in a
-            // future request. The item must exist (but not necessarily loaded in the cache before
-            // the POI can be initialized). We don't log an error, because the API of the
-            // MapViewImageCache allows the user to delete an image from the cache. The geometry
-            // creation is handled using some task queue which means that in some cases, a POI entry
-            // may be removed before the geometry is created. See queueGeometryCreation, which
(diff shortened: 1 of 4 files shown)


@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -492,8 +485,12 @@ export class PoiRenderer {
         if (poiInfo === undefined) {
             return false;
         }
+        const { imageItem, imageTexture } = this.imageCached(poiInfo);
```
nzjony, line 488: Here we prevent creation of the image isn't in the cache.

@here/harp-mapview/test/PoiRendererTest.ts
```diff
@@ -97,4 +100,93 @@ describe("PoiRenderer", function () {
...
+            const testCache = new MapViewImageCache();
+            // This is the difference to the test above, we don't start loading, we expect the
+            // prepareRender function to do this.
+            addRandomImageToCache(testCache, testImageName, false);
+            const caches = [testCache];
+            const poiRenderer = new PoiRenderer(webGLRenderer, poiManager, caches);
+            const pointLabel = createPointLabel(testImageName);
+            const waitingForLoad = poiRenderer.prepareRender(pointLabel, mapEnv);
+            const result = await Promise.resolve(waitingForLoad);
+            expect(result).true;
```
nzjony, line 169: this is the test which required the change to return a Promise<boolean>

karma.options.js
```diff
@@ -170,6 +170,7 @@ const options = function (isCoverage, isMapSdk, prefixDirectory) {
                 allowJs: true
             },
             coverageOptions: {
+                instrumentation: isCoverage ? true : false,
```
nzjony, line 173: This is required to debug locally when running `yarn test-browser`

atomicsulfate:
```suggestion
                instrumentation: isCoverage,
```

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -373,37 +373,21 @@ export class PoiBatchRegistry {
...
-            '${imageTexture?.image ?? imageTextureName}'`);
-        poiInfo.isValid = false;
-        if (missingTextureName.get(imageTextureName) === undefined) {
-            missingTextureName.set(imageTextureName, true);
-            logger.error(`preparePoi: No imageTexture with name '${imageTextureName}' found`);
+            return imageItem;
         }
     }
-    return imageItem;
+    return undefined;
```
atomicsulfate, line 390: How do we catch actual errors, i.e. an image missing from cache that should actually be there, e.g. outdated atlas or style.

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -486,14 +470,29 @@ export class PoiRenderer {
...
             return false;
         }
+        const { imageItem, imageTexture } = this.imageCached(poiInfo.imageTextureName);
+        if (!imageItem) {
+            // Set the POI to be invalid, so it will not be rendered, nor will it be re-checked in a
+            // future request. The item must exist (but not necessarily loaded in the cache before
+            // the POI can be initialized). We don't log an error, because the API of the
+            // MapViewImageCache allows the user to delete an image from the cache. The geometry
+            // creation is handled using some task queue which means that in some cases, a POI entry
+            // may be removed before the geometry is created. See queueGeometryCreation, which
```
atomicsulfate, line 488: I would avoid to make explicit references to code outside this file that is not even part of the API. They become obsolete quickly.

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -486,14 +470,29 @@ export class PoiRenderer {
...
+     * or a Promise if the image has to be loaded first. This happens when the user explicitly
+     * doesn't load the image when it is put into the cache.
      */
-    prepareRender(pointLabel: TextElement, env: Env): boolean {
+    prepareRender(pointLabel: TextElement, env: Env): boolean | Promise<boolean> {
         const poiInfo = pointLabel.poiInfo;
         if (poiInfo === undefined) {
             return false;
         }
+        const { imageItem, imageTexture } = this.imageCached(poiInfo.imageTextureName);
```
atomicsulfate, line 481: There's no need to do this if the poiInfo.buffer is already created (see the if below). Why did you move this out of preparePoi?

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -486,14 +470,29 @@ export class PoiRenderer {
...
+        if (!imageItem) {
+            // Set the POI to be invalid, so it will not be rendered, nor will it be re-checked in a
+            // future request. The item must exist (but not necessarily loaded in the cache before
+            // the POI can be initialized). We don't log an error, because the API of the
+            // MapViewImageCache allows the user to delete an image from the cache. The geometry
+            // creation is handled using some task queue which means that in some cases, a POI entry
+            // may be removed before the geometry is created. See queueGeometryCreation, which
+            // appends some item to the list. In such cases, the POI will just be abandoned and the
+            // geometry will be recreated with the new ImageItem.
+            poiInfo.isValid = false;
```
atomicsulfate, line 491: In the Venues case, the image is not in cache cause the theme is not yet loaded. Here we'll mark pois referencing to images that are not yet loaded as invalid, and they'll never be rendered.

Do you plan to fix that on JSAPI side by waiting for the theme to load?

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -539,7 +538,10 @@ export class PoiRenderer {
 
         if (opacity > 0) {
             if (!poiInfo.buffer) {
-                this.preparePoi(poiInfo.textElement, env);
+                const { imageItem, imageTexture } = this.imageCached(poiInfo.imageTextureName);
```
atomicsulfate, line 541: ditto, why do we need to move this here? Furthermore, now it's in two places, prepareRender() and addPoi()

@here/harp-mapview/test/PoiRendererTest.ts
```diff
@@ -97,4 +103,107 @@ describe("PoiRenderer", function () {
...
+            // This is false because the image is still being loaded in the background, notice the
+            // true flag passed to the call to testCache.addImage.
+            expect(waitingForLoad).false;
+
+            // Promise.resolve used to convert the type `ImageItem | Promise<ImageItem>` to
+            // `Promise<ImageItem>`.
+            await Promise.resolve(testImageItem);
+
+            const imageLoaded = poiRenderer.prepareRender(pointLabel, mapEnv);
+            expect(imageLoaded).true;
```
atomicsulfate, line 161: Why does this pass? Looking at the code, it semes that preparePoi would return false at line 675, cause poinInfo.isValid would have been set to false in the previous call.

@here/harp-mapview/test/PoiRendererTest.ts
```diff
@@ -97,4 +103,107 @@ describe("PoiRenderer", function () {
...
+            const waitingForLoad = poiRenderer.prepareRender(pointLabel, mapEnv);
+            const result = await Promise.resolve(waitingForLoad);
+            expect(result).true;
+
+            // Check that a second attempt to prepareRender succeeds.
+            const imageLoadedAgain = poiRenderer.prepareRender(pointLabel, mapEnv);
+            expect(imageLoadedAgain).true;
+        });
+
+        it("poi is invalid when not in cache, adding to cache doesn't help", async function () {
```
atomicsulfate, line 187: Ditto, this test is saying that the Venues issue is not solved (an image not found because the theme is still being loaded). Will we solve that on JSAPI side?

@here/harp-mapview/test/PoiRendererTest.ts
atomicsulfate: If you added the promise just for testing, there's another option that would leave the production code intact. You can wait for the async tasks to be executed doing:

`const wait = (ms: number = 0) => new Promise(res => setTimeout(res, ms));

await wait();
`
It's used in TileGeometryLoaderTest, TileGeometryCreatorTest and DecodedTilesHelperTest.

atomicsulfate (changes requested):
review done

@here/harp-mapview/lib/poi/PoiRenderer.ts
nzjony: The problem is, with the current architecture / API, we don't know when this situations occurs.

The current task queue geometry creation makes the assumption that the image cache is fixed until the geometry is created, but the user can modify the cache before the task has executed.

I break that assumption and say we can create geometry for something which doesn't exist in the image cache, but the POI is invalid, and it is up to the user to make the tiles dirty and re-request them (such is the responsibility of using a custom cache).

@here/harp-mapview/lib/poi/PoiRenderer.ts
nzjony: I'll remove it.

@here/harp-mapview/lib/poi/PoiRenderer.ts
nzjony: I don't think it is needed, if I check the JSAPI side, the `setTheme`, then `await addDataSource`, which internally anyway `await`s on the theme.

@here/harp-mapview/lib/poi/PoiRenderer.ts
nzjony: I wanted to centralise the point where the poiInfo.isValid = false is set, it is just set in this place, as opposed to being set in two places previously.

And it simplifies the `preparePoi` a bit.

But good point about the fact that I don't need this if the buffer is already created, I'll think about this.

@here/harp-mapview/test/PoiRendererTest.ts
nzjony: There are two things, an item not being in the cache and an item in the cache not being loaded.

The first thing will cause the test to fail, the second won't, because it will be loaded.

@here/harp-mapview/test/PoiRendererTest.ts
nzjony: As discussed, I tried this, but it doesn't work, so will resolve.

@here/harp-mapview/lib/poi/PoiRenderer.ts
nzjony: I moved it, so will resolve.

@here/harp-mapview/test/PoiRendererTest.ts
atomicsulfate: @nzjony You missed this comment.

@here/harp-mapview/lib/poi/PoiRenderer.ts
atomicsulfate: I was thinking about the cache used to store theme images. The user has no control over that one. In theory we should always find images referenced in the style. 
It's quite possible that at some point we use an outdated atlas we an updated style, specially now that we don't generate it automatically. Without even an error in the log, the missing resources might go unnoticed for a long time.

I think it's better to keep logging the errors. We can make them warnings instead of errors, maybe also say in the message that it might just be a transient error.

atomicsulfate (commented):
still some questions open

PR #2284, opened 2021-08-19
MAPSJS-2983 Write to the console if we need too many attempts getting…

nzjony:
… the image from the cache

Some failed attempts is ok, because there are cases where the DataSource is attached and ready
and the map has started to render, but the theme isn't fully set.

There are other ways to fix the problem, but they are generally more invasive and could introduce
regressions, some options are:
- In the TileGeometryCreator::createAllGeometries method, await on the getTheme() Promise
- Don't allow new TextElements to be added while the theme is updating
- Investigate and remove all calls to update the mapview while the method `addDataSource` is called
  and see if this can be removed, because no update calls means we don't render and request new tiles
  which may contain references to POI's that don't yet exist on the cache.

But, for the sake of simplicity, this is the preferred method.

### DIFF
--- a/@here/harp-mapview/lib/poi/PoiRenderer.ts
+++ b/@here/harp-mapview/lib/poi/PoiRenderer.ts
@@ -393,7 +393,7 @@ function findImageItem(
         }
     }
 
-    // There is a texture messing in the cache, we attempt again, and then error out.
+    // There is a texture missing in the cache, we attempt again, and then error out.
     const missingTextureCount = missingTextureName.get(imageTextureName);
     missingTextureName.set(imageTextureName, missingTextureCount ? missingTextureCount + 1 : 0);
     if (missingTextureName.get(imageTextureName)! >= SEARCH_CACHE_ATTEMPTS) {
--- a/@here/harp-mapview/lib/poi/PoiRenderer.ts
+++ b/@here/harp-mapview/lib/poi/PoiRenderer.ts
@@ -261,8 +261,7 @@ export class PoiBatchRegistry {
         const { imageItem, imageTexture } = poiInfo;
 
         if (!imageItem) {
-            // No image -> invisible -> ignore
-            poiInfo.isValid = false;
+            // No image found, therefore just return undefined. It will probably come in soon?
             return undefined;
         }
 
@@ -373,8 +372,10 @@ export class PoiBatchRegistry {
     }
 }
 
-// keep track of the missing textures, but only warn once
-const missingTextureName: Map<string, boolean> = new Map();
+// keep track of the missing textures, we throw an error if the number of attempts goes over some
+// threshold.
+const missingTextureName: Map<string, number> = new Map();
+const SEARCH_CACHE_ATTEMPTS = 5;
 
 function findImageItem(
     poiInfo: PoiInfo,
@@ -382,28 +383,26 @@ function findImageItem(
     imageTexture?: ImageTexture
 ): ImageItem | undefined {
     assert(poiInfo.imageTextureName !== undefined);
-    const imageTextureName = poiInfo.imageTextureName!;
+    const imageTextureName = imageTexture ? imageTexture.image : poiInfo.imageTextureName!;
 
-    let imageItem: ImageItem | undefined;
     for (const imageCache of imageCaches) {
-        imageItem = imageTexture
-            ? imageCache.findImageByName(imageTexture.image)
-            : imageCache.findImageByName(imageTextureName);
+        const imageItem = imageCache.findImageByName(imageTextureName);
         if (imageItem) {
-            break;
+            missingTextureName.delete(imageTextureName);
+            return imageItem;
         }
     }
 
-    if (!imageItem) {
-        logger.error(`init: No imageItem found with name
-            '${imageTexture?.image ?? imageTextureName}'`);
+    // There is a texture messing in the cache, we attempt again, and then error out.
+    const missingTextureCount = missingTextureName.get(imageTextureName);
+    missingTextureName.set(imageTextureName, missingTextureCount ? missingTextureCount + 1 : 0);
+    if (missingTextureName.get(imageTextureName)! >= SEARCH_CACHE_ATTEMPTS) {
+        logger.error(`PoiRenderer::findImageItem: No imageItem found with name:
+            '${imageTexture?.image ?? imageTextureName}'
+            after ${SEARCH_CACHE_ATTEMPTS} attempts.`);
         poiInfo.isValid = false;
-        if (missingTextureName.get(imageTextureName) === undefined) {
-            missingTextureName.set(imageTextureName, true);
-            logger.error(`preparePoi: No imageTexture with name '${imageTextureName}' found`);
-        }
     }
-    return imageItem;
+    return undefined;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -373,37 +372,37 @@ export class PoiBatchRegistry {
...
-            break;
+            missingTextureName.delete(imageTextureName);
+            return imageItem;
         }
     }
 
-    if (!imageItem) {
-        logger.error(`init: No imageItem found with name
-            '${imageTexture?.image ?? imageTextureName}'`);
+    // There is a texture messing in the cache, we attempt again, and then error out.
```
nzjony, line 396:
```suggestion
    // There is a texture missing in the cache, we attempt again, and then error out.
```

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -373,37 +372,37 @@ export class PoiBatchRegistry {
...
-    if (!imageItem) {
-        logger.error(`init: No imageItem found with name
-            '${imageTexture?.image ?? imageTextureName}'`);
+    // There is a texture missing in the cache, we attempt again, and then error out.
+    const missingTextureCount = missingTextureName.get(imageTextureName);
+    missingTextureName.set(imageTextureName, missingTextureCount ? missingTextureCount + 1 : 0);
+    if (missingTextureName.get(imageTextureName)! >= SEARCH_CACHE_ATTEMPTS) {
+        logger.error(`PoiRenderer::findImageItem: No imageItem found with name:
+            '${imageTexture?.image ?? imageTextureName}'
+            after ${SEARCH_CACHE_ATTEMPTS} attempts.`);
```
atomicsulfate, line 402: The problem may still happen as long as we mark the poiInfo as invalid. Consider that the same image might be referenced by many POIs an hence we might get 5 cache misses in a single frame (or many more).
It should be ok to remove the line below. We'll just won't stop trying to load missing resources.

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -373,37 +372,37 @@ export class PoiBatchRegistry {
...
         }
     }
 
-    if (!imageItem) {
-        logger.error(`init: No imageItem found with name
-            '${imageTexture?.image ?? imageTextureName}'`);
+    // There is a texture missing in the cache, we attempt again, and then error out.
+    const missingTextureCount = missingTextureName.get(imageTextureName);
+    missingTextureName.set(imageTextureName, missingTextureCount ? missingTextureCount + 1 : 0);
+    if (missingTextureName.get(imageTextureName)! >= SEARCH_CACHE_ATTEMPTS) {
```
atomicsulfate, line 399: We could log the error just once to avoid cluttering the console. Not sure if it makes sense to do it after n attempts or at the first one. It's difficult to say when "n attempts" is enough, cause it depends on how many POIs use that same image.

@here/harp-mapview/test/PoiBatchRegistryTest.ts
```diff
@@ -25,12 +25,11 @@ describe("PoiBatchRegistry", () => {
         registry = new PoiBatchRegistry(renderer.capabilities);
     });
     describe("registerPoi", function () {
-        it("marks PoiInfo as invalid if it has no image item", () => {
+        it("return undefined if poiInfo has no image item", () => {
```
atomicsulfate, line 28: force here that the poi is still valid.
Also, add some test to check that a poi is valid after a miss on findImageItem()

atomicsulfate (changes requested):
I left some comments.

@here/harp-mapview/lib/poi/PoiRenderer.ts
nzjony: I would argue that if there is an error after 5 attempts, there is something fundamentally wrong and a new approach should be taken (like the ones I mentioned in the commit message). At some point, trying to present the error in a nice way is counter productive.

nzjony: In saying that, I can reduce cluttering the console by changing the `>=` to be just `=`, then we avoid any complex state management.

@here/harp-mapview/lib/poi/PoiRenderer.ts
nzjony: Ok, I'll remove.

@here/harp-mapview/test/PoiBatchRegistryTest.ts
nzjony: Done

@here/harp-mapview/lib/poi/PoiRenderer.ts
```diff
@@ -396,11 +396,10 @@ function findImageItem(
     // There is a texture missing in the cache, we attempt again, and then error out.
     const missingTextureCount = missingTextureName.get(imageTextureName);
     missingTextureName.set(imageTextureName, missingTextureCount ? missingTextureCount + 1 : 0);
-    if (missingTextureName.get(imageTextureName)! >= SEARCH_CACHE_ATTEMPTS) {
+    if (missingTextureName.get(imageTextureName)! === SEARCH_CACHE_ATTEMPTS) {
```
atomicsulfate, line 399: minor: we could do `missingTextureName.get(imageTextureName)` only once.

@here/harp-mapview/test/PoiRendererTest.ts
```diff
@@ -97,4 +103,88 @@ describe("PoiRenderer", function () {
...
+    describe("search for cached images", function () {
+        const webGLRenderer = {
+            capabilities: {
+                isWebGL2: false
+            }
+        } as THREE.WebGLRenderer;
+        const mapViewStub = sinon.createStubInstance(MapView);
+        const poiManager = new PoiManager((mapViewStub as any) as MapView);
+        const testImageName = "testImage";
+        const mapEnv = new Env();
```
atomicsulfate, line 116: move this to a beforeEach() function. Each test should start with a clean slate.

@here/harp-mapview/test/PoiRendererTest.ts
```diff
@@ -97,4 +103,88 @@ describe("PoiRenderer", function () {
...
+        let y = 1;
+
+        const addRandomImageToCache = (
+            testCache: MapViewImageCache,
+            name: string,
+            load: boolean
+        ) => {
+            // Note, the images must be unique, otherwise the test will fail, because the internal
+            // caching mechanism of the ImageCache will have already loaded the image.
+            return testCache.addImage(name, `https://picsum.photos/${x++}/${y++}`, load);
```
atomicsulfate, line 140: Unit tests must not have external dependencies. We can instead stub testCache's findImageByName method and make it return what we want whenever we want in the test.

atomicsulfate (changes requested):

@here/harp-mapview/lib/poi/PoiRenderer.ts
nzjony: Yes, but I would need to then do the +1 twice. What is preferred?

@here/harp-mapview/test/PoiRendererTest.ts
nzjony: Yes, very good point, I will fix.

nzjony: Done, I could use a local resource so I don't need to stub.

atomicsulfate (approved):

atomicsulfate (approved):

PR #2300, opened 2021-08-30
MAPSJS-2998: Issues with the elevated markers

ThFischer:
Signed-off-by: Fischer, Thomas <thomas.fischer@here.com>


atomicsulfate (changes requested):
Please add some testing:
- Best option would be to add an IBCT test. See test "renders elevated point using marker technique" in ScreenSpaceRenderingTests.ts. We could add a new one with 0 tilt, with the camera placed far enough so that the markers are culled without your changes.
- Additionally there could be some unit test in OmvDecodedTileEmitter checking that the min/max height are updated with point elevations.

@here/harp-vectortile-datasource/lib/VectorTileDataEmitter.ts
```diff
@@ -376,6 +376,13 @@ export class VectorTileDataEmitter {
                         scaleHeights
                     );
                 }
+
+                if (tmpV3.z > this.m_maxGeometryHeight) {
+                    this.m_maxGeometryHeight = tmpV3.z
+                } else if (tmpV3.z < this.m_minGeometryHeight) {
+                    this.m_minGeometryHeight = tmpV3.z
+                }
```
atomicsulfate, line 384:
```suggestion
                // For planar projections take the world z coordinate (potentially scaled), for spherical use the z coordinate 
               // in the source projection (no scaling needed).
                const height = this.projection.type === ProjectionType.Planar ? tmpV3.z : pos.z;
                if (height > this.m_maxGeometryHeight) {
                    this.m_maxGeometryHeight = height;
                } else if (height < this.m_minGeometryHeight) {
                    this.m_minGeometryHeight = height;
                }
```

PR #2302, opened 2021-08-30
MAPSJS-2660: Clean MapViewUtils namespace

atomicsulfate:
Remove from MapViewUtils functions that are unrelated to MapView.
MapViewUtils namespace has now only functions related to view transformations, camera setup, and view bounds computation.

### DIFF
--- a/@here/harp-geoutils/lib/tiling/TileKeyUtils.ts
+++ b/@here/harp-geoutils/lib/tiling/TileKeyUtils.ts
@@ -11,62 +11,10 @@ import { Vector3Like } from "../math/Vector3Like";
 import { TileKey } from "./TileKey";
 import { TilingScheme } from "./TilingScheme";
 
-/** @hidden */
-export const powerOfTwo = [
-    0x1,
-    0x2,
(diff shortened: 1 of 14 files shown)


@here/harp-geoutils/lib/tiling/TileKeyUtils.ts
```diff
@@ -11,19 +11,76 @@ import { Vector3Like } from "../math/Vector3Like";
...
+    0x100000000000,
+    0x200000000000,
+    0x400000000000,
+    0x800000000000,
+    0x1000000000000,
+    0x2000000000000,
+    0x4000000000000,
+    0x8000000000000,
+    0x10000000000000
+];
```
ThFischer, line 69: More efficient:
let val = .5;
export const powerOfTwo = Array(53).fill(0).map(() => val *= 2);

atomicsulfate: good catch!

ThFischer (approved):

PR #2350, opened 2021-12-20
HARP-17877: Fixes Style Evaluator expression for cached step expression

FraukeF:

# CI Check

Happy contributing!

### DIFF
--- a/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
+++ b/@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2021 HERE Europe B.V.
+ * Copyright (C) 2019-2022 HERE Europe B.V.
  * Licensed under Apache 2.0, see full license in LICENSE
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -879,8 +879,8 @@ describe("StyleSetEvaluator", function () {
             assert.strictEqual(techniques.length, 0);
         });
     });
-    //not supported at the moment, see HARP-17877
-    it.skip("lookup expressions and lookup table are cached", function () {
+
+    it("lookup expressions and lookup table are cached", function () {
         const lookupTable: JsonArray = [
             "literal",
             [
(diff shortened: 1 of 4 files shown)


gairik (approved):

@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
```diff
@@ -989,8 +990,8 @@ describe("StyleSetEvaluator", function () {
                 const technique = techniques[0] as PoiTechnique;
 
                 const context: AttrEvaluationContext = {
-                    env,
-                    cachedExprResults: styleSetEvaluator.expressionEvaluatorCache
+                    env
+                    // cachedExprResults: styleSetEvaluator.expressionEvaluatorCache
```
AndriiHeonia, line 994: I'd propose to remove this commented out line. You removed "styleSetEvaluator.expressionEvaluatorCache" so here you refer to the property that doesn't exist in the code anymore, it will be confusing for future us.

@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
```diff
@@ -879,7 +879,8 @@ describe("StyleSetEvaluator", function () {
             assert.strictEqual(techniques.length, 0);
         });
     });
-    it("lookup expressions and lookup table are cached", function () {
+    //not supported at the moment, see HARP-17877
+    it.skip("lookup expressions and lookup table are cached", function () {
```
AndriiHeonia, line 883: I'd propose to enable the test back and comment out only `expect(lookupSpy.callCount).equals(entries.length);` and write a comment about HARP-17877 above that line. Because this test also checks that the lookup table, that has been converted from an Array to Map on the first call, is reused again and that it is actually a Map instance.

AndriiHeonia (changes requested):

@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
FraukeF: done

@here/harp-datasource-protocol/test/StyleSetEvaluatorTest.ts
FraukeF: done

AndriiHeonia (approved):

